PR #26, opened 2018-11-27
[HOLD] feat:  implementation of merkle tree and proof

quake:


quake:
@kilb prefer a different leafs sequence to this implementation, let's hold this PR.

doitian:
RFC posted here https://github.com/nervosnetwork/rfcs/pull/35

PR #52, opened 2018-12-04
refactor: rpc and miner

quake:
- [x] refactor: split rpc trait
- [x] refactor: miner


src/cli/app.yml
```diff
@@ -0,0 +1,85 @@
+name: ckb
```
doitian: Clap YAML extension is no longer used.

shared/src/index.rs
```diff
@@ -150,6 +154,25 @@ impl<T: 'static + KeyValueDB> ChainIndex for ChainKVStore<T> {
...
+
+    fn get_candidate_uncles(&self) -> Vec<H256> {
+        self.get(COLUMN_META, META_CANDIDATE_UNCLES_KEY)
+            .map(|raw| deserialize(&raw[..]).unwrap())
+            .unwrap_or_else(Vec::new)
+    }
+
+    fn insert_candidate_uncle(&self, hash: &H256) -> Result<(), SharedError> {
+        let mut uncles = self.get_candidate_uncles();
+        uncles.push(*hash);
```
zhangsoledad: candidate_uncles increase infinity?

miner/src/client.rs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                Ok(new) => {
+                    let is_new_job = |new: &BlockTemplate, old: &BlockTemplate| {
+                        new.raw_header.number() != old.raw_header.number()
+                            || new.commit_transactions.len()
+                                >= old.commit_transactions.len()
+                                    + self.config.new_transactions_threshold as usize
+                    };
+                    let inner = self.shared.inner.upgradable_read();
+                    if inner.clone().map_or(true, |old| is_new_job(&new, &old)) {
```
zhangsoledad: Probably should be `inner.as_ref()`.
Maybe you should use `WriteGuard` directly, those conditional statements need to acquire an exclusive lock.

shared/src/index.rs
quake: > candidate_uncles increase infinity?
Planning to do after this change: https://github.com/nervosnetwork/ckb-dev/issues/47.

How about adding a size limit as a temp solution?

doitian:
Review feedback:

* Rename miner to miner_agent 
* Chain/Pool —notify→ MinerAgent
* MinerAgent is responsible to prepare block template
* RPC gets the template from MinerAgent, and no longer depends on notify.
* Create a module miner. It is a standalone miner, which gets work from the node via RPC.

quake:
@doitian @zhangsoledad pushed mod, please review again, thanks.

miner/src/agent.rs
```diff
@@ -236,8 +236,8 @@ impl<CI: ChainIndex + 'static> Agent<CI> {
                 block.header().number() / self.shared.consensus().difficulty_adjustment_interval();
 
             // uncle must be same difficulty epoch with tip
-            if !block.header().difficulty() == header.difficulty()
```
doitian: ... 😂

pool/src/tests/pool.rs
```diff
@@ -464,8 +464,8 @@ struct TestPool<CI> {
 impl<CI: ChainIndex + 'static> TestPool<CI> {
     fn simple() -> TestPool<ChainKVStore<MemoryKeyValueDB>> {
         let (_handle, notify) = NotifyService::default().start::<&str>(None);
-        let new_tip_receiver = notify.subscribe_new_tip(TXS_POOL_SUBSCRIBER);
-        let switch_fork_receiver = notify.subscribe_switch_fork(TXS_POOL_SUBSCRIBER);
+        let new_tip_receiver = notify.subscribe_new_tip("txs_pool");
+        let switch_fork_receiver = notify.subscribe_switch_fork("txs_pool");
```
zhangsoledad:
```
#[cfg(test)]
......
const TXS_POOL_SUBSCRIBER: &str = "txs_pool";
```
declare here, but is not used

quake: I missed this comment, will fix it later.

pool/src/txs_pool/pool.rs
```diff
@@ -19,6 +19,8 @@ use std::thread::{self, JoinHandle};
 #[cfg(test)]
 use ckb_core::BlockNumber;
 
+const TXS_POOL_SUBSCRIBER: &str = "txs_pool";
```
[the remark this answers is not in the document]
quake: It's used in pool#start function
```
        let new_tip_receiver = self.notify.subscribe_new_tip(TXS_POOL_SUBSCRIBER);
        let switch_fork_receiver = self.notify.subscribe_switch_fork(TXS_POOL_SUBSCRIBER);
```

PR #64, opened 2018-12-05
fix: fix IBD sync process

jjyr:
Fix duplicated Headers messages between peers.

Fix the logic by correct the IBD and `sync` flag to their original intention.


sync/src/synchronizer/peers.rs
```diff
@@ -168,7 +168,7 @@ impl Peers {
     pub fn connected(&self, peer: PeerIndex) {
```
zhangsoledad: Unused method, just delete it.

PR #67, opened 2018-12-07
replace ethereum-types with numext

zhangsoledad:
Main changes:
1. uint and hash types no longer implement `Copy` and `Deref`
2. the minimum supported stable Rust version is 1.30, cause proc-macro

Furthermore:
bump rust-toolchain from 1.29.2 to 1.31.0, 
but not unlock 2018 edition, there will be another PR bring it


rpc/src/service.rs
```diff
@@ -254,10 +256,10 @@ impl<CI: ChainIndex + 'static> RpcService<CI> {
                 block.header().number() / self.shared.consensus().difficulty_adjustment_interval();
 
             // uncle must be same difficulty epoch with tip
-            if !block.header().difficulty() == header.difficulty()
-                || !block_difficulty_epoch == tip_difficulty_epoch
+            if block.header().difficulty() != header.difficulty()
+                || block_difficulty_epoch != tip_difficulty_epoch
```
zhangsoledad: notice here

core/src/transaction.rs
```diff
@@ -177,8 +181,8 @@ impl Transaction {
         self.inputs.len() == 1 && self.inputs[0].previous_output.is_null()
     }
 
-    pub fn hash(&self) -> H256 {
-        sha3_256(serialize(self).unwrap()).into()
+    pub fn hash(&self) -> &H256 {
```
quake: We need to clone this ref in most cases, can we change it to returning owned copy?

zhangsoledad: It's meant to be, you should use borrow as you can, 
If you need a copy, explicitly clone it.

The reason for `Hash` and `Uint` types no longer implement `Copy` is prevent unconscious copy. Essentially, `Hash` is an array filled with byte,  when we use the array, we prefer to using `slice` pointer to access it instead of copy everywhere.

Most other languages copy and allocate all the time by default, while Rust gives a choice of avoiding that in several ways. the same thought here.

README.md
```diff
@@ -31,13 +31,13 @@ The contribution workflow is described in [CONTRIBUTING.md](CONTRIBUTING.md), an
...
-CKB is currently tested mainly with `stable-1.29.2` on Linux and Mac OSX.
+CKB is currently tested mainly with `stable-1.31.0` on Linux and Mac OSX.
 
 We recommend installing Rust through [rustup](https://www.rustup.rs/)
 
 ```bash
 # Get rustup from rustup.rs, then in your `ckb` folder:
-rustup override set 1.29.2
+rustup override set 1.31.0
 rustup component add rustfmt-preview
```
doitian: Should we removed `-preview` here?

README.md
```diff
@@ -31,13 +31,13 @@ The contribution workflow is described in [CONTRIBUTING.md](CONTRIBUTING.md), an
...
 rustup component add clippy-preview
```
doitian: and here

network/src/network.rs
```diff
@@ -741,8 +743,8 @@ impl Network {
         Ok(service_futures)
     }
 
-    #[cfg_attr(feature = "cargo-clippy", allow(type_complexity))]
-    pub fn new(
+    #[allow(clippy::type_complexity)]
+    pub fn try_new(
```
jjyr: Maybe we should use `build` and `inner_build` to replace `try_new` and `build`.

janx:
finally :clap: :clap: :clap:

PR #80, opened 2018-12-11
fix: fix submit block using a temporary solution

xxuejie:


xxuejie:
Close this since it doesn't solve all the issues.

PR #76, opened 2018-12-11
Rc/v0.2.0

doitian:
Changelog: [diff](https://github.com/nervosnetwork/ckb/pull/76/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e) | [preview](https://github.com/nervosnetwork/ckb/blob/rc/v0.2.0/CHANGELOG.md)

PR #84, opened 2018-12-12
fix: remove hash caches to avoid JSON deserialization bug

### LINKED ISSUE #79: Block submit by miner fails in verification
And I've seen this error from the miner log: 

```
ERROR miner  rpc call submit_block error: Rpc(RpcError { code: -32603, message: "Internal error", data: None })
```

It seems the verification fails at here: https://github.com/nervosnetwork/ckb/blob/develop/verification/src/block_verifier.rs#L200

xxuejie:
fixes #79

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -632,10 +632,7 @@ pub mod test {
         } else {
             chain2
         };
-        assert_eq!(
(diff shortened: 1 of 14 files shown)


doitian:
Will be cherry picked to #76

PR #88, opened 2018-12-12
fix: unchecked subtract overflow

zhangsoledad:
* fix: unchecked subtract overflow
* turn off `kad` temporarily
* fix debug log format

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -155,13 +155,17 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
             };
 
             self.shared.store().insert_block(batch, block);
-            self.shared.store().insert_output_root(batch, &block.header().hash(), &root);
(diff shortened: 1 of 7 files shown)


chain/src/chain.rs
```diff
@@ -155,13 +155,17 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+                .store()
+                .insert_output_root(batch, &block.header().hash(), &root);
+            self.shared
+                .store()
+                .insert_block_ext(batch, &block.header().hash(), &ext);
 
             let current_total_difficulty = tip_header.total_difficulty();
             debug!(
-                "difficulty diff = {}; current = {}, cannon = {}",
-                &cannon_total_difficulty - current_total_difficulty,
```
zhangsoledad: unchecked overflow

PR #95, opened 2018-12-16
feat: implement network group and inbound peer eviction

jjyr:
This PR implement network group and inbound peer eviction mechanics mention in [RFC-0007](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0007-scoring-system-and-network-security/0007-scoring-system-and-network-security.md)


network/src/peers_registry.rs
```diff
@@ -112,235 +113,283 @@ impl PeerConnection {
...
+    let mut groups: FnvHashMap<Group, Vec<&'a PeerId>> =
+        FnvHashMap::with_capacity_and_hasher(16, Default::default());
+    let mut largest_group_len = 0;
+    let mut largest_group: Group = Default::default();
+
+    for (peer_id, peer) in peers {
+        let group_name = peer.network_group();
+        let group = groups.entry(group_name.clone()).or_insert_with(Vec::new);
+        group.push(peer_id);
+        if group.len() > largest_group_len {
```
quake: we can use `max_by` to simplify code here

network/src/peers_registry.rs
```diff
@@ -112,235 +113,283 @@ impl PeerConnection {
...
+                .iter()
+                .filter(|(peer_id, peer)| peer.is_inbound() && !self.is_reserved(peer_id));
+            let candidate_peers = find_most_peers_in_same_network_group(inbound_peers);
+            let peer_store = self.peer_store.read();
+
+            let mut lowest_score = std::i32::MAX;
+            let mut low_score_peers = Vec::new();
+            for peer_id in candidate_peers {
+                if let Some(score) = peer_store.peer_score(peer_id) {
+                    if score > lowest_score {
```
quake: please use `min_by`

network/src/peers_registry.rs
```diff
@@ -112,235 +113,283 @@ impl PeerConnection {
...
+            (
+                low_score_peers[..]
+                    .choose(&mut rng)
+                    .unwrap()
+                    .to_owned()
+                    .to_owned(),
+                lowest_score,
+            )
+        };
+        if score < candidate_score {
```
quake: Should we move this condition to the first filter step?

network/src/peers_registry.rs
```diff
@@ -112,235 +113,283 @@ impl PeerConnection {
...
+                if peer_store.is_banned(&peer_id) {
+                    return Err(
+                        ErrorKind::InvalidNewPeer(format!("peer {:?} is denied", peer_id)).into(),
+                    );
+                }
+                peer_store.peer_score_or_default(&peer_id)
+            };
+            let connection_status = self.connection_status();
+            // check peers connection limitation
+            if connection_status.unreserved_inbound >= self.max_inbound
```
quake: `connection_status` function is only used here, I think it's better to extract to a new function `over_max_inbound_limit` and remove `connection_status`

network/src/peers_registry.rs
```diff
@@ -112,235 +113,283 @@ impl PeerConnection {
...
                 ))
                 .into());
             }
-            if self.deny_list.is_denied(&peer_id) {
+            if self.peer_store.read().is_banned(&peer_id) {
                 return Err(
                     ErrorKind::InvalidNewPeer(format!("peer {:?} is denied", peer_id)).into(),
                 );
             }
             let connection_status = self.connection_status();
```
quake: same here

network/src/tests/peers_registry.rs
```diff
@@ -0,0 +1,145 @@
...
+    reserved_peers: Vec<PeerId>,
+) -> PeersRegistry {
+    PeersRegistry::new(
+        peer_store,
+        max_inbound,
+        max_outbound,
+        reserved_only,
+        reserved_peers,
+    )
+}
```
TheWaWaR: 没看出来这个函数的必要性

network/src/peer_store.rs
```diff
@@ -23,31 +19,98 @@ pub enum Status {
...
+
+pub struct ScoringSchema {
+    schema: FnvHashMap<Behaviour, Score>,
+    peer_init_score: Score,
+    ban_score: Score,
+    default_ban_timeout: Duration,
+}
+
+impl ScoringSchema {
+    pub fn new_default() -> Self {
```
quake: should impl `Default` trait

network/src/tests/peers_registry.rs
TheWaWaR: Seems an unnecessary function?

network/src/memory_peer_store.rs
```diff
@@ -1,42 +1,34 @@
...
 struct PeerInfo {
     addresses: Vec<Multiaddr>,
     last_updated_at: Instant,
-    score: u32,
+    score: Score,
     status: Status,
 }
 
 // NOTICE MemoryPeerStore is used for test environment only!!!
 pub struct MemoryPeerStore {
```
quake: we can add `Default` derive to simplify test code

network/src/network_group.rs
```diff
@@ -0,0 +1,49 @@
+use libp2p::core::{AddrComponent, Multiaddr};
+use std::net::IpAddr;
+
+pub type Group = Vec<u8>;
```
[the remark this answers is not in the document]
jjyr: u8 * 4 -> u32

Since the only usage of Group is to categorize addresses, I believe Vec is more clearly and would never cause a collision between difference sized Vector,

network/src/memory_peer_store.rs
jjyr: SQLite based peer store should be the default implementation when it's ready.

network/src/peers_registry.rs
jjyr: We have several unimplemented eviction strategies which described in RFC-0007, this function would be helpful when applying that logic.

network/src/peers_registry.rs
jjyr: Seem can't simplify code here, use `max_by` can remove only two lines but introduce an unnecessary implied loop.

network/src/peers_registry.rs
jjyr: What we want here is a random choice when there're multiple peers have the same score, `min_by` not guaranteed this.

network/src/peers_registry.rs
jjyr: What?

network/src/peers_registry.rs
quake: we can avoid extra variable allocating and simplify code to one line by using `max_by`:
```
peers.fold(FnvHashMap::default(), |mut acc, (peer_id, peer)| {
        acc.entry(peer.network_group()).or_insert_with(Vec::new).push(peer_id);
        acc
    }).values().max_by_key(|peers| peers.len())
```

jjyr: Cool! updated

jjyr:
Implement the new inbound peer eviction strategy https://github.com/nervosnetwork/rfcs/pull/54

network/src/network_group.rs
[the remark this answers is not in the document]
doitian: This is a waste of %50 space in the most conditions today.

network/src/network.rs
```diff
@@ -50,20 +50,20 @@ pub struct SessionInfo {
...
     #[inline]
-    pub fn is_outgoing(&self) -> bool {
+    pub fn is_outbound(&self) -> bool {
         self.endpoint_role == Endpoint::Dialer
     }
 
     #[inline]
     pub fn is_incoming(&self) -> bool {
-        !self.is_outgoing()
+        !self.is_outbound()
```
zhangsoledad: why not `inbound`

network/src/peers_registry.rs
```diff
@@ -88,259 +92,350 @@ type ProtocolConnec = (ProtocolId, UniqueConnec<(UnboundedSender<Bytes>, u8)>);
...
             reserved_only,
-            deny_list,
         }
     }
 
     #[allow(clippy::needless_lifetimes)]
     #[inline]
     pub fn get_peer_id<'a>(&'a self, peer_index: PeerIndex) -> Option<&'a PeerId> {
-        self.peer_connections.get_peer_id(peer_index)
+        self.peers.get_peer_id(peer_index)
```
zhangsoledad:
```
    #[inline]
    fn get_peer_id(&self, peer_index: PeerIndex) -> Option<&PeerId> {
        self.peer_id_by_index.get(&peer_index)
    }

    #[inline]
    fn get_mut(&mut self, peer_id: &PeerId) -> Option<&mut PeerConnection> {
        self.peers.get_mut(peer_id)
    }

    #[inline]
    fn iter(&self) -> impl Iterator<Item = (&PeerId, &PeerConnection)> {
        self.peers.iter()
    }
```
any problem with clippy lint here？ why  dismiss it.

network/src/peers_registry.rs
```diff
@@ -88,259 +92,350 @@ type ProtocolConnec = (ProtocolId, UniqueConnec<(UnboundedSender<Bytes>, u8)>);
...
     pub fn get<'a>(&'a self, peer_id: &PeerId) -> Option<&'a PeerConnection> {
-        self.peer_connections.get(peer_id)
+        self.peers.get(peer_id)
     }
 
     #[inline]
     pub fn get_mut<'a>(&'a mut self, peer_id: &PeerId) -> Option<&'a mut PeerConnection> {
-        self.peer_connections.get_mut(peer_id)
+        self.peers.get_mut(peer_id)
     }
```
zhangsoledad: same as above, needless_lifetimes.

network/src/peers_registry.rs
```diff
@@ -88,259 +92,350 @@ type ProtocolConnec = (ProtocolId, UniqueConnec<(UnboundedSender<Bytes>, u8)>);
...
 
     #[inline]
     pub fn connected_peers_indexes<'a>(&'a self) -> impl Iterator<Item = PeerIndex> + 'a {
-        Box::new(
-            self.peer_connections
-                .peer_id_by_index
-                .iter()
-                .map(|(k, _v)| *k),
-        )
+        Box::new(self.peers.peer_id_by_index.iter().map(|(k, _v)| *k))
```
zhangsoledad: with impl trait, you don't have to box it.
just return iterator directly
`self.peers.peer_id_by_index.iter().map(|(k, _v)| *k)`

network/src/peers_registry.rs
```diff
@@ -88,259 +92,350 @@ type ProtocolConnec = (ProtocolId, UniqueConnec<(UnboundedSender<Bytes>, u8)>);
...
-        let mut deny_list = self.deny_list.lock();
-        if let Some(timeout) = deny_list.get(peer_id).cloned() {
-            if timeout > Instant::now() {
-                return true;
-            } else {
-                deny_list.remove(peer_id);
-            }
-        }
-        false
+        self.peers = Default::default();
```
zhangsoledad: I suggest impl  clear for `PeerConnections`

sync/src/synchronizer/mod.rs
```diff
@@ -1034,7 +1034,7 @@ mod tests {
                 peer_id: random_peer_id().unwrap(),
                 endpoint_role: Endpoint::Dialer,
                 last_ping_time: None,
-                remote_addresses: vec![],
+                connected_addr: "/ip4/127.0.0.1".to_multiaddr().unwrap(),
```
zhangsoledad: Give priority to `expect` instead of `unwrap`

network/src/peers_registry.rs
jjyr: Clippy will output a false warning in the previous version, seem fixed in this version

Issue #54, opened 2018-12-04
Consensus rule on cellbase outputs maturity period


doitian:
- [x] Propose an RFC for this feature.
- [x] Add a consensus spec configuration `COINBASE_MATURITY`
- [x] Index whether a TX is a cellbase in the chain.
- [x] Cellbase outputs must mature 100 blocks before they can be spent. If a tx uses an output of cellbase in block N, it can only be committed in block M where `M - N >= COINBASE_MATURITY`

u2:
This issue can be assigned to me. :)

u2:
https://github.com/nervosnetwork/ckb/blob/develop/core/src/transaction_meta.rs#L20

In TransactionMeta, we already have a `BitVec` for `output_spent`, there are two options:
1. Add a bool indicating whether it's a cellbase;
2. As https://github.com/paritytech/parity-bitcoin/blob/master/storage/src/transaction_meta.rs#L14, the first bit for the cellbase tx.

quake:
@u2 we will remove `TransactionMeta` in favor of `MMR`, may need alternative solution.

u2:
add on: 

> Make this a configurable option in chain spec.
> Only enable it in testnet chain spec. Disable it in dev and integration.

u2:
#481 I think this issue can be closed.

PR #112, opened 2018-12-20
chore: add cellbase maturity

u2:
https://github.com/nervosnetwork/ckb/issues/54


spec/src/consensus.rs
```diff
@@ -31,6 +32,7 @@ pub struct Consensus {
     pub transaction_propagation_timeout: BlockNumber,
     pub pow: Pow,
     pub verification: bool,
+    pub cellbase_maturity: usize,
```
doitian: Add some doc?

u2: Ok

PR #114, opened 2018-12-20
docs: update README for using a simple shell command to modify the config file

rainchen:
according to the subject

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,12 @@ Update `data_dir` configuration in config file to a different directory.
 "data_dir": "node2"
 ```
 
+or using a simple shell command to modify the config file:
+
+```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -156,6 +156,12 @@ Update `data_dir` configuration in config file to a different directory.
 "data_dir": "node2"
 ```
 
+or using a simple shell command to modify the config file:
+
+```shell
+sed -i '' 's/\"data_dir\": \"default\"/\"data_dir\": \"node2\"/g' nodes/node2.json
```
doitian: Well, `sed -i ''` is only valid for FreeBSD sed, this command fails in Linux with GNU sed.

rainchen: right, any suggestion? using pipeline `>` is not DRY to me since the config file is repeating.

rainchen: changed to use `ex` command

PR #116, opened 2018-12-21
fix: Regulate parameters used in syscalls

xxuejie:
This starts as a discussion in one [RFC](https://github.com/nervosnetwork/rfcs/pull/53#discussion_r243179777), turns out the syscall parameters used right now do not make too much sense. Since now we are still in development phase, it's better we fix this now than living with it forever.


doitian:
Oh, my eyes 😂

PR #119, opened 2018-12-24
feat: add transaction filter

quake:
This PR is an implementation of  [transaction filter protocol](https://github.com/nervosnetwork/rfcs/pull/50), but need further improvement:
- Merkle tree proof generation is not ready, waiting for MMR PR and reuse the generation function.
- Sync/Relay protocol share the `Peers` information data, should refactor and move it to network protocol.
- Unify H256 ser/dser.


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -1,5 +1,5 @@
-use super::header_view::HeaderView;
 use crate::synchronizer::{BlockStatus, Synchronizer};
+use crate::types::HeaderView;
```
doitian: I don't see that the mod `header_view` is removed, so now we have two different `HeaderVIew`?

quake: ops, I removed mod ref but forgot to delete this file. will push it later.

sync/src/relayer/transaction_process.rs
```diff
@@ -38,7 +38,15 @@ where
...
 
             for peer_id in self.nc.connected_peers() {
-                if peer_id != self.peer {
+                if peer_id != self.peer
+                    && self
+                        .relayer
+                        .peers()
+                        .transaction_filters
+                        .read()
+                        .get(&self.peer)
```
jjyr: ???? should be peer_id?

sync/src/synchronizer/filter_process.rs
```diff
@@ -0,0 +1,89 @@
...
+    peer: PeerIndex,
+}
+
+impl<'a, CI> AddFilterProcess<'a, CI>
+where
+    CI: ChainIndex + 'a,
+{
+    pub fn new(
+        message: &'a AddFilter,
+        synchronizer: &'a Synchronizer<CI>,
```
jjyr: should receive transaction_fitlers instead the whole Synchronizer?

src/cli/run_impl.rs
```diff
@@ -61,6 +61,7 @@ pub fn run(setup: Setup) {
         chain_controller.clone(),
         shared.clone(),
         tx_pool_controller.clone(),
+        synchronizer.peers(),
```
jjyr: I have notice Peers have been moved out from synchronizer module, here is better to move out the initialize of Peer from Synchronizer struct to keep logical consistent.

protocol/src/builder.rs
```diff
@@ -340,6 +340,47 @@ impl<'a> SyncMessage<'a> {
...
+        fbb: &mut FlatBufferBuilder<'b>,
+        block: &Block,
+        transactions_index: &[usize],
+    ) -> WIPOffset<FilteredBlock<'b>> {
+        let transactions = transactions_index
+            .iter()
+            .map(|ti| {
+                let fbs_transaction = FbsTransaction::build(fbb, &block.commit_transactions()[*ti]);
+                let mut builder = IndexTransactionBuilder::new(fbb);
+                builder.add_index(*ti as u32);
```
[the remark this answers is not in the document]
jjyr: There're other protocol struct builders also use `usize` as the parameter, should be fixed in a standalone PR.

PR #121, opened 2018-12-24
refactor: several rename

zhangsoledad:
- [x] send PR to https://github.com/nervosnetwork/mruby-contracts 
       https://github.com/nervosnetwork/mruby-contracts/pull/3
- [x] send PR to https://github.com/nervosnetwork/ckb-demo-ruby-sdk
       https://github.com/nervosnetwork/ckb-demo-ruby-sdk/pull/27

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -21,7 +21,7 @@ impl From<CrypError> for Error {
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum TxError {
-    OutofBound,
+    OutOfBound,
     NotMatch,
     EmptyGroup,
     WrongFormat,
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -18,11 +18,3 @@ impl From<CrypError> for Error {
         Error::InvalidSignature(e)
     }
 }
-
-#[derive(Debug, PartialEq, Eq)]
-pub enum TxError {
-    OutOfBound,
-    NotMatch,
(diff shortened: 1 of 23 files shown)


core/src/error.rs
```diff
@@ -21,7 +21,7 @@ impl From<CrypError> for Error {
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum TxError {
-    OutofBound,
+    OutOfBound,
```
xxuejie: Can we rename this `OutOfBound` error to something else? I think it's only used now when capacity doesn't match. But it happens to collide with an error with the same name at VM side: https://github.com/nervosnetwork/ckb-vm/blob/79bc6e33a42902c54a5479c96efbfb8898bb7cdc/src/lib.rs#L72 I suggest we rename one of them to avoid confusion in logs.

zhangsoledad: Agreed, `OutOfBound` is ambiguous to describe exceed capacity error.

core/src/transaction.rs
```diff
@@ -79,16 +79,17 @@ pub struct CellOutput {
     pub capacity: Capacity,
     pub data: Vec<u8>,
     pub lock: H256,
-    pub contract: Option<Script>,
+    #[serde(rename = "type")]
```
zhangsoledad: :cry:

zhangsoledad:
@doitian  fixed

PR #125, opened 2018-12-25
refactor: unify h256 and ProposalShortId serialization

quake:
This PR did code refactoring for H256 serialization. Needs rebase after https://github.com/nervosnetwork/ckb/pull/119 

TODO:
- [x] 1. Should we unify these `Bytes` field also?
*NO* `difficulty` / U256 
*YES*`proposal_transactions` and `short_ids` / [u8; 10]

- [x] 2. Client SDK change

### DIFF
--- a/protocol/src/protocol.fbs
+++ b/protocol/src/protocol.fbs
@@ -81,19 +81,9 @@ table CellInput {
 
 table CellOutput {
     capacity:       uint64;
-<<<<<<< HEAD
-<<<<<<< HEAD
     data:           Bytes;
-    lock:           Bytes;
-    type:           Script;
-=======
-    data:           [ubyte];
-=======
-    data:           Bytes;
->>>>>>> Revert "refactor: replace flatten Bytes with [ubyte]"
     lock:           H256;
-    contract:       Script;
->>>>>>> refactor: replace flatten Bytes with [ubyte]
+    type:           Script;
 }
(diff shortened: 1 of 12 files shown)


protocol/src/protocol.fbs
```diff
@@ -65,28 +69,28 @@ table Transaction {
...
-    data:           Bytes;
-    lock:           Bytes;
+    data:           [ubyte];
+    lock:           H256;
     contract:       Script;
 }
 
 table Script {
     version:        uint8;
     args:           [Bytes];
```
xxuejie: If there isn't a way to use something like `[[ubyte]]`, I suggest we still keep `Bytes` instead of `[ubyte]` for consistency.

quake: regarding `Bytes`, if we change `short_ids ` to a new  [u8; 10] struct `ShortId` like `H256`, we can rename it to 'Arg', it's only used in `args` and `signed_args`

doitian: Let's keep Bytes instead of `[ubyte]` for now.

doitian:
Are you plan to change other fixed array in this PR?

quake:
> Are you plan to change other fixed array in this PR?

Yes, plan to change `ShortId` to `[u8; 10]`, will push it later.

PR #130, opened 2018-12-27
define jsonrpc types

zhangsoledad:
- [x] send PR ruby sdk: https://github.com/nervosnetwork/ckb-demo-ruby-sdk

### DIFF
--- a/rpc/src/types/blockchain.rs
+++ b/rpc/src/types/blockchain.rs
@@ -346,3 +346,110 @@ impl From<Block> for CoreBlock {
             .build()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ckb_core::transaction::ProposalShortId as CoreProposalShortId;
+    use proptest::{collection::size_range, prelude::any_with, proptest, proptest_helper};
+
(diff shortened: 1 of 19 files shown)


rpc/src/types/blockchain.rs
```diff
@@ -0,0 +1,455 @@
...
+            reference,
+            binary.map(|b| b.into_vec()),
+            signed_args.into_iter().map(|arg| arg.into_vec()).collect(),
+        )
+    }
+}
+
+impl From<CoreScript> for Script {
+    fn from(core: CoreScript) -> Script {
+        let (version, args, reference, binary, signed_args) = core.destruct();
```
jjyr: `destruct` is only called in `From` and the data is already consumed by borrowship, maybe we can remove `destruct` method?

zhangsoledad: Core types do not expose field directly, so you can't and shouldn't just destruct outside such as 
`let CoreScript{ version, args, reference, binary, signed_args } = core;`
`destruct` are a temporary solution, for now. If we decide to implement zero-copy types in the future, they may change.

PR #134, opened 2018-12-29
feat: past blocks median time based header timestamp verification

jjyr:
1. Shared: Add new method `block_median_time` on `Shared`, get block-median-time from past 11 blocks.
2. HeaderVerifier: New block's timestamp must be greater than parent block's block_metian_time.


shared/src/shared.rs
```diff
@@ -276,6 +278,25 @@ impl<CI: ChainIndex> ChainProvider for Shared<CI> {
...
+        let count = self.consensus.median_time_block_count();
+        let mut block_times = Vec::with_capacity(count);
+        let mut current_header = self.block_header(hash);
+        for _ in 0..count {
+            let header = match current_header {
+                Some(ref header) => header,
+                None => break,
+            };
+            block_times.push(header.timestamp());
+            if block_times.len() < count {
```
doitian: Use `hash` as the iterator control variable, and call `block_header` at the beginning of the loop, you can eliminate this extra condition check.

xxuejie:
Can we document the logic implemented here somewhere? Maybe as an RFC?

spec/src/consensus.rs
```diff
@@ -12,6 +12,8 @@ pub const MAX_UNCLE_AGE: usize = 6;
 pub const TRANSACTION_PROPAGATION_TIME: BlockNumber = 1;
 pub const TRANSACTION_PROPAGATION_TIMEOUT: BlockNumber = 10;
 pub const CELLBASE_MATURITY: usize = 100;
+// TODO: should consider with block time
```
doitian: I cannot catch what does the comment says.

spec/src/consensus.rs
```diff
@@ -36,6 +38,7 @@ pub struct Consensus {
     // it must have at least `cellbase_maturity` confirmations;
     // else reject this transaction.
     pub cellbase_maturity: usize,
+    pub median_time_block_count: usize,
```
doitian: Adds a simple comment like `cellbase_maturity`

verification/src/error.rs
```diff
@@ -95,8 +95,8 @@ pub enum PowError {
 
 #[derive(Debug, PartialEq, Clone, Copy, Eq)]
 pub enum TimestampError {
-    ZeroBlockTime { min: u64, found: u64 },
-    FutureBlockTime { max: u64, found: u64 },
+    BlockTimeTooEarly { min: u64, found: u64 },
```
doitian: I think `BlockTimeTooOld` is better.

verification/src/header_verifier.rs
```diff
@@ -41,38 +44,44 @@ impl<T: HeaderResolver> Verifier for HeaderVerifier<T> {
...
     pub fn verify(&self) -> Result<(), Error> {
-        let min = self.parent.timestamp() + 1;
-        if self.header.timestamp() < min {
-            return Err(Error::Timestamp(TimestampError::ZeroBlockTime {
+        let min = match self
+            .chain_provider
+            .block_median_time(self.header.parent_hash())
+        {
+            Some(time) => time,
+            None => return Err(Error::UnknownParent(self.header.parent_hash().clone())),
```
doitian: Will we run verifier on genesis block?

shared/src/shared.rs
jjyr: Because we release headers inside the loop, we need to call 10 times clone on `header.hash` to keep the outside `hash` variable.

jjyr:
> Can we document the logic implemented here somewhere? Maybe as an RFC?

Current implementation refers to Bitcoin [BIP-0113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki), I' ll make a PR to CKB RFC next week/year.

shared/src/block_median_time_context.rs
```diff
@@ -0,0 +1,23 @@
...
+            match self.timestamp(&current_hash) {
+                Some(timestamp) => block_times.push(timestamp),
+                None => break,
+            }
+            match self.parent_hash(&current_hash) {
+                Some(hash) => current_hash = hash,
+                None => break,
+            }
+        }
+        block_times.sort_by(|a, b| b.cmp(a));
```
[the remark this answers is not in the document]
doitian: It's a reversed sorting.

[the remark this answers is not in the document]
quake: > It's a reversed sorting.

it's same for `median`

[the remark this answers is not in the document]
jjyr: It is a reverse sort and it's purpose is to reduce elements swap, notice the block_times is reversed ordering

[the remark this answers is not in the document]
quake: We can use `sort_unstable` here for the purpose of optimization

PR #157, opened 2019-01-04
feat: Add block level script cycle limit

xxuejie:
An RFC for this PR will be added soon.

### DIFF
--- a/verification/src/block_verifier.rs
+++ b/verification/src/block_verifier.rs
@@ -6,6 +6,7 @@ use ckb_core::block::Block;
 use ckb_core::cell::{CellProvider, CellStatus};
 use ckb_core::header::Header;
 use ckb_core::transaction::{Capacity, CellInput, OutPoint};
+use ckb_core::Cycle;
 use ckb_shared::shared::ChainProvider;
 use fnv::{FnvHashMap, FnvHashSet};
 use merkle_root::merkle_root;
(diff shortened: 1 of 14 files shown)


verification/src/block_verifier.rs
```diff
@@ -447,26 +448,30 @@ impl<P: ChainProvider + CellProvider> TransactionsVerifier<P> {
...
         // make verifiers orthogonal
         // skip first tx, assume the first is cellbase, other verifier will verify cellbase
-        let err: Vec<(usize, TransactionError)> = block
+        let results: Vec<Result<Cycle, TransactionError>> = block
             .commit_transactions()
             .par_iter()
             .skip(1)
             .map(|x| wrapper.resolve_transaction_at(x, &parent_hash))
+            .map(|tx| TransactionVerifier::new(&tx).verify(max_cycles))
+            .collect();
```
doitian: Do we need to collect all the errors here? I think [while_some](https://docs.rs/rayon/1.0.3/rayon/iter/trait.ParallelIterator.html#method.while_some) can be used here to return early. However, this is not related to this PR.

For this PR, we also can use some advanced methods from rayon, such as [try_fold](https://docs.rs/rayon/1.0.3/rayon/iter/trait.ParallelIterator.html#method.try_fold) to return early when the total cycles exceed the limit. Also, see other methods starting with `try_` if `try_fold` does not work.

xxuejie: @doitian @zhangsoledad How about now?

verification/src/block_verifier.rs
```diff
@@ -447,26 +447,50 @@ impl<P: ChainProvider + CellProvider> TransactionsVerifier<P> {
...
+                                    if new_cycles > max_cycles {
+                                        Err(Error::ExceededMaximumCycles)
+                                    } else {
+                                        Ok(new_cycles)
+                                    }
+                                })
+                        })
+                },
+            )
+            .try_reduce(
```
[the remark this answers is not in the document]
xxuejie: I think that's a pattern in rayon: `try_fold` only gives a `TryFold` struct, which might contain different portions of the processed data, `try_reduce` will do a final reduction and reduce the results into a single value. My understanding here, is this is more like a `map-reduce` pattern: `try_fold` does the map phase, and `try_reduce` does the final reduce phase.

PR #158, opened 2019-01-06
feat: using delayed verify

kilb:


core/src/cell.rs
```diff
@@ -56,90 +55,81 @@ pub struct ResolvedTransaction {
...
-            transaction: transaction.clone(),
-            input_cells,
-            dep_cells,
-        }
+        let mut seen_inputs = FnvHashSet::default();
+        resolve_transaction(transaction, &mut seen_inputs, |x| self.cell(x))
     }
 
-    fn resolve_transaction_at(
+    fn resolve_transaction_at<F: Fn(&OutPoint) -> Option<bool>>(
```
quake: Unused function?

core/src/cell.rs
```diff
@@ -56,90 +55,81 @@ pub struct ResolvedTransaction {
...
-            input_cells,
-            dep_cells,
-        }
+        let mut seen_inputs = FnvHashSet::default();
+        resolve_transaction(transaction, &mut seen_inputs, |x| {
+            self.cell_at(x, |o| is_spent(o))
+        })
     }
 
     fn resolve_transaction_unknown_inputs(&self, resolved_transaction: &mut ResolvedTransaction) {
```
quake: Unused function?

core/src/cell.rs
```diff
@@ -56,90 +55,81 @@ pub struct ResolvedTransaction {
...
-        ) {
-            if *state == CellStatus::Unknown {
-                *state = self.cell(out_point);
+        resolve_transaction_unknown_inputs(resolved_transaction, |x| self.cell(x))
+    }
+}
+
+pub fn resolve_transaction<F: Fn(&OutPoint) -> CellStatus>(
+    transaction: &Transaction,
+    seen_inputs: &mut FnvHashSet<OutPoint>,
```
quake: `seen_inputs` is always initialized as an empty set, can we remove it from function params?

core/src/header.rs
```diff
@@ -244,3 +244,26 @@ impl HeaderBuilder {
         self.inner
     }
 }
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug, Default)]
+pub struct RichHeader {
```
quake: Weird name... need documentation or better naming.

core/src/transaction_meta.rs
```diff
@@ -44,7 +44,7 @@ impl TransactionMeta {
     }
 
     pub fn is_spent(&self, index: usize) -> bool {
-        self.output_spent.get(index).unwrap_or(false)
+        self.output_spent.get(index).unwrap_or(true)
```
quake: Why return `true` when index is out of bounds?

shared/src/txo_set.rs
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    pub fn mark_unspent(&mut self, o: &OutPoint) {
+        if let Some(meta) = self.inner.get_mut(&o.hash) {
+            meta.unset_spent(o.index as usize);
+        }
+    }
+
+    pub fn roll_back(&mut self, inputs: Vec<OutPoint>, outputs: Vec<H256>) {
```
quake: this fn should be private and it's better to use `rollback` as fn name

shared/src/txo_set.rs
```diff
@@ -0,0 +1,81 @@
...
+        for h in outputs {
+            self.remove(&h);
+        }
+
+        for o in inputs {
+            self.mark_unspent(&o);
+        }
+    }
+
+    pub fn forward(&mut self, inputs: Vec<OutPoint>, outputs: Vec<(H256, usize)>) {
```
quake: should be private fn

shared/src/txo_set.rs
```diff
@@ -0,0 +1,81 @@
...
+        self.inner.remove(hash)
+    }
+
+    pub fn mark_spent(&mut self, o: &OutPoint) {
+        if let Some(meta) = self.inner.get_mut(&o.hash) {
+            meta.set_spent(o.index as usize);
+        }
+    }
+
+    pub fn mark_unspent(&mut self, o: &OutPoint) {
```
quake: should be private fn

chain/src/chain.rs
```diff
@@ -60,12 +63,7 @@ pub struct BlockInsertionResult {
 
 impl<CI: ChainIndex + 'static> ChainService<CI> {
     pub fn new(shared: Shared<CI>, notify: NotifyController) -> ChainService<CI> {
-        let block_verifier = BlockVerifier::new(shared.clone());
-        ChainService {
-            shared,
-            notify,
-            block_verifier,
-        }
+        ChainService { shared, notify }
```
quake: Why PR #140 is reverted here?

kilb: the `consensus` is Arc<> now, `clone()` is not a problem. PR#140 should not be approved.

core/src/transaction_meta.rs
kilb: This output will not appear, and can not be spent, just like being spent.

core/src/cell.rs
kilb: No, seen_inputs is not always initialized(in block_verify.rs).

core/src/cell.rs
kilb: may be used in future.

core/src/cell.rs
quake: > No, seen_inputs is not always initialized(eg in block_verify.rs).

It's initialized as an empty set also in block_verify.rs line 395:

```
let mut seen_inputs = FnvHashSet::default();
```

kilb: but it is passed in `resolve_transaction` many times.

shared/src/txo_set.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    pub fn get(&self, h: &H256) -> Option<&TransactionMeta> {
+        self.inner.get(h)
+    }
+
+    pub fn insert(&mut self, hash: H256, outputs_len: usize) {
+        self.inner.insert(hash, TransactionMeta::new(outputs_len));
+    }
+
+    pub fn remove(&mut self, hash: &H256) -> Option<TransactionMeta> {
```
doitian: The fully spent transaction can be removed to save memory usage.

chain/src/chain.rs
```diff
@@ -281,52 +244,223 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+            exts.push((hash, ext));
+        }
+
+        //TODO: remove this clone
+        new_blocks.push(block.clone());
+
+        hash = block.header().parent_hash().clone();
+
         loop {
             if let Some(old_hash) = self.shared.block_hash(number) {
```
doitian: Better to add an assert/expect here. If number is less than or equal to tip_number, it must return a hash. If number is larger than tip_number, we don't need to bother looking up the hash from the store.

chain/src/chain.rs
```diff
@@ -281,52 +244,223 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
                 if old_hash == hash {
+                    verified = true;
```
doitian: I still not quite understand what you mean `verified` here. If the new fork contains only pending verification blocks, the loop finally stops at the common ancestor and will return `verified` as true.

chain/src/chain.rs
```diff
@@ -281,52 +244,223 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+        let mut new_outputs = FnvHashMap::default();
+
+        if verified {
+            let new_blocks_iter = new_blocks.iter().rev();
+            let old_blocks_iter = old_blocks.iter().rev();
+            let new_blocks_len = new_blocks.len();
+
+            let verified_len = new_blocks_len - exts.len();
+
+            for b in old_blocks_iter.skip(verified_len) {
```
doitian: Why skip `verified_len` old blocks here?

chain/src/chain.rs
```diff
@@ -281,52 +244,223 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+        if number < tip_number {
+            for n in number..=tip_number {
+                let hash = self.shared.block_hash(n).unwrap();
+                let old_block = self.shared.block(&hash).unwrap();
+
+                old_blocks.push(old_block);
+            }
+        }
+
+        let mut verified = ext.valid.is_some();
```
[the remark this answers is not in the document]
kilb: if block is invalid, the result will also be saved.

chain/src/chain.rs
kilb: Block number is not sufficient to estimate which chain is better. It is easy to construct a chain which is longer but has much lower difficulty.

chain/src/chain.rs
kilb: skip the meaning of verified, I will change the name. It just meaning no error happened.

chain/src/chain.rs
kilb: Because they have been verified, we do not need verify it again.

chain/src/chain.rs
doitian: My point is:

- If `number > tip_number`, it must be absent, we don't need to look up it in the store.
- If `number <= tip_number`, the store must return its hash, we should assert it.

chain/src/chain.rs
doitian: But doesn't all old blocks have been verified? Since these old blocks are in the previous main chain. Also `verified_len` is counted based on new blocks and the new blocks pending verification, I don't see a relationship between old blocks here.

doitian:
@kilb conflict with #223 , could you resolve the conflicts?

chain/src/chain.rs
```diff
@@ -111,116 +116,94 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+            "difficulty current = {}, cannon = {}",
+            current_total_difficulty,
+            cannon_total_difficulty,
+        );
 
+        let ext = BlockExt {
+            received_at: unix_time_as_millis(),
+            total_difficulty: cannon_total_difficulty.clone(),
+            total_uncles_count: parent_ext.total_uncles_count + block.uncles().len() as u64,
+            valid: if parent_ext.valid == Some(false) {
```
quake: Why continue processing the new block here if its parent block validation result is false?

core/src/header.rs
quake: I still don't understand why to define a new struct in core mod with a weird name, is it only used in sync mod?  can we move it to sync mod and just use a type alias?

shared/src/shared.rs
```diff
@@ -95,77 +107,86 @@ impl<CI: ChainIndex> Shared<CI> {
...
-                    CellStatus::Live(transaction.outputs()[index].clone())
-                } else {
-                    CellStatus::Dead
+    pub fn init_txo_set(store: &CI, mut hash: H256) -> TxoSet {
+        let mut txo_set = TxoSet::new();
+        let mut blocks = Vec::new();
+        while let Some(block) = store.get_block(&hash) {
+            let number = block.header().number();
+            hash = block.header().parent_hash().clone();
+            blocks.push(block);
```
quake: need to resolve memory issue here, e.g. it will consume 10G memory when restarting a node with 10,000 * 1M block.

doitian: Good catch

chain/src/chain.rs
kilb: we should save it

shared/src/shared.rs
kilb: fixed

core/src/header.rs
doitian: @zhangsoledad is working on this.

chain/src/chain.rs
doitian: We can just save this ext and quit.

PR #173, opened 2019-01-08
fix: turn on shared option for SqlitePeerStore memory mode

jjyr:


zhangsoledad:
https://www.sqlite.org/inmemorydb.html

PR #179, opened 2019-01-09
feat: collect clock time offset from network peers

### LINKED ISSUE #170: Bitcoin-like local time warning
Bitcoin-like local time warning

jjyr:
resolve #170

### DIFF
--- a/sync/src/net_time_checker.rs
+++ b/sync/src/net_time_checker.rs
@@ -1,0 +1,161 @@
+use ckb_network::{CKBProtocolContext, CKBProtocolHandler, PeerIndex};
+use ckb_protocol::TimeMessage;
+use ckb_util::RwLock;
+use flatbuffers::{get_root, FlatBufferBuilder};
+use log::{debug, warn};
+use std::collections::VecDeque;
+
+const TOLERANT_OFFSET: u64 = 7_200_000;
+const MIN_SAMPLES: usize = 5;
+const MAX_SAMPLES: usize = 11;
+
+/// Collect and check time offset samples
+pub struct NetTimeChecker {
+    /// Local clock should has less offset than this value.
+    tolerant_offset: u64,
+    max_samples: usize,
+    min_samples: usize,
+    samples: VecDeque<i64>,
+}
+
+impl NetTimeChecker {
+    pub fn new(min_samples: usize, max_samples: usize, tolerant_offset: u64) -> Self {
+        NetTimeChecker {
+            min_samples,
+            max_samples,
+            tolerant_offset,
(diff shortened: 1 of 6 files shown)


sync/src/net_time_checker.rs
```diff
@@ -0,0 +1,165 @@
...
+const MIN_SAMPLES: usize = 5;
+const MAX_SAMPLES: usize = 11;
+
+/// Collect and check time offset samples
+pub struct NetTimeChecker {
+    /// Local clock should has less offset than this value.
+    tolerant_offset: u64,
+    max_samples: usize,
+    min_samples: usize,
+    samples: VecDeque<i64>,
```
quake: check out: https://leetcode.com/problems/find-median-from-data-stream/
solution tab: `Approach #3 Two Heaps! [Accepted]`

jjyr: It's too complex. We only need to maintain 8~1x samples here.

sync/src/net_time_checker.rs
```diff
@@ -0,0 +1,165 @@
...
+        if nc.session_info(peer).map(|s| s.peer.is_outbound()) == Some(true) {
+            let now = faketime::unix_time_as_millis();
+            let timestamp = match get_root::<TimeMessage>(&data)
+                .payload()
+                .map(|time| time.timestamp())
+            {
+                Some(timestamp) => timestamp,
+                None => return,
+            };
+            let offset: i64 = if now > timestamp {
```
quake:
```suggestion
            let offset = (now  as i64) - (timestamp as i64);
```

sync/src/net_time_checker.rs
quake: the lines of code are similar to the current algorithm, and we can extract it to a util lib and reuse it in `BlockMedianTimeContext`, anyway, I will submit a PR for this improvement

sync/src/net_time_checker.rs
jjyr: `now` is u64, direct cast it to i64 may cause overflow, instead, we cast the calculation result.

sync/src/net_time_checker.rs
jjyr: Notice below code we also need to maintain last 11 offset samples, so we need a FIFO queue here, it is definitely more complex to implement this in the two-heap algorithm.

sync/src/net_time_checker.rs
zhangsoledad: `overflowing_sub`

doitian:
ping @quake

PR #186, opened 2019-01-12
docs: change PascalCase parameters to snake_case style

classicalliu:


rpc/doc.md
```diff
@@ -182,9 +181,9 @@ Returns the information about cells collection by type_hash.
 
 ## Parameters
 
-    Type_hash - Cell type_hash.
-    From - Start BlockNumber.
-    To - End BlockNumber.
+    type_hash - Cell type_hash.
```
doitian:
```suggestion
    type_hash - Cell type hash.
```

The part after `-` is a description, I don't think we need the underscore in it. So does the following two lines.

rpc/doc.md
```diff
@@ -182,9 +181,9 @@ Returns the information about cells collection by type_hash.
...
+    from - Start block_number.
```
doitian:
```suggestion
    from - Start block number.
```

rpc/doc.md
```diff
@@ -182,9 +181,9 @@ Returns the information about cells collection by type_hash.
...
+    to - End block_number.
```
doitian:
```suggestion
    to - End block number.
```

rpc/doc.md
classicalliu: Nice, I'll update it.

Issue #221, opened 2019-01-22
flag "-c" not working on command "ckb run"


mohanson:
```
$ /src/ckb/target/release/ckb run -c /tmp/nodes/node2.json 
$ 2019-01-22 23:59:16.821 +08:00 main INFO main  Setup with config /tmp/nodes/default.json
```
Still "/tmp/nodes/default.json" but "-c nodes/node2.json" has been given.

**Environment**
 - OS: Ubuntu
 - Version: develop branch on github

clairecuijing:
I tried in v0.6.0 "-c" is already working on command "ckb run".

PR #222, opened 2019-01-23
fix: cli arg matches

zhangsoledad:


zhangsoledad:
https://github.com/nervosnetwork/ckb/issues/221

PR #223, opened 2019-01-24
feat: use new merkle tree

kilb:


kilb:
> Please setup CI in https://github.com/nervosnetwork/merkle-tree

Only admin can setup CI.

util/merkle-tree/Cargo.toml
```diff
@@ -1,9 +1,9 @@
...
+name = "ckb-merkle-tree"
 version = "0.5.0-pre"
 license = "MIT"
 authors = ["Nervos Core Dev <dev@nervos.org>"]
 edition = "2018"
 
-[dev-dependencies]
-proptest = "0.8"
+[dependencies]
+merkle-tree = { git = "https://github.com/nervosnetwork/merkle-tree", rev="79a00ed", features = ["sha3"] }
```
quake: Would you publish it to crates.io and use versioned lib here?

kilb: maybe

PR #224, opened 2019-01-24
refactor chain spec config

janx:
Chain config should be at the same level as `network`, as they're the same level modules.

This PR changes chain config to:

```json
"chain": {
    "spec": "path/to/chain/spec.json"
}
```


janx:
Force-pushed updated commit.

Issue #55, opened 2018-11-07
Make rocksdb configurable via config file


doitian:

doitian:
可以挑一些重要的选项方便来调整

doitian:
研究下 rocksdb 是不是自己支持某种格式的配置文件

doitian:
Review Log
**Last Reviewed At** — 2019-01-11

- Add options on demand, when we really need to tweak the options.
- Check whether rocksdb already supports some kind of config file.

janx:
RocksDB already supports config file. I propose to add following options to node json and leave RocksDB options in a separate file:

```json
"db": {
    "rocksdb": "path/to/the/config/file"
}
```

janx:
RocksDB supports config file but rust-rocksdb doesn't expose its load API. Instead, rust-rocksdb provides its own option interface. I'll add rust-rocksdb options to config file.

zhangsoledad:
Closed by https://github.com/nervosnetwork/ckb/pull/230

PR #230, opened 2019-01-26
feat: configurable rocksdb

janx:
Implemente https://github.com/nervosnetwork/ckb/issues/55.


quake:
The code would be much simpler if we sync our forked `rust-rocksdb` to upstream, there are `set_options` function in the new version,  a string key/value map can be used in the config file, and no need to define RocksDBConfig struct in our code.

janx:
Refactored to use `set_options` and upgraded `rust-rocksdb` library.

janx:
Conflicts resolved.

PR #238, opened 2019-01-29
[WIP] graceful shutdown

zhangsoledad:


zhangsoledad:
1. when received SIGTERM, RPC shutdown first, network shutdown next, then all controller get dropped,  service will close automatically.
2. verification MUST NOT be consensus params.
3. refactory service interface.

util/stop-handler/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+
+#[derive(Debug)]
+struct Handler<T> {
+    signal: SignalSender,
+    thread: JoinHandle<T>,
+}
+
+#[derive(Clone, Debug)]
+pub struct StopHandler<T> {
+    inner: Option<Arc<Mutex<Option<Handler<T>>>>>,
```
doitian: Why need two `Option`?

doitian:
Please resolve the conflicts.

util/stop-handler/src/lib.rs
zhangsoledad: Usually we call `try_send(&mut self)` in `Drop::drop(&mut self)`,
but we need take ownership of inner structs, `Option` is for that.

the outer `Option`  take ownership for `Arc::try_unwrap`
```
 pub fn try_unwrap(this: Self) -> Result<T, Self>
```

the inner `Option `  take ownership for `oneshot::Sender` and `JoinHandle`

doitian: Looks weird, better add this as a comment.

TheWaWaR: I think use `Option` is a common technique to take ownership from a `&mut self`

PR #239, opened 2019-01-30
feat: skip verifing those transactions that from pool

kilb:


chain/src/chain.rs
```diff
@@ -117,7 +126,11 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
         }
     }
 
-    fn process_block(&mut self, block: Arc<Block>) -> Result<(), ProcessBlockError> {
+    fn process_block(
+        &mut self,
+        block: Arc<Block>,
+        cycles_set: Vec<Cycle>,
```
jjyr: This `cycles_set` parameter is confused me, it spent my lots of time to understand the behaviour:
if cycles_set[index] equals to Cycle::default() then it will use max_cycles to verify then rewrite the cycles to cycles_set, otherwise it use cycles_set[index] as max_cycles to execute validation.

It is strange, do we really need such behavior?
A better and simple solution is computing the actual cycles in chain and pool and return the cycles to let the invoker to decide whether it is valid or not.

jjyr: We only need to compare the actual txs cycles to the txs cycles we received from network peers to determine a peer is honest or lying. The Pool and Chain should cache the cycles, but it is not necessary to pass a cycles limit as a parameter.

kilb: The cycles_set records the cycles actually used of all transactions, if it is Cycle::default(), it means that the transaction has not been executed and the value is unknown. If it is not Cycle::default(), the verification will skiped. We need a cycles_set because if we want to readd the transaction in the pool when switch fork, we need know the cycles.

kilb: The verification behavior is different.

chain/src/chain.rs
```diff
@@ -117,7 +126,11 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
         }
     }
 
-    fn process_block(&mut self, block: Arc<Block>) -> Result<(), ProcessBlockError> {
+    fn process_block(
```
quake: This function intends to verify and insert a new block to store, I think that `cycles_set` is always passed as an empty vector, we should remove it from here and `insert_block` function below.

jjyr:
Pls move the cache into Verifier as we discussed.

chain/src/chain.rs
```diff
@@ -491,52 +482,31 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
         // update exts
-        for (ext, b) in fork
-            .open_exts
-            .iter_mut()
-            .zip(fork.new_blocks.iter())
-            .rev()
-            .take(new_len)
-        {
-            ext.valid = Some(true);
+        for (ext, b) in fork.open_exts.iter().zip(fork.new_blocks.iter()).rev() {
```
zhangsoledad: Can we merge two `Iterator`？ Basically the same。

PR #242, opened 2019-02-02
Upgrade secp256k1

driftluo:
Upgrade secp256k1

### DIFF
--- a/util/crypto/src/secp/generator.rs
+++ b/util/crypto/src/secp/generator.rs
@@ -1,4 +1,3 @@
-use super::error::Error;
 use super::privkey::Privkey;
 use super::pubkey::Pubkey;
 use super::SECP256K1;
@@ -13,23 +12,21 @@ impl Generator {
     }
 
     pub fn random_privkey(&self) -> Privkey {
-        let mut random_slice: [u8; secp256k1::constants::SECRET_KEY_SIZE] = rand::random();
         loop {
-            match SecretKey::from_slice(&random_slice) {
-                Ok(sec) => return sec.into(),
-                Err(_) => {
-                    random_slice = rand::random();
-                }
+            if let Ok(sec) = SecretKey::from_slice(&rand::random::<
+                [u8; secp256k1::constants::SECRET_KEY_SIZE],
+            >()) {
+                return sec.into();
             }
         }
     }
 
-    pub fn random_keypair(self) -> Result<(Privkey, Pubkey), Error> {
+    pub fn random_keypair(self) -> (Privkey, Pubkey) {
         let privkey = self.random_privkey();
-        let sec = SecretKey::from_slice(privkey.as_bytes())?;
+        let sec = SecretKey::from_slice(privkey.as_bytes()).unwrap();
         let publ = PublicKey::from_secret_key(&SECP256K1, &sec);
 
-        Ok((privkey, publ.into()))
+        (privkey, publ.into())
     }
 }
 
--- a/util/crypto/src/secp/generator.rs
+++ b/util/crypto/src/secp/generator.rs
@@ -13,10 +13,15 @@ impl Generator {
     }
 
     pub fn random_privkey(&self) -> Privkey {
-        let random_slice: [u8; secp256k1::constants::SECRET_KEY_SIZE] = rand::random();
-        SecretKey::from_slice(&random_slice)
-            .expect("slice has the right size")
-            .into()
+        let mut random_slice: [u8; secp256k1::constants::SECRET_KEY_SIZE] = rand::random();
+        loop {
+            match SecretKey::from_slice(&random_slice) {
+                Ok(sec) => return sec.into(),
+                Err(_) => {
+                    random_slice = rand::random();
+                }
+            }
+        }
     }
 
     pub fn random_keypair(self) -> Result<(Privkey, Pubkey), Error> {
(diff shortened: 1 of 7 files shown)


util/crypto/src/secp/generator.rs
```diff
@@ -13,14 +13,21 @@ impl Generator {
...
+        }
     }
 
     pub fn random_keypair(self) -> Result<(Privkey, Pubkey), Error> {
-        let mut rng = rand::thread_rng();
-
-        let (sec, publ) = SECP256K1.generate_keypair(&mut rng)?;
+        let random_slice: [u8; secp256k1::constants::SECRET_KEY_SIZE] = rand::random();
+        let sec = SecretKey::from_slice(&random_slice)?;
+        let publ = PublicKey::from_secret_key(&SECP256K1, &sec);
```
jjyr: Use the `random_privkey` above to generate a valid secret_key to maintain the consistency of this function.

https://github.com/nervosnetwork/rust-secp256k1/blob/master/src/lib.rs#L526

util/crypto/src/secp/generator.rs
```diff
@@ -13,16 +13,23 @@ impl Generator {
     }
 
     pub fn random_privkey(&self) -> Privkey {
-        let mut rng = rand::thread_rng();
-        key::SecretKey::new(&SECP256K1, &mut rng).into()
+        let mut random_slice: [u8; secp256k1::constants::SECRET_KEY_SIZE] = rand::random();
```
quake: secp256k1 re-exports dep crate after version 0.11.5:
https://github.com/rust-bitcoin/rust-secp256k1/blob/master/CHANGELOG.md#0115---2018-11-09

We can use this feature directly:
```rust
SecretKey::new(&mut secp256k1::rand::thread_rng()).into()
```

driftluo: @quake Why do we fork it? just because of `rand` 0.4 0.5 0.6 coexistence will conflict.

By writing this, I can guarantee that we are using the latest secp256 without affecting the use of the latest `rand` elsewhere.

I don't think we need to fork, nor do we need to force downgrade the `rand` crate.

quake: not forking, just enable the `rand` feature in Cargo.toml:
```
secp256k1 = { version = "0.12.2", optional = true, features = ["rand"] }
```

driftluo: Secp256k1 exports a version of 0.4, but we don't really need this feature, so I don't use it.

I remember that there was a problem with `rand` crate multi-version conflicts in history. In order to reduce this unnecessary trouble, I think really does not need to use this feature

jjyr:
It is strange that we hold H256 instead of `secp256k1::Secretkey`, why don't direct hold SecretKey?

util/crypto/src/secp/generator.rs
quake: Yes, secp256k1 has a version conflict issue before 0.11.5 and therefore it exports dep crate to resolve this issue:

https://github.com/rust-bitcoin/rust-secp256k1/pull/80

driftluo: We use `rand` 0.6 on `ckb-protocol`、`ckb-shared`、`bloom-filters` and lot of crate.

Enable secp256k1 `rand` feature doesn't give us any benefit other than letting us compile more than one library.

util/crypto/src/secp/generator.rs
```diff
@@ -13,16 +13,23 @@ impl Generator {
     }
 
     pub fn random_privkey(&self) -> Privkey {
```
quake: I prefer `loop` to `while`:
```
    pub fn random_privkey(&self) -> Privkey {
        loop {
            let sec = SecretKey::from_slice(&rand::random::<[u8; SECRET_KEY_SIZE]>());
            if sec.is_ok() {return sec.unwrap().into()}
        }
    }
```

util/crypto/src/secp/generator.rs
```diff
@@ -13,16 +13,23 @@ impl Generator {
...
+            match SecretKey::from_slice(&random_slice) {
+                Ok(sec) => return sec.into(),
+                Err(_) => {
+                    random_slice = rand::random();
+                }
+            }
+        }
     }
 
     pub fn random_keypair(self) -> Result<(Privkey, Pubkey), Error> {
```
quake: Can we return turple here? since we have changed the inner logic of `random_privkey` which always return a valid key.

util/crypto/src/secp/generator.rs
driftluo: maybe it's better:
```rust
   pub fn random_privkey(&self) -> Privkey {
        loop {
            let sec = SecretKey::from_slice(&rand::random::<[u8; SECRET_KEY_SIZE]>());
           if let Ok(sec) = sec {
              return sec.into()
           }
        }
    }
```

util/crypto/src/secp/generator.rs
driftluo: It really works, but since I have to modify all the code that uses this API, I am hesitant.

util/crypto/src/secp/generator.rs
quake: LGTM

util/crypto/src/secp/generator.rs
quake: `random_keypair ` is only used in testing code if I remember correctly.

driftluo: Oh，let‘s do it right now

zhangsoledad:
Welcome to contribute, although we have planned to remove this when SDK prepare to take the job.

PR #263, opened 2019-02-11
Revert "Upgrade secp256k1"

doitian:
Reverts nervosnetwork/ckb#242 since it breaks Linux


driftluo:
This is due to the link failure caused by two versions of the c library.

There are now two solutions to this problem:
1. revert upgrade pr
2. upgrade secp256k1 on `https://github.com/nervosnetwork/rust-libp2p/blob/master/protocols/secio/Cargo.toml#L22`

I personally think that upgrading the version of libp2p is the normal solution.

If choose to reverted, there will need to upgrade again on replace `libp2p` to `p2p`. This is why I want to upgrade.

PR #268, opened 2019-02-12
fix: fork logic

zhangsoledad:
fix issue: `get_forks` may push genesis in old_blocks, cause tx_pool infinite loop

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -51,10 +51,32 @@ pub struct BlockInsertionResult {
     pub new_best_block: bool,
 }
 
-struct Fork {
-    new_blocks: Vec<Block>,
-    old_blocks: Vec<Block>,
-    open_exts: Vec<BlockExt>,
+#[derive(Debug, Default)]
+pub(crate) struct Fork {
+    pub(crate) new_blocks: Vec<Block>,
+    pub(crate) old_blocks: Vec<Block>,
+    pub(crate) open_exts: Vec<BlockExt>,
+}
+
+pub(crate) struct GlobalIndex {
+    pub(crate) number: BlockNumber,
+    pub(crate) hash: H256,
+    pub(crate) unseen: bool,
+}
+
+impl GlobalIndex {
+    pub(crate) fn new(number: BlockNumber, hash: H256, unseen: bool) -> GlobalIndex {
+        GlobalIndex {
+            number,
+            hash,
+            unseen,
+        }
+    }
+
+    pub(crate) fn forward(&mut self, hash: H256) {
+        self.number -= 1;
+        self.hash = hash;
+    }
 }
 
 pub struct ChainService<CI> {
@@ -119,7 +141,7 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
 
     // process_block will do block verify
     // but invoker should guarantee block header be verified
-    fn process_block(&mut self, block: Arc<Block>) -> Result<(), ProcessBlockError> {
+    pub(crate) fn process_block(&mut self, block: Arc<Block>) -> Result<(), ProcessBlockError> {
         debug!(target: "chain", "begin processing block: {}", block.header().hash());
         if self.verification {
             let block_verifier = BlockVerifier::new(self.shared.clone());
@@ -136,7 +158,7 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
     }
 
     #[allow(clippy::op_ref)]
-    fn insert_block(&self, block: &Block) -> Result<BlockInsertionResult, SharedError> {
+    pub(crate) fn insert_block(&self, block: &Block) -> Result<BlockInsertionResult, SharedError> {
         let mut new_best_block = false;
         let mut total_difficulty = U256::zero();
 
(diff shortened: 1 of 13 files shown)


chain/src/chain.rs
```diff
@@ -280,74 +289,86 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
+        let mut is_new_record = true;
 
-        // The old fork may longer than new fork
-        if number < current_tip_number {
-            for n in number..=current_tip_number {
-                let hash = self.shared.block_hash(n).unwrap();
+        // new_blocks = forks[latest_common + 1 .. new_tip]
+        // old_blocks = chain[latest_common + 1 .. old_tip]
+        new_blocks.push(new_tip_block.clone());
+        let mut index = new_tip_number - 1;
```
jjyr: I believe `common_parent_number` is more accurate for this variable and can help a lot for others who read this code.

chain/src/chain.rs
```diff
@@ -280,74 +289,86 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
-                if old_hash == hash {
-                    break;
-                }
-
-                let old_block = self.shared.block(&old_hash).unwrap();
-                old_blocks.push(old_block);
+            if index == 0 {
+                break;
+            }
+            let old_hash = self.shared.block_hash(index).unwrap();
```
jjyr: `common_parent_hash` or something like it is a more related name.

chain/src/chain.rs
jjyr: or common_block_number

chain/src/chain.rs
```diff
@@ -280,74 +289,86 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
 
-        // The old fork may longer than new fork
-        if number < current_tip_number {
-            for n in number..=current_tip_number {
-                let hash = self.shared.block_hash(n).unwrap();
+        // new_blocks = forks[latest_common + 1 .. new_tip]
+        // old_blocks = chain[latest_common + 1 .. old_tip]
+        new_blocks.push(new_tip_block.clone());
+        let mut index = new_tip_number - 1;
+        let mut index_hash = new_tip_block.header().parent_hash().clone();
```
jjyr: Use some name like `new_chain_index` to distinguish it with the `index` variable, those two value is confused.

chain/src/chain.rs
```diff
@@ -280,74 +289,86 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
...
 
-            if is_open {
-                let ext = self.shared.block_ext(&hash).unwrap();
+            let new_block = self.shared.block(&index_hash).unwrap();
+            index_hash = new_block.header().parent_hash().clone();
+            new_blocks.push(new_block);
+
+            if is_new_record {
+                let ext = self.shared.block_ext(&index_hash).unwrap();
                 if ext.valid.is_none() {
```
jjyr: I still can't get it, what is the intention here?

rpc/src/module/miner.rs
```diff
@@ -49,23 +51,35 @@ impl<CI: ChainIndex + 'static> MinerRpc for MinerRpcImpl<CI> {
...
+                Ok(Some(block.header().hash().clone()))
+            } else {
+                debug!(target: "rpc", "submit_block process_block {:?}", ret);
+                Ok(None)
+            }
         } else {
-            debug!(target: "rpc", "submit_block process_block {:?}", ret);
-            Err(Error::internal_error())
+            debug!(target: "rpc", "submit_block header verifier {:?}", header_verify_ret);
+            Ok(None)
```
jjyr: Why return an Option?

chain/src/chain.rs
zhangsoledad: It's NOT `common_parent_hash`

chain/src/chain.rs
zhangsoledad: np, `common_block_number` `common_parent_number` make no sense.
it's just an index of an enumerator.
check https://github.com/bitcoin/bitcoin/blob/v0.17.1/src/chain.cpp#L51. almost like this.

rpc/src/module/miner.rs
zhangsoledad: It's NOT an `internal_error`,  submit_block SHOULD provide share rejection reasons.
but this needs much further work, temporarily return an Option.

this patch fixes We haven't verified the block header

chain/src/chain.rs
jjyr: It is ok to use a variable like `index` in a short code snippet, but for here it confuses.

Moreover, we have two confused variables, at least we need a name to be enough to distinguish the two variables usage. 

I think `new_chain_index` and `block_number`, or other names definitely could be great than `index` and `index_hash`.

chain/src/chain.rs
zhangsoledad: It's an unseen mark, maybe name it unseen directly.

zhangsoledad: btw. It's a hotfix, not a refactory,

zhangsoledad:
cc @jjyr , have some updates.

jjyr:
The `chain` module should be refactoring in further.

PR #289, opened 2019-02-24
chore: code cleanup for db trait

quake:


benches/benches/process_block.rs
```diff
@@ -164,7 +164,7 @@ fn new_chain() -> (
     let shared = SharedBuilder::<CacheDB<RocksDB>>::default()
         .db(&DBConfig {
             path: db_dir.path().to_owned(),
-            rocksdb: None,
+            options: None,
```
doitian: In my opinion, `rocksdb` is better.

quake: I want to reuse the `DBConfig` in other backend db, e.g. LMDB: `SharedBuilder::<CacheDB<Lmdb>>::default()...`, so renamed it to `options`

PR #296, opened 2019-03-01
fix(miner): uncles in solo mining

zhangsoledad:
Always try to update block_template after submit_block.

NEW CONFIG: block_on_submit control whether miner block on submit_block.
BREAKING CHANGE: miner config poll_interval unit: second -> millisecond


janx:
nice

PR #297, opened 2019-03-01
Use toml

janx:
Use toml format for configurations as it's more friendly to devops/sysadmins than json.


README.md
```diff
@@ -153,13 +153,13 @@ Update `data_dir` configuration in config file to a different directory.
 or using a simple shell command to modify the config file:
 
 ```shell
-ex -sc '%s/"data_dir": "default"/"data_dir": "node2"/|x' nodes/node2.json
+ex -sc '%s/"data_dir": "default"/"data_dir": "node2"/|x' nodes/node2.toml
```
quake: The replace command should be changed also.

janx: Removed the `ex` as it's overshooting for a single line change.

PR #305, opened 2019-03-04
test: add basic integration test framework

quake:


quake:
> Is there a version of `jsonrpc-client-core` that depends on jsonrpc 10.0? If I remember correctly, the old version of jsonrpc has security vulnerabilities.

Checked cargo lock file, no such dependency

PR #310, opened 2019-03-06
fix: remove use of upgradable reads

zhangsoledad:
Upgradable reads block plain reads, this behavior does not match the docs


zhangsoledad:
https://github.com/Amanieu/parking_lot/issues/101

PR #318, opened 2019-03-11
docs: add 'version is always 0' [skip ci]

Mine77:
Add notice to explicitly inform that the version field is always 0, which is different from Bitcoin, where version is always 1.


doitian:
PR to master is forbidden

PR #320, opened 2019-03-12
docs: update version description and add hash of transaction [skip ci]

Mine77:
two updates:
* Add notice to explicitly inform that the version field is always 0, which is different from Bitcoin, where version is always 1.
* Add hash of transaction to the data structure.


zhangsoledad:
The `version` field is for doing parallel soft forking deployments. It's not always 0 or 1.

PR #335, opened 2019-03-15
feat: enforce `type` field of a cellbase output cell must be absent

zhangsoledad:
* remove redundant EmptyVerifier
* enforce `type` field of a cellbase output cell must be absent

### DIFF
--- a/verification/src/block_verifier.rs
+++ b/verification/src/block_verifier.rs
@@ -18,8 +18,6 @@ use std::collections::HashSet;
 //TODO: cellbase, witness
 #[derive(Clone)]
 pub struct BlockVerifier<P> {
-    // Verify if the committed transactions is empty
-    empty: EmptyVerifier,
     // Verify if the committed and proposed transactions contains duplicate
     duplicate: DuplicateVerifier,
     // Verify the cellbase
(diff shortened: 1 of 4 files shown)


verification/src/block_verifier.rs
```diff
@@ -99,6 +90,16 @@ impl<CP: ChainProvider + Clone> CellbaseVerifier<CP> {
         {
             return Err(Error::Cellbase(CellbaseError::InvalidInput));
         }
+
+        // currently, we enforce`type` field of a cellbase output cell must be absent
```
janx: why do we enforce this?

zhangsoledad: Currently, we have an involuntary issue that CKB-VM execution skipped `cellbase`. It's a temporary solution until we implement `cellbase` type script verifiction.

verification/src/block_verifier.rs
```diff
@@ -75,9 +69,6 @@ impl<CP: ChainProvider + Clone> CellbaseVerifier<CP> {
     }
 
     pub fn verify(&self, block: &Block) -> Result<(), Error> {
-        if block.commit_transactions().is_empty() {
```
doitian: Is cellbase required now? Has it been added in the verification RFC doc?

verification/src/block_verifier.rs
janx: Feels like something we should fix before testnet.

verification/src/block_verifier.rs
zhangsoledad: Cellbase is always required. RFC updated .

PR #336, opened 2019-03-15
feature: index whether a tx is a cellbase in the chain

u2:
Index whether a tx is a cellbase in the chain, prepare for the cellbase outputs maturity checking.

Now saving the cellbase index and block number in the `TransactionMeta`, there is another implementation which creates a `HashMap<tx_hash, number>`. The second one may be a little memory saving, but this one is more simple. I think both are ok.

https://github.com/nervosnetwork/ckb/issues/54

### DIFF
--- a/core/src/transaction_meta.rs
+++ b/core/src/transaction_meta.rs
@@ -16,15 +16,31 @@ impl From<BitVecSerde> for BitVec {
 
 #[derive(Default, Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct TransactionMeta {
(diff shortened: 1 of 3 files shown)


doitian:
@quake is refactoring the storage, we'll going to hold this PR.

quake:
LGTM, however, I would like to hold this PR after https://github.com/nervosnetwork/ckb/pull/343 merged, and I will rebase this PR later.

u2:
> LGTM, however, I would like to hold this PR after #343 merged, and I will rebase this PR later.

updated

PR #343, opened 2019-03-20
refactor: use CellProvider

quake:
This refactoring is intended to remove closure in ChainService and duplicate code in ChainState. And fix bugs in block processing and add some test cases.


shared/src/cell_set.rs
```diff
@@ -1,15 +1,49 @@
...
+use ckb_core::cell::{CellProvider, CellStatus};
 use ckb_core::transaction::OutPoint;
 use ckb_core::transaction_meta::TransactionMeta;
-use fnv::FnvHashMap;
+use fnv::{FnvHashMap, FnvHashSet};
 use numext_fixed_hash::H256;
 use serde_derive::{Deserialize, Serialize};
 
 #[derive(Default, Debug, Clone)]
 pub struct CellSetDiff {
```
u2: It's better to add some tests for `CellSetDiff`.

core/src/cell.rs
```diff
@@ -54,32 +56,81 @@ pub struct ResolvedTransaction {
...
-        _is_dead: F,
-    ) -> CellStatus {
-        unreachable!()
+impl<'a> BlockCellProvider<'a> {
+    pub fn new(block: &'a Block) -> Self {
+        let mut output_indexs = FnvHashMap::default();
+        for (i, tx) in block.commit_transactions().iter().enumerate() {
+            output_indexs.insert(tx.hash(), i);
+        }
+        Self {
```
zhangsoledad: It would be better if us  `with_capacity` to avoid reallocating. 
Moreover,  use `from_iter`

```
use std::iter::FromIterator;
FnvHashMap::from_iter(
   block.commit_transactions().iter().enumerate().map(|(idx, tx)| (tx.hash(), idx))
)
```

core/src/cell.rs
```diff
@@ -54,32 +56,81 @@ pub struct ResolvedTransaction {
...
 
-    fn cell_at<F: Fn(&OutPoint) -> Option<bool>>(
-        &self,
-        _out_point: &OutPoint,
-        _is_dead: F,
-    ) -> CellStatus {
-        unreachable!()
+impl<'a> BlockCellProvider<'a> {
+    pub fn new(block: &'a Block) -> Self {
+        let mut output_indexs = FnvHashMap::default();
```
zhangsoledad: typo indexs

shared/src/cell_set.rs
quake: added some basic tests, it's covered in https://github.com/nervosnetwork/ckb/pull/343/commits/22355a19b0897514610839c118b1c10812e4b03b

core/src/cell.rs
quake: Resolved, thanks

core/src/cell.rs
quake: Resolved, thanks

quake:
bors try

quake:
bors try

quake:
bors try

chain/src/chain.rs
```diff
@@ -547,21 +505,13 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
                 .insert_block_ext(batch, &b.header().hash(), ext);
         }
 
-        if found_error {
-            Err(SharedError::InvalidTransaction)?;
+        if found_error.is_some() {
+            Err(SharedError::InvalidTransaction(
+                found_error.unwrap().to_string(),
+            ))?;
         }
```
zhangsoledad: can be written more concisely as
```rust 
        if let Some(err) = found_error {
            return Err(SharedError::InvalidTransaction(err.to_string()));
        }
```

shared/src/cell_set.rs
```diff
@@ -52,28 +87,36 @@ impl CellSet {
...
+        });
 
-        for o in inputs {
-            self.mark_dead(&o);
-        }
+        diff.new_inputs.iter().for_each(|o| {
+            self.mark_dead(o);
+        });
     }
+}
```
zhangsoledad: `fn insert` should keep `move semantics`
 pub fn insert(&mut self, hash: H256, outputs_len: usize) 

leverage destructuring assignment in `fn update` keeps logic same as original code, avoid unnecessary clone.
```rust 
        let CellSetDiff {
            old_inputs,
            old_outputs,
            new_inputs,
            new_outputs,
        } = diff;

        old_outputs.iter().for_each(|h| {
            self.remove(h);
        });

        old_inputs.iter().for_each(|o| {
            self.mark_live(o);
        });

        new_outputs.into_iter().for_each(|(hash, len)| {
            self.insert(hash, len);
        });

        new_inputs.iter().for_each(|o| {
            self.mark_dead(o);
        });
```

chain/src/tests/util.rs
```diff
@@ -36,43 +40,64 @@ pub(crate) fn start_chain(
...
-    let header = HeaderBuilder::default()
+    let header_builder = HeaderBuilder::default()
         .parent_hash(parent_header.hash().clone())
         .timestamp(unix_time_as_millis())
         .number(number)
         .difficulty(difficulty)
-        .nonce(nonce)
-        .build();
+        .cellbase_id(cellbase.hash())
+        .nonce(rand::random());
```
zhangsoledad: I oppose introducing randomness, may cause some involuntary bug.

quake: After review original nonce value usage here, I think it can be removed completely, pushed a new commit, please review again, thanks.

PR #344, opened 2019-03-20
refactor: Revise script structure

xxuejie:
* [x] rfc changes: https://github.com/nervosnetwork/rfcs/pull/91
* [x] ckb-system-scripts: https://github.com/nervosnetwork/ckb-system-scripts/pull/2
* [x] mruby-contracts: https://github.com/nervosnetwork/mruby-contracts/pull/7
* [x] ruby sdk: https://github.com/nervosnetwork/ckb-demo-ruby/pull/76
* [x] ruby scripts: https://github.com/nervosnetwork/ckb-ruby-scripts/pull/9
* [ ] swift sdk
* [ ] java sdk
* [ ] js sdk

---

This is a **BREAKING CHANGE**.


core/src/script.rs
```diff
@@ -13,54 +13,11 @@ use std::mem;
...
-        match self.0 {
-            Some(ref v) => write!(f, "Some({})", v),
-            None => write!(f, "None"),
-        }
-    }
+    // The reference hash here can be used to refer to binary in any of the
+    // following locations:
+    // 1. Data part of a dep cell in current transaction
+    // 2. An embed item in current transaction
+    pub reference: H256,
```
Mine77: Could you please elaborate what exact hash should be used here to refer a dep cell or an embed item in tx?

Is it the hash of the `data` part of a cell?

xxuejie: Let me update the comment to be more clear

core/src/script.rs
```diff
@@ -13,54 +13,13 @@ use std::mem;
...
-        match self.0 {
-            Some(ref v) => write!(f, "Some({})", v),
-            None => write!(f, "None"),
-        }
-    }
+    // The reference hash here can be used to refer to binary in any of the
+    // following locations:
+    // 1. Data part of a dep cell in current transaction, reference hash
+    // must be the hash of the cell data so as to reference this cell
+    // 2. An embed item in current transaction, reference hash must be the
```
u2: So the embed item is a one-time data?

xxuejie: Yep, it's like the original binary but at a different place with more potential use cases

verification/src/transaction_verifier.rs
```diff
@@ -70,18 +70,8 @@ impl<'a> InputVerifier<'a> {
...
-                if let Some(ref input_cell) = cs.get_live() {
-                    // TODO: remove this once VM mmap is in place so we can
-                    // do P2SH within the VM.
-                    if input_cell.lock != input.unlock.type_hash() {
-                        return Err(TransactionError::InvalidScript);
-                    }
-                }
-            } else if cs.is_dead() {
+        for cs in &self.resolved_transaction.input_cells {
+            if cs.is_dead() {
```
u2: :+1:

u2:
Add some description for this PR. The linked issue is private.

core/src/script.rs
doitian: Is `binary_hash` a better name?

script/src/syscalls/mod.rs
```diff
@@ -29,6 +29,7 @@ enum CellField {
     Capacity = 0,
     Data = 1,
     DataHash = 2,
+    Lock = 6,
```
jjyr: Move this item after `TypeHash = 5` for consistency.

xxuejie: There are 2 ways to understand consistency here: first, we can sort them by the value of the enums, of course. But there's also another way which we can group the values based on the meeting, so we can do Data, DataHash, Lock, LockHash, this way it's easier to see what options are available here. Thoughts?

doitian:
@xxuejie this PR conflicts with the #350

xxuejie:
@doitian Conflict is resolved now

ashchan:
SDKs haven't caught up yet, but I guess we can plan this for `v0.8.0` which is RC frozen today (meaning develop during the RC period). How about that? @doitian @xxuejie

Mine77:
are you considering updating the [how-to-write-contract doc](https://github.com/nervosnetwork/ckb-demo-ruby/blob/develop/docs/how-to-write-contracts.md) as well?

doitian:
Integration test stuck, cc @quake

xxuejie:
bors try

PR #345, opened 2019-03-20
feat: Add random_peers function to PeerStore

jjyr:


network/src/peer_store/db.rs
```diff
@@ -186,6 +186,27 @@ impl PeerAddr {
...
+
+    let mut peers = Vec::with_capacity(count as usize);
+    for row in rows {
+        let (id, peer_id) = row?;
+        let mut addrs = PeerAddr::get_addrs(conn, id, 1)?;
+        if let Some(addr) = addrs.pop() {
+            peers.push((peer_id, addr));
+        }
+    }
+    Ok(peers)
```
TheWaWaR: Will get different peers to the same address. The address is discovered from different peers.

TheWaWaR: Will get different peers to the same address. The address is discovered from different peers. Should get just different addresses no matter the address from which peer.

PR #356, opened 2019-03-25
Unify network peer scoring

jjyr:


network/src/behaviour.rs
```diff
@@ -0,0 +1,21 @@
...
+// Define behaviours and scores
+define_behaviour! {
+    CONNECT => 10,
+    PING => 10,
+    FAILED_TO_PING => -20,
+    TIMEOUT => -20,
+    SYNC_USELESS => -50,
+    UNEXPECTED_MESSAGE => -50,
+    UNEXPECTED_DISCONNECT => -10
+}
```
TheWaWaR: https://doc.rust-lang.org/stable/book/ch06-02-match.html#the-match-control-flow-operator

A match is enough for this.

jjyr: They're not same, behaviors have the same structure and may have the same score, an enum can't represent this directly, it is more like a k-v pair than an enum, however k-v pair lack of static check.

quake: I prefer enum:
```rust
enum Behaviour {
    Connect,
    Ping,
}

impl Behaviour {
    fn score(&self) -> Score {
        match self {
            Behaviour::Connect => 10,
            Behaviour::Ping => 10,
        }
    }
}
```

jjyr:
bor r+

PR #360, opened 2019-03-26
Feeler connection

jjyr:
Use a temporary workaround to registry feeler connection protocol, should refactor after removing CKBProtocolHandler.


jjyr:
bors try

network/src/network.rs
```diff
@@ -315,39 +297,59 @@ impl Network {
...
+            protocols.extend(
+                // TODO remove the filter after refactor
+                self.ckb_protocols
+                    .iter()
+                    .map(|p| p.0.id())
+                    .filter(|id| *id != FEELER_PROTOCOL_ID)
+                    .collect::<Vec<_>>(),
+            );
+            DialProtocol::Multi(protocols)
+        })
```
TheWaWaR: Have you tested in your machine? This dial will not work out. We have more protocols than those 4.

jjyr: This code dial more than 4 protocols on my machine.

jjyr:
bors try-
bors try

doitian:
It seems the node stuck in the integration test.

network/src/network.rs
```diff
@@ -315,39 +297,59 @@ impl Network {
...
+        let p2p_phase =
+            Multihash::from_bytes(expected_peer_id.as_bytes().to_vec()).map_err(|_| {
+                Error::P2P(format!("peer_id is invalid {:?}", expected_peer_id).to_string())
+            })?;
+        addr.append(multiaddr::Protocol::P2p(p2p_phase));
+        self.p2p_control
+            .write()
+            .dial(addr, dial_protocol)
+            .map_err(|_| Error::P2P("dial error".to_string()))
+    }
```
TheWaWaR: Assume I got a peer address, and it's not include a `/p2p` section, than I can't dial it?

network/src/network.rs
```diff
@@ -315,39 +297,59 @@ impl Network {
...
+            DialProtocol::Multi(protocols)
+        })
+    }
+
+    pub fn dial_feeler(&self, expected_peer_id: &PeerId, addr: Multiaddr) -> Result<(), Error> {
+        self.inner_dial(
+            expected_peer_id,
+            addr,
+            DialProtocol::Single(FEELER_PROTOCOL_ID),
+        )
```
TheWaWaR: Where is the code use this function?

network/src/network.rs
TheWaWaR: And very likely one will update his private key, than `/p2p` section is invalid, how to deal with it?

jjyr: CKB network security and scoring base on `peer_id` and `address`(RFC0007), we must enforce connection encrypt.

jjyr: If a peer updated with a new key, In the current implementation, the address(without p2p phase) will consider as invalid, until node rediscovery the new address(with new p2p phase).

network/src/network.rs
jjyr: For feeler connection, see `protocol/feeler`

network/src/network.rs
TheWaWaR: `connection encrypt` is not require peer_id, you just dial, `secio` will know target peer's peer_id.

TheWaWaR: And, `/p2p` section required in `bootnodes` is not user friendly.

network/src/network.rs
TheWaWaR: You can try `Ctrl + F` on this page, than type: `dial_feeler`

network/src/network.rs
jjyr: Of cause we need `/p2p/` phase to make sure bootnodes' ips not controlled be bad crackers. Pls see RFC0007 first.

network/src/network.rs
jjyr: My mistake, fixed.

network/src/network.rs
TheWaWaR: Yes, `bootnode` must require peer_id, but RFC0007 is not show how peer_id is contribute to the security.

As a normal user, he may change his private key from time to time, but he always use the same address. I think it's a OK behavior, but will he be banned?

jjyr: He **will not be banned**.

The reason we need to deny the old address is as same as the `bootnode` security, a user with good behaviors may earn more scores in his peers' side, once the private key changed, we can't prove the peer's IP is controlled by the same good people or a bad cracker, we must reset the trust(score) to this address.

A normal user may change his key once or twice, but not every time he start the node.

By the way, even the user do it, node outbound connections work as well; only inbound peers will not trust the old score, these peers consider the node as a new node.

There's nothing about * user-friendly*; there's a high probability the user can't feel different after reset key.

jjyr:
bors try

PR #361, opened 2019-03-26
BREAKCHANGE: refactor network config

jjyr:


jjyr:
bors try

jjyr:
bors try

doitian:
It seems the node stuck in the integration test.

jjyr:
bors try

network/src/network_config.rs
```diff
@@ -1,93 +1,102 @@
...
         }
     }
+}
 
-    pub fn new() -> Self {
-        Self::default()
+impl NetworkConfig {
+    pub fn secret_key_path(&self) -> PathBuf {
+        let mut path = self.path.clone();
+        path.set_file_name("secret_key");
```
quake: I think that we should use `push` here, `set_file_name` will replace the folder, please check:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=0bb6b29be629e34476bce4d091c6669d

test/fixtures/nodes_template/default.toml
```diff
@@ -12,15 +12,17 @@ filter = "info,network=trace,rpc=debug,sync=debug,relay=debug"
 color = true
 
 [network]
+path = "default/network"
 listen_addresses = ["/ip4/0.0.0.0/tcp/P2P_PORT"]
+public_addresses = []
```
quake: A question may not be relevant to this PR: why need `public_adresses` option in the network config?

jjyr: In a scenario, a node is behind some LB or gateway, this option allows the user to put the LB address.

network/src/network_config.rs
jjyr: fixed

jjyr: fixed

test/fixtures/nodes_template/default.toml
quake: The public address list should be updated by identify protocol through remote peer connection like bitcoin. This way, we don’t need to worry about LB or gateway.

jjyr: It is different, identity protocol can find the address if peer's outbound IP is the same as LB address in this case, but it is high depends on the network setup.
Otherwise, identity can't find LB address.

PR #363, opened 2019-03-27
remove uncle cellbase

zhangsoledad:
- [ ] update RFC#0019


docs/data-structures.md
```diff
@@ -1,6 +1,6 @@
 # Data Structures of Cell/Script/Transaction/Block
```
doitian: Pick the changes to RFC. We'll remove this file.

miner/src/block_assembler.rs
```diff
@@ -489,18 +484,17 @@ mod tests {
             // uncles_count_limit,
         } = block_template;
 
-        let (cellbase_id, cellbase) = {
-            let CellbaseTemplate { hash, data, .. } = cellbase;
-            (hash, data)
+        let cellbase = {
+            let CellbaseTemplate { data, .. } = cellbase;
```
quake: Do we still need `CellbaseTemplate` after removing cellbase id from header? Can we add it to `Vec<TransactionTemplate>` ?

zhangsoledad:
```
pub struct TransactionTemplate {
    pub hash: H256,
    pub required: bool,
    pub cycles: Option<Cycle>,
    pub depends: Option<Vec<u32>>,
    pub data: Transaction, // temporary
}
```
`cellbase` has no depends, and it's required no doubt, so perfer keep it.

docs/data-structures.md
doitian: Posted here https://github.com/nervosnetwork/rfcs/pull/91#pullrequestreview-223619365

doitian:
bors retry

doitian:
Need to fix the merged generated flatbuffers file.

doitian:
> Need to fix the merged generated flatbuffers file.

Regenerate verifier: https://github.com/nervosnetwork/ckb/pull/409

PR #364, opened 2019-03-27
Verify flatbuffer messages

### LINKED ISSUE #328: Generate code to verify flatbuffers

doitian:
The Rust code generated by `flatc` does not verify the buffer, it panics on invalid data because of out of bands memory access.

This PR integrates the verifier generated by [cfbc](https://github.com/nervosnetwork/cfb) to fix this issue. Peers sending malformed messages are marked as `Behaviour::UnexpectedMessage`.

Closes #328


sync/src/relayer/mod.rs
```diff
@@ -302,7 +302,15 @@ where
 
     fn received(&self, nc: Box<CKBProtocolContext>, peer: PeerIndex, data: Bytes) {
         // TODO use flatbuffers verifier
```
quake: This comment can be removed now.

zhangsoledad:
```
get_root::<Block>(&[0x0,0x0,0x0,0x0,0xa])
```
`panicked at 'attempt to subtract with overflow'`

https://github.com/google/flatbuffers/blob/master/rust/flatbuffers/src/vtable.rs#L43

doitian:
> ```
> get_root::<Block>(&[0x0,0x0,0x0,0x0,0xa])
> ```
> 
> `panicked at 'attempt to subtract with overflow'`
> 
> 
> https://github.com/google/flatbuffers/blob/master/rust/flatbuffers/src/vtable.rs#L43

Resolved

PR #367, opened 2019-03-28
🐞✋ feat: add cellbase maturity verification

u2:
Draft
https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0010-cellbase-maturity-period/0010-cellbase-maturity-period.md

### DIFF
--- a/verification/src/transaction_verifier.rs
+++ b/verification/src/transaction_verifier.rs
@@ -1,12 +1,15 @@
 use crate::error::TransactionError;
 use ckb_core::transaction::{Capacity, Transaction, TX_VERSION};
+use ckb_core::transaction_meta::TransactionMeta;
 use ckb_core::{
     cell::{CellMeta, CellStatus, ResolvedTransaction},
     BlockNumber, Cycle,
 };
 use ckb_script::TransactionScriptsVerifier;
 use ckb_traits::BlockMedianTimeContext;
+use fnv::FnvHashMap;
 use lru_cache::LruCache;
+use numext_fixed_hash::H256;
 use occupied_capacity::OccupiedCapacity;
 use std::cell::RefCell;
 use std::collections::HashSet;
@@ -15,6 +18,7 @@ pub struct TransactionVerifier<'a, M> {
     pub version: VersionVerifier<'a>,
     pub null: NullVerifier<'a>,
     pub empty: EmptyVerifier<'a>,
+    pub maturity: MaturityVerifier<'a>,
     pub capacity: CapacityVerifier<'a>,
     pub duplicate_inputs: DuplicateInputsVerifier<'a>,
     pub inputs: InputVerifier<'a>,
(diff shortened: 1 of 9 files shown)


quake:
@u2 I proposal a refactoring: move `max_cycles` to `ChainState` struct fields, which will simplify `add_tx_to_pool / verify_rtx / etc` functions' parameters list. I think that `cellbase_maturity` can be handled same way, what's your thoughts on this?

u2:
> @u2 I proposal a refactoring: move `max_cycles` to `ChainState` struct fields, which will simplify `add_tx_to_pool / verify_rtx / etc` functions' parameters list. I think that `cellbase_maturity` can be handled same way, what's your thoughts on this?

Agree.

u2:
@quake updated

quake:
Please hold until this #372 is merged, after that, we can refactor `verify_rtx` fn to:
```rust
chain_state.verify_rtx(&rtx);
```

chain/src/tests/basic.rs
```diff
@@ -73,7 +73,8 @@ fn test_genesis_transaction_spend() {
 
 #[test]
 fn test_transaction_spend_in_same_block() {
-    let (chain_controller, shared) = start_chain(None, true);
+    let consensus = Consensus::default().set_cellbase_maturity(0);
```
quake: I perfer setting default test `cellbase_maturity` value in `start_chain`  helper.

u2: Updated.

verification/src/transaction_verifier.rs
```diff
@@ -124,6 +137,41 @@ impl<'a> EmptyVerifier<'a> {
...
+    pub fn verify(
+        &self,
+        block_number: BlockNumber,
+        cell_set: &FnvHashMap<H256, TransactionMeta>,
+        cellbase_maturity: usize,
+    ) -> Result<(), TransactionError> {
+        let immature_spend = self.transaction.inputs().iter().any(|input| {
+            match cell_set.get(&input.previous_output.hash) {
+                Some(ref meta)
+                    if meta.is_cellbase()
```
jjyr: Not every cell outpoint in cellbase need to be maturity, I think we need more discussion on this?

Consider the UDT tx fee design, maybe we should leave more rooms for upper UDT contracts?

u2: > Not every cell outpoint in cellbase need to be maturity, I think we need more discussion on this?
> 
> Consider the UDT tx fee design, maybe we should leave more rooms for upper UDT contracts?

I think it should be because of the block confirmation. The UDT tx fee would be included in the cellbase?

jjyr: Not decide yet.

jjyr: Seem your implement is correct due to the explanation of maturity in this link https://bitcoin.stackexchange.com/questions/1991/what-is-the-block-maturation-time

Anyway please hold this PR until #372 merged, 372 introduce few refactoring about tx verification.

doitian:
@jjyr is going to take a look at this PR.

@quake Do you prefer to merge it after your store refactoring, or it doesn't matter.

quake:
it doesn't matter.

u2:
Updated, 
https://github.com/nervosnetwork/ckb/pull/367/files#diff-b2197becbc6452c42618e284e3dbf52eR226
But I am not sure whether the cellbase_maturity and cell_set are correct, because the `max_cycles` coming from the function params, and the `cellbase_maturity` coming from the `self.consensus`.

doitian:
I'm sorry but it seems that @github is not going to fix the 500 issue. The team could not open this PR in desktop browser, it is inconvenient to discuss in this PR.  Could you please close this and create a new PR instead?

PR #368, opened 2019-03-28
feat: segregated witness

zhangsoledad:
* [ ] rfc#0019
* [ ] mruby-contracts https://github.com/nervosnetwork/mruby-contracts/pull/9
* [ ] ruby sdk
* [ ] ruby scripts https://github.com/nervosnetwork/ckb-ruby-scripts/pull/10
* [ ] ckb-system-scripts https://github.com/nervosnetwork/ckb-system-scripts/pull/5
* [ ] ruby-demo https://github.com/nervosnetwork/ckb-demo-ruby/pull/78
* [ ] swift sdk
* [ ] java sdk
* [ ] js sdk

---

This is a **BREAKING CHANGE**.


doitian:
Is this PR going to be merged now, or after all the SDKs have been upgraded?

core/src/transaction.rs
```diff
@@ -141,91 +141,47 @@ impl CellOutput {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Debug, Default, OccupiedCapacity)]
+pub struct Witness(pub u32, pub Vec<Vec<u8>>);
+
+impl Witness {
+    // Index of relevant input
+    pub fn index(&self) -> u32 {
```
doitian: Witness is not matched by position?

zhangsoledad: >A non-witness program (defined hereinafter) txin MUST be associated with an empty witness field, represented by a 0x00

Some scripts don't need to use or just don't want to use witness.
Bitcoin enforce every txin associate with one witness(could by empty), and fixed the witness program format，such as

>* If the version byte is 0, and the witness program is 20 bytes:
The witness must consist of exactly 2 items (≤ 520 bytes each). The first one a signature, and the second one a public key.

It is not very suitable to port this rules.

zhangsoledad:
> Is this PR going to be merged now, or after all the SDKs have been upgraded?

Either is Ok.

ashchan:
> > Is this PR going to be merged now, or after all the SDKs have been upgraded?
> 
> Either is Ok.

SDKs are catching up (planed to be include in `v0.9.0` as the previous Script refactoring) but I don't think they should block core. Personally I think we should merge this and allow toolchain team a few days to implement the SDKs.

core/src/block.rs
```diff
@@ -142,6 +142,17 @@ impl BlockBuilder {
                 .collect::<Vec<_>>(),
         );
 
+        // The witness hash of cellbase transaction is assumed to be zero 0x0000....0000
+        let mut witnesses = vec![H256::zero()];
```
quake: We may refactor these code to a function of `Block`, e.g. `commit_transactions_ witnesses_root`, it will eliminate duplicate code in builder and verifier, especially hardcoded zero hash for cellbase. And previous code also can be refactored to `commit_transactions_root`.

doitian:
Trace SDK and RFC changes in a new issue.

* [ ] rfc#0019
* [ ] mruby-contracts https://github.com/nervosnetwork/mruby-contracts/pull/9
* [ ] ruby sdk
* [ ] ruby scripts https://github.com/nervosnetwork/ckb-ruby-scripts/pull/10
* [ ] ckb-system-scripts https://github.com/nervosnetwork/ckb-system-scripts/pull/5
* [ ] ruby-demo https://github.com/nervosnetwork/ckb-demo-ruby/pull/78
* [ ] swift sdk
* [ ] java sdk
* [ ] js sdk

zhangsoledad:
need regenerate flatbuffers verifier,
wait https://github.com/nervosnetwork/ckb/pull/409

PR #369, opened 2019-03-28
feat: Embed testnet chain spec in compiled binary

xxuejie:


spec/src/lib.rs
```diff
@@ -16,13 +19,99 @@ use numext_fixed_hash::H256;
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum SpecPath {
+    Testnet,
+    Local(PathBuf),
+}
+
+struct SpecPathVisitor;
+
+impl<'de> serde::de::Visitor<'de> for SpecPathVisitor {
```
quake: How about using toml's enum deserializing directly?

```toml
spec = "Testnet"
spec = { Local = "spec/dev.toml" }
```

spec/src/lib.rs
```diff
@@ -81,18 +169,20 @@ pub(self) fn build_system_cell_transaction(
...
+
         Ok(spec)
     }
 
     pub fn pow_engine(&self) -> Arc<dyn PowEngine> {
         self.pow.engine()
     }
 
-    pub fn to_consensus(&self) -> Result<Consensus, Box<Error>> {
+    pub fn to_consensus(&self, spec_path: &SpecPath) -> Result<Consensus, Box<Error>> {
```
quake: Refactoring to the builder pattern may simplify the code of `Consensus` creation.

spec/chainspecs/testnet/testnet.toml
```diff
@@ -0,0 +1,36 @@
...
+edge_bits = 15
+
+# length of the cycle to be found, must be an even number, a minimum of 12 is
+# recommended
+cycle_length = 12
+
+# An array list paths to system cell files, which is absolute or relative to
+# the directory containing this config file.
+[[system_cells]]
+path = "cells/secp256k1_blake2b_lock"
```
doitian: Does it support path like `../cells`. We may want to share the identical cells between specs.

xxuejie: The crate we use doesn't support it, since internally what it does is just a key lookup process in a map. However, I believe we can [canonicalize](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) in our code to solve this problem.

xxuejie: The crate we use here doesn't support this, since all it does is a hashmap lookup. However, I think we can [clean](https://github.com/danreeves/path-clean) the path outside of the crate to support this behavior.

spec/src/lib.rs
xxuejie: I actually tried that path, but we are talking about 2 parameters which are both required to build `Consensus`, I'm not sure using a builder pattern would simplify this code. It might end up just with more code and no obvious gain.

Tho I can refactor this to a separate function instead of method on ChainSpec.

quake: OK, I can try it later (someday... :)

PR #372, opened 2019-03-29
feat: tx valid since

jjyr:


shared/src/shared.rs
```diff
@@ -81,13 +82,14 @@ impl<CI: ChainIndex> Shared<CI> {
                 proposal_ids,
                 TxPool::new(tx_pool_config),
                 LruCache::new(txs_verify_cache_size),
+                Arc::clone(&consensus),
```
quake: I think that we can refactor ChainState construct function now:
```rust
ChainState::new(&store, tx_pool_config, Arc::clone(&consensus))
```

and move all init code to construct function.

verification/src/transaction_verifier.rs
```diff
@@ -25,6 +37,7 @@ impl<'a> TransactionVerifier<'a> {
             script: ScriptVerifier::new(rtx),
             capacity: CapacityVerifier::new(rtx),
             inputs: InputVerifier::new(rtx),
+            valid_since: ValidSinceVerifier::new(&rtx.transaction, &block_context),
```
u2: Call `self.valid_since.verify()?;` in `verify`.

https://github.com/nervosnetwork/ckb/pull/372/files#diff-d0ea0a76c2e66e05fbcf6f957b6eccf4R44

jjyr:
bors try

jjyr:
RFC update https://github.com/nervosnetwork/rfcs/pull/97

shared/src/chain_state.rs
```diff
@@ -39,14 +39,35 @@ impl<CI: ChainIndex> ChainState<CI> {
...
     pub fn new(
         store: &Arc<CI>,
-        tip_header: Header,
-        total_difficulty: U256,
-        cell_set: CellSet,
-        proposal_ids: TxProposalTable,
-        tx_pool: TxPool,
-        txs_verify_cache: LruCache<H256, Cycle>,
         consensus: Arc<Consensus>,
+        txs_verify_cache_size: usize,
```
quake: `txs_verify_cache_size` is one field of `TxPoolConfig`, we can omit it here.

shared/src/chain_state.rs
```diff
@@ -39,14 +39,35 @@ impl<CI: ChainIndex> ChainState<CI> {
     #[allow(clippy::too_many_arguments)]
```
quake: Please remove this clippy hint

shared/src/chain_state.rs
jjyr: Logically `txs_verify_cache_size` is not related with `TxPoolConfig`, you can see we use `txs_verify_cache` in `chain` or other modules, it is more like a shared object.

I'd rather leave it here for now.

quake: Please check this line:
https://github.com/nervosnetwork/ckb/blob/61855c678d7806e5dc733e3806757d2f47a0a47b/shared/src/shared.rs#L305

`txs_verify_cache_size` is copied from `TxPoolConfig`

jjyr: Fair enough

chain/src/chain.rs
```diff
@@ -191,10 +225,10 @@ impl<CI: ChainIndex + 'static> ChainService<CI> {
         let current_total_difficulty = chain_state.total_difficulty().clone();
 
         debug!(
-            target: "chain",
-            "difficulty current = {}, cannon = {}",
-            current_total_difficulty,
-            cannon_total_difficulty,
+        target: "chain",
```
zhangsoledad: meaningless indent

zhangsoledad:
The transaction with a future `valid_since`: 
* RPC:  be discarded directly
* Relay:  staging failed and then be discarded.

Any plan for this?

jjyr:
> The transaction with a future `valid_since`:
> 
> * RPC:  be discarded directly
> * Relay:  staging failed and then be discarded.
> 
> Any plan for this?

We just reject to send immature tx in RPC, which is bitcoin currently behavior.

core/src/transaction.rs
```diff
@@ -63,32 +63,36 @@ impl OutPoint {
...
+    pub fn new(previous_output: OutPoint, valid_since: u64, args: Vec<Vec<u8>>) -> Self {
         CellInput {
             previous_output,
+            valid_since,
             args,
         }
     }
 
-    pub fn new_cellbase_input(block_number: BlockNumber) -> Self {
+    pub fn new_cellbase_input(block_number: BlockNumber, valid_since: u64) -> Self {
```
quake: I think that `valid_since` of Cellbase should always be 0, no need to pass the parameter here.

jjyr: Fixed

verification/src/transaction_verifier.rs
```diff
@@ -209,3 +225,161 @@ impl<'a> CapacityVerifier<'a> {
...
     }
 }
+
+const LOCK_TYPE_FLAG: u64 = 1 << 63;
+const TIME_TYPE_FLAG: u64 = 1 << 62;
+const VALUE_MUSK: u64 = 0x00ff_ffff_ffff_ffff;
+
+/// RFC 0017
+#[derive(Copy, Clone, Debug)]
+struct ValidSince(u64);
```
quake: I prefer enum to struct:
```rust
enum ValidSince {
  None,
  AbsoluteNumber(u64),
  AbsoluteTimestamp(u64),
  RelativeNumber(u64),
  RelativeTimestamp(u64),
}
```
And impl From/To u64 trait, that will remove these helper methods and simplify code.

jjyr: tomayto tomato, I can't see big improvement, it just change the conditions check from function to pattern match,

quake: It's a common refactoring practice in other OOP language, it is called `Replace Conditional with Polymorphism`.

jjyr: Can't agree, If we support more bits flags, the enum will increment exponentially, there's no clear improvement.

jjyr: Do not have a conclusion, just use the current implementation for now.

doitian:
https://travis-ci.com/nervosnetwork/ckb/builds/107753855

protocol/src/protocol_generated_verifier.rs
```diff
@@ -52,19 +52,14 @@ pub mod ckb {
 
                 let vtab = tab.vtable();
                 let vtab_num_bytes = vtab.num_bytes();
-                let object_inline_num_bytes = vtab.object_inline_num_bytes();
```
doitian: This is generated by an old version of cfbc, please install 0.19 and regenerate it

PR #376, opened 2019-03-31
fix: Repeat dial failed addresses

TheWaWaR:


network/src/network.rs
```diff
@@ -361,14 +378,36 @@ impl ServiceHandle for EventHandler {
...
-    fn handle_event(&mut self, _context: &mut ServiceContext, event: ServiceEvent) {
+    fn handle_event(&mut self, context: &mut ServiceContext, event: ServiceEvent) {
         info!(target: "network", "p2p service event: {:?}", event);
+        // When session disconnect update status anyway
+        if let ServiceEvent::SessionClose { session_context } = event {
+            let peer_id = session_context
+                .remote_pubkey
+                .as_ref()
+                .map(|pubkey| pubkey.peer_id())
+                .expect("Secio must enabled");
```
jjyr: If peer disables sec-io, will this work or panic?

network/src/network.rs
```diff
@@ -361,14 +378,36 @@ impl ServiceHandle for EventHandler {
...
+
+            let mut peer_store = self.network_state.peer_store().write();
+            if peer_store.peer_status(&peer_id) != Status::Disconnected {
+                peer_store.report(&peer_id, Behaviour::UnexpectedDisconnect);
+                peer_store.update_status(&peer_id, Status::Disconnected);
```
jjyr: Why move this from Disconnect event?

PR #387, opened 2019-04-02
feat: bundle app config

doitian:
#### Proposal: [[PREVIEW]](https://github.com/doitian/ckb/blob/config-rework/docs/configure.md)

- Add `ckb_resource::ResourceLocator` to load config files, which fallbacks to the bundled files.
- Add a new subcommand `ckb init` to dump the bundled config files.
- The bundled config files support template to set the different options for different specs in the same file.
- Add many checks before starting the app.
- Refactor the code organization in `src`.
- Options `logger.file`, `db.path` and `network.path` are set automatically from the data dir.

**Breaking Changes**:

- Remove command line argument `-c`, and add command line argument `-C`.
- No longer search the config file `nodes/default.toml`.
- Remove `logger.file`, `db.path` and `network.path` from config file. Add config option `logger.copy_to_stdout`.

**Future Enhancements**:

- Better exit message.
- Verify genesis hash when starting the app.
- Verify database version when starting the app.

---
**TODO**

- [x] Add tests to load bundled config files.
- [x] Add tests to load the config files generated by `ckb init`.
- [x] `ckb init` should quit if any file already exists unless `--force` is specified.
- [x] `ckb init` should write file to temp file first, and move to the target location to avoid generating partial file because of interrupts.


doitian:
Rebased onto rc/v0.9.0

https://github.com/nervosnetwork/ckb/pull/425

PR #414, opened 2019-04-08
fix: clear tx verfy cache when chain reorg

zhangsoledad:


doitian:
Need to pick to v0.9.0

doitian:
> Need to pick to v0.9.0

Done

PR #415, opened 2019-04-08
chore: regenerate flatbuffers verifier

zhangsoledad:


protocol/src/protocol_generated_verifier.rs
```diff
@@ -52,19 +52,14 @@ pub mod ckb {
 
                 let vtab = tab.vtable();
                 let vtab_num_bytes = vtab.num_bytes();
-                let object_inline_num_bytes = vtab.object_inline_num_bytes();
```
doitian: This is generated by an old version of cfbc, see https://github.com/nervosnetwork/ckb/pull/386

PR #418, opened 2019-04-09
fix: try to repair a corrupted rocksdb automatically

yangby-cryptape:
I have test this patch with a corrupted database (from @TheWaWaR).

But I can not reproduce the bug by myself, so I did not write any test case.

The database works, but the data is still corrupted (**because `header` add new fields?**):

```
2019-04-09 07:39:24.154 +00:00 main WARN ckb_db::diskdb  Try repairing the rocksdb since Corruption: SST file is ahead of WALs ...
2019-04-09 07:39:24.384 +00:00 main WARN ckb_db::diskdb  Try opening the repaired rocksdb ...
thread 'main' panicked at 'header deserializing should be ok: Custom("invalid length 3, expected a 0x-prefixed hex string with 64 digits")', src/libcore/result.rs:997:5
stack backtrace:
```

Ref:
- [RocksDB: Repairer](https://github.com/facebook/rocksdb/wiki/RocksDB-Repairer)
- [RocksDB: WAL Recovery Modes](https://github.com/facebook/rocksdb/wiki/WAL-Recovery-Modes)
- [`pub fn set_wal_recovery_mode(&mut self, mode: DBRecoveryMode)`](https://github.com/rust-rocksdb/rust-rocksdb/blob/267d92cbf9fdcbce9052ef06d7cb90b3111c6c54/src/db_options.rs#L1042-L1046)

### DIFF
--- a/db/src/diskdb.rs
+++ b/db/src/diskdb.rs
@@ -1,6 +1,7 @@
 use crate::batch::{Batch, Col, Operation};
 use crate::config::DBConfig;
 use crate::kvdb::{ErrorKind, KeyValueDB, Result};
+use log::warn;
 use rocksdb::{ColumnFamily, Options, WriteBatch, DB};
 use std::ops::Range;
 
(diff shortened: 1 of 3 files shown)


db/src/diskdb.rs
```diff
@@ -21,7 +22,18 @@ impl RocksDB {
 
         let cfnames: Vec<_> = (0..columns).map(|c| format!("c{}", c)).collect();
         let cf_options: Vec<&str> = cfnames.iter().map(|n| n as &str).collect();
-        let db = DB::open_cf(&opts, &config.path, &cf_options).expect("Failed to open rocksdb");
+        let db = DB::open_cf(&opts, &config.path, &cf_options).unwrap_or_else(|err| {
+            warn!("Failed to open rocksdb: {}", err);
```
quake: Only `Corruption` error is repairable, we may need to check err message string before run repair function.

yangby-cryptape: > Only `Corruption` error is repairable, ...

Yes!

I used to be doubtful should we just [test a string to classify the error kind](https://github.com/rust-rocksdb/rust-rocksdb/blob/ea770309c8ccf99e7e2d9be0194f476bbf008346/src/lib.rs#L107-L110):
- If the error was not a `Corruption` error, repairer will do nothing. 
- If the error message was changed in upstream, we won't be noticed unless we read the ChangeLog (or source code).

PR #434, opened 2019-04-10
feat: Change all u64 fields in RPC to String

xxuejie:
This PR actually contains 2 parts:

* refactor: Use TryFrom when parsing RPC request data
* feat: Change capacity field to String type in RPC

### DIFF
--- a/miner/src/miner.rs
+++ b/miner/src/miner.rs
@@ -2,6 +2,7 @@ use crate::client::Client;
 use crate::Work;
 use ckb_core::block::{Block, BlockBuilder};
 use ckb_core::header::{HeaderBuilder, RawHeader, Seal};
+use ckb_core::BlockNumber;
 use ckb_pow::PowEngine;
 use ckb_util::TryInto;
 use crossbeam_channel::Receiver;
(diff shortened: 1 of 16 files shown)


miner/src/miner.rs
```diff
@@ -70,11 +71,34 @@ impl Miner {
                 .timestamp(current_time)
                 .parent_hash(parent_hash);
 
+            let uncles = uncles.into_iter().map(TryInto::try_into).collect();
+            if let Err(ref e) = uncles {
+                error!(target: "miner", "error parsing uncles: {:?}", e);
+            }
+            let cellbase = cellbase.try_into();
+            if let Err(ref e) = cellbase {
+                error!(target: "miner", "error parsing cellbase: {:?}", e);
```
jjyr: Should return a Result instead continue execution?
The below code will panic

xxuejie: Ouch I meant to return after printing log here, somehow forgot to add the code 😂 Thanks for pointing this out

jjyr: Consider using the `?` operator to return Result? which is a more general way to do error handling in rust, current code looks like golang style..

xxuejie: How about now?

PR #435, opened 2019-04-10
refactor: store module

quake:
BREAKCHANGE: This PR is a break change since we changed the rocksdb's column family name and column list, need to delete old database.

This PR refactor db and chain store, make it easy to switch different key-value store implementation.
- [x] Refactor `Batch` to associated type
- [x] Refactor `ChainIndex` 
- [x] Remove some static lifetime bound for `ChainIndex`
- [x] Change `Col` from Option to u32
- [x] Move `Cachedb` to db module
- [ ] Merge `ChainIndex` and `ChainStore` into one trait (plan to do it in separate PR)


db/src/lib.rs
```diff
@@ -3,13 +3,37 @@
...
 pub mod memorydb;
+pub mod rocksdb;
 
 pub use crate::config::DBConfig;
-pub use crate::diskdb::RocksDB;
-pub use crate::kvdb::KeyValueDB;
 pub use crate::memorydb::MemoryKeyValueDB;
+pub use crate::rocksdb::RocksDB;
+
+pub type Col = Option<u32>;
```
doitian: Indeed, we never use `None` as a col. It is simpler to just define Col as u32.

db/src/rocksdb.rs
```diff
@@ -32,62 +26,97 @@ impl RocksDB {
...
         }
 
-        let inner = Inner {
-            db,
-            cfnames: cfnames.clone(),
-        };
-        RocksDB { inner }
-    }
-
-    fn cf_handle(&self, col: Option<u32>) -> Result<Option<ColumnFamily>> {
```
doitian: It can be defined as a global function `cf_handle(db, col) -> Result` to avoid the duplicated code in read/insert/delete.

shared/src/store.rs
```diff
@@ -38,68 +38,38 @@ impl<T: 'static + KeyValueDB> ChainKVStore<T> {
...
+impl<T: KeyValueDB> ChainStore for ChainKVStore<T> {
+    type Batch = DefaultStoreBatch<T::Batch>;
 
-    fn size_hint(&self) -> (usize, Option<usize>) {
-        match self.head {
-            Some(ref h) => (1, Some(h.number() as usize + 1)),
-            None => (0, Some(0)),
+    fn new_batch(&self) -> Self::Batch {
+        DefaultStoreBatch {
+            inner: self.db.batch().expect("new db batch should be ok"),
```
doitian: If it should be OK, why let the interface returns a result? The two interfaces should be consistent, both use `Result` or neither.

shared/src/store.rs
```diff
@@ -161,63 +131,99 @@ impl<T: 'static + KeyValueDB> ChainStore for ChainKVStore<T> {
...
-        batch.insert(COLUMN_BLOCK_BODY, hash.clone(), block_data);
-        batch.insert(
+    fn insert_serialize<T: Serialize + ?Sized>(&mut self, col: Col, key: &[u8], item: &T) {
+        self.inner
+            .insert(
+                col,
+                key,
+                &serialize(item).expect("serializing should be ok"),
+            )
+            .expect("batch insert_serialize should be ok")
```
doitian: Same with `new_batch`

shared/src/store.rs
```diff
@@ -161,63 +131,99 @@ impl<T: 'static + KeyValueDB> ChainStore for ChainKVStore<T> {
...
+    }
+
+    fn delete(&mut self, col: Col, key: &[u8]) {
+        self.inner
+            .delete(col, key)
+            .expect("batch delete should be ok")
```
doitian: Same with `new_batch`

zhangsoledad:
bors try

shared/src/store.rs
quake: Agree, I will change it to use `Result` later

quake: Resolved, please review again, thanks

PR #462, opened 2019-04-12
✋[ᚬmaster] develop

doitian:
**DO NOT MERGE**

[CI status](https://github.com/nervosnetwork/ckb/pull/462/checks)

- Travis CI - Branch: Integration Test
- Travis CI - Pull Request: cargo test

PR #464, opened 2019-04-13
refactor: don't have to build rocksdb when use all jsonrpc types

yangby-cryptape:
Move `ckb_shared::tx_pool::{Action, TxTrace}` to `jsonrpc_types::{Action, TxTrace}`.

There is a JSON-RPC protocol returns `TxTrace`:
https://github.com/nervosnetwork/ckb/blob/1afbd869eafcaaf14e2b9c672dff8f8ea21f9a3a/rpc/src/module/trace.rs#L25-L26

But if users introduce this struct from [`ckb_shared`](https://github.com/nervosnetwork/ckb/blob/1afbd869eafcaaf14e2b9c672dff8f8ea21f9a3a/shared/src/tx_pool/trace.rs#L17-L22), they have to build rocksdb.

After this PR, when users want to introduce all jsonrpc types to their rust applications, they can use `jsonrpc_types` instead of compile rocksdb or make a copy of this struct.


doitian:
> bors r+

RIP bors

doitian:
We no longer use bors. Integration test runs in `Travis CI - Branch` in this PR https://github.com/nervosnetwork/ckb/pull/462

PR #473, opened 2019-04-15
feature: cellbase maturity verify

u2:
#367 

> Updated,
> https://github.com/nervosnetwork/ckb/pull/367/files#diff-b2197becbc6452c42618e284e3dbf52eR226
> But I am not sure whether the cellbase_maturity and cell_set are correct, because the `max_cycles` coming from the function params, and the `cellbase_maturity` coming from the `self.consensus`.

> 
> @jjyr is going to take a look at this PR.@quake Do you prefer to merge it after your store refactoring, or it doesn't matter.

> Please rebase this PR to latest develop and ensure:
> 
> Make this a configurable option in chain spec.
> Only enable it in testnet chain spec. Disable it in dev and integration.

### DIFF
--- a/verification/src/transaction_verifier.rs
+++ b/verification/src/transaction_verifier.rs
@@ -15,6 +15,7 @@ pub struct TransactionVerifier<'a, M> {
     pub version: VersionVerifier<'a>,
     pub null: NullVerifier<'a>,
     pub empty: EmptyVerifier<'a>,
+    pub maturity: MaturityVerifier<'a>,
     pub capacity: CapacityVerifier<'a>,
     pub duplicate_inputs: DuplicateInputsVerifier<'a>,
     pub inputs: InputVerifier<'a>,
@@ -30,11 +31,13 @@ where
         rtx: &'a ResolvedTransaction,
         median_time_context: &'a M,
         tip_number: BlockNumber,
+        cellbase_maturity: usize,
     ) -> Self {
         TransactionVerifier {
             version: VersionVerifier::new(&rtx.transaction),
             null: NullVerifier::new(&rtx.transaction),
             empty: EmptyVerifier::new(&rtx.transaction),
+            maturity: MaturityVerifier::new(&rtx, tip_number, cellbase_maturity),
             duplicate_inputs: DuplicateInputsVerifier::new(&rtx.transaction),
             script: ScriptVerifier::new(rtx),
             capacity: CapacityVerifier::new(rtx),
@@ -47,6 +50,7 @@ where
         self.version.verify()?;
         self.empty.verify()?;
         self.null.verify()?;
+        self.maturity.verify()?;
         self.inputs.verify()?;
         self.capacity.verify()?;
         self.duplicate_inputs.verify()?;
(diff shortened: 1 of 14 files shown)


verification/src/transaction_verifier.rs
```diff
@@ -140,6 +153,50 @@ impl<'a> EmptyVerifier<'a> {
     }
 }
 
+pub struct MaturityVerifier<'a> {
+    transaction: &'a Transaction,
```
jjyr: Use ResolvedTransaction instead

verification/src/transaction_verifier.rs
```diff
@@ -140,6 +153,50 @@ impl<'a> EmptyVerifier<'a> {
...
+            transaction,
+            tip_number,
+            cell_set,
+            cellbase_maturity,
+        }
+    }
+
+    pub fn verify(&self) -> Result<(), TransactionError> {
+        let immature_spend = self.transaction.inputs().iter().any(|input| {
+            match self.cell_set.get(&input.previous_output.hash) {
```
jjyr: You can get `CellMeta` from `ResolvedRransaction#inputs`.
https://github.com/nervosnetwork/ckb/blob/develop/core/src/cell.rs#L10

Add `is_cellbase` to CellMeta, and use CellMeta to instead the directly reference to cell_set and transaction_meta

verification/src/transaction_verifier.rs
```diff
@@ -140,6 +153,50 @@ impl<'a> EmptyVerifier<'a> {
...
+    tip_number: BlockNumber,
+    cell_set: &'a FnvHashMap<H256, TransactionMeta>,
+    cellbase_maturity: usize,
+}
+
+impl<'a> MaturityVerifier<'a> {
+    pub fn new(
+        transaction: &'a Transaction,
+        tip_number: BlockNumber,
+        cell_set: &'a FnvHashMap<H256, TransactionMeta>,
```
jjyr: See below comment, we should avoid directly access the inner struct

verification/src/transaction_verifier.rs
u2: I see. Maybe I need to add a field called `is_cellbase` for `CellMeta`.

verification/src/transaction_verifier.rs
u2: Updated.

One question: should we add `cellbase maturity verify` for `dep_cells?

verification/src/transaction_verifier.rs
```diff
@@ -140,6 +144,51 @@ impl<'a> EmptyVerifier<'a> {
...
+        let immature_spend =
+            self.transaction
+                .input_cells
+                .iter()
+                .any(|cell_status| match cell_status.get_live() {
+                    Some(ref meta)
+                        if meta.is_cellbase()
+                            && self.tip_number
+                                < meta.block_number.expect(
+                                    "cell meta should have block number when transaction verify",
```
jjyr:
```suggestion
                                    "cellbase cell meta must have block number when transaction verify",
```

jjyr:
Pls skip CellbaseImmature error in relay protocol, https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/transaction_process.rs#L84

verification/src/transaction_verifier.rs
jjyr: https://bitcoin.stackexchange.com/questions/10825/why-must-generated-coins-mature-120-blocks-before-they-can-be-spent

Due to this explain, we should also add this verify for dep_cells

u2: updated

u2:
> Please rebase this PR to latest develop and ensure:
> 
> Make this a configurable option in chain spec.
> Only enable it in testnet chain spec. Disable it in dev and integration.

@doitian I will create another PR for it a few days later, ok?

u2:
#54

jjyr:
seems it broke the integration test https://travis-ci.com/nervosnetwork/ckb/jobs/193211555

jjyr:
Template solution https://github.com/nervosnetwork/ckb/pull/479

PR #476, opened 2019-04-15
Modify jsonrpc types

yangby-cryptape:
#### Changes
+ chore: let all jsonrpc types be public (for client)
+ feat: change all `u64` fields in RPC to String and hide internal `Script` struct
+ chore: replace unnecessary `TryFrom` by `From`
+ docs: fix README.md for JSON-RPC protocols
  - It's too difficult to distinguish differences between old and new versions for all protocols, so I rewrite all examples after do tests.
  - [`shell` is not a keyword for highlight](https://github.com/github/linguist/blob/2734ca0134c0afe3c62ecb1ec8c721a28c8b9e63/lib/linguist/languages.yml#L4626-L4633)

#### Note
  https://github.com/nervosnetwork/ckb/blob/dbfb81cea902ba4d923f78685ed0b0bfddce7a3a/util/jsonrpc-types/src/cell.rs#L11-L15

This piece of code wasn't list for review.

But, `Capacity` was changed from `u64` to `String`, and the `Script` was changed from `ckb-core::Script` to `jsronrpc-tpyes::Script`.

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -263,7 +263,7 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
     fn transform_cellbase(tx: &Transaction, cycles: Option<Cycle>) -> CellbaseTemplate {
         CellbaseTemplate {
             hash: tx.hash(),
-            cycles,
+            cycles: cycles.map(|c| c.to_string()),
             data: tx.into(),
         }
     }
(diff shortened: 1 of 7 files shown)


doitian:
`rpc/README.md` is outdated, could you please update it as well. For example block number parameter in `get_block_hash` is still a number in the doc.

yangby-cryptape:
@doitian Done!

It's a bit difficult to distinguish differences between old and new versions for all protocols, so I rewrite all examples after do tests.

There are three protocols I did not write (2 for miner, 1 for test).

PR #477, opened 2019-04-16
fix: mine depend txs in one block

jjyr:


miner/src/block_assembler.rs
```diff
@@ -324,8 +374,13 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
         // bytes, they really serve the same purpose at the moment
         let block_reward = self.shared.block_reward(header.number() + 1);
         let mut fee = 0;
-        for pe in pes {
-            fee += self.shared.calculate_transaction_fee(&pe.transaction)?;
+        for (i, pe) in pes.iter().enumerate() {
+            // depends cells may produced from previous tx
+            let fee_calculator = FeeCalculator {
```
quake: It's better to use `HashMap` in FeeCalculator as a previously calculated txs capacity cache, and change the `calculate_transaction_fee` function to mut, which will simplify code here.

test/src/specs/transaction/depend_tx_in_same_block.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{sleep, Net, Spec};
+use log::info;
+
+pub struct DepentTxInSameBlock;
```
quake: Please move this spec to specs/mining.rs

jjyr: I'm not sure, even this bug caused by miner, in general, this is a transaction/block verifier related spec.

quake: Verifier uses different logic to verify cellbase capacity, this PR only fix the bug in miner's `create_cellbase_transaction` function, so it's better to move this spec to `mining.rs`

PR #491, opened 2019-04-17
chore: update lock cell for segwit and address format

classicalliu:
Replace `resource/specs/cells/secp256k1_blake2b_lock` to `secp256k1_blake160_sighash_all` for support segwit and [address format](https://github.com/nervosnetwork/ckb/wiki/Common-Address-Format)

BREAKING CHANGE: Testchain spec has changed the system cells, which leads to different genesis block hash.


doitian:
Some tests depend on the genesis hash, you have to fix them.

classicalliu:
> Some tests depend on the genesis hash, you have to fix them.

👌

PR #496, opened 2019-04-18
✋ feat: add two wallet rpc

quake:
add `get_live_cells_by_lock_hash` and `get_transactions_by_lock_hash` rpc

### DIFF
--- a/rpc/src/module/wallet.rs
+++ b/rpc/src/module/wallet.rs
@@ -1,14 +1,14 @@
 use bincode::{deserialize, serialize};
 use ckb_core::block::Block;
 use ckb_core::header::BlockNumber;
-use ckb_core::transaction::{CellOutput, OutPoint as CoreOutPoint};
+use ckb_core::transaction::{CellOutput, OutPoint};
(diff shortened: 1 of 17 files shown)


rpc/src/module/wallet.rs
```diff
@@ -0,0 +1,563 @@
...
+        DefaultWalletStore { db }
+    }
+
+    pub fn start<S: ToString>(self, thread_name: Option<S>, notify: &NotifyController) {
+        let mut thread_builder = thread::Builder::new();
+        if let Some(name) = thread_name {
+            thread_builder = thread_builder.name(name.to_string());
+        }
+
+        let new_tip_receiver = notify.subscribe_new_tip(WALLET_SUBSCRIBER);
```
doitian: Since the DB is updated async, if the node crashes after data having been written to main DB, but before writing to the index DB, then the two DBs are out of sync. We need to store the tip in index DB, so that we know when a new tip change event comes, whether or not we should catch up the old tip first.

quake: aha, I totally forgot to handle this scenario, may need to add a cli command to sync wallet db with chain db.

Issue #389, opened 2019-04-03
Make dummy pow time configurable


u2:
https://github.com/nervosnetwork/ckb/blob/develop/pow/src/dummy.rs#L29

Adjust dummy pow time to a smaller value for facilitating test or add some config.

quake:
We may change it to config value.

doitian:
https://github.com/nervosnetwork/ckb/pull/501

PR #501, opened 2019-04-18
feat: add parameters to control the block generation rate in dummy mode

yangby-cryptape:
When I do tests in my laptop, the speed of product block is too slow.

Ref: #389


doitian:
Need to update the doc in Ruby demo.
cc @xxuejie @classicalliu

yangby-cryptape:
These parameters was not required.

If these parameters wasn't provided, the default option will be used:
- Delay 5000 milliseconds to send back the result to miners.

PR #503, opened 2019-04-18
feat: rpc resolve tx from pending and staging

driftluo:


rpc/src/module/pool.rs
```diff
@@ -37,7 +37,7 @@ impl<CS: ChainStore + 'static> PoolRpc for PoolRpcImpl<CS> {
         let tx: CoreTransaction = tx.try_into().map_err(|_| Error::parse_error())?;
 
         let mut chain_state = self.shared.chain_state().lock();
-        let rtx = chain_state.resolve_tx_from_pool(&tx, &chain_state.tx_pool());
+        let rtx = chain_state.rpc_resolve_tx_from_pool(&tx, &chain_state.tx_pool());
         let tx_result = chain_state.verify_rtx(&rtx, self.shared.consensus().max_block_cycles());
         debug!(target: "rpc", "send_transaction add to pool result: {:?}", tx_result);
         match tx_result {
```
jjyr: We can remove the special error `TransactionError::Unknown` now.

PR #505, opened 2019-04-18
fix: verify genesis

zhangsoledad:


src/setup/mod.rs
```diff
@@ -199,12 +201,19 @@ fn is_daemon(subcommand_name: &str) -> bool {
 }
 
 fn consensus_from_spec(spec: &ChainSpec) -> Result<Consensus, ExitCode> {
-    spec.to_consensus().map_err(|err| {
+    spec.to_consensus().and_then(verify_genesis).map_err(|err| {
```
doitian: `scope.set_tag("genesis", consensus.genesis_block.header().hash());` can use `consensus.genesis_hash` instead.

PR #510, opened 2019-04-19
hotfix: fix VM hang bug for certain invalid programs

xxuejie:
This is a hotfix for VM bug fixed at here: https://github.com/nervosnetwork/ckb-vm/pull/47

Since it would result in CKB hanging there forever, it's better we handle this earlier as a hotfix than next VM upgrade.


xxuejie:
@doitian Can we cherry-pick this to 0.9.0 if we can get it merged in time?

PR #514, opened 2019-04-19
[WIP] refactor: tx_pool

zhangsoledad:

PR #515, opened 2019-04-20
✋ fix: tx pool cellbase immature check

jjyr:
* fix immature check
* refactor ckb config to allow rewrite configs in tests
* add an immature check test case


quake:
Is it possible to add config file override? like this:

```toml
include = "ckb-integration-test"

[params]
cellbase_maturity = 0
```

It may be easier to add more different spec config value for future integration tests in this way than serializing / deserializing the full config file.

zhangsoledad:
I will take this to https://github.com/nervosnetwork/ckb/pull/514

jjyr:
> I will take this to #514

review integration tests part, I'll rebase this PR after #514 get merged

Issue #527, opened 2019-04-22
The results of some tests is not deterministic


yangby-cryptape:
### Issue

These tests in my laptop can get random results:

The latest version:
https://github.com/nervosnetwork/ckb/blob/a7c27df3b8b2f7d0216b9ad0b27085d1fd1f4232/chain/src/tests/delay_verify.rs#L94
https://github.com/nervosnetwork/ckb/blob/a7c27df3b8b2f7d0216b9ad0b27085d1fd1f4232/chain/src/tests/basic.rs#L247

The second latest version:
https://github.com/nervosnetwork/ckb/blob/4a8533377c85a1594460def46a822bfd12707e3c/chain/src/tests/delay_verify.rs#L94
https://github.com/nervosnetwork/ckb/blob/4a8533377c85a1594460def46a822bfd12707e3c/chain/src/tests/basic.rs#L198

### Further

Caused by:
https://github.com/nervosnetwork/ckb/blob/ef1e7ff156e9e021f100168d89a4530f06cea0a7/verification/src/block_verifier.rs#L370-L375

I used `regex` to fix it (first commit in #528).

But, a better solution is required.

quake:
It's deterministic, the change is from this PR:
https://github.com/nervosnetwork/ckb/pull/520/files#diff-365798ee66cdaead4971104be0b60a52L94

Before #520 , we are using a hashmap to check duplicate inputs, which will report the error index of the second duplicate transaction inputs. After refactoring, it will report the first index.

yangby-cryptape:
But, I ran `make ci` for 63060826224aefceea7a0a085b68eae01d4a9529, again and again.

I got both two results.

PR #528, opened 2019-04-22
feat: capacity uses unit shannon which is 10e-8 CKBytes

yangby-cryptape:
### Commits

- fix: some tests about transactions are not deterministic
- refactor: refactor the `Capacity`
  - refactor: rename the occupied capacity derive to fix naming conflict
  - refactor: use the last 8 digits of capacity as the fractional part
    **BREAKING CHANGE**:
    The value of capacity is `10^8` times of the old value, since the last 8 digits of capacity are used as the fractional part.
  - refactor: apply the new capacity to all crates
  - ci: apply the new capacity to all tests
  - ci: test if only `Capacity` was changed
  - ci: remove test code and update some hardcode hash in tests

### Summary

- I split this PR into several commits, since review these commits one by one is more easy than review them all at once.
- The first part has only one commit(432256bea073a1c01c4a146bd3e23a10c6e99171) which fixed a bug (#527). I had to fix it before I push other commits.
- In the second part:
  - The second commit(3bb7109d940de4afbe8a6c6bbafce20ddb6f6a8e) and the third commit(85f485426e3a30781ee6417acf355c78c891019d) is the core commits of the refactor.
  - The fifth commit(deee2c2266d2d62916643deb7f1856f4a0e26ad5) proves the changes about hardcode hashes in the last commit is right.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -419,7 +419,6 @@ dependencies = [
  "numext-fixed-hash 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "numext-fixed-uint 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "occupied-capacity 0.11.0-pre",
- "occupied-capacity-derive 0.11.0-pre",
  "serde 1.0.90 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_bytes 0.11.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.90 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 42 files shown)


yangby-cryptape:
cc @ashchan @classicalliu 

- `shannon` is the smallest indivisible unit in `Capacity`
- `byte` is the basic unit in `Capacity`
- `1 byte = 10^8 shannons`
- The value of capacity is `10^8` times of the old value, since the value of `Capacity` was changed from `byte` to `shannon`
- The range of `Capacity` is not changed, it still is `[0, 2^64-1]`.
- The `JSON-RPC` protocols are not changed.  

I will update the RFCs to add more details, and I will mention you in that PR.

PR #529, opened 2019-04-23
✋ refactor: rpc service in actor

jjyr:
Use an event loop agent handle RPC calls,  to prevent holding lock too long.

### DIFF
--- a/rpc/src/agent.rs
+++ b/rpc/src/agent.rs
@@ -1,0 +1,622 @@
+#![allow(clippy::type_complexity)]
+use build_info::{get_version, Version};
+use ckb_chain::chain::ChainController;
+use ckb_core::service::{Request, DEFAULT_CHANNEL_SIZE, SIGNAL_CHANNEL_SIZE};
+use ckb_core::{
+    block::Block as CoreBlock,
+    cell::CellProvider,
+    cell::CellStatus,
+    header::Header as CoreHeader,
+    transaction::OutPoint as CoreOutPoint,
+    transaction::{ProposalShortId, Transaction as CoreTransaction},
+    BlockNumber, Cycle,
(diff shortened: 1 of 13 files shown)


rpc/src/agent.rs
```diff
@@ -0,0 +1,622 @@
...
+            block_assembler,
+        }
+    }
+
+    #[allow(clippy::cyclomatic_complexity)]
+    pub fn start<S: ToString>(self, thread_name: Option<S>) -> RpcAgentController {
+        let (signal_sender, signal_receiver) =
+            crossbeam_channel::bounded::<()>(SIGNAL_CHANNEL_SIZE);
+        let (local_node_info_sender, local_node_info_receiver) =
+            crossbeam_channel::bounded(SIGNAL_CHANNEL_SIZE);
```
quake: should be `DEFAULT_CHANNEL_SIZE` ?

jjyr: Good catch

keroro520:
Agent is a single point. It locks every time processing rpc-requests like before. So I think it has the same lock holding time with before.
I cannot figure out the point of why event loop agent prevent holding lock ?

And what about using RAII guard rule to optimize this problem ? @jjyr

jjyr:
> Agent is a single point. It locks every time processing rpc-requests like before. So I think it has the same lock holding time with before.
> I cannot figure out the point of why event loop agent prevent holding lock ?
> 
> And what about using RAII guard rule to optimize this problem ? @jjyr

It can reduce the number of RPC lock waiters from n to 1 agent.

doitian:
Need some benchmark to prove it worths to apply the changes.

doitian:
Close for now.

PR #534, opened 2019-04-23
chore: Change default ping interval and timeout

TheWaWaR:
This is an experience value, refer to [bitcoin](https://github.com/bitcoin/bitcoin/blob/master/src/net.h#L42-L45).


TheWaWaR:
I understand this as experience value, refer to [bitcoin](https://github.com/bitcoin/bitcoin/blob/master/src/net.h#L42-L45)

PR #539, opened 2019-04-23
chore: change default max_outbound_peers to 8

jjyr:


doitian:
@TheWaWaR @zhangsoledad please confirm this change

TheWaWaR:
That is what [bitcoin](https://github.com/bitcoin/bitcoin/blob/master/src/net.h#L59) does, confirmed!

PR #541, opened 2019-04-23
fix: apply the new capacity to all configuration files

yangby-cryptape:
Fix the [integration tests](https://travis-ci.com/nervosnetwork/ckb/jobs/194822193).

BREAKING CHANGE:
- `initial_block_reward` in configuration files have to be updated.
  The unit of it is changed from "CK Byte" to "CK Shannon".

### DIFF
--- a/resource/specs/dev.toml
+++ b/resource/specs/dev.toml
@@ -14,7 +14,7 @@ nonce = 0
 proof = [0]
 
 [params]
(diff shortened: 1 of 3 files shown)


yangby-cryptape:
cc @ashchan @classicalliu 
- `initial_block_reward` in configuration files have to be updated.

PR #547, opened 2019-04-24
[HOLD] [ᚬtestnet] feat: Use DNS txt records for peer address seeding

TheWaWaR:
Use DNS txt records for peer address seeding. 

### When start DNS seeding?
* If peer store is empty, DNS seeding will immediately start
* If peer store is not empty, wait for [11 seconds ](https://github.com/bitcoin/bitcoin/blob/master/src/net.cpp#L1530)
  - After 11 seconds, if there are less than 2 outbound peers, start DNS seeding

### How DNS seeding works?
* Read seeding DNS address from `ckb.toml` config file `network` section `bootnodes` field.
* Query every `bootnode` for TXT record and try to parse and verify them by hard coded **pubkey** (secret key is from domain owner).
  - If verify success, add the address to peer store
  - If verify failed, try to covert `bootnode` to Multiaddr, and add the success value to peer store
* Wait outbound peer service to dial those added address

### The record format:
```
TXT = {ip};{port};{peer_id(base58)};{valid_until(base10)};{signature(base58)}
```
A **typical** TXT record:
```
x.x.x.x;x;QmdkPCdJCYHFSwyhEPqskrnEnUk37HyPGmZxC98Ld5gG3T;1890561609;vadxC7jQCaUDKw3kRbKhM2zFw9o7mNYxcJGZcqyGQXDj5nAratAF8TFtMxRMoESLM48sMwus2vodHEkeESUUqsEw
```

### How signature built?
```
hash = keccak::sha3_256([ip, port, peer_id, valid_until].join(';'))
signature = Secp256k1.sign_recoverable(hash, privkey)
```

### Generate the TXT records
NOTE: As generate a DNS TXT record is not a common user use case, please use this [separated tool](https://github.com/TheWaWaR/p2p-address-seeding) to generate the record.

**Usage** :
```
./target/debug/p2p-address-seeding \
    generate \
    -p ../bob-privkey \
    --addr "/ip4/x.x.x.x/tcp/x/p2p/QmdkPCdJCYHFSwyhEPqskrnEnUk37HyPGmZxC98Ld5gG3T" \
    --valid-until "2029-11-28T21:00:09+09:00"
```


network/src/services/dns_seeding/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+                                }
+                            },
+                            Err(err) => {
+                                debug!(target: "network", "get dns txt record error: {:?}", err);
+                            }
+                        }
+                    }
+                }
+                Err(_) => {
+                    if let Ok(addr) = seed.to_multiaddr() {
```
doitian: DNS may fail temporarily. Can we explicitly differentiate between a peer address and a DNS in `bootnodes`, such as using a special format so that DNS domain should not be recognized as a valid multiaddr.

network/src/services/dns_seeding/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+        self.network_state.with_peer_store_mut(|peer_store| {
+            for mut addr in addrs {
+                match addr.pop() {
+                    Some(Protocol::P2p(key)) => {
+                        if let Ok(peer_id) = PeerId::from_bytes(key.into_bytes()) {
+                            peer_store.add_discovered_addr(&peer_id, addr);
+                        }
+                    }
+                    _ => {
+                        debug!(target: "network", "Got addr without peer_id: {}", addr);
```
doitian: This should be an `error` level log.

network/src/services/dns_seeding/seed_record.rs
```diff
@@ -0,0 +1,273 @@
...
+            .as_secs()
+            > self.valid_until
+        {
+            return Err(SeedRecordError::SeedTimeout);
+        }
+        Ok(())
+    }
+
+    pub fn decode(record: &str) -> Result<SeedRecord, SeedRecordError> {
+        if record.split(SEP).count() != 5 {
```
doitian: I would rather collect the result into a `Vec<&str>` here.

network/src/config.rs
```diff
@@ -23,7 +23,7 @@ pub struct NetworkConfig {
     pub connect_outbound_interval_secs: u64,
     pub listen_addresses: Vec<Multiaddr>,
     pub public_addresses: Vec<Multiaddr>,
-    pub bootnodes: Vec<Multiaddr>,
+    pub bootnodes: Vec<String>,
```
quake: Multiaddr supports dns format, is it possible to keep Multiaddr here? User can mix different bootnodes in config:
```
bootnodes = [
"/ip4/47.111.175.189/tcp/8111/p2p/QmQqYVVigwiogARK4UgB5scgEd9tUfhSiZkLK7bM82cVCu",
"/dns/crystal-nova.rylai.nervos.org/"
]
```

network/src/services/dns_seeding/mod.rs
TheWaWaR: OK

network/src/services/dns_seeding/seed_record.rs
TheWaWaR: Sounds good

network/src/config.rs
TheWaWaR: I'll separate them

network/src/services/dns_seeding/mod.rs
TheWaWaR: PeerId is not required for dial a peer, but the connection still can be encrypted, just the dialer can not sure the listener is the listener (identify by ip:port).
So if worried about someone hack the network between dialer and listener here should be an `error` level log, if not I think debug is enough!

network/src/config.rs
TheWaWaR: `/dns` is not an official Protocol, and I don't what choose between `/dns4` and `/dns6`.

network/src/services/dns_seeding/seed_record.rs
```diff
@@ -0,0 +1,267 @@
...
+            .as_secs()
+            > self.valid_until
+        {
+            return Err(SeedRecordError::SeedTimeout);
+        }
+        Ok(())
+    }
+
+    pub fn decode(record: &str) -> Result<SeedRecord, SeedRecordError> {
+        let parts = record.split(SEP).collect::<Vec<&str>>();
```
doitian:
```suggestion
        let parts = record.splitn(6, SEP).collect::<Vec<&str>>();
```

TheWaWaR:
Will fire PR to `develop` branch.

PR #557, opened 2019-04-25
fix: hash => tx_hash in RPC of OutPoint

classicalliu:


classicalliu:
> Could u update the [documents](https://github.com/nervosnetwork/ckb/blob/e1283e4fbaf65afbb98fa3ff103598f3d190f96b/rpc/README.md), please?

👌

PR #558, opened 2019-04-25
Testnet hotfix 

zhangsoledad:


rpc/src/module/miner.rs
```diff
@@ -57,9 +57,15 @@ impl<CS: ChainStore + 'static> MinerRpc for MinerRpcImpl<CS> {
...
             None => None,
         };
-        self.block_assembler
-            .get_block_template(cycles_limit, bytes_limit, max_version)
-            .map_err(|_| Error::internal_error())
+        let r = self
+            .block_assembler
+            .get_block_template(cycles_limit, bytes_limit, max_version);
+        //r.map_err(|_| Error::internal_error())
+        r.map_err(|err| Error {
```
jjyr: Remove this debug log

rpc/src/module/pool.rs
```diff
@@ -34,29 +31,24 @@ pub(crate) struct PoolRpcImpl<CS> {
...
-                }
+                let fbb = &mut FlatBufferBuilder::new();
+                let message = RelayMessage::build_transaction(fbb, &tx, cycles);
+                fbb.finish(message, None);
+                let data = fbb.finished_data().to_vec();
+                self.network_controller
+                    .broadcast(NetworkProtocol::RELAY.into(), data);
+                Ok(tx_hash)
             }
+            Err(e) => Err(RPCError::custom(RPCError::Staging, e.to_string())),
```
jjyr: should return `RPCError::Invalid`, Staging error is removed on develop

shared/src/chain_state.rs
```diff
@@ -177,138 +176,63 @@ impl<CS: ChainStore> ChainState<CS> {
...
-                let cycles = TransactionVerifier::new(
-                    &rtx,
-                    &self,
-                    self.tip_number(),
-                    self.consensus().cellbase_maturity,
-                )
-                .verify(max_cycles)?;
-                // write cache
-                self.txs_verify_cache.borrow_mut().insert(tx_hash, cycles);
+        match self.verify_rtx(&rtx, None) {
```
quake: Use Result#map:

```suggestion
 self.verify_rtx(&rtx, None).map(|cycles| {
    if self.contains_proposal_id(&short_id) {
        // if tx is proposed, we resolve from staging, verify again
        self.staging_tx_and_descendants(&mut tx_pool, Some(cycles), tx);
    } else {
        tx_pool.enqueue_tx(Some(cycles), tx);
    }
    cycles
})
```

shared/src/tx_pool/pending.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    pub(crate) fn get_tx(&self, id: &ProposalShortId) -> Option<&Transaction> {
+        self.get(id).map(|x| &x.transaction)
+    }
+
+    pub(crate) fn remove(&mut self, id: &ProposalShortId) -> Option<PoolEntry> {
+        self.inner.remove(id)
+    }
+
+    pub(crate) fn fetch(&self, n: usize) -> Vec<ProposalShortId> {
```
quake: Unused fn?

sync/src/synchronizer/mod.rs
```diff
@@ -734,6 +721,23 @@ impl<CS: ChainStore> CKBProtocolHandler for Synchronizer<CS> {
...
+        info!(target: "sync", "SyncProtocol.connected peer={}", peer_index);
+        self.on_connected(nc.as_ref(), peer_index);
+    }
+
+    fn disconnected(&mut self, _nc: Box<CKBProtocolContext>, peer_index: PeerIndex) {
+        info!(target: "sync", "SyncProtocol.disconnected peer={}", peer_index);
+        let mut state = self.peers.state.write();
+        if let Some(peer_state) = state.get(&peer_index) {
+            if peer_state.sync_started && self.n_sync.fetch_sub(1, Ordering::Release) == 0 {
+                panic!("Synchronizer n_sync overflow");
```
quake: could you please explain in more detail when will panic?

quake:
I noticed that there are several disabled lint, please double check and cleanup:
```
#![allow(dead_code)]
#[allow(unreachable_patterns)]
#[allow(unused_imports)] 
```

pow/src/dummy.rs
```diff
@@ -4,13 +4,13 @@ use rand::{
...
-use serde_derive::Deserialize;
+use serde_derive::{Deserialize, Serialize};
 use std::{fmt, thread, time};
 
-#[derive(Deserialize, Copy, Clone, Eq, PartialEq, Hash, Debug, Default)]
+#[derive(Serialize, Deserialize, Copy, Clone, Eq, PartialEq, Hash, Debug, Default)]
 pub struct DummyPowParams {
     // Delay offset (in milliseconds)
-    #[serde(default)]
+    #[serde(skip)]
```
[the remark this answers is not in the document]
zhangsoledad: ref https://github.com/nervosnetwork/ckb/commit/a302643669ad405a60c494ca2fd9ab2ddfe64692
need derive  Serialize on DummyPowParams,`#[serde(default)]`

zhangsoledad:
> I noticed that there are several disabled lints, please double check and cleanup:
> 
> ```
> #![allow(dead_code)]
> #[allow(unreachable_patterns)]
> #[allow(unused_imports)] 
> ```

All of this is mute for noise or incorrect lint

sync/src/synchronizer/mod.rs
zhangsoledad: if our code running as we expect, it shouldn't happen.
 ` fetch_sub` wraps around on overflow, we need check manually, panic here to prevent some bug be hidden silently.

PR #559, opened 2019-04-25
feat: add panic logger hook

keroro520:
The Rust built-in panic hook only prints panics to stdout. Here add
a hook to print to logfile as well.


util/logger/src/lib.rs
```diff
@@ -174,3 +177,32 @@ pub fn init(config: Config) -> Result<LoggerInitGuard, SetLoggerError> {
...
+        log::error!(
+            target: "panic", "thread '{}' panicked at '{}': {}:{}{:?}",
+            name,
+            msg,
+            location.file(),
+            location.line(),
+            backtrace,
+        );
+    };
+    panic::set_hook(Box::new(move |info| {
```
doitian: This conflicts with sentry:

> Registers a custom panic hook, replacing any that was previously registered.

doitian: https://github.com/getsentry/sentry-rust/blob/bdffebe356212dab7968e72941bed77a56556e59/src/integrations/panic.rs#L84

TheWaWaR: > This conflicts with sentry:
> 
> > Registers a custom panic hook, replacing any that was previously registered.

It says dispatches (take the current hook and call after report sent)
``` rust
pub fn register_panic_handler() {
    let next = panic::take_hook();
    panic::set_hook(Box::new(move |info| {
        panic_handler(info);
        next(info);
    }));
}
```

TheWaWaR: Maybe we can do the same as `[sentry](https://github.com/getsentry/sentry-rust/blob/bdffebe356212dab7968e72941bed77a56556e59/src/integrations/panic.rs#L84)` do, so we do not care about the order.

keroro520: > Maybe we can do the same as `[sentry](https://github.com/getsentry/sentry-rust/blob/bdffebe356212dab7968e72941bed77a56556e59/src/integrations/panic.rs#L84)` do, so we do not care about the order.

Yes, I am considering about it. I added a comment [here](https://github.com/nervosnetwork/ckb/pull/559/files#diff-15ed784c2cdfaaa8f2b56f024da15f55R206), the disadvantage is that stdout will display twice panic info since logger's output redirect to stdout as well.

keroro520: @TheWaWaR @doitian I decide to follow @TheWaWaR 's addvice to wrap previous hook inside panic_logger. It is a safer way.

As for displaying twice in stdout, I think, that is ok, just ignore it.

util/logger/src/lib.rs
```diff
@@ -174,3 +177,32 @@ pub fn init(config: Config) -> Result<LoggerInitGuard, SetLoggerError> {
...
+        panic_logger(info);
+        next(info);
```
doitian: We don't want the default hook. This will print the panic twice. Just override the default hook, and add a comment in `Setup` that logger must init first before other panic hooks.

doitian:
Test failed because of #565

util/logger/src/lib.rs
keroro520: Updated:

* Remove default hook
* Add comments in `Setup::setup_app` and `setup_panic_logger`

doitian:
Please rebase on to latest develop:

- Need to resolve the conflicts, `Setup` has been moved to `util/app-config/src/lib.rs`.
- #565 is merged, which has fixed the CI failures.

keroro520:
@doitian Rebased and CI passed

PR #560, opened 2019-04-25
✋ fix: revert secp256k1_blake160_sighash_all

doitian:
A wrong version is selected when merged into develop from v0.9.0.

---
Hold because I want to write a tool to dump various hashes into a file and commit that file into the repository.

- [ ] Add a tool to dump various hashes into a meta file and commit that file into repo.
- [ ] Add CI job to check the meta file is up-to-dated.
- [ ] Mention the file in README so the SDK team can easily traces the changes.


doitian:
wrong base, will recreate

PR #561, opened 2019-04-25
[ᚬrc/v0.10.0] fix: revert secp256k1_blake160_sighash_all

doitian:
A wrong version is selected when merged into develop from v0.9.0.


doitian:
I will write a tool to dump various hashes into a file and commit that file into the repository.

- [ ] Add a tool to dump various hashes into a meta file and commit that file into repo.
    - https://github.com/nervosnetwork/ckb/blob/v0.9.0/spec/src/lib.rs#L231
- [ ] Add CI job to check the meta file is up-to-dated.
- [ ] Mention the file in README so the SDK team can easily traces the changes.

PR #563, opened 2019-04-25
[WIP] feat: max_proposals and block_size limit

zhangsoledad:


core/src/header.rs
```diff
@@ -90,6 +90,15 @@ impl RawHeader {
     pub fn mut_uncles_count(&mut self) -> &mut u32 {
         &mut self.uncles_count
     }
+
+    // temp
+    pub const fn serialized_size() -> usize {
```
janx: Let's improve this to remove hard-coded calculations.

janx: I suggest the name `normalized_size`, as this size is used for block size limit calculation, not a size of real serialization.

janx: Would it be better to abstract the logic into `NormalizedSizeCalculator` to calculate `normalized_size` for different structs?

doitian: We'll use CFB later, let's use the temp quick solution first to verify our assumption.

doitian:
I'll continue the work

PR #552, opened 2019-04-24
[HOLD] feat: verify cellbase output type

quake:
This PR allows the miner to set cellbase's output type, and some refactoring:

1.  Remove unused `NullVerifier` and  `DuplicateInputsVerifier `
2. Revert changes of `get_cell_status` and `LiveCell`
3. Refactoring `ChainState`

PR #567, opened 2019-04-26
[HOLD] refactor: CellProvider and TransactionVerifier

quake:
I split this PR to 5 small commits, make sure they pass the unit test individuallly, and it's easier to review one by one than whole change log.

1. Move resolve_transaction fn to CellProvider and avoid some clone

2. Return short circuit `UnresolvableError` in resolve_transaction fn

3. ResolvedTransaction contains live cell only, simplify code and remove some unnecessary verifiers

4. ResolvedTransaction contains `CellMeta` only

5. Add lifetime to ResolvedTransaction, avoiding transaction clone

After this PR is merged, I will rebase https://github.com/nervosnetwork/ckb/pull/552, it's just 2 or 3 lines to enable cellbase script type verification.

### DIFF
--- a/core/src/cell.rs
+++ b/core/src/cell.rs
@@ -94,6 +94,12 @@ pub struct ResolvedTransaction {
     pub input_cells: Vec<CellStatus>,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum UnresolvableError {
+    Dead,
+    Unknown,
+}
+
 pub trait CellProvider {
     fn cell(&self, out_point: &OutPoint) -> CellStatus;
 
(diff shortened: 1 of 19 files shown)


spec/src/lib.rs
```diff
@@ -290,9 +290,9 @@ pub mod test {
             "55a809b92c5c404989bfe523639a741f4368ecaa3d4c42d1eb8854445b1b798b"
         );
 
-        assert!(
-            chain_spec.to_consensus().is_ok(),
-            format!("{:?}", chain_spec.to_consensus())
-        );
+        // assert!(
```
quake: Current develop branch failed to pass CI here, I will rebase this assert later.

core/src/cell.rs
```diff
@@ -70,39 +70,67 @@ impl CellStatus {
...
+    Unknown(OutPoint),
 }
 
 pub trait CellProvider {
     fn cell(&self, out_point: &OutPoint) -> CellStatus;
 
-    fn get_cell_status(&self, out_point: &OutPoint) -> CellStatus {
-        if out_point.is_null() {
-            CellStatus::Live(LiveCell::Null)
+    fn resolve_transaction<'a>(
```
xxuejie: Please keep `resolve_transaction` as a separate function outside of this. The problem is that in order to generalize OutPoint, we will add a separate HeaderProvider to `resolve_transaction` so it can also resolve headers. Putting `resolve_transaction` here makes it hard to deal with HeaderProvider, and we might still need to extract `resolve_transaction` out of CellProvider.

quake: We can add a helper fn for `HeaderProvider` if needed:

```rust
fn resolve_transaction(tx: &Transaction, cp: &CellProvider, hp: &HeaderProvider) -> ResolvedTransaction {
    cp.resolve_transaction(tx).and_then(|hp| hp.resolve_transaction(tx))
}
```

Or a `TransactionResolver` trait/struct like this:

```rust
struct TransactionResolver {
    cp: &CellProvider,
    hp: &HeaderProvider,
}

impl TransactionResolver {
    fn resolve(tx: &Transaction) {
        
    }
}
```

xxuejie: Personally I don't see how this is better than a plain function and the need to put `resolve_transaction` function within CellProvider trait, but it's not a big deal here, I don't have strong feelings.

quake: Before this PR, `resolve_transaction` fn needs a mutable `seen_inputs` map as the parameter, this PR removes it, and all the codes are related to `CellProvider` inner logic now, so I encapsulate them to `CellProvider`.

doitian: `BlockCellProvider` is highly coupled with the block verification logic in `ChainService::reconcile_main_chain`. This is really a problem. `BlockCellProvider` is not a cell set status after applied the transactions in Block, instead, it is partially applied. And the partially applied status just happens to be working in the block verification.

doitian: At least, `BlockCellProvider` should be renamed, and it should reside inside crate `chain`. It does not look a struct can be shared, it is designed dedicated to the block verification. So does `TransactionCellProvider`.

quake:
as discussed, I will split this PR into the small one and revert `Move resolve_transaction fn to CellProvider and avoid some clone`, please hold this.

quake:
Replaced by #576 and #592

PR #569, opened 2019-04-26
feat: check if genesis hash in config file was same as 0th block hash in db

yangby-cryptape:
At startup:
- Read genesis hash from configuration files.
- Try to fetch tip header from database.
- (new) If tip header is existed, then fetch the 0th block hash from database.
- (new) Check if the 0th block hash was same as the genesis hash.

Replace several `expect(..)` by `Result`.
Add a lot of `unwrap()` in tests.

### DIFF
--- a/benches/benches/process_block.rs
+++ b/benches/benches/process_block.rs
@@ -198,7 +198,8 @@ fn new_chain(
             options: None,
         })
         .consensus(consensus)
-        .build();
+        .build()
+        .unwrap();
     let notify = NotifyService::default().start::<&str>(None);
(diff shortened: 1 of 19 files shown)


u2:
Why do we need to check the 0th block hash was the same as the genesis?

yangby-cryptape:
> Why do we need to check the 0th block hash was the same as the genesis?

If someone updates settings about genesis after he/she ran the program for a while, then he/she restarts the program, the program should give he/she a warning message and shutdown.

It often happens during the development phase.

doitian:
It also happens when switching env, e.g., from testnet to mainnet, without
creating a new ckb directory.

PR #570, opened 2019-04-26
feat: check if the data in database is compatible with the program

yangby-cryptape:
At startup:
- If database is not existed, create it and write a version into it, then **do as normal**.
- If database is existed, check if it has a version.
  - If the database doesn't has a version, **panic with a message**.
  - If the database has a version, check if it is matched with current program.
    - If match, **do as normal**.
    - If not match, check if the database can be migrated automatically.
      - If it can be migrated, **do migration**.
      - **Panic with a message**.

### DIFF
--- a/db/src/rocksdb.rs
+++ b/db/src/rocksdb.rs
@@ -1,36 +1,67 @@
 use crate::{Col, DBConfig, DbBatch, Error, KeyValueDB, Result};
-use log::warn;
+use log::{info, warn};
 use rocksdb::{ColumnFamily, Error as RdbError, Options, WriteBatch, DB};
 use std::ops::Range;
 use std::sync::Arc;
+
+// If any data format in database was changed, we have to update this constant manually.
+//      - If the data can be migrated at startup automatically: update "x.y.z1" to "x.y.z2".
+//      - If the data can be migrated manually: update "x.y1.z" to "x.y2.0".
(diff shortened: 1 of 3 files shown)


db/src/rocksdb.rs
```diff
@@ -1,49 +1,116 @@
...
 
         let cfnames: Vec<_> = (0..columns).map(|c| c.to_string()).collect();
         let cf_options: Vec<&str> = cfnames.iter().map(|n| n as &str).collect();
-        let db = DB::open_cf(&opts, &config.path, &cf_options).unwrap_or_else(|err| {
-            if err.as_ref().starts_with("Corruption:") {
-                warn!("Try repairing the rocksdb since {} ...", err);
+
+        let db = DB::open_cf(&opts, &config.path, &cf_options).or_else(|err| {
+            let err_str = err.as_ref();
+            if err_str.starts_with("Invalid argument:")
```
doitian: It is not reliable to depend on the error string. It is easier to just check whether the key is present in the database.

yangby-cryptape: No, there are two situation:
- (passed) database is not existed
- (failed) database is existed but no version is present

doitian: It's OK to assume all the guys using the older program will clear their database.

PR #580, opened 2019-04-28
feat: max_proposals and block_size limit

doitian:


miner/src/block_assembler.rs
```diff
@@ -275,14 +284,30 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
         }
     }
 
+    fn cal_txs_size_limit(
+        &self,
+        bytes_limit: u64,
+        uncles: &[UncleBlock],
+        proposals: &[ProposalShortId],
+    ) -> usize {
+        bytes_limit as usize
```
keroro520: It would be better to check overflowing_sub

doitian:
Feature completed, need to add some tests.

zhangsoledad:
change base to https://github.com/nervosnetwork/ckb/pull/619

Issue #562, opened 2019-04-25
Write a tool to dump well known hashes into a readable file


doitian:
I will write a tool to dump various hashes into a file and commit that file into the repository.

- [ ] Add a tool to dump various hashes into a meta file and commit that file into repo.
    - https://github.com/nervosnetwork/ckb/blob/v0.9.0/spec/src/lib.rs#L231
- [ ] Add CI job to check the meta file is up-to-dated.
- [ ] Mention the file in README so the SDK team can easily traces the changes.

_Originally posted by @doitian in https://github.com/nervosnetwork/ckb/pull/561#issuecomment-486657519_

PR #582, opened 2019-04-28
feat: Verify genesis every time starting

keroro520:
* Verify genesis block by `BlockVerifier` every time starting node
* Fix capacity calculation and inputs of genesis block
* Verify occupied_capacity of celbase transactions 
* Refactor `tests/uncle_verifier.rs` and add several tests
* **BREAKING CHANGE**: Update genesis hash


spec/src/lib.rs
```diff
@@ -247,12 +256,6 @@ pub mod test {
         let chain_spec = dev.unwrap();
         let tx = chain_spec.build_system_cell_transaction().unwrap();
 
-        // Tx and Output hash will be used in some test cases directly, assert here for convenience
-        assert_eq!(
-            format!("{:x}", tx.hash()),
-            "48168c5b2460bfa698f60e67f08df5298b1d43b2da7939a219ffd863e1380d11"
-        );
-
```
keroro520: This test makes no sense.

util/app-config/src/lib.rs
```diff
@@ -217,17 +215,7 @@ fn is_daemon(subcommand_name: &str) -> bool {
...
-        ExitCode::Config
-    })
-}
-
-fn verify_genesis(consensus: Consensus) -> Result<Consensus, Box<Error>> {
-    MerkleRootVerifier::new()
-        .verify(consensus.genesis_block())
-        .map_err(Box::new)?;
-    Ok(consensus)
+    spec.to_consensus().map_err(|_| ExitCode::Config)
```
keroro520: `spec/src/lib.rs` only ensures that genesis is matched with the specs, recorded in `specs/*.toml`.

`src/subcommand/run.rs` do the real block validation (Unable to do this inside `spec/src/lib.rs` cause circular dependency of `ckb_shared` and `ckb-chain-specs`).

verification/src/block_verifier.rs
```diff
@@ -97,6 +96,16 @@ impl<CP: ChainProvider + Clone> CellbaseVerifier<CP> {
             return Err(Error::Cellbase(CellbaseError::InvalidOutput));
         }
 
+        let invalid_capacity = cellbase_transaction.outputs().iter().any(|op| {
```
quake: I think that it's better to add a function to `CellOutput`
```
pub fn is_occupied_capacity_overflow(&self) {
    
}
```

And refactoring this code block to:
```
    if cellbase_transaction.outputs().iter().any(CellOutput::is_occupied_capacity_overflow) {
        return Err(Error::CapacityOverflow);
    }
```

Similar code of `CapacityVerifier` can be refactored to use this function also.

spec/src/lib.rs
doitian: This is for developers who need to hard-coded `always_success` hash. However, it is going to be removed in #562 .

spec/src/lib.rs
```diff
@@ -168,19 +169,27 @@ impl ChainSpec {
                 c.get()
                     .map_err(|err| Box::new(err) as Box<Error>)
                     .and_then(|data| {
-                        let data = data.into_owned();
                         // TODO: we should provide a proper lock script here so system cells
                         // can be updated.
-                        Capacity::bytes(data.len())
-                            .map(|cap| CellOutput::new(cap, data, Script::default(), None))
-                            .map_err(|err| Box::new(err) as Box<Error>)
+                        let data = data.into_owned();
```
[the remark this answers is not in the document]
keroro520: Thanks! It is more clear. Applied and rebased the suggested code.

verification/src/block_verifier.rs
keroro520: Thanks. Updated and rebased.

spec/src/lib.rs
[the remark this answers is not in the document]
yangby-cryptape: IMHO, we shouldn't pass a `capacity` to `CellOutput` as an parameter, `capacity` should be computed in `CellOutput::new()`.
(Not in this PR, we can do it later.)

keroro520:
> Need update `test/integration.toml` genesis hash.

Updated.
And added monitor for child process in `test/` https://github.com/nervosnetwork/ckb/pull/582/files#diff-750b6a3c46becef8048e261f521c1d93

PR #583, opened 2019-04-28
fix: Return early for non-existent block

keroro520:
* fix: Return early for non-existent block
* feat: Limit page size of `get_cells_by_lock_hash`


rpc/src/module/chain.rs
```diff
@@ -77,28 +80,44 @@ impl<CS: ChainStore + 'static> ChainRpc for ChainRpcImpl<CS> {
         let to = to
             .parse::<BlockNumber>()
             .map_err(|_| Error::parse_error())?;
+        if from > to {
```
quake:
```suggestion
        if from >= to {
```

PR #584, opened 2019-04-29
fix: Disconnect wrong peer when process getheaders message

TheWaWaR:
Should only disconnect outbound not protected peer and the message length is less `MAX_HEADERS_LEN`


zhangsoledad:
https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L1680-L1697

PR #587, opened 2019-04-29
feat: lazy load cell output

jjyr:
* delay cell output load until the system call
* fix cachedb


core/src/cell.rs
```diff
@@ -3,20 +3,35 @@ use crate::transaction::{CellOutput, OutPoint, Transaction};
...
+}
+
+impl From<&CellOutput> for CellMeta {
+    fn from(output: &CellOutput) -> Self {
+        CellMeta {
+            cell_output: Some(output.clone()),
+            capacity: output.capacity,
+            ..Default::default()
+        }
+    }
```
TheWaWaR: Just write a `new` function, or use the builder pattern.

db/src/cachedb.rs
```diff
@@ -35,13 +112,13 @@ impl<T> KeyValueDB for CacheDB<T>
 where
     T: KeyValueDB,
 {
-    type Batch = T::Batch;
+    type Batch = CacheDBBatch<T::Batch>;
 
     fn read(&self, col: Col, key: &[u8]) -> Result<Option<Vec<u8>>> {
-        let cache_guard = self.cache.read();
+        let mut cache_guard = self.cache.lock();
```
quake: each read will hold the lock of cache and block other read operations, even it's reading the none cachable cols, we may need to change `CacheTable`
```
type CacheTable = FnvHashMap<Col, Mutex<LruCache<Vec<u8>, Vec<u8>>>>;
```

script/proptest-regressions/syscalls/mod.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
quake: I think that this file is generated by proptest in your processing of refactoring. Could you please double check whether we have a unit test of this case?

core/src/transaction.rs
```diff
@@ -123,7 +123,7 @@ impl fmt::Debug for CellOutput {
 }
 
 impl CellOutput {
-    pub fn new(capacity: Capacity, data: Vec<u8>, lock: Script, type_: Option<Script>) -> Self {
+    pub fn new(capacity: Capacity, data: Bytes, lock: Script, type_: Option<Script>) -> Self {
```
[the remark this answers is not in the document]
jjyr: We should use Bytes instead of `Vec<u8>` to avoid meaningless clone in the future, the invoker should clearly aware the convert between `D` to Bytes.

script/src/syscalls/load_cell.rs
```diff
@@ -67,6 +75,7 @@ impl<'a, Mac: SupportMachine> Syscalls<Mac> for LoadCell<'a> {
             return Ok(true);
         }
         let cell = cell.unwrap();
+        let output = self.store.lazy_load_cell_output(&cell);
```
xxuejie: This is just a minor thing so feel free to ignore it and leave it to a future PR: I noticed right now we are still using `flat_serializer` for serializing CellOutput in the store. And right now, the life of a CellOutput is:

1. When inserting the block, a CellOutput is serialized using `flat_serializer` and put into store
2. Here when we are loading the cell, we first load it and deserialize it using `flat_serializer`
3. Then the CellOutput is serialized again using flatbuffers, and VM reads the section it needs.

What if we just use flatbuffer to serialize CellOutput? Pairing it with partial read in rocksdb, we can only read the exact bytes required by CKB VM in the syscall from rocksdb, and return those directly to the VM, no serialization/deserialization steps are needed here in this way.

I cannot promise the percentage of speedup we can get by implementing this, but what I can promise is: if we can make the change here, we won't need the comments below saying load cell is an expensive operation which should be avoided, and we can revise the cycle costs of this syscall, enabling totally new paradigms in our scripts which is both easier to write and cost effective.

Thoughts?

jjyr: Totally agree, except CFB is not ready yet. In fact, we're using `bincode` for serialize and I have no idea how to achieve this with `bincode`.

janx: Sounds like a good future PR candidate.

PR #590, opened 2019-04-29
chore: enable aborting on panic

jjyr:


keroro520:
What is the purpose of enabling this ?
`panic as abort` option will lose the panic stacktraces, while I think these info may help us debug testnet panic problems.

jjyr:
> What is the purpose of enabling this ?
> `panic as abort` option will lose the panic stacktraces, while I think these info may help us debug testnet panic problems.

The intention is to crash the whole program when a panic occurs, in the current implementation,  some threads panic silence.

backtraces are important, seems we need to find another way to achieve this.

https://doc.rust-lang.org/std/panic/index.html

PR #591, opened 2019-04-29
refactor: specify different structs for JSON-RPC requests and responses

yangby-cryptape:
### Changes

Since `Transaction`, `Header`, `Block`, `UncleBlock` are different in JSON-RPC requests and
responses, we should not use same structs to present them.

**BREAKING CHANGE**: Redundant fields (only hash) in JSON-RPC requests are not allowed anymore.

### Reason

- Should prevent users to submit `hash` for `Transaction` or `Header`.

  Now, they are able to do that, and they can submit any things, even a random string.

- Users will be confused:

  What is `hash` in `Transaction`? What should I send?

  https://github.com/nervosnetwork/ckb/blob/42bd97cb93d047c537eae39456dbe9fd585d0532/util/jsonrpc-types/src/blockchain.rs#L168-L176

- The crate `jsonrpc-types` are not able to reuse for client.

  For example, if client fetch a `Header` string through JSON-RPC and deserialize it to `Header` in `jsonrpc-types`, the client will lost the `hash`.

  https://github.com/nervosnetwork/ckb/blob/42bd97cb93d047c537eae39456dbe9fd585d0532/util/jsonrpc-types/src/blockchain.rs#L335-L349


keroro520:
Could we just remove `skip_deserializing` ?

As for rpc client and server:

* When client sends transaction with `hash`, server can reuse its `hash`, or recalculate. It depends by server itself.
* When server response transaction with `hash`, client can reuse its `hash`, or verify by recalculating. It depends by client itself.

Seems trivial that skip deserializing `hash`, is it ? Waiting for answer ...... 😃

yangby-cryptape:
> * When client sends transaction with `hash`, server can reuse its `hash`, or recalculate. It depends by server itself.

Servers should never reuse `hash` from clients.

jjyr:
XXXFull cannot describe the intention of data struct, can we consider change struct name to `Block`, `BlockResponse`, `Transaction`, `TransactionResponse`?

yangby-cryptape:
> XXXFull cannot describe the intention of data struct, can we consider change struct name to `Block`, `BlockResponse`, `Transaction`, `TransactionResponse`?

I don't insist on names. But `Response` is not suitable, either.

Each of them is just a part of `Response`.

I don't want to name the Http Response of `TransactionResponse` to `TransactionResponseResponse`, and the Http Response of `Transaction` to `TransactionResponse.`

quake:
How about changing the hash to `Option`?
```
     pub hash: Option<H256>, 
```

Different server and client implementations can choose to provide this value or not, if this value is present, it will be verified with the calculated value. It's helpful for client / server debug.

yangby-cryptape:
> How about changing the hash to `Option`?

- I'm not sure, but IMHO, a provided `hash` can't help computing hash to be faster.

- **Security** I think if we don't specify them clearly, it probably makes troubles.

  If someone use the hashes from the clients (forget verifying), it will be very dangerous.

- **Confused** These hashes were not provided by users in almost all cases.

  Not only users, but also new ckb developers will be confused for that.

  In fact, up to now, the output `Transaction` has two more fields than input `Transaction`: `status` and `hash`. They are really different things. (And, maybe we have to add a third or fourth fields in future.)

  p.s. I think `TransactionWithStatus` wasn't required, if we decouple them a little earlier, we can just add `status` into `TransactionFull`.

doitian:
> **I don't insist on names.** But `Response` is not suitable, either.

How about `TransactionView`

yangby-cryptape:
> How about `TransactionView`

It's better than `XxxxFull` (the original names I posted) or `FullXxxx` (I updated this name yesterday afternoon after talked with folks in office).

I will update the names today.

PR #592, opened 2019-04-30
refactor: block processing and transaction verification

quake:
1. Returns short circuit `UnresolvableError` in chain block processing before verification
2. ResolvedTransaction contains live cell only, simplify code and remove some unnecessary verifiers

### DIFF
--- a/core/src/cell.rs
+++ b/core/src/cell.rs
@@ -181,11 +181,25 @@ impl<'a> CellProvider for BlockCellProvider<'a> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum UnresolvableError {
+    Dead(Vec<OutPoint>),
+    Unknown(Vec<OutPoint>),
+}
+
+impl std::error::Error for UnresolvableError {}
+
+impl std::fmt::Display for UnresolvableError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        std::fmt::Debug::fmt(&self, f)
+    }
+}
+
 pub fn resolve_transaction<'a, CP: CellProvider>(
     transaction: &'a Transaction,
     seen_inputs: &mut FnvHashSet<OutPoint>,
     cell_provider: &CP,
-) -> ResolvedTransaction<'a> {
+) -> Result<ResolvedTransaction<'a>, UnresolvableError> {
     let input_cells = transaction
         .input_pts()
         .iter()
@@ -210,11 +224,11 @@ pub fn resolve_transaction<'a, CP: CellProvider>(
         })
         .collect();
 
-    ResolvedTransaction {
+    Ok(ResolvedTransaction {
         transaction,
         input_cells,
         dep_cells,
-    }
+    })
 }
 
 impl<'a> ResolvedTransaction<'a> {
--- a/core/src/cell.rs
+++ b/core/src/cell.rs
@@ -3,6 +3,7 @@ use crate::transaction::{CellOutput, OutPoint, Transaction};
 use crate::Capacity;
 use fnv::{FnvHashMap, FnvHashSet};
 use numext_fixed_hash::H256;
+use serde_derive::{Deserialize, Serialize};
 
 #[derive(Clone, Eq, PartialEq, Debug, Default, Deserialize, Serialize)]
 pub struct CellMeta {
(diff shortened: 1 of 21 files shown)


core/src/cell.rs
```diff
@@ -168,65 +136,69 @@ impl<'a> CellProvider for BlockCellProvider<'a> {
...
-        self.cells_iter().all(CellStatus::is_live)
+        match cell_status {
+            CellStatus::Dead => dead_out_points.push(out_point.clone()),
+            CellStatus::Unknown => unknown_out_points.push(out_point.clone()),
+            CellStatus::Live(cell_meta) => dep_cells.push(cell_meta),
+        }
+    });
+
+    if !dead_out_points.is_empty() {
+        Err(UnresolvableError::Dead(dead_out_points))
```
jjyr: Do we really need to save all failed resolved out_points? why not return an error on the first failure?

What happened if we have both Dead and Unknown resolves?

quake: We can early return `Dead` as a short circuit error, but cannot do for `Unknown`, which is used in the tx pool:
https://github.com/nervosnetwork/ckb/blob/cc16678426b76309967f4c8880a4fc9368434741/shared/src/chain_state.rs#L242

quake: `Dead` is a higher priority error than `Unknown`, however, to make code easy to understand and debug, I choose to return all error `OutPoint` instead of first one.

verification/src/transaction_verifier.rs
```diff
@@ -262,32 +195,6 @@ impl<'a> DuplicateDepsVerifier<'a> {
     }
 }
 
-pub struct NullVerifier<'a> {
```
keroro520: Why remove `NullVerifier` ?

core/src/cell.rs
jjyr: IMO short-cut return Dead error is more efficient and clear.

verification/src/transaction_verifier.rs
quake: All resolved transactions contain live cell only now, so `NullVerifier` is unnecessary

core/src/cell.rs
janx: I think it's better to return error asap in `resolve_transaction`.

verification/src/transaction_verifier.rs
```diff
@@ -40,13 +40,10 @@ where
 
 pub struct TransactionVerifier<'a, M> {
     pub version: VersionVerifier<'a>,
-    pub null: NullVerifier<'a>,
     pub empty: EmptyVerifier<'a>,
     pub maturity: MaturityVerifier<'a>,
     pub capacity: CapacityVerifier<'a>,
-    pub duplicate_inputs: DuplicateInputsVerifier<'a>,
     pub duplicate_deps: DuplicateDepsVerifier<'a>,
```
jjyr: Should also remove this?

shared/src/lib.rs
```diff
@@ -16,8 +16,11 @@ pub mod store;
 pub mod tx_pool;
 mod tx_proposal_table;
 
-#[cfg(test)]
-mod tests;
+// These tests are from testnet data dump, they are hard to maintenance.
+// Although they pass curruent unit tests, we still comment out here,
+// Keep the code for future unit test refactoring reference.
+// #[cfg(test)]
+// mod tests;
```
jjyr: I think we still need these tests. some problem is hard to reproduce with unit tests, we need to maintain testnet/mainnet data that once broken network.

verification/src/transaction_verifier.rs
quake: `DuplicateDepsVerifier` is needed here, because `resolve_transaction` fn only checks the duplicate inputs with the `seen_inputs` set.

verification/src/error.rs
```diff
@@ -135,20 +135,15 @@ pub enum DifficultyError {
 
 #[derive(Debug, PartialEq, Clone, Copy, Eq)]
 pub enum TransactionError {
-    NullInput,
-    NullDep,
     /// Occur output's bytes_len exceed capacity
     CapacityOverflow,
-    DuplicateInputs,
     DuplicateDeps,
```
jjyr: Why remove the DuplicateInputs but leave the DuplicateDeps?

jjyr:
LGTM, this PR change the tx validation, PLS wait for another reviewer double check.

PR #593, opened 2019-04-30
[ᚬmaster] sync: improve the performance of get_locator

lerencao:
Use the last header as the `base` to avoid scan from tip.


lerencao:
Close this to stay on branch develop

PR #596, opened 2019-04-30
sync: improve the performance of get_locator

lerencao:
use the last header as the `base` to avoid scan from tip.


TheWaWaR:
`sync` is not a valid commit prefix: https://github.com/commitizen/cz-cli

lerencao:
@TheWaWaR thx

Updated and  PTAL.

PR #614, opened 2019-05-04
feat: verify compact block

keroro520:
* Implement `CompactBlockVerifier` to verify the validity of CompactBlock, which be passed to `Relayer::reconstruct_block` function. The correctness of reconstructed block will be verified by `BlockVerifier` and `TransactionsVerifier`.
* test: Add unit tests for `Relayer::reconstruct_block` and `CompactBlockVerifier`
* perf: Use `HashSet::contains` to check whether a short id be contained in staging pool. It will decrease the time-complexity.

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -95,7 +95,10 @@ impl<'a, CS: ChainStore> CompactBlockProcess<'a, CS> {
                     Arc::clone(&self.relayer.shared.consensus().pow_engine()),
                 );
                 let compact_block_verifier = CompactBlockVerifier::new();
-                header_verifier.verify(&resolver)?;
+                if let Err(err) = header_verifier.verify(&resolver) {
+                    debug!(target: "relay", "unexpected header verify failed: {}", err);
+                    return Ok(());
+                }
                 compact_block_verifier.verify(&compact_block)?;
             }
 
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -1,4 +1,5 @@
-use super::compact_block::CompactBlock;
+use crate::relayer::compact_block::CompactBlock;
+use crate::relayer::compact_block_verifier::CompactBlockVerifier;
 use crate::relayer::Relayer;
 use ckb_core::{header::Header, BlockNumber};
 use ckb_network::{CKBProtocolContext, PeerIndex};
@@ -58,11 +59,10 @@ impl<'a, CS: ChainStore> CompactBlockProcess<'a, CS> {
                 );
                 return Ok(());
             }
-        } else {
+        } else if self.relayer.shared.is_initial_block_download() {
             // If self is in the IBD state, do nothing
-            if self.relayer.shared.is_initial_block_download() {
-                return Ok(());
-            }
+            return Ok(());
+        } else {
             debug!(target: "relay", "UnknownParent: {}, send_getheaders_to_peer({})", block_hash, self.peer);
             self.relayer.shared.send_getheaders_to_peer(
                 self.nc,
@@ -72,12 +72,16 @@ impl<'a, CS: ChainStore> CompactBlockProcess<'a, CS> {
             return Ok(());
         }
 
+        // The new arrived has greater difficulty than local best known chain
         let mut missing_indexes: Vec<usize> = Vec::new();
         {
+            // Verify compact block
             let mut pending_compact_blocks = self.relayer.state.pending_compact_blocks.lock();
-            if pending_compact_blocks.get(&block_hash).is_none()
-                && self.relayer.shared.get_block(&block_hash).is_none()
+            if pending_compact_blocks.get(&block_hash).is_some()
+                || self.relayer.shared.get_block(&block_hash).is_some()
             {
+                debug!(target: "relay", "already processed compact block {}", block_hash);
+            } else {
                 let resolver = HeaderResolverWrapper::new(
                     &compact_block.header,
                     self.relayer.shared.shared().to_owned(),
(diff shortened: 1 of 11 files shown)


sync/src/relayer/compact_block_verifier.rs
```diff
@@ -0,0 +1,96 @@
+use crate::relayer::compact_block::CompactBlock;
+use crate::relayer::error::Error;
+use ckb_protocol::{short_transaction_id, short_transaction_id_keys};
+use std::collections::HashSet;
+
+// TODO move `CompactBlock` into core types, and `CompactBlockVerifier`
```
zhangsoledad: `CompactBlock` is not chain data, should never move to core

sync/src/relayer/compact_block_verifier.rs
```diff
@@ -0,0 +1,94 @@
...
+        Self {}
+    }
+
+    pub(crate) fn verify(&self, block: &CompactBlock) -> Result<(), Error> {
+        let prefilled_transactions = &block.prefilled_transactions;
+        let short_ids = &block.short_ids;
+        let txs_len = prefilled_transactions.len() + short_ids.len();
+
+        // Check indices order of prefilled transactions
+        let mut prev = prefilled_transactions
```
quake: We can use `windows` here
```
prefilled_transactions.iter().windows(2).any(|pt| pt[0].index > pt[1].index)
```

sync/src/relayer/compact_block_verifier.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl ShortIdsVerifier {
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+
+    pub(crate) fn verify(&self, block: &CompactBlock) -> Result<(), Error> {
+        let prefilled_transactions = &block.prefilled_transactions;
+        let short_ids = &block.short_ids;
+        let short_ids_set: HashSet<[u8; 6]> = short_ids.iter().map(Clone::clone).collect();
```
quake: Can we use `HashSet<&[u8; 6]>` here to remove clone?

sync/src/relayer/mod.rs
```diff
@@ -233,26 +239,34 @@ impl<CS: ChainStore> Relayer<CS> {
     ) -> Result<Block, Vec<usize>> {
         let (key0, key1) =
             short_transaction_id_keys(compact_block.header.nonce(), compact_block.nonce);
+        let mut short_ids_set: HashSet<[u8; 6]> =
+            compact_block.short_ids.iter().map(Clone::clone).collect();
```
quake: Same `clone` issue here

sync/src/relayer/compact_block_verifier.rs
keroro520: Thanks. Change to use `tuple_window` instead.

sync/src/relayer/compact_block_verifier.rs
keroro520: Changed to `HashSet<&[u8; 6]>`

sync/src/relayer/compact_block_verifier.rs
quake: `tuple_windows` will clone the iterator element if I remember correctly, and I prefer the std collection than itertools crate.

sync/src/relayer/compact_block_verifier.rs
```diff
@@ -0,0 +1,94 @@
...
+
+impl ShortIdsVerifier {
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+
+    pub(crate) fn verify(&self, block: &CompactBlock) -> Result<(), Error> {
+        let prefilled_transactions = &block.prefilled_transactions;
+        let short_ids = &block.short_ids;
+        let short_ids_set: HashSet<&[u8; 6]> = short_ids.iter().collect();
```
quake: Use type alias:
```suggestion
        let short_ids_set: HashSet<&ShortTransactionID> = short_ids.iter().collect();
```

sync/src/relayer/mod.rs
```diff
@@ -233,26 +239,33 @@ impl<CS: ChainStore> Relayer<CS> {
     ) -> Result<Block, Vec<usize>> {
         let (key0, key1) =
             short_transaction_id_keys(compact_block.header.nonce(), compact_block.nonce);
+        let mut short_ids_set: HashSet<&[u8; 6]> = compact_block.short_ids.iter().collect();
```
quake:
```suggestion
        let mut short_ids_set: HashSet<&ShortTransactionID> = compact_block.short_ids.iter().collect();
```

sync/src/relayer/compact_block_verifier.rs
keroro520: Oh, `windows` is only implemented for `Slice`, but not `Iterator`.
Changed to use `Slice::windows` instead.

```
        let unordered = prefilled_transactions
            .as_slice()
            .windows(2)
            .any(|pt| pt[0].index >= pt[1].index);
```

sync/src/relayer/compact_block_process.rs
```diff
@@ -90,37 +94,31 @@ impl<'a, CS: ChainStore> CompactBlockProcess<'a, CS> {
...
-                                pending_compact_blocks
-                                    .insert(block_hash.clone(), compact_block.clone());
-                            }
-                        }
-                    }
-                    Err(err) => {
-                        debug!(target: "relay", "unexpected header verify failed: {}", err);
-                    }
+                let compact_block_verifier = CompactBlockVerifier::new();
+                header_verifier.verify(&resolver)?;
```
TheWaWaR: Return error here will ban this peer, but verify failed happens frequently (UnknownParent error for example)

keroro520: It's my fault.

Errors for invalid block format should be banned, 
some errors like `UnknownParent` should be warned.

I will change here to log errors without returning outside. And refactoring score system later is required, so that we can handle errors gracefully.

doitian: Yes, we need to dispatch different errors to different punishment. Maybe add a trait Punish on the errors.

keroro520: Updated.

But error thrown from `CompactBlockVerifier` is still return outside. The reason is that this kind of errors must be invalid format error, we can ban it.

PR #619, opened 2019-05-04
Block size limit

zhangsoledad:


spec/src/consensus.rs
```diff
@@ -21,6 +21,7 @@ pub(crate) const POW_SPACING: u64 = 15 * 1000; //15s
 
 pub(crate) const MAX_BLOCK_CYCLES: Cycle = 100_000_000;
 pub(crate) const MAX_BLOCK_BYTES: u64 = 10_000_000; // 10mb
+pub(crate) const MAX_BLOCK_PROPOSALS_LIMIT: u64 = 5_000;
```
jjyr: Change the limit to 100_000

PR #620, opened 2019-05-04
feat: Generalize OutPoint struct to reference headers as well

xxuejie:
TODO:

* [x] tests

This is still lacking tests, but since the PR is already quite big, I'm sending it out first so we can start the review process earlier.


core/src/cell.rs
```diff
@@ -69,14 +72,92 @@ impl CellStatus {
...
+        self == &HeaderStatus::Unknown
+    }
+
+    pub fn is_unspecified(&self) -> bool {
+        self == &HeaderStatus::Unspecified
+    }
+}
+
+#[derive(Debug)]
+pub struct ResolvedOutPoint {
```
doitian: Both `Header` and `CellMeta` are big in memory, especially `Header`. Considering most `OutPoint` will resolve to cell only, an `Option<Header>` just wastes abundant memory. We need to consider whether using `Option<Box<Header>>` here.

core/src/cell.rs
```diff
@@ -237,7 +455,7 @@ impl<'a> ResolvedTransaction<'a> {
     pub fn inputs_capacity(&self) -> ::occupied_capacity::Result<Capacity> {
         self.input_cells
             .iter()
-            .map(CellMeta::capacity)
+            .map(|o| o.cell.as_ref().map_or(Capacity::zero(), CellMeta::capacity))
```
doitian:
```suggestion
            .map(|o| o.cell.as_ref().map_or_else(Capacity::zero, CellMeta::capacity))
```

doitian:
lgtm

core/src/cell.rs
```diff
@@ -69,14 +72,92 @@ impl CellStatus {
...
+    }
 }
 
 /// Transaction with resolved input cells.
 #[derive(Debug)]
 pub struct ResolvedTransaction<'a> {
     pub transaction: &'a Transaction,
-    pub dep_cells: Vec<CellMeta>,
-    pub input_cells: Vec<CellMeta>,
+    pub dep_cells: Vec<ResolvedOutPoint>,
```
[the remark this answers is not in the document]
xxuejie: I've renamed `dep_cells` to `resolved_deps`, `input_cells` to `resolved_inputs`

PR #622, opened 2019-05-04
feat: cellbase type script

quake:
This PR allows the miner to set cellbase's output type, and cellbase will be verified with the same rule as a normal transaction.

### DIFF
--- a/verification/src/block_verifier.rs
+++ b/verification/src/block_verifier.rs
@@ -4,7 +4,7 @@ use crate::{TransactionVerifier, Verifier};
 use ckb_core::cell::ResolvedTransaction;
 use ckb_core::extras::EpochExt;
 use ckb_core::header::Header;
-use ckb_core::transaction::{Capacity, CellInput, CellOutput, Transaction};
+use ckb_core::transaction::{Capacity, CellInput, Transaction};
 use ckb_core::Cycle;
 use ckb_core::{block::Block, BlockNumber};
 use ckb_store::ChainStore;
@@ -97,23 +97,6 @@ impl CellbaseVerifier {
         if cellbase_input != &CellInput::new_cellbase_input(block.header().number()) {
             return Err(Error::Cellbase(CellbaseError::InvalidInput));
         }
-
-        // currently, we enforce`type` field of a cellbase output cell must be absent
-        if cellbase_transaction
-            .outputs()
-            .iter()
-            .any(|op| op.type_.is_some())
-        {
-            return Err(Error::Cellbase(CellbaseError::InvalidOutput));
-        }
-
-        if cellbase_transaction
-            .outputs()
-            .iter()
-            .any(CellOutput::is_occupied_capacity_overflow)
-        {
-            return Err(Error::CapacityOverflow);
-        };
 
         Ok(())
     }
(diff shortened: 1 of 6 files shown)


verification/src/transaction_verifier.rs
```diff
@@ -211,26 +211,31 @@ impl<'a> CapacityVerifier<'a> {
...
-            .outputs()
-            .iter()
-            .try_fold(Capacity::zero(), |acc, output| {
-                acc.safe_add(output.capacity)
-            })?;
-
-        if inputs_total < outputs_total {
-            return Err(TransactionError::OutputsSumOverflow);
+        // skip OutputsSumOverflow verification for resolved cellbase
+        // cellbase's outputs are verified by TransactionsVerifier#InvalidReward
```
doitian: I think it's better to use `Transaction::is_cellbase` here.

quake: Resolved, please review again.

verification/src/tests/block_verifier.rs
```diff
@@ -127,22 +127,6 @@ pub fn test_cellbase_with_fee() {
...
-        .input(CellInput::new_cellbase_input(0))
-        .output(CellOutput::new(
-            capacity_bytes!(5),
-            vec![1, 2, 3, 4, 5, 6, 7, 8, 9].into(),
-            Script::default(),
-            None,
-        ))
-        .build();
-    let block = BlockBuilder::default().transaction(cellbase).build();
-    let verifier = CellbaseVerifier::new();
```
doitian: We can keep the test by using `TransactionsVerifier` here.

quake: @doitian it's tested in
https://github.com/nervosnetwork/ckb/blob/584952d42e8d13e21cb15bc1eb166fa11224c7e6/verification/src/tests/transaction_verifier.rs#L21

quake:
> Need to support the changes in #627
> 
> The cellbase used cycles should be excluded when getting transactions from pool.

Resolved, thanks.

verification/src/block_verifier.rs
```diff
@@ -407,7 +389,8 @@ impl TransactionsVerifier {
             })
             .collect::<Result<Vec<_>, _>>()?;
 
-        let sum: Cycle = cycles_set.iter().sum();
+        // excludes cellbase cycle
+        let sum: Cycle = cycles_set.iter().skip(1).sum();
```
doitian: I think we need to exclude the cycles of cellbase in block assembler. For example, the limit is 1000, and cellbase has used 200, then the block assembler can only fetch transactions from pool up to 800 cycles.

quake: I don't know how to implement this, because we need to fetch transactions with cycle limit before creating a cellbase:
https://github.com/nervosnetwork/ckb/blob/960f0c9059ab579d6e5ca3becd762fb278968cb4/miner/src/block_assembler.rs#L361

doitian: Let's revert it first, include cellbase in the limit. Add a comment in `BlockAssembler::get_block_template` that

```
// It is assumed that cellbase transaction consumes 0 cycles, so it is not excluded when getting transactions from pool.
```
```

doitian:
Please resolve conflicts

quake:
> Please resolve conflicts

rebased and squashed

PR #633, opened 2019-05-06
feat: remove cycles config from miner

zhangsoledad:


doitian:
CI passed, GitHub not notified

PR #647, opened 2019-05-07
ci: [only integration] control CI checks by the last commit message

yangby-cryptape:
### Why Need This

For example, #541 and #646 don't have to run unit tests, and both of them should have to run integration tests.

Now, each reviewers have to run integration tests in his own laptop (not in CI) for reviewing (or approving).

### Solution

Use the last commit message to control CI checks.

#### Format

- Only one tag is allowed for **only one** commit message.

  There are only few tags at present:

  - `[only integration]`: only run integration tests

  - `[skip tests]`: skip all tests

  - `[all tests]`: run all tests

- The tag should be written **at start** of the commit message, so reviewers can see it clearly.

  Examples:

  - `ci: [only integration] control CI checks by the last commit message`

### DIFF
--- a/devtools/ci/script.sh
+++ b/devtools/ci/script.sh
@@ -8,13 +8,27 @@ RUN_TEST=false
 # Run integration only in master, develop and rc branches
 RUN_INTEGRATION=false
 if [ "$TRAVIS_PULL_REQUEST" != false ]; then
-  RUN_TEST=true
+  LAST_COMMIT_MSG="$(git log --max-count 1 --skip 1 --format="%s")"
+  echo "Last commit message is \"${LAST_COMMIT_MSG}\""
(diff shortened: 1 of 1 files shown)


doitian:
It seems the script does not work.

https://travis-ci.com/nervosnetwork/ckb/jobs/198116658

doitian:
FYI
```
bash version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin17)
Copyright (C) 2007 Free Software Foundation, Inc.
```

yangby-cryptape:
> It seems the script does not work.

Yep, so I add the `hold` label.

Because the last commit is `Merge xxx into xxx` for PR (created by Travis).

I will force-push after tests.

PR #650, opened 2019-05-07
chore: export build_tx

rink1969:
ckb-fake-env need this function, but it's private


script/src/lib.rs
```diff
@@ -6,6 +6,7 @@ mod verify;
 use ckb_vm::Error as VMInternalError;
 
 pub use crate::verify::TransactionScriptsVerifier;
+pub use crate::syscalls::build_tx;
```
doitian: consecutive `use` should be sorted

PR #651, opened 2019-05-07
feat: Add syscall to load current script hash

xxuejie:
All the other methods to load current cell's information, such as
capacity, data, etc. are also removed in this commit. This way we can
enable group based transaction validation which saves cycles for
duplicate scripts.

TODO:

* [x] RFC: https://github.com/nervosnetwork/rfcs/pull/112


doitian:
lgtm

PR #655, opened 2019-05-07
fix: Use the String alias type EpochNumber

ashchan:
Hide the real `String` type just as the way we use `BlockNumber`.

### DIFF
--- a/core/src/extras.rs
+++ b/core/src/extras.rs
@@ -1,4 +1,4 @@
-use crate::{BlockNumber, Capacity};
+use crate::{EpochNumber, BlockNumber, Capacity};
 use failure::Error as FailureError;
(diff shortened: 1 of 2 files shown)


core/src/extras.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{BlockNumber, Capacity};
+use crate::{EpochNumber, BlockNumber, Capacity};
```
yangby-cryptape: Items in `use` should be sorted.
(If you have `rust toolchain`, you can execute `cargo fmt -- --emit files` to fix this.)

ashchan: Holly I have thought about that but thought we would have a precommit hook. Updating.

PR #656, opened 2019-05-07
feat: add rpc get_epoch_by_number

keroro520:
* add rpc `get_epoch_by_number`

### DIFF
--- a/rpc/src/module/chain.rs
+++ b/rpc/src/module/chain.rs
@@ -1,6 +1,6 @@
 use crate::error::RPCError;
 use ckb_core::cell::{CellProvider, CellStatus};
-use ckb_core::{transaction::ProposalShortId, BlockNumber};
+use ckb_core::{transaction::ProposalShortId, BlockNumber, EpochNumber};
 use ckb_shared::shared::Shared;
 use ckb_store::ChainStore;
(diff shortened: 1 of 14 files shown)


rpc/src/module/chain.rs
```diff
@@ -118,6 +121,17 @@ impl<CS: ChainStore + 'static> ChainRpc for ChainRpcImpl<CS> {
...
+        Ok(self
+            .shared
+            .block_hash(
+                number
+                    .parse::<BlockNumber>()
+                    .map_err(|_| Error::parse_error())?,
+            )
+            .and_then(|hash| self.shared.get_epoch_ext(&hash).map(Into::into)))
+    }
+
```
zhangsoledad: It's incorrect implementation.
Currently, there is no ready-made epoch_number-> epoch_ext index.
First step, add the index, then leverage it in RPC.

keroro520: I've made fixes accordingly:

* Create database entry `epoch_number => epoch_index`
* refactor `ChainStore::get_epoch_ext` to receive epoch index as parameter instead
* add `ChainStore::get_epoch_index`
* add `ChainStore::get_block_epoch_index`

PTAL

Issue #549, opened 2019-04-24
Memorize hash


zhangsoledad:
- [x] Cache hashes in the source structs (`Header` and `Transaction`) since they are constructed.
- [x] Do not calculate Header hash when fetching header from database.
- [x] Optimize `ProposalShortId::hash()`.
- [x] Save hashes related to `Transaction` into database.
      So, we can just fetch them from database instead of compute them.

yangby-cryptape:
### Summary

I think, only `Header` and `Transaction` are worth memorizing (I did).

All others can be found be a shell command:

```console
yangby@laptop:ckb> grep -r 'fn [a-z_]*hash(' [^t]* | grep "[-]> [^&]*H256"
core/src/header.rs:    pub fn pow_hash(&self) -> H256 {
core/src/header.rs:    pub fn pow_hash(&self) -> H256 {
core/src/script.rs:    pub fn hash(&self) -> H256 {
core/src/block.rs:    pub fn cal_uncles_hash(&self) -> H256 {
core/src/uncle.rs:pub fn uncles_hash(uncles: &[UncleBlock]) -> H256 {
core/src/transaction.rs:    pub fn data_hash(&self) -> H256 {
core/src/transaction.rs:    pub fn hash(&self) -> H256 {
rpc/src/module/chain.rs:    fn get_block_hash(&self, _number: String) -> Result<Option<H256>>;
rpc/src/module/chain.rs:    fn get_block_hash(&self, number: String) -> Result<Option<H256>> {
shared/src/shared.rs:    fn block_hash(&self, number: BlockNumber) -> Option<H256> {
store/src/store.rs:    fn get_block_hash(&self, number: BlockNumber) -> Option<H256>;
store/src/store.rs:    fn get_block_hash(&self, number: BlockNumber) -> Option<H256> {
sync/src/types.rs:    pub fn block_hash(&self, number: BlockNumber) -> Option<H256> {
verification/src/tests/dummy.rs:    fn block_hash(&self, _height: u64) -> Option<H256> {
```

### Further

I think we can optimize the hash of `ProposalShortId`.

https://github.com/nervosnetwork/ckb/blob/de1cfc9af1fe9df3a7c24c08183523a8e6d2d779/core/src/transaction.rs#L408-L409
https://github.com/nervosnetwork/ckb/blob/de1cfc9af1fe9df3a7c24c08183523a8e6d2d779/core/src/transaction.rs#L457-L459

My suggest is:

```rust
[...]

PR #657, opened 2019-05-07
perf: remove ProposalShortId hash and Proposals root

yangby-cryptape:
### BREAKING CHANGE

Since the format of Header is changed, use proposals hash to replace proposals root:
- JSON-RPC protocols which are related to ProposalShortId are changed.
- Header in database is changed.
- Genesis Hash is changed.

### Associated Issue

#549


doitian:
It also breaks database.

core/src/block.rs
```diff
@@ -32,6 +31,20 @@ fn cal_witnesses_root(vec: &[Transaction]) -> H256 {
...
+    if vec.is_empty() {
+        H256::zero()
+    } else {
+        let mut bytes = Vec::with_capacity(vec.len() * 10);
+        for id in vec.iter() {
+            bytes
+                .write_all(&(&id as &[u8; 10])[..])
+                .expect(VEC_WRITE_ALL_EXPECT);
+        }
+        blake2b_256(bytes).into()
```
doitian: It's better to use `Blake2b::update` on the object created by `hash::new_blake2b` instead of constructing the vector first then hashing it.

yangby-cryptape: I rebase from the latest develop branch (to sync code of `Script::hash()`), replace two `blake2b_256` to `new_blake2b`, then force-push.

yangby-cryptape:
cc @ashchan @classicalliu

PR #658, opened 2019-05-07
✋ chore: script to run CI locally

doitian:


doitian:
Travis uses checks API to submit the result, which is only available as a GitHub App. So this script does not work, and there's no way to override the Travis CI status via personal API.

PR #671, opened 2019-05-08
feat: rpc get blockchain info

keroro520:
* Add rpc `get_blockchain_info`
* Add rpc `get_peers_state`, currently only return the info of blocks synchronizing in flight.

### DIFF
--- a/util/jsonrpc-types/src/chain_info.rs
+++ b/util/jsonrpc-types/src/chain_info.rs
@@ -1,0 +1,19 @@
+use crate::EpochNumber;
+use numext_fixed_uint::U256;
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Debug)]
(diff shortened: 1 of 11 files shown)


util/jsonrpc-types/src/chain_info.rs
```diff
@@ -0,0 +1,18 @@
...
+use serde_derive::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct ChainInfo {
+    // network name
+    pub chain: String,
+    // median time for the current tip block
+    pub median_time: String,
+    // the current epoch number
+    pub epoch: String,
```
yangby-cryptape: https://github.com/nervosnetwork/ckb/blob/f86214b111f3bf7bf716250cd38aa0dbc607a5bf/util/jsonrpc-types/src/lib.rs#L13

PR #672, opened 2019-05-08
style: use more handsome constructors for hashes

yangby-cryptape:
Remove a useless test. Initially, some const hashes are constructed by `[u8; N]`, so there need some tests to ensure correctness. Now, they are constructed in the same way as the tests, so the tests are redundant.

### DIFF
--- a/util/crypto/src/secp/signature.rs
+++ b/util/crypto/src/secp/signature.rs
@@ -147,19 +147,3 @@ impl FromStr for Signature {
             .map_err(|_| Error::InvalidSignature)
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    #[test]
-    fn test_n() {
-        let half: H256 =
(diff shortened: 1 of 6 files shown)


util/crypto/src/secp/signature.rs
```diff
@@ -147,19 +147,3 @@ impl FromStr for Signature {
...
-    fn test_n() {
-        let half: H256 =
-            H256::from_hex_str("7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF5D576E7357A4501DDFE92F46681B20A0")
-                .unwrap();
-        let n: H256 =
-            H256::from_hex_str("fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141")
-                .unwrap();
-        assert_eq!(half, HALF_N);
-        assert_eq!(n, N);
-    }
```
yangby-cryptape: https://github.com/nervosnetwork/ckb/blob/f86214b111f3bf7bf716250cd38aa0dbc607a5bf/util/crypto/src/secp/signature.rs#L16-L19

PR #688, opened 2019-05-09
fix: initial block download blocked

TheWaWaR:


sync/src/synchronizer/mod.rs
```diff
@@ -411,12 +411,20 @@ impl<CS: ChainStore> Synchronizer<CS> {
...
     fn start_sync_headers(&self, nc: &CKBProtocolContext) {
+        let now = unix_time_as_millis();
         let peers: Vec<PeerIndex> = self
             .peers
             .state
             .read()
             .iter()
-            .filter(|(_, state)| !state.sync_started)
+            .filter(|(_, state)| {
+                !state.sync_started
```
jjyr: should use a function to detect sync_status

PR #690, opened 2019-05-09
feat: Use script to generate rpc doc

keroro520:
* Add a rpc description file in json format
* Add a script to generate README.md based on rpc description file below
* Use json to test rpc. Several cases would be skipped:
  - `get_peers_state`
  - `local_node_info`
  - `get_peers`


rpc/Cargo.toml
```diff
@@ -37,6 +37,11 @@ ckb-traits = { path = "../traits" }
 ckb-util = { path = "../util" }
 faketime = "0.2.0"
 sentry = "^0.15.4"
+jsonrpc-client-core = "0.5.0"
```
quake: Is it only be used in test code, can we move it to `dev-dependencies` section?

keroro520: Got! It's useless dependencies, which I introduced when test. I will remove it.

PR #695, opened 2019-05-10
fix: metric transaction header mem size

zhangsoledad:


core/src/header.rs
```diff
@@ -99,7 +99,7 @@ impl RawHeader {
         mem::size_of::<Version>()
             + H256::size_of() * 5
             + U256::size_of()
-            + mem::size_of::<u64>() * 2
+            + mem::size_of::<u64>() * 3
             + mem::size_of::<u32>()
```
zhangsoledad: epoch number

core/src/transaction.rs
```diff
@@ -539,7 +539,11 @@ impl Transaction {
                 .iter()
                 .map(CellOutput::serialized_size)
                 .sum::<usize>()
-            + self.witnesses.iter().map(Vec::len).sum::<usize>()
+            + self
+                .witnesses
+                .iter()
+                .map(|witness| witness.iter().map(Vec::len).sum::<usize>())
```
quake:
```suggestion
                .flat_map(|witness| witness.iter().map(Vec::len))
```

doitian:
@nervos-bot ci-status 97f1d0c3ad5a1d21be92d65a6baa1bf243d51ae1
CI: success

PR #704, opened 2019-05-11
fix: fix default config code hash

xxuejie:


quake:
Will merge with #687 
https://github.com/nervosnetwork/ckb/pull/687/commits/7c92d300a9fc6b4d3c61e274cc87f4b8c3743818

PR #708, opened 2019-05-11
chore: use symlink for integration test

quake:
reduce integration test compiling time and disk usage


doitian:
I'll update locally

PR #727, opened 2019-05-12
chore: add verification level to config

quake:
This PR adds verification level to the config file, some integration test needs this feature.

### DIFF
--- a/chain/src/tests/util.rs
+++ b/chain/src/tests/util.rs
@@ -1,4 +1,4 @@
-use crate::chain::{ChainBuilder, ChainController};
+use crate::chain::{ChainBuilder, ChainController, VerificationLevel};
 use ckb_chain_spec::consensus::Consensus;
 use ckb_core::block::Block;
 use ckb_core::block::BlockBuilder;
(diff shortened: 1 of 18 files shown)


chain/src/tests/util.rs
```diff
@@ -48,7 +48,11 @@ pub(crate) fn start_chain(
 
     let notify = NotifyService::default().start::<&str>(None);
     let chain_service = ChainBuilder::new(shared.clone(), notify)
-        .verification(verification)
+        .verification_level(if verification {
+            VerificationLevel::Full
+        } else {
+            VerificationLevel::None
+        })
```
keroro520: Why not change `verification: bool` to `CheckLevel` ?

src/subcommand/run.rs
```diff
@@ -34,7 +34,11 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
 
     let notify = NotifyService::default().start(Some("notify"));
 
-    let chain_controller = setup_chain(shared.clone(), notify.clone());
+    let chain_controller = setup_chain(
+        shared.clone(),
+        notify.clone(),
+        args.config.verification_level,
```
keroro520: FYI. I prefer to pass `args.config` to `setup_chain` instead.

util/app-config/src/app_config.rs
```diff
@@ -38,6 +39,7 @@ pub enum AppConfigContent {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct CKBAppConfig {
     pub data_dir: PathBuf,
+    pub verification_level: Option<VerificationLevel>,
```
keroro520: Move it into `ChainConfig` ?

quake:
> It's not a good idea to make verification level configurable. We should find other ways to make writing integration tests more convenient.

New PR: https://github.com/nervosnetwork/ckb/pull/732

PR #732, opened 2019-05-13
refactor: move `verification` field from ChainService struct to `process_block` fn params

quake:
This PR moves`verification` field from ChainService struct to `process_block` fn params, it provides fine control over the block processing verification, it's a required feature for integration test and other unimplemented features, e.g. bitcoins's assumevalid / minimumchainwork / checkpoints etc.

Please note that `ChainBuilder` is removed due to the redundancy function of `ChainService`, and  compact block unit test ( `sync/src/tests/relayer.rs` ) is removed due to higher coverage of integration test https://github.com/nervosnetwork/ckb/pull/725


chain/src/chain.rs
```diff
@@ -198,24 +190,29 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
...
             warn!(target: "chain", "receive 0 number block: {}-{:x}", block.header().number(), block.header().hash());
         }
-        if self.verification {
+        self.insert_block(block, need_verify)?;
+        debug!(target: "chain", "finish processing block");
+        Ok(())
+    }
+
+    fn insert_block(&self, block: Arc<Block>, need_verify: bool) -> Result<(), FailureError> {
+        if need_verify {
```
keroro520: Don't move block verification inside `insert_block`, it is confused.

chain/src/chain.rs
```diff
@@ -43,13 +43,14 @@ impl Drop for ChainController {
 }
 
 impl ChainController {
-    pub fn process_block(&self, block: Arc<Block>) -> Result<(), FailureError> {
-        Request::call(&self.process_block_sender, block).expect("process_block() failed")
+    pub fn process_block(&self, block: Arc<Block>, need_verify: bool) -> Result<(), FailureError> {
+        Request::call(&self.process_block_sender, (block, need_verify))
+            .expect("process_block() failed")
```
keroro520: Suggest to add another test-used-only interface like `process_block_without_verifying`.

PR #735, opened 2019-05-13
feat: panic if it's likely to reach a deadlock

yangby-cryptape:
### Commits

- chore: use `try_lock_for(...)` to replace `lock()` for ChainState

  `sed -i "s/[.]chain_state()[.]lock()/.lock_chain_state()/g" $(grep -rl "[.]chain_state()[.]lock()" [^t]*)`

- chore: use `try_lock_for(...)` to replace `lock()` in TxsVerifyCache

  `sed -i "s/[.]txs_verify_cache()[.]lock()/.lock_txs_verify_cache()/g" $(grep -rl "[.]txs_verify_cache()[.]lock()" [^t]*)`

### Summary

Panic if it's likely to reach a deadlock (wait for 300 seconds).

https://github.com/nervosnetwork/ckb/blob/4b890724d375f845773b57b5ee45b82bee9785f7/util/src/lib.rs#L18-L23

### DIFF
--- a/util/src/lib.rs
+++ b/util/src/lib.rs
@@ -1,6 +1,7 @@
 mod linked_hash_set;
 
 use std::fmt;
+use std::time::Duration;
 
 pub use fnv::{FnvBuildHasher, FnvHashMap, FnvHashSet};
 pub use linked_hash_map::{Entries as LinkedHashMapEntries, LinkedHashMap};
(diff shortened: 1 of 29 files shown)


util/src/lib.rs
```diff
@@ -14,6 +14,14 @@ pub use parking_lot::{
     self, Condvar, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard,
 };
 
+const TRY_LOCK_SECS: u64 = 300;
```
quake:
```suggestion
const TRY_LOCK_SECS: Duration = Duration::from_secs(300);
```

PR #740, opened 2019-05-13
feat: remove always_success script

xxuejie:
Now always_success is only used in tests, CKB won't bundle
always_success script in the binary


xxuejie:
> Please update the misleading comment in ckb.toml.
> 
> ```
> [block_assembler]
> # value is set as always success binary hash
> ```

This is updated

doitian:
Included in #745

PR #743, opened 2019-05-13
fix: increase protocols time event interval

jjyr:


sync/src/synchronizer/mod.rs
```diff
@@ -41,7 +41,7 @@ use std::time::{Duration, Instant};
 pub const SEND_GET_HEADERS_TOKEN: u64 = 0;
 pub const BLOCK_FETCH_TOKEN: u64 = 1;
 pub const TIMEOUT_EVICTION_TOKEN: u64 = 2;
-const SYNC_NOTIFY_INTERVAL: Duration = Duration::from_millis(200);
+const SYNC_NOTIFY_INTERVAL: Duration = Duration::from_millis(2000);
```
TheWaWaR: Should be bigger

PR #751, opened 2019-05-14
fix: token unreachable bug

TheWaWaR:


sync/src/synchronizer/mod.rs
```diff
@@ -568,7 +568,7 @@ impl<CS: ChainStore> CKBProtocolHandler for Synchronizer<CS> {
                 TIMEOUT_EVICTION_TOKEN => {
                     self.eviction(nc.as_ref());
                 }
-                _ => unreachable!(),
+                _ => {}
```
doitian: Add a comment why skip this branch.

PR #762, opened 2019-05-14
feat: Live cell block hash

keroro520:
* Return `block_hash` for `get_cells_by_lock_hash`
* Add `make gen-doc` command

BREAKING CHANGE: RPC `get_cells_by_lock_hash` response changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ doc: ## Build the documentation for the local package.
 doc-deps: ## Build the documentation for the local package and all dependencies.
 	cargo doc --all
 
+gen-doc:  ## Generate rpc documentation
+	./devtools/doc/rpc.py rpc/json/rpc.json > rpc/README.md
+
 ##@ Building
 check: setup-ckb-test ## Runs all of the compiler's checks.
 	cargo check ${VERBOSE} --all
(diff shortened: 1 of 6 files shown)


keroro520:
> I think the response is misleading that we must include block hash when refer to a cell in the transaction.

I agree.
But I think if our response contain `block_hash`, we should fill it with specific value.
Or remove `block_hash` field from response ?

Makefile
```diff
@@ -61,7 +64,7 @@ security-audit: ## Use cargo-audit to audit Cargo.lock for crates with security
 ##@ Continuous Integration
 
 ci: ## Run recipes for CI.
-ci: cargo-license fmt clippy security-audit test
+ci: cargo-license fmt check-dirty-doc
```
doitian: `clippy security-audit test` should not be removed

Makefile
```diff
@@ -72,6 +75,9 @@ info: ## Show environment info.
 cargo-license:
 	FILES="$$(find . -name Cargo.toml | xargs grep -L '^license')"; if [ -n "$$FILES" ]; then echo "Missing license in: $${FILES}"; false; fi
 
+check-dirty-doc:
+	GEN="$$(./devtools/doc/rpc.py rpc/json/rpc.json)"; CUR="$$(cat rpc/README.md)"; if [ "$$GEN" != "$$CUR" ]; then echo "Dirty RPC document, please check"; false; fi
```
doitian: Just generate the file into target file, and check whether it is dirty via `git diff --exit-code rpc/README.md`.

PR #771, opened 2019-05-15
fix(deps): Upgrade p2p fix gracefully shutdown network service

TheWaWaR:
✨ Silky smooth `Ctrl + C` experience ✨


doitian:
Please send PR to base rc/v0.12.0

PR #777, opened 2019-05-15
[ᚬrylai-testbed] fix: fix block template cache

keroro520:
* Use block hash as one of the keys of block template cache

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -98,7 +98,7 @@ pub struct BlockAssembler<CS> {
     config: BlockAssemblerConfig,
     work_id: AtomicUsize,
     last_uncles_updated_at: AtomicU64,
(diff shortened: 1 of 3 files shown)


miner/src/block_assembler.rs
```diff
@@ -266,11 +266,12 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
 
         let header = chain_state.tip_header().to_owned();
         let number = chain_state.tip_number() + 1;
+        let hash = header.hash().to_owned();
         let current_time = cmp::max(unix_time_as_millis(), header.timestamp() + 1);
 
         let mut template_caches = self.template_caches.lock();
 
-        if let Some(template_cache) = template_caches.get(&(cycles_limit, bytes_limit, version)) {
+        if let Some(template_cache) = template_caches.get(&(hash.to_owned(), cycles_limit, bytes_limit, version)) {
```
jjyr: remove number argument from `is_outdate`

PR #784, opened 2019-05-15
✋ feat: compute code hash

keroro520:
* Add rpc `compute_code_hash`, `compute_script_hash`
* Return rpc parse error detail


doitian:
Once CI passes, I'll merge it into rc/v0.12.0

doitian:
Merged to rc/v0.12.0

PR #785, opened 2019-05-15
[HOLD] feat: Group rpc APIs by module

keroro520:
* Group rpc APIs by module


keroro520:
Should be rebased after https://github.com/nervosnetwork/ckb/pull/784

PR #788, opened 2019-05-15
[ᚬrc/v0.12.0] fix: correct block_median_count

keroro520:
* Correct block_median_count when verifying a compact block


doitian:
It's indeed a breaking change to change consensus rule. However, since the online data should pass both the old and new rule, we can assume this feature can be delivered gracefully.

PR #789, opened 2019-05-15
[ᚬrc/v0.12.0] doc: Group rpc APIs [skip ci]

keroro520:
* Group RPC APIs by modules

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -3,26 +3,594 @@
 
 ## JSON-RPC
 
-* [`_compute_code_hash`](#_compute_code_hash)
-* [`_compute_script_hash`](#_compute_script_hash)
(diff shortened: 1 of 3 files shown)


rpc/README.md
```diff
@@ -3,230 +3,31 @@
...
-```json
-{
-    "id": 2,
-    "jsonrpc": "2.0",
-    "result": {
-        "cycles": "0"
-    }
-}
-```
+* [`chain`](#chain)
```
doitian: As commented in the original PR, the title link is broken, and the name should match the one in config file, a.k.a, capitalizing it.

keroro520: Oops. I didn't see any comments in https://github.com/nervosnetwork/ckb/pull/785

Updated rpc.py: capitalize module name, and link module to the first API inside it.

PR #793, opened 2019-05-15
[ᚬrc/v0.12.0] fix: Outbound peer service and discovery service

TheWaWaR:
* Fix outbound service interval cause network thread CPU 100% bug
* Fix discovery service broadcast local address


TheWaWaR:
https://github.com/cryptape/ckb-internal/issues/975

PR #802, opened 2019-05-16
feat: Flush logger at shutdown

keroro520:
* Flush logger at shutdown. Avoid losing logs which are not flushed into the disk.


doitian:
Already flushed in logger guard, duplicated flush will cause panic.

keroro520:
> Already flushed in logger guard, duplicated flush will cause panic.

You are right. Thanks!
I have to figure out the real reason why my logs lost

PR #810, opened 2019-05-17
fix: Discard invalid orphan blocks

keroro520:
* Discard invalid orphan blocks
    When accepts a new block, its descendants should be accepted too if valid. So if an error occurs when we try to accept its descendants, the descendants are invalid.

### DIFF
--- a/sync/src/synchronizer/mod.rs
+++ b/sync/src/synchronizer/mod.rs
@@ -242,59 +242,57 @@ impl<CS: ChainStore> Synchronizer<CS> {
         Ok(())
     }
 
-    //FIXME: guarantee concurrent block process
+    // FIXME: guarantee concurrent block process
+    // TODO: limit and prune orphan pool
     fn insert_new_block(&self, peer: PeerIndex, block: Block) {
-        let block = Arc::new(block);
-        if self
-            .shared
-            .block_header(&block.header().parent_hash())
-            .is_some()
-        {
-            let accept_ret = self.accept_block(peer, &block);
-            if accept_ret.is_ok() {
-                let pre_orphan_block = self
-                    .orphan_block_pool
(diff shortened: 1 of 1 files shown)


sync/src/synchronizer/mod.rs
```diff
@@ -257,25 +257,26 @@ impl<CS: ChainStore> Synchronizer<CS> {
                     .remove_blocks_by_parent(&block.header().hash());
                 for block in pre_orphan_block {
                     let block = Arc::new(block);
-                    if self
+                    let ret = if self
```
quake: we can use rust idiomatic for option and result, it's easier to read:
```
if let Some(Err(e)) = self.shared.block_header(&block.header().parent_hash())
    .and(|| self.accept_block(peer, &block).err())
    .or(|| Err(err_msg(...))) {
     debug!(...)
}
```

keroro520: I have tried to change it more rust idiomatic but failed...
Could you give me the complete code snips? Thanks ~

quake: Could you try this?  my previous comment is something wrong, sorry.
```
                    match self
                        .shared
                        .block_header(&block.header().parent_hash())
                        .and(Some(self.accept_block(peer, &block)))
                    {
                        None => {
                            debug!(
                                target: "sync", "[Synchronizer] accept orphan block {:?}, parent not found",
                                block
                            );
                        }
                        Some(Err(e)) => {
                            debug!(
                                target: "sync", "[Synchronizer] accept orphan block {:?} error {:?}",
                                block,
                                e
                            );
                        }
                        Some(Ok(_)) => {}
                    }
```

keroro520: Thanks.
I have totally rewritten this function. PTAL

PR #827, opened 2019-05-18
[ᚬrc/v0.12.0] [ᚬmaster] docs: upadte the docker document

Mine77:
add a step to change the default configuration

### DIFF
--- a/docs/run-ckb-with-docker.md
+++ b/docs/run-ckb-with-docker.md
@@ -34,6 +34,12 @@ docker run --rm -it \
   nervos/ckb:latest init --spec testnet --force
 ```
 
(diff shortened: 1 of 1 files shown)


docs/run-ckb-with-docker.md
```diff
@@ -34,6 +34,12 @@ docker run --rm -it \
   nervos/ckb:latest init --spec testnet --force
 ```
 
+Find the rpc configuration in `ckb.toml`, and change the `listen_address` to `0.0.0.0:8114`:
```
doitian: This should be added after files have been copied from container. 

I'll close it, since I plan to fix it by setting it to 0.0.0.0 in docker.

PR #832, opened 2019-05-18
refactor: peer_store db::PeerInfoDB interface

jjyr:


network/src/peer_store/sqlite/db.rs
```diff
@@ -51,52 +51,31 @@ pub fn create_tables(conn: &Connection) -> DBResult<()> {
...
-                    conn,
-                    peer_id,
-                    addr,
-                    session_type,
-                    score,
-                    last_connected_at_ms,
-                )?;
-                Self::get_by_peer_id(conn, peer_id)
+                Self::insert_or_update(conn, peer)?;
+                Self::get_by_peer_id(conn, &peer.peer_id)
```
doitian: It is a bit weird, why not just return the peer info directly? Do we have any database generated field here?

PR #833, opened 2019-05-18
feat: improve `ckb init` subcommand.

doitian:
- force editing `ckb.toml` to enable miner.
    Add a `enable` flag in `block_assembler` option in `ckb.toml`. Set it to
    false except in the development environment. When user generates config
    files for online chain environment, he/she is required to edit the config
    file to enable miner.

    BREAKING CHANGE: exiting ckb.toml must add `enable = true` under
    `block_assembler`.

- feat: update options for `ckb init`.

    This is a breaking change.

    - Export `specs/dev.toml` when init for dev chain.
    - Remove option `--export-specs`.
    - Rename `spec` to `chain` in options.
       - Option `--spec` is renamed to `--chain`
       - Option `--list-specs` is renamed to `--list-chains`
    - Rename `export` to `create` in messages.

- feat: limit chain spec name

   Only `ckb_dev` is allowed in the expected chain spec file.


spec/src/lib.rs
```diff
@@ -87,48 +87,43 @@ pub struct IssuedCell {
...
 #[derive(Debug)]
-pub struct FileNotFoundError;
-
-impl FileNotFoundError {
-    fn boxed() -> Box<Self> {
-        Box::new(FileNotFoundError)
-    }
+pub enum SpecLoadError {
+    FileNotFound,
+    SpecNameNotAllowed(String),
```
keroro520: If change the commandline option `--spec` to `--chain`, then may adjust the error messages.

doitian:
I'll split it into 4 PRs, since they are really unrelated changes.

PR #841, opened 2019-05-19
feat: apply tx_pool limit

zhangsoledad:
### Features
1. apply tx_pool limit
2. tx size verify, enforce tx size below  block size limit

#### config ckb.toml

```diff
[tx_pool]
- max_pool_size = 10000
- max_orphan_size = 10000
- max_proposal_size = 10000
- max_cache_size = 1000
- max_pending_size = 10000
- txs_verify_cache_size = 100000
+ max_mem_size = 20_000_000 # 20mb
+ max_cycles = 200_000_000_000
+ max_verfify_cache_size = 100_000
```

#### rpc tx_pool_info

```diff
+ "total_tx_cycles": "2",
+ "total_tx_size": "156",
```

### DIFF
--- a/rpc/json/rpc.json
+++ b/rpc/json/rpc.json
@@ -570,7 +570,9 @@
             "last_txs_updated_at": "0",
             "orphan": "0",
             "pending": "1",
-            "proposed": "0"
+            "proposed": "0",
+            "total_tx_cycles": "2",
+            "total_tx_size": "156"
         }
     },
     {
--- a/shared/src/chain_state.rs
+++ b/shared/src/chain_state.rs
@@ -243,23 +243,34 @@ impl<CS: ChainStore> ChainState<CS> {
         self.tx_pool.borrow().get_tx_with_cycles(short_id)
     }
 
+    pub(crate) fn reach_tx_pool_limit(&self, tx_size: usize, cycles: Cycle) -> bool {
+        let tx_pool = self.tx_pool.borrow();
+        tx_pool.reach_size_limit(tx_size) || tx_pool.reach_cycles_limit(cycles)
+    }
+
     // Add a verified tx into pool
     // this method will handle fork related verifications to make sure we are safe during a fork
     pub fn add_tx_to_pool(&self, tx: Transaction, cycles: Cycle) -> Result<Cycle, PoolError> {
         let short_id = tx.proposal_short_id();
+        let tx_size = tx.serialized_size();
+        if self.reach_tx_pool_limit(tx_size, cycles) {
+            return Err(PoolError::LimitReached);
+        }
         match self.resolve_tx_from_pending_and_proposed(&tx) {
             Ok(rtx) => {
-                self.verify_rtx(&rtx, Some(cycles)).map(|cycles| {
+                self.verify_rtx(&rtx, Some(cycles)).and_then(|cycles| {
                     let mut tx_pool = self.tx_pool.borrow_mut();
                     if self.contains_proposal_id(&short_id) {
                         // if tx is proposed, we resolve from proposed, verify again
-                        if let Err(e) = self.proposed_tx_and_descendants(&mut tx_pool, Some(cycles), tx.serialized_size(), tx) {
-                            debug!(target: "tx_pool", "Failed to add proposed tx {:?}, reason: {:?}", short_id, e)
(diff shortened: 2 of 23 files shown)


rpc/json/rpc.json
```diff
@@ -576,15 +576,19 @@
       "last_txs_updated_at": "0",
       "orphan": "0",
       "pending": "1",
-      "proposed": "0"
+      "proposed": "0",
+      "total_tx_cycles": "2",
+      "total_tx_size": "156"
```
doitian: Please revert the format changes below, it will cause problems when merged back to develop. I'll set up a CI job to enforce this file is formatted using prettier.

shared/src/chain_state.rs
```diff
@@ -455,42 +489,42 @@ impl<CS: ChainStore> ChainState<CS> {
...
             if self.contains_proposal_id(entry.key()) {
                 let entry = entry.remove();
-                entries.push((entry.cycles, entry.transaction));
+                entries.push((entry.cycles, entry.size, entry.transaction));
             }
         }
 
         // try move pending to proposed
         for entry in tx_pool.pending.entries() {
             if self.contains_proposal_id(entry.key()) {
```
jjyr: why

shared/src/tx_pool/pool.rs
```diff
@@ -55,38 +61,76 @@ impl TxPool {
...
+        self.total_tx_cycles += cycles;
+    }
+
+    // cycles overflow is possible, currently cycles is not precise
+    pub fn update_statics_for_remove_tx(&mut self, tx_size: usize, cycles: Cycle) {
+        let (total_tx_size, overflow) = self.total_tx_size.overflowing_sub(tx_size);
+        if overflow {
+            log::error!(target: "tx_pool", "total_tx_size {} overflow by sub {}", self.total_tx_size, tx_size);
+        }
+        let (total_tx_cycles, overflow) = self.total_tx_cycles.overflowing_sub(cycles);
```
jjyr: https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_sub

shared/src/chain_state.rs
```diff
@@ -455,42 +489,42 @@ impl<CS: ChainStore> ChainState<CS> {
...
+                gaps.push((entry.cycles, entry.size, entry.transaction));
             }
         }
 
-        for (cycles, tx) in entries {
+        for (cycles, size, tx) in entries {
             let tx_hash = tx.hash().to_owned();
-            if let Err(e) = self.proposed_tx_and_descendants(&mut tx_pool, cycles, tx) {
+            if let Err(e) = self.proposed_tx_and_descendants(&mut tx_pool, cycles, size, tx) {
                 debug!(target: "tx_pool", "Failed to add proposed tx {:x}, reason: {:?}", tx_hash, e);
```
jjyr: Missing the decrease of pool size here? Entries are removed but pool size is not change

zhangsoledad: see  `proposed_tx`

zhangsoledad:
> IMO, it is too easy to miss size calculation in the current implementation, better to move size calculation inside the tx_pool basic interfaces.

It's not quite easy,  It's associated with error handle on different levels.
It's not considered at the beginning.
We have the plan to deal with this.  :shipit:  Don't waste time on this.

verification/src/transaction_verifier.rs
```diff
@@ -103,20 +108,43 @@ impl<'a> VersionVerifier<'a> {
...
+pub struct SizeVerifier<'a> {
+    transaction: &'a Transaction,
+    block_bytes_limit: u64,
+}
+
+impl<'a> SizeVerifier<'a> {
+    pub fn new(transaction: &'a Transaction, block_bytes_limit: u64) -> Self {
+        SizeVerifier {
+            transaction,
+            block_bytes_limit,
```
quake: If we use consensus `max_block_bytes` here, the transaction which size approaching bytes limit will be accepted and broadcasted to peers, but cannot be assembled in `BlockTemplate` because of the extra cellbase size.

How about adding cellbase size limit to consensus and minus this value here as gap?

doitian: Let's address this in https://github.com/nervosnetwork/ckb/issues/885

Issue #228, opened 2019-01-26
Limit TXO set memory usage


doitian:
In #158, TXO stores all the history transaction, although each transaction only occupies about 1 byte, it will keep increasing.

At some time point, the general nodes will have no enough memory to store all the TXO spent flags. A simple optimization is pruning the fully spent transactions from the TXO set. It is OK to assume that the transaction using dead cells is orphan.

Even the number of transactions with unspent outputs [is increasing](https://statoshi.info/dashboard/db/unspent-transaction-output-set). To make the node friendly to small memory devices, this should be hybrid storage, where the memory usage can be limited via config.

_Originally posted by @doitian in https://github.com/nervosnetwork/ckb/pull/158/review_comment/create_

PR #845, opened 2019-05-19
feat: limit TXO set memory usage

yangby-cryptape:
### Commits

- feat: add a function to traverse all keys for database

- feat: save cell set into database (but not used)

  BREAKING CHANGE: Add new column to store cell set into database.

- refactor: loading CellSet do not have to load all blocks

- perf: delete transactions whose all cells are dead

- fix: the order to apply cell set diff is very important

### Issue

#228

### Notes

- I insist on the opinion: save live cells is better than save transactions which include at least one live cell.

  We can use a only-key-and-no-value database (for example, Redis) to store live cells (`CellOutPoint` as key). 

  I tried, but it's too difficult to adapt to current logic, so I gave up.

- I didn't change current design about how to use `CellSet`, I just made them be persisted.

- **All transactions** which include at least one live cell, **are still kept in memory**.

  But **the transactions which don't have any live cells are removed from memory**.

### DIFF
--- a/core/src/transaction.rs
+++ b/core/src/transaction.rs
@@ -10,6 +10,7 @@ use hash::blake2b_256;
 use numext_fixed_hash::{h256, H256};
 use occupied_capacity::{HasOccupiedCapacity, OccupiedCapacity};
 use serde_derive::{Deserialize, Serialize};
+use std::convert::TryInto;
 use std::fmt;
 use std::hash::{Hash, Hasher};
 use std::mem;
@@ -25,6 +26,50 @@ pub const TX_VERSION: Version = 0;
 // unlocking process of the DAO cell. The hex used here is actually
 // "NERVOSDAOINPUT0001" in hex mode.
 pub const ISSUING_DAO_HASH: H256 = h256!("0x4e4552564f5344414f494e50555430303031");
+
+pub struct CellKey([u8; 36]);
+
+impl CellKey {
+    pub fn calculate(tx_hash: &H256, index: u32) -> Self {
+        let mut key: [u8; 36] = [0; 36];
+        key[..32].copy_from_slice(tx_hash.as_bytes());
+        key[32..36].copy_from_slice(&index.to_le_bytes());
+        CellKey(key)
+    }
+
+    pub fn recover(&self) -> CellOutPoint {
+        Self::deconstruct(&self.0)
+    }
+
+    pub fn deconstruct(bytes: &[u8]) -> CellOutPoint {
+        let tx_hash = H256::from_slice(&bytes[..32]).expect("should not be failed");
+        let le_bytes: [u8; 4] = bytes[32..36].try_into().expect("should not be failed");
+        let index = u32::from_le_bytes(le_bytes);
+        CellOutPoint { tx_hash, index }
+    }
+}
+
+impl AsRef<[u8]> for CellKey {
+    fn as_ref(&self) -> &[u8] {
+        &self.0[..]
+    }
+}
+
+#[derive(Eq, PartialEq)]
+pub struct CellOutPointRef<'a> {
+    tx_hash: &'a H256,
+    index: u32,
+}
+
+impl<'a> CellOutPointRef<'a> {
+    pub fn new(tx_hash: &'a H256, index: u32) -> CellOutPointRef<'a> {
+        CellOutPointRef { tx_hash, index }
+    }
+
+    pub fn cell_key(&self) -> CellKey {
+        CellKey::calculate(self.tx_hash, self.index)
+    }
+}
 
 #[derive(Clone, Serialize, Deserialize, Eq, PartialEq, Hash, HasOccupiedCapacity)]
 pub struct CellOutPoint {
@@ -60,6 +105,17 @@ impl CellOutPoint {
 
     pub const fn serialized_size() -> usize {
         H256::size_of() + mem::size_of::<u32>()
+    }
+
+    pub fn cell_key(&self) -> CellKey {
+        CellKey::calculate(&self.tx_hash, self.index)
+    }
+
+    pub fn to_ref(&self) -> CellOutPointRef {
+        CellOutPointRef {
+            tx_hash: &self.tx_hash,
+            index: self.index,
+        }
     }
 }
 
--- a/core/src/transaction_meta.rs
+++ b/core/src/transaction_meta.rs
@@ -22,36 +22,47 @@ impl From<BitVecSerde> for BitVec {
 pub struct TransactionMeta {
     block_number: u64,
     epoch_number: u64,
-    /// first bit indicate if transaction is a cellbase transaction
-    /// next bits indicate if transaction has dead cells
+    cellbase: bool,
+    /// each bits indicate if transaction has dead cells
     #[serde(with = "BitVecSerde")]
     dead_cell: BitVec,
 }
 
 impl TransactionMeta {
-    pub fn new(block_number: u64, epoch_number: u64, outputs_count: usize) -> TransactionMeta {
+    pub fn new(
+        block_number: u64,
+        epoch_number: u64,
+        outputs_count: usize,
+        all_dead: bool,
+    ) -> TransactionMeta {
         TransactionMeta {
             block_number,
             epoch_number,
-            dead_cell: BitVec::from_elem(outputs_count + 1, false),
+            cellbase: false,
+            dead_cell: BitVec::from_elem(outputs_count, all_dead),
         }
     }
 
     /// New cellbase transaction
-    pub fn new_cellbase(block_number: u64, epoch_number: u64, outputs: usize) -> Self {
-        let mut result = Self::new(block_number, epoch_number, outputs);
-        result.dead_cell.set(0, true);
+    pub fn new_cellbase(
+        block_number: u64,
+        epoch_number: u64,
+        outputs_count: usize,
+        all_dead: bool,
+    ) -> Self {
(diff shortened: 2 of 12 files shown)


core/src/transaction_meta.rs
```diff
@@ -66,27 +77,23 @@ impl TransactionMeta {
...
-    pub fn is_new(&self) -> bool {
-        self.dead_cell.none()
+    pub fn is_dead(&self, index: usize) -> Option<bool> {
+        self.dead_cell.get(index)
     }
 
-    pub fn is_all_dead(&self) -> bool {
+    pub fn all_dead(&self) -> bool {
         self.dead_cell.all()
     }
```
yangby-cryptape: **Note**
This is wrong: because `first bit indicate if transaction is a cellbase transaction`.
So I removed this bit and add a `bool` to store whether it's a cellbase transaction.

core/src/transaction.rs
```diff
@@ -26,6 +26,51 @@ pub const TX_VERSION: Version = 0;
...
+        key[..32].copy_from_slice(tx_hash.as_bytes());
+        key[32..36].copy_from_slice(&index.to_le_bytes());
+        CellKey(key)
+    }
+
+    pub fn recover(&self) -> CellOutPoint {
+        Self::deconstruct(&self.0)
+    }
+
+    pub fn deconstruct(bytes: &[u8]) -> CellOutPoint {
```
jjyr: `to_cell_outpoint` is better?

core/src/transaction.rs
```diff
@@ -61,6 +106,17 @@ impl CellOutPoint {
     pub const fn serialized_size() -> usize {
         H256::size_of() + mem::size_of::<u32>()
     }
+
+    pub fn cell_key(&self) -> CellKey {
+        CellKey::calculate(&self.tx_hash, self.index)
+    }
+
+    pub fn to_ref(&self) -> CellOutPointRef {
```
jjyr: `as_ref`?

shared/src/cell_set.rs
```diff
@@ -124,69 +132,67 @@ impl CellSet {
...
+
+    pub(crate) fn insert_cell(
+        &mut self,
+        cell: &CellOutPoint,
+        number: u64,
+        epoch: u64,
+        cellbase: bool,
+        outputs_len: usize,
+    ) -> TransactionMeta {
+        let mut meta = if cellbase {
```
jjyr: what if TransactionMeta is already existed?

core/src/transaction.rs
yangby-cryptape: `as_ref()` is used by `AsRef<T>`, and `as_ref()` will return a reference (`&T`), but `to_ref()` return `R` (not `&R`).

shared/src/cell_set.rs
yangby-cryptape: - If we care this, then we should use `insert_transaction(...)`.

- In our scene, the check has already did before we call this function (I call it after `try_mark_live(..)`).

core/src/transaction.rs
yangby-cryptape: There are more than two methods returns `CellOutPoint`, which should be `to_cell_outpoint`?

- `recover(..)`: we have a `CellKey`, we recover the `CellOutPoint` from the `CellKey`

- `deconstruct(..)`: we don't have a `CellKey`, we can deconstruct / parse / deserialize a `CellOutPoint` from a slice input. `CellKey` is more like a tool or a utility struct.

- I wrote two methods, because we can avoid to create a temporary `[u8; N]` from a slice.

shared/src/cell_set.rs
```diff
@@ -124,69 +132,67 @@ impl CellSet {
...
-            new_outputs,
-        } = diff;
-
-        old_outputs.iter().for_each(|h| {
-            self.remove(h);
-        });
-
-        old_inputs.iter().for_each(|o| {
-            self.mark_live(o);
-        });
```
yangby-cryptape: The order in this function is not right: If an old input is also an old output, this mark live will not work as we expected.

core/src/transaction_meta.rs
```diff
@@ -22,36 +22,47 @@ impl From<BitVecSerde> for BitVec {
...
 
     /// New cellbase transaction
-    pub fn new_cellbase(block_number: u64, epoch_number: u64, outputs: usize) -> Self {
-        let mut result = Self::new(block_number, epoch_number, outputs);
-        result.dead_cell.set(0, true);
+    pub fn new_cellbase(
+        block_number: u64,
+        epoch_number: u64,
+        outputs: usize,
```
quake: inconsistent naming in fn `new`
```suggestion
        outputs_count: usize,
```

shared/src/cell_set.rs
```diff
@@ -65,6 +65,11 @@ pub struct CellSet {
     pub(crate) inner: FnvHashMap<H256, TransactionMeta>,
 }
 
+pub(crate) enum CellSetOpr {
+    Delete,
+    Update(TransactionMeta),
```
quake: is it possible to use lifetime here and voiding clone in `mark_dead`?

core/src/transaction.rs
```diff
@@ -26,6 +26,51 @@ pub const TX_VERSION: Version = 0;
...
+
+    pub fn recover(&self) -> CellOutPoint {
+        Self::deconstruct(&self.0)
+    }
+
+    pub fn deconstruct(bytes: &[u8]) -> CellOutPoint {
+        let tx_hash = H256::from_slice(&bytes[..32]).expect("should not be failed");
+        let mut le_bytes = [0u8; 4];
+        le_bytes.copy_from_slice(&bytes[32..36]);
+        let index = u32::from_le_bytes(le_bytes);
```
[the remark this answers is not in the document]
yangby-cryptape: I will test if it works.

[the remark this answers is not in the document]
yangby-cryptape:
```
   Compiling ckb-core v0.13.0-pre (/demo/core)
warning: unused import: `std::convert::TryInto`
 --> core/src/transaction_meta.rs:3:5
  |
3 | use std::convert::TryInto;
  |     ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default

error[E0599]: no method named `try_into` found for type `[u8]` in the current scope
  --> core/src/transaction.rs:45:47
   |
45 |         let le_bytes: [u8; 4] = bytes[32..36].try_into().expect("should not be failed");
   |                                               ^^^^^^^^
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           `use std::convert::TryInto;`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `ckb-core`.
```

shared/src/cell_set.rs
yangby-cryptape: I can return `E<&T>`, but I can't put some `&T` from `E<&T>` into `Vec<&T>`.

Or maybe I can use `&T` immediately, avoid `Vec<&T>`.

But the code will be **very hard to understand**: I have to mix `insert` and `remove` both in memory and database, in one closure.

I tried before I pulled this PR, but I failed, I asked @driftluo for help, and he think that is impossible, too.

yangby-cryptape: Order is important, all `remove` have to be after all `insert`, so, I have to collect all `insert` and `remove` before do it.

So, I can't avoid `Vec<&T>`.

quake: Get, thanks.

core/src/transaction_meta.rs
```diff
@@ -22,36 +22,47 @@ impl From<BitVecSerde> for BitVec {
 pub struct TransactionMeta {
     block_number: u64,
     epoch_number: u64,
-    /// first bit indicate if transaction is a cellbase transaction
-    /// next bits indicate if transaction has dead cells
+    cellbase: bool,
+    /// each bits indicate if transaction has dead cells
     #[serde(with = "BitVecSerde")]
     dead_cell: BitVec,
```
keroro520: It names as `dead_cell` but actually means "output cells"? 
:(

store/src/store.rs
```diff
@@ -341,8 +363,11 @@ impl<T: KeyValueDB> ChainStore for ChainKVStore<T> {
     }
 
     fn get_cell_meta(&self, tx_hash: &H256, index: u32) -> Option<CellMeta> {
-        self.get(COLUMN_CELL_META, &cell_store_key(tx_hash, index))
-            .map(|raw| deserialize(&raw[..]).unwrap())
+        self.get(
+            COLUMN_CELL_META,
+            CellKey::calculate(tx_hash, index).as_ref(),
+        )
+        .map(|raw| deserialize(&raw[..]).unwrap())
```
keroro520: Please use `expect`.
Although I assume the items fetched from the database should be OK. In fact, Sentry has reported several "unwrap" events, which seems related to deserializing database raw data.

core/src/transaction_meta.rs
yangby-cryptape: `rename` can be another PR (or commit). I just keep the old name. :)

store/src/store.rs
yangby-cryptape: :cry: I feel a bit wronged. I just keep the old code~ but, that's OK, I will update this after lunch.

yangby-cryptape:
The last force push (f472b42):
- Rebase from the latest develop.
- Add a commit to change all `unwrap()` in `store.rs` to `expect(..)` with a significant message. (For [this discussion](https://github.com/nervosnetwork/ckb/pull/845#discussion_r287919723))

PR #848, opened 2019-05-20
chore: improve command line messages

ashchan:
* Fix typos
* Display messages in the 3rd person

This is an attempt to unify the message style. Personally I like to NOT use the 3rd parson, but `clap`'s `help` and `version` uses that. (Anyway to override that?)

BTW, would `perf` be a better subcommand name than `prof`? @jjyr 

## Before

Specifically, now we have three styles: `Prints`, `Import`, `Profling` (should be `Profiling`).

```shell
ckb 0.13.0-pre
Nervos Core Dev <dev@nervos.org>
Nervos CKB - The Common Knowledge Base

USAGE:
    ckb [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -C <path>        Runs as if ckb was started in <path> instead of the current working directory.

SUBCOMMANDS:
    cli       CLI tools
    export    Export ckb data
    help      Prints this message or the help of the given subcommand(s)
    import    Import ckb data
    init      Create a CKB direcotry or reinitialize an existing one
    miner     Run ckb miner
    prof      Profling ckb node
              Example: Process 1..500 blocks then output flagme graph
              cargo flamegraph --bin ckb -- -C <dir> prof 1 500
    run       Run ckb node
```

## After

```shell
ckb 0.13.0-pre
Nervos Core Dev <dev@nervos.org>
Nervos CKB - The Common Knowledge Base

USAGE:
    ckb [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -C <path>        Run as if ckb was started in <path> instead of the current working directory.

SUBCOMMANDS:
    cli       CLI tools
    export    Exports ckb data
    help      Prints this message or the help of the given subcommand(s)
    import    Imports ckb data
    init      Creates a CKB direcotry or reinitialize an existing one
    miner     Runs ckb miner
    prof      Profiles ckb node
              Example: Process 1..500 blocks then output flagme graph
              cargo flamegraph --bin ckb -- -C <dir> prof 1 500
    run       Runs ckb node
```

### DIFF
--- a/util/app-config/src/cli.rs
+++ b/util/app-config/src/cli.rs
@@ -167,7 +167,7 @@ fn init() -> App<'static, 'static> {
             Arg::with_name(ARG_FORCE)
                 .short("f")
                 .long(ARG_FORCE)
-                .help("Force overwrits existing files"),
+                .help("Forces overwriting existing files"),
         )
         .arg(
             Arg::with_name(ARG_RPC_PORT)
--- a/util/app-config/src/cli.rs
+++ b/util/app-config/src/cli.rs
@@ -39,7 +39,7 @@ pub fn get_matches(version: &Version) -> ArgMatches<'static> {
                 .value_name("path")
                 .takes_value(true)
                 .help(
-                    "Run as if ckb was started in <path> instead of the current working directory.",
+                    "Runs as if ckb was started in <path> instead of the current working directory.",
                 ),
         )
         .subcommand(run())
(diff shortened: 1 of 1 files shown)


doitian:
missing `Run as if ckb was started in <path> instead of the current working directory.`

doitian:
clap message can be overrides by setting `help_message` and `version_message` on `App`.

```
App::new("ckb")
  .help_message("Prints help information")
  .version_message("Prints version information")
```

ashchan:
> missing `Run as if ckb was started in <path> instead of the current working directory.`

That was actually included in the changes but I guess the `after` snippet was old. Updated that in description.

ashchan:
> clap message can be overrided by setting `help_message` and `version_message` on `App`.
> 
> ```
> App::new("ckb")
>   .help_message("Prints help information")
>   .version_message("Prints version information")
> ```
> 
> The problem of this solution is that you have to set the message for all the sub commands as well.

OK then I think we should use 3rd person for the time being.

util/app-config/src/cli.rs
```diff
@@ -167,24 +167,24 @@ fn init() -> App<'static, 'static> {
             Arg::with_name(ARG_FORCE)
                 .short("f")
                 .long(ARG_FORCE)
-                .help("Force overwriting existing files"),
+                .help("Force overwrits existing files"),
```
xxuejie: Oops

ashchan: Oops

PR #854, opened 2019-05-20
feat: Calculate median time by tracing parents

keroro520:
At present, the way calculating the passed median time is that collects block timestamps one by one by block_number. This PR change to collects blocks timestamps by tracing parents. The new way is more robust.

In addition to this, I use assert-style to rewrite the calculation of passed median time.

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -1,7 +1,7 @@
 use crate::relayer::compact_block::CompactBlock;
 use crate::relayer::compact_block_verifier::CompactBlockVerifier;
 use crate::relayer::Relayer;
-use ckb_core::{header::Header, BlockNumber};
+use ckb_core::header::Header;
 use ckb_network::{CKBProtocolContext, PeerIndex};
 use ckb_protocol::{CompactBlock as FbsCompactBlock, RelayMessage};
 use ckb_shared::shared::Shared;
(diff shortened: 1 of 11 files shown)


traits/src/block_median_time_context.rs
```diff
@@ -1,24 +1,47 @@
...
+    /// let parent_header = current_header.timestamp_and_parent().header();
+    /// return (parent_header.timestamp(), parent_header.hash());
+    /// ```
+    fn timestamp_and_parent(&self, block_hash: &H256) -> (u64, H256);
+
+    /// Return past block median time, **including the timestamp of the given one**
+    ///
+    /// It's duty for outside caller to ensure that the block_number and block_hash are matched
+    fn block_median_time(&self, block_number: BlockNumber, block_hash: &H256) -> u64 {
+        let median_time_span = min(block_number + 1, self.median_block_count());
```
jjyr: Do we really need the block_number argument?

keroro520: Without `block_number`, we must assume the parent of genesis block is "0x0000000000000".
If we accept this assumption, it's ok to remove this parameter here, in another PR.
I am not sure about whether we should make this assumption.

jjyr: IMO it is more like a part of consensus rule rather than an assumption.
However, we are missing some part to verify special rules for the genesis block.

sync/src/relayer/compact_block_process.rs
```diff
@@ -180,25 +181,29 @@ where
...
-        let mut block_hash = self.header.parent_hash().to_owned();
-        let mut timestamps: Vec<u64> = Vec::with_capacity(count as usize);
-        for _ in 0..count {
-            let header = match self.get_header(&block_hash) {
-                Some(h) => h,
-                None => break,
-            };
-            timestamps.push(header.timestamp());
-            block_hash = header.parent_hash().to_owned();
+    fn get_block_hash(&self, block_number: BlockNumber) -> Option<H256> {
```
keroro520: @jjyr Please take a look at this method

PR #856, opened 2019-05-20
[ᚬrylai] feat: revamp the secp256k1 support in CKB

doitian:
- CKB does not provide key generation feature.
- Add `ckb cli blake160` and `ckb cli blake256` utilities to compute hash.
- Add `ckb cli secp256k1-lock` to print block assembler config from
a secp256k1 pubkey.


resource/ckb.toml
```diff
@@ -88,16 +88,37 @@ cell_output_cache_size  = 128
...
+#
 # The `code_hash` identifies different cryptography algorithm. Read the manual
 # of the lock script provider about how to generate this config.
 #
 # CKB provides an secp256k1 implementation, it requires a public key hash in
-# the args.
+# the args. The hash algorithm is blake2b, with personal "ckb-default-hash".
+# The first 160 bits (20 bytes) are used as the only arg.
+#
+# The public key is compatible with Bitcoin, use any tool you trust to generate
```
Mine77: "The public key is compatible with Bitcoin" is quite confusing to me.

my advise:

You can use any tool you trust to generate a Bitcoin private key and public key pairs, which can be used in CKB as well. CKB CLI provides the function for you to convert your public key into block assembler configuration parameters.

Here is an example using OpenSSL to generate the key pair:
```
....
```

You can then generate the block assembler lock with:
```
.....
```

Here is an example output:
```
.....
```

PR #858, opened 2019-05-21
chore(network): decrease network error log level

TheWaWaR:
* Decrease error log level from warning to debug
* Or send it to sentry

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -32,6 +32,7 @@ use p2p::{
         TargetSession,
     },
     traits::ServiceHandle,
+    utils::extract_peer_id,
     SessionId,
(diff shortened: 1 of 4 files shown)


network/src/network.rs
```diff
@@ -429,7 +431,7 @@ impl ServiceHandle for EventHandler {
                 ref address,
                 ref error,
             } => {
-                warn!(target: "network", "DialerError({}) {}", address, error);
+                debug!(target: "network", "DialerError({}) {}", address, error);
```
driftluo: The other errors here also need to be processed, and when they time out, delete their own dail timeout timer.

PR #859, opened 2019-05-21
feat: Use snappy to compress large messages

driftluo:
BREAKING CHANGE: protocol communication break

On the test net monitoring, the bandwidth usage is often in a full state. We try to use the snappy compression algorithm to reduce network transmission consumption.

After testing, the compression yield of flatbuffer format is very high, cpu consumption is relatively acceptable.

The following is the data transmission on the test net:

```
2019-05-20 16:27:41.875 +08:00 tokio-runtime-worker-7 DEBUG compress  raw_data len: 625400, compress used time: 3.635121ms, compress_data size: 335401, compression ratio: 0.536298369043812, decompress used time: 1.496667ms
2019-05-20 16:27:42.128 +08:00 tokio-runtime-worker-6 DEBUG compress  raw_data len: 633544, compress used time: 3.789752ms, compress_data size: 335462, compression ratio: 0.5295007134468955, decompress used time: 1.490144ms
2019-05-20 16:27:42.340 +08:00 tokio-runtime-worker-6 DEBUG compress  raw_data len: 633216, compress used time: 3.998678ms, compress_data size: 333458, compression ratio: 0.5266101930462906, decompress used time: 1.593165ms
2019-05-20 16:27:42.558 +08:00 tokio-runtime-worker-5 DEBUG compress  raw_data len: 632992, compress used time: 3.453616ms, compress_data size: 333552, compression ratio: 0.5269450482786512, decompress used time: 1.052606ms
2019-05-20 16:27:42.740 +08:00 tokio-runtime-worker-2 DEBUG compress  raw_data len: 633760, compress used time: 1.256847ms, compress_data size: 340022, compression ratio: 0.5365154001514769, decompress used time: 545.473µs
2019-05-20 16:37:43.934 +08:00 tokio-runtime-worker-1 DEBUG compress  raw_data len: 186912, compress used time: 659.317µs, compress_data size: 42640, compression ratio: 0.22812874507789763, decompress used time: 515.287µs
2019-05-20 16:37:47.338 +08:00 tokio-runtime-worker-3 DEBUG compress  raw_data len: 186520, compress used time: 189.079µs, compress_data size: 42334, compression ratio: 0.22696761741368218, decompress used time: 150.644µs
2019-05-20 16:37:50.729 +08:00 tokio-runtime-worker-3 DEBUG compress  raw_data len: 186520, compress used time: 197.656µs, compress_data size: 42336, compression ratio: 0.22697834012438345, decompress used time: 145.5µs
2019-05-20 16:38:52.549 +08:00 tokio-runtime-worker-4 DEBUG compress  raw_data len: 95904, compress used time: 217.968µs, compress_data size: 33801, compression ratio: 0.3524461961961962, decompress used time: 95.818µs
2019-05-20 16:39:32.522 +08:00 tokio-runtime-worker-0 DEBUG compress  raw_data len: 47320, compress used time: 418.183µs, compress_data size: 17183, compression ratio: 0.363123415046492, decompress used time: 252.148µs
```

Note that this is a **break change**, the data is modified as follows:

By default, data above 40k enters compressed mode.

From the current point of view, the high bit 1 is the compressed format and the high bit 0 is the uncompressed format.

If you want to support multiple compression formats in the future, you can simply think that 0b1000 is in snappy format and 0b0000 is in uncompressed format.

```
 # Message in Bytes:

 +---------------------------------------------------------------+
 | Bytes | Type | Function                                       |
 |-------+------+------------------------------------------------|
 |   0   |  u1  | Compress: true 1, false 0                      |
 |       |  u7  | Reserved                                       |
 +-------+------+------------------------------------------------+
 |  1~   |      | Payload (Serialized Data with Compress)        |
 +-------+------+------------------------------------------------+
```

### DIFF
--- a/network/src/compress.rs
+++ b/network/src/compress.rs
@@ -1,4 +1,5 @@
 use bytes::{Bytes, BytesMut};
+use log::debug;
 use snap::{Decoder as SnapDecoder, Encoder as SnapEncoder};
 use tokio::codec::{Decoder, Encoder, LengthDelimitedCodec};
 
@@ -27,80 +28,71 @@ struct Message {
 }
 
 impl Message {
-    /// Init
-    fn init() -> Self {
-        Message {
-            inner: BytesMut::from(&[0u8][..]),
-        }
+    /// create from raw data
+    fn from_raw(data: Bytes) -> Self {
+        let mut inner = BytesMut::from(&[0u8][..]);
+        inner.unsplit(BytesMut::from(data));
+        Self { inner }
+    }
+
+    /// create from compressed data
+    fn from_compressed(data: BytesMut) -> Self {
+        Self { inner: data }
     }
 
     /// Compress message
-    fn compress(&mut self, input: Bytes) {
+    fn compress(mut self) -> Bytes {
+        let input = self.inner.split_off(1);
         if input.len() > SKIP_COMPRESS_SIZE {
             match SnapEncoder::new().compress_vec(&input) {
                 Ok(res) => {
                     self.inner.unsplit(BytesMut::from(res));
                     self.set_compress_flag(true);
                 }
-                Err(_) => {
-                    self.inner.unsplit(BytesMut::from(input));
+                Err(e) => {
+                    debug!(target: "network", "snappy compress error: {}", e);
+                    self.inner.unsplit(input);
                     self.set_compress_flag(false);
                 }
             }
         } else {
             self.set_compress_flag(false);
-            self.inner.unsplit(BytesMut::from(input));
+            self.inner.unsplit(input);
         }
+        self.inner.freeze()
     }
 
     /// Decompress message
-    fn decompress(mut self) -> Option<BytesMut> {
-        if self.inner.len() <= 1 {
-            None
+    fn decompress(mut self) -> Result<BytesMut, io::Error> {
+        if self.inner.is_empty() {
+            Err(io::ErrorKind::InvalidData.into())
         } else if self.compress_flag() {
             match SnapDecoder::new().decompress_vec(&self.inner[1..]) {
-                Ok(res) => Some(BytesMut::from(res)),
-                Err(_) => None,
+                Ok(res) => Ok(BytesMut::from(res)),
+                Err(e) => {
+                    debug!(target: "network", "snappy error: {:?}", e);
+                    Err(io::ErrorKind::InvalidData.into())
+                }
             }
         } else {
             self.inner.split_to(1);
-            Some(self.inner.take())
+            Ok(self.inner.take())
         }
     }
 
     fn set_compress_flag(&mut self, flag: bool) {
         let compress_flag = if flag { 0b1000_0000 } else { 0b0000_0000 };
-        self.inner[0] = (self.inner[0] & 0b0111_1111) + (compress_flag & 0b1000_0000);
+        self.inner[0] = (self.inner[0] & 0b0111_1111) + compress_flag;
     }
 
     fn compress_flag(&self) -> bool {
         (self.inner[0] & 0b1000_0000) != 0
     }
-
-    fn into_inner(self) -> Bytes {
-        self.inner.freeze()
-    }
-}
-
-impl From<BytesMut> for Message {
-    fn from(src: BytesMut) -> Self {
-        Message { inner: src }
-    }
-}
-
-impl From<Bytes> for Message {
-    fn from(src: Bytes) -> Self {
-        Message {
-            inner: BytesMut::from(src),
-        }
-    }
 }
 
 /// Compress data
 pub fn compress(src: Bytes) -> Bytes {
-    let mut msg = Message::init();
-    msg.compress(src);
-    msg.into_inner()
+    Message::from_raw(src).compress()
 }
 
 /// Decompression structure for Codec
@@ -126,9 +118,13 @@ impl Decoder for LengthDelimited {
     type Error = io::Error;
 
     fn decode(&mut self, src: &mut bytes::BytesMut) -> Result<Option<Self::Item>, Self::Error> {
-        self.0
-            .decode(src)
-            .map(|item| item.and_then(|item| Into::<Message>::into(item).decompress()))
+        self.0.decode(src).and_then(|item| {
+            match item {
+                // if decompress error, just return error to disconnect
+                Some(inner) => Message::from_compressed(inner).decompress().map(Some),
+                None => Ok(None),
+            }
+        })
     }
 }
 
@@ -138,8 +134,9 @@ mod test {
 
     #[test]
     fn test_no_need_compress() {
-        let mut msg = Message::init();
-        msg.compress(Bytes::from("1222"));
+        let cmp_data = Message::from_raw(Bytes::from("1222")).compress();
+
+        let msg = Message::from_compressed(cmp_data.into());
 
         assert!(!msg.compress_flag());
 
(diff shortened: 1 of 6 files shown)


network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+impl Message {
+    /// Init
+    fn init() -> Self {
+        Message {
+            inner: BytesMut::from(&[0u8][..]),
+        }
+    }
+
+    /// Compress message
+    fn compress(&mut self, input: Bytes) {
```
doitian: The way to create an empty message, then compress the input into the message looks strange to me. For example, we may forget to create new message, and call `compress` multiple times on the message. It's better to be a constructor.

network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        } else {
+            self.inner.split_to(1);
+            Some(self.inner.take())
+        }
+    }
+
+    fn set_compress_flag(&mut self, flag: bool) {
+        let compress_flag = if flag { 0b1000_0000 } else { 0b0000_0000 };
+        self.inner[0] = (self.inner[0] & 0b0111_1111) + (compress_flag & 0b1000_0000);
```
doitian: The last `&` is redundant.

network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Compress message
+    fn compress(&mut self, input: Bytes) {
+        if input.len() > SKIP_COMPRESS_SIZE {
+            match SnapEncoder::new().compress_vec(&input) {
+                Ok(res) => {
+                    self.inner.unsplit(BytesMut::from(res));
+                    self.set_compress_flag(true);
+                }
+                Err(_) => {
```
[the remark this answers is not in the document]
driftluo: There is only one situation: 
> The total number of bytes to compress exceeds 2^32 - 1.

I don't think we will encounter this kind of problem, so I ignored the past. If there is not enough memory, it will panic

network/src/compress.rs
driftluo: Therefore, the method exported by this module is `compress` and a `codec`, and other modules cannot use the `Message` struct.

If it is really necessary to change it is also feasible

network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn encode(&mut self, item: Self::Item, dst: &mut bytes::BytesMut) -> Result<(), Self::Error> {
+        self.0.encode(item, dst)
+    }
+}
+
+impl Decoder for LengthDelimited {
+    type Item = bytes::BytesMut;
+    type Error = io::Error;
+
+    fn decode(&mut self, src: &mut bytes::BytesMut) -> Result<Option<Self::Item>, Self::Error> {
```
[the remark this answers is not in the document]
driftluo: Yes, this is risky, but if writing asymmetric codec will waste a lot of CPU to compression

We have a lot of messages in broadcast mode. This type of message only needs to be compressed once.

In this scenario, I wrap the compression method at the exit of the protocol message. Whether it is unicast or broadcast, I just do compression operation once.

network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+                }
+            }
+        } else {
+            self.set_compress_flag(false);
+            self.inner.unsplit(BytesMut::from(input));
+        }
+    }
+
+    /// Decompress message
+    fn decompress(mut self) -> Option<BytesMut> {
```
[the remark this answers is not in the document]
driftluo: There are only two cases where decompress has an error:

> Invalid compressed Snappy data was seen.
> The total space required for decompression exceeds 2^32 - 1.

If the error is returned here, the protocol will be disconnected. Are we sure we need this behavior? Or ignore this message directly?

network/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+        if self.inner.len() <= 1 {
```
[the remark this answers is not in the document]
driftluo: Whether the message is compressed or not, it will add a bytes identifier to the header, so messages less than 1 bytes are illegal.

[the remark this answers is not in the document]
driftluo: Oh, here should be written `<

driftluo:
@doitian done

network/src/compress.rs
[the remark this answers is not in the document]
doitian: If so, I prefer `expect` here to let it panic.

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
...
+                    Err(io::ErrorKind::InvalidData.into())
+                }
+            }
+        } else {
+            self.inner.split_to(1);
+            Ok(self.inner.take())
+        }
+    }
+
+    fn set_compress_flag(&mut self, flag: bool) {
```
TheWaWaR: More clear with 
```
fn set_compress_flag(&mut self, compress: bool)
```

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
...
+        let compress_flag = if flag { 0b1000_0000 } else { 0b0000_0000 };
```
TheWaWaR: Use `0b1000_0000` from a `const`

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
...
+/// +-------+------+------------------------------------------------+
+#[derive(Clone, Debug)]
+struct Message {
+    inner: BytesMut,
+}
+
+impl Message {
+    /// create from raw data
+    fn from_raw(data: Bytes) -> Self {
+        let mut inner = BytesMut::from(&[0u8][..]);
```
TheWaWaR: Use `0u8` from as `const`

network/src/compress.rs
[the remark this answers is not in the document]
doitian: If the app has a loop and sends messages to p2p framework one by one, I can accept this trick. But from the changes in this PR, I only see this pattern: compress the data in a function, send it to p2p, and drop the compressed data. If my observation is right, I think we should change p2p to ensure codec is called only once on message handed to p2p framework.

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
+use bytes::{Bytes, BytesMut};
+use log::debug;
+use snap::{Decoder as SnapDecoder, Encoder as SnapEncoder};
+use tokio::codec::{Decoder, Encoder, LengthDelimitedCodec};
+
+use std::io;
+
+const SKIP_COMPRESS_SIZE: usize = 40 * 1024;
```
quake:
```suggestion
const COMPRESSION_SIZE_THRESHOLD: usize = 40 * 1024;
```

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// create from compressed data
+    fn from_compressed(data: BytesMut) -> Self {
+        Self { inner: data }
+    }
+
+    /// Compress message
+    fn compress(mut self) -> Bytes {
+        let input = self.inner.split_off(1);
```
quake: Checking the bytes len first, which avoids `split/unsplit` for small size message.
```
if self.inner.len() > SKIP_COMPRESS_SIZE {
}
```

network/src/compress.rs
```diff
@@ -0,0 +1,160 @@
...
+/// +-------+------+------------------------------------------------+
+/// |  1~   |      | Payload (Serialized Data with Compress)        |
+/// +-------+------+------------------------------------------------+
+#[derive(Clone, Debug)]
+struct Message {
+    inner: BytesMut,
+}
+
+impl Message {
+    /// create from raw data
```
quake:
```suggestion
    /// create from uncompressed raw data
```

driftluo:
I adjusted the compression threshold: from 40kb to 1kb.

The 20m bandwidth transmission speed is 2.5 m/s, and the 1 kb data requires 0.4 ms.

Snappy compression is 500m/s, our data compression rate is 2x and above, pressing 500b requires 0.2 ms transmission, but the compression time is only 0.002 ms.

The data may be jittered when considering the rate to test, but all cases are 2-5 times lower in compression speed, which does not affect the overall bandwidth utilization. Especially our data compression effect may be much larger than 50%.

However, compression is the time to squeeze the CPU, and we should not compress all the data, which will affect the execution of other tasks.

So, at present, I set the target of compression to be more than 1kb. This threshold can be adjusted in the future. We can watch while walking.

network/src/compress.rs
```diff
@@ -0,0 +1,132 @@
...
+                    self.inner.unsplit(BytesMut::from(res));
+                    self.set_compress_flag(true);
+                }
+                Err(e) => {
+                    debug!(target: "network", "snappy compress error: {}", e);
+                    self.inner.unsplit(input);
+                    self.set_compress_flag(false);
+                }
+            }
+        } else {
```
quake: can we omit this branch since the default value of compress flag is zero?

network/src/compress.rs
```diff
@@ -0,0 +1,128 @@
...
+            self.inner.split_to(1);
+            Ok(self.inner.take().freeze())
+        }
+    }
+
+    fn set_compress_flag(&mut self) {
+        self.inner[0] = (self.inner[0] & 0b0111_1111) + COMPRESS_FLAG;
```
quake:
```suggestion
        self.inner[0] = COMPRESS_FLAG;
```

PR #871, opened 2019-05-22
feat: includes full version in RPC `local_node_info`

doitian:
This is a breaking change for clients and SDKs that rely on the version
detailed content.

### DIFF
--- a/rpc/src/module/net.rs
+++ b/rpc/src/module/net.rs
@@ -23,7 +23,7 @@ pub(crate) struct NetworkRpcImpl {
 impl NetworkRpc for NetworkRpcImpl {
     fn local_node_info(&self) -> Result<Node> {
         Ok(Node {
-            version: self.network_controller.node_version().to_string(),
+            version: self.network_controller.node_version().long(),
             is_outbound: None,
             node_id: self.network_controller.node_id(),
             addresses: self


rpc/src/module/net.rs
```diff
@@ -23,7 +23,7 @@ pub(crate) struct NetworkRpcImpl {
 impl NetworkRpc for NetworkRpcImpl {
     fn local_node_info(&self) -> Result<Node> {
         Ok(Node {
-            version: self.network_controller.node_version().to_string(),
```
driftluo: These two writing effects are the same.

ref: https://doc.rust-lang.org/src/alloc/string.rs.html#2122-2132

`to_string` is `format_args!("{}", self)`, the same as `format!("{}", self)`

doitian: The `Display` impl uses `short`, we changed to use long version instead.

Response before:

    0.12.0

Now:

    0.12.0 (v0.12.0 2019-05-18)

driftluo: Our `Display` is to determine whether `commit_describe` exists. If it exists, it is long. If it does not exist, it is a revision of short. 

That is to say, even if you call long directly, there may be a short display.

doitian: Confirmed

PR #873, opened 2019-05-22
chore: rename the target in log for rpc module

yangby-cryptape:
The old target is same as the target in [`jsonrpc-core`](https://github.com/paritytech/jsonrpc/blob/master/core/src/io.rs#L177).

I rename it to `rpc-server` since we have `rpc` client in some tests.

### DIFF
--- a/resource/ckb-miner.toml
+++ b/resource/ckb-miner.toml
@@ -16,7 +16,7 @@ spec = "specs/dev.toml" # {{
 
 [logger]
 filter = "info" # {{
-# integration => filter = "info,network=trace,rpc=debug,sync=debug,relay=debug"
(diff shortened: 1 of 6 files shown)


doitian:
@quake Should we change the default logger filter for integration?

resource/ckb-miner.toml
```diff
@@ -16,7 +16,7 @@ spec = "specs/dev.toml" # {{
 
 [logger]
 filter = "info" # {{
-# integration => filter = "info,network=trace,rpc=debug,sync=debug,relay=debug"
+# integration => filter = "info,network=trace,sync=debug,relay=debug"
```
yangby-cryptape: Miner don't need `rpc-server`, so I delete this.

PR #874, opened 2019-05-22
feat: revise uncle rule

zhangsoledad:
1. get rid uncle age limit
2. try include disconnected block as uncle

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -14,7 +14,7 @@ use ckb_shared::error::SharedError;
 use ckb_shared::shared::Shared;
 use ckb_store::{ChainStore, StoreBatch};
 use ckb_traits::ChainProvider;
-use ckb_verification::{BlockVerifier, ContextualBlockVerifier, Verifier};
+use ckb_verification::{BlockVerifier, ContextualBlockVerifier, ForkContext, Verifier};
 use crossbeam_channel::{self, select, Receiver, Sender};
 use dao::calculate_dao_data;
 use failure::Error as FailureError;
@@ -310,6 +310,12 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
                 fork.detached_proposal_id().iter(),
                 &mut txs_verify_cache,
             );
+            if !fork.detached_blocks().is_empty() {
+                for detached_block in fork.detached_blocks() {
+                    self.notify
+                        .notify_new_uncle(Arc::new(detached_block.clone()));
+                }
+            }
             if log_enabled!(target: "chain", log::Level::Debug) {
                 self.print_chain(&chain_state, 10);
             }
@@ -484,14 +490,20 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
         let attached_blocks_iter = fork.attached_blocks().iter().rev();
         let detached_blocks_iter = fork.detached_blocks().iter().rev();
 
-        let unverified_len = fork.attached_blocks.len() - dirty_exts.len();
+        let mut fork_context = ForkContext {
+            attached_blocks: Vec::with_capacity(fork.attached_blocks().len()),
+            detached_blocks: fork.detached_blocks().iter().rev().collect(),
+            provider: self.shared.clone(),
+        };
+
+        let verified_len = fork.attached_blocks.len() - dirty_exts.len();
 
         for b in detached_blocks_iter {
             cell_set_diff.push_old(b);
             block_headers_provider.push_detached(b);
         }
 
-        for b in attached_blocks_iter.take(unverified_len) {
+        for b in attached_blocks_iter.take(verified_len) {
             cell_set_diff.push_new(b);
             outputs.extend(
                 b.transactions()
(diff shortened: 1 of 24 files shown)


doitian:
Add `b:consensus` if this is a breaking change.

chain/src/chain.rs
```diff
@@ -310,6 +310,12 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
                 fork.detached_proposal_id().iter(),
                 &mut txs_verify_cache,
             );
+            if !fork.detached_blocks().is_empty() {
```
doitian: Seems a redundant check.

chain/src/chain.rs
```diff
@@ -484,31 +490,36 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
...
         let detached_blocks_iter = fork.detached_blocks().iter().rev();
 
-        let unverified_len = fork.attached_blocks.len() - dirty_exts.len();
+        let mut fork_context = ForkContext {
+            attached_blocks: Vec::with_capacity(fork.attached_blocks().len()),
+            detached_blocks: fork.detached_blocks().iter().rev().collect(),
+            provider: self.shared.clone(),
+        };
+
+        let verified_len = fork.attached_blocks.len() - dirty_exts.len();
```
doitian: 🤣

chain/src/chain.rs
```diff
@@ -484,31 +490,36 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
         let attached_blocks_iter = fork.attached_blocks().iter().rev();
         let detached_blocks_iter = fork.detached_blocks().iter().rev();
 
-        let unverified_len = fork.attached_blocks.len() - dirty_exts.len();
+        let mut fork_context = ForkContext {
+            attached_blocks: Vec::with_capacity(fork.attached_blocks().len()),
+            detached_blocks: fork.detached_blocks().iter().rev().collect(),
```
doitian: Bad smell, similar structure, same field name, but have a different order.

miner/src/block_assembler.rs
```diff
@@ -135,31 +132,41 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
...
-                        Ok(uncle_block) => {
-                            let hash = uncle_block.header().hash();
-                            self.candidate_uncles.insert(hash.to_owned(), uncle_block);
-                            self.last_uncles_updated_at
-                                .store(unix_time_as_millis(), Ordering::SeqCst);
-                        }
-                        _ => {
-                            log::error!(target: "miner", "new_uncle_receiver closed");
+            .spawn(move || {
+                let mut candidate_uncles = LruCache::new(MAX_CANDIDATE_UNCLES);
```
doitian: Why move from the field to a local variable? Work around the conflict mutable borrow?

verification/src/contextual_block_verifier.rs
```diff
@@ -20,38 +20,111 @@ use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIter
...
+        }
+        let detached = fork
+            .detached_blocks
+            .iter()
+            .map(|b| b.header().hash())
+            .collect();
+        let detached_uncles = fork
+            .detached_blocks
+            .iter()
+            .map(|b| b.uncles().iter().map(UncleBlock::hash))
```
doitian: can use `flat_map`

doitian:
How can we verify PoW if the uncle is disconnected?

miner/src/block_assembler.rs
zhangsoledad: `Move out `,   `RefCel `l， `Mutex `
I choose  `Move out`

zhangsoledad:
> How can we verify PoW if the uncle is disconnected?
Disconnected block all are PoW verified. 
We process block by order, the orphan block won't appear here.

chain/src/chain.rs
zhangsoledad: I unified the order. haven't idea about name, still .

miner/src/block_assembler.rs
```diff
@@ -429,80 +433,41 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
...
+        loop {
+            if candidate_uncles.is_empty() || uncles.len() == max_uncles_num {
                 break;
             }
-
-            let epoch_number = current_epoch_ext.number();
-
-            // uncle must be same difficulty epoch with candidate
+            let (hash, block) = candidate_uncles
+                .pop_front()
```
quake: It doesn't insert back the poped uncle block to `candidate_uncles`, this uncle block will be excluded in the next `get_block_template` function invoking.

verification/src/contextual_block_verifier.rs
```diff
@@ -49,11 +49,83 @@ impl<'a, CS: ChainStore> BlockMedianTimeContext for ForkContext<'a, CS> {
...
+
+impl<'a, P: ChainProvider> UncleProvider for UncleVerifierContext<'a, P> {
+    fn double_inclusion(&self, hash: &H256) -> bool {
+        if self.excluded.contains(hash) {
+            return true;
+        }
+
+        // main chain
+        if self.provider.store().get_block_number(hash).is_some() {
+            if self.detached.contains(hash) {
```
quake:
```suggestion
            return !self.detached.contains(hash);
```

verification/src/contextual_block_verifier.rs
```diff
@@ -49,11 +49,83 @@ impl<'a, CS: ChainStore> BlockMedianTimeContext for ForkContext<'a, CS> {
...
+                return false;
+            }
+            return true;
+        }
+
+        if self.provider.store().is_uncle(hash) {
+            if self.detached_uncles.contains(hash) {
```
quake:
```suggestion
            return !self.detached_uncles.contains(hash);
```

verification/src/transaction_verifier.rs
```diff
@@ -24,13 +25,13 @@ where
     pub fn new(
         rtx: &'a ResolvedTransaction,
         median_time_context: &'a M,
-        tip_number: BlockNumber,
+        block_number: BlockNumber,
         tip_epoch_number: EpochNumber,
```
keroro520: Suggest correcting the variable name `tip_epoch_number` together, and the type should be `EpochNumber`.

PR #876, opened 2019-05-22
feat: tweak consensus params

zhangsoledad:


spec/src/consensus.rs
```diff
@@ -23,10 +23,10 @@ pub(crate) const ORPHAN_RATE_TARGET_RECIP: u64 = 20;
 
 const MAX_BLOCK_INTERVAL: u64 = 60 * 1000; // 60s
 const MIN_BLOCK_INTERVAL: u64 = 5 * 1000; // 5s
-pub(crate) const EPOCH_DURATION_TARGET: u64 = 4 * 60 * 60 * 1000; // 4hours
+pub(crate) const EPOCH_DURATION_TARGET: u64 = 8 * 60 * 60 * 1000; // 4hours
```
doitian:
```suggestion
pub(crate) const EPOCH_DURATION_TARGET: u64 = 8 * 60 * 60 * 1000; // 8 hours
```

spec/src/consensus.rs
```diff
@@ -97,7 +97,7 @@ impl Default for Consensus {
             id: "main".to_owned(),
             max_uncles_age: MAX_UNCLE_AGE,
             max_uncles_num: MAX_UNCLE_NUM,
-            epoch_reward: capacity_bytes!(5_000_000),
+            epoch_reward: capacity_bytes!(500_000),
```
doitian: The config in `testnet.toml` is 5,000,000 bytes, but the default value here is 500,000 bytes. Is it by design that the default value is for mainnet?

doitian: `DEFAULT_SECONDARY_EPOCH_REWARD` is the same, the default value is 1/10 of the one in `testnet.toml`.

doitian: And we'd better keep consistent here, all the other fields use constants, only this one uses hard-coded number.

resource/specs/testnet.toml
```diff
@@ -33,7 +33,7 @@ code_hash = "0xb35557e7e9854206f7bc13e3c3a7fa4cf8892c84a09237fb0aab40aab3771eee"
 args = []
 
 [params]
-epoch_reward = 5_000_000_000_000_000
+epoch_reward = 500_000_000_000_000
```
doitian: It's hard to read a number in unit Shannon like this, I suggest group the last 8 digits like this:

```suggestion
epoch_reward = 5_000_000_00000000
```

It should also be applied to `secondary_epoch_reward`

spec/src/consensus.rs
```diff
@@ -11,6 +11,7 @@ use std::sync::Arc;
 
 // TODO: add secondary reward for miner
 pub(crate) const DEFAULT_SECONDARY_EPOCH_REWARD: Capacity = capacity_bytes!(5);
```
doitian:
```suggestion
pub(crate) const DEFAULT_SECONDARY_EPOCH_REWARD: Capacity = capacity_bytes!(50);
```

Keep consistent

PR #881, opened 2019-05-22
chore: improve build time

doitian:
This PR includes two commits, I'm not sure it worths to include the second one.

The first one is f26fde99, which avoids rebuilding when build info changed. It
relies on git to determine whether build info should be refreshed. When git is
not available, it falls back to the default behaviour, rebuild when any file
changes in the directory.

The second one is 176f4292. It moves most the code in the current crate into
a child crate `ckb-bin`. When there's only build info change in the repository,
we don't need to rebuild the crate execpt the version info. This change
requires a large code reorganized, and only reduces the build time from 16
seconds to 12 seconds, since most of the time spends in linking.


doitian:
Oops, I messed up the order of the two commits in the description. Fixed now.

doitian:
@nervos-bot-user ci status f26fde99f6b6c02a48d3f9e1cd843526ad198a5b
CI: success

PR #887, opened 2019-05-23
feat: Alert system

jjyr:
Introduce: 
https://en.bitcoin.it/wiki/Alert_system

Purpose: 
Implement the alert system in CKB for urgent situation,
In CKB early stage we may meet the same crisis bugs that Bitcoin meet,
in urgent case, CKB core team can send an alert message across CKB P2P network,
the client will show the alert message, the other behaviors of CKB node will not change.

The Alert System will be removed soon once the CKB network is considered mature.

Implement:
1. Add a new network protocol `alt`.
2. Add a default disabled RPC `send_alert`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -310,6 +310,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "ckb-alert-system"
+version = "0.1.0"
+dependencies = [
+ "bytes 0.4.12 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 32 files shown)


jjyr:
> Will the node relay the alert message? Is the message signature checked before relay?

https://github.com/nervosnetwork/ckb/pull/887/files#diff-f9056d164d152ac4dad0a349994e8c30R138

All alert messages need to be verified with signatures before the relay.

jjyr:
I decide split multisig to another PR.

PR #889, opened 2019-05-23
feat: add codename in version

doitian:
- Add codename in version
- BREAKING CHANGE: Bump genesis hash


src/main.rs
```diff
@@ -35,11 +35,13 @@ fn get_version() -> Version {
     #[cfg(docker)]
     let commit_describe = commit_describe.map(|s| s.replace("-dirty", ""));
     let commit_date = option_env!("COMMIT_DATE").map(ToString::to_string);
+    let code_name = Some("rylai-v2".to_string());
```
keroro520: I don't understand, we have defined `code_name` in spec/testnet.toml, so why here give code_name `rylai-v2" one more time?

What is the use of `code_name`?

doitian: The code name is for the app. The genesis message can be anything, it just happens to be code name now.

doitian: code name is a readable nickname for a `0.y.*` and `x.*.*`

PR #891, opened 2019-05-23
feat: secp256k1 multisig

jjyr:
Used in alert system https://github.com/nervosnetwork/ckb/pull/887

### DIFF
--- a/util/multisig/Cargo.toml
+++ b/util/multisig/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "multisig"
+version = "0.1.0"
+license = "MIT"
+authors = ["Nervos Core Dev <dev@nervos.org>"]
+edition = "2018"
+
+[dependencies]
+crypto = { path = "../crypto" }
+failure = "0.1.5"
+log = "0.4"
+
+[dev-dependencies]
+rand = "0.6.5"
+
--- a/util/multisig/src/error.rs
+++ b/util/multisig/src/error.rs
@@ -1,0 +1,36 @@
+use failure::Context;
+
+#[derive(Debug)]
+pub struct Error {
+    inner: Context<ErrorKind>,
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Fail)]
+pub enum ErrorKind {
+    #[fail(display = "The count of sigs should less than pks.")]
+    SigCountOverflow,
+    #[fail(display = "The count of sigs less than threshold.")]
+    SigNotEnough,
+    #[fail(display = "Failed to meet threshold {:?}.", _0)]
+    Threshold { threshold: usize, pass_sigs: usize },
+}
+
+impl Error {
+    pub fn kind(&self) -> ErrorKind {
+        *self.inner.get_context()
+    }
+}
+
+impl From<ErrorKind> for Error {
+    fn from(kind: ErrorKind) -> Error {
+        Error {
+            inner: Context::new(kind),
+        }
+    }
+}
+
+impl From<Context<ErrorKind>> for Error {
+    fn from(inner: Context<ErrorKind>) -> Error {
+        Error { inner }
+    }
+}
--- a/util/multisig/src/secp256k1.rs
+++ b/util/multisig/src/secp256k1.rs
@@ -1,0 +1,170 @@
+use crate::error::{Error, ErrorKind};
+pub use crypto::secp::{Error as Secp256k1Error, Message, Privkey, Pubkey, Signature};
+use log::{debug, trace};
+use std::collections::HashSet;
+use std::hash::BuildHasher;
+
+/// verify m of n signatures
+/// Example 2 of 3 sigs: [s1, s3], pks: [pk1, pk2, pk3]
+pub fn verify_m_of_n<S>(
+    message: &Message,
+    m_threshold: usize,
+    sigs: &[Signature],
+    pks: &HashSet<Pubkey, S>,
+) -> Result<(), Error>
+where
+    S: BuildHasher,
+{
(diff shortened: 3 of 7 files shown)


util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+    if m_threshold > sigs.len() {
+        Err(ErrorKind::SigNotEnough)?;
+    }
+    let verified_sig_count = sigs
+        .iter()
+        .zip(pks.iter())
+        .filter_map(|(sig, pk)| {
+            sig.and_then(|sig| {
+            trace!(target: "multisig", "verify sig {:x?} with pubkey {:x?}", &sig.serialize_compact()[..], &pk.serialize()[..]);
```
doitian: bad format

util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,146 @@
...
+                match SECP256K1.verify(&message, &sig, pk) {
+                Ok(()) => Some(()),
```
doitian: bad format

doitian: Still not indented 😂

util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,146 @@
...
+pub fn sign(message: &Message, sk: &SecretKey) -> Signature {
+    SECP256K1.sign(message, sk)
+}
+
+/// position of each sigs must according to the pk that sined the sig
+/// Example 2 of 3 sigs: [s1, None, s3], pks: [pk1, pk2, pk3]
+pub fn verify_m_of_n(
+    message: &Message,
+    m_threshold: usize,
+    sigs: Vec<Option<Signature>>,
```
quake: it's better to use slice since we don't need to add or remove elements from the vector.
```suggestion
    sigs: &[Option<Signature>],
    pks: &[PublicKey],
```

util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,146 @@
...
+    pks: Vec<PublicKey>,
+) -> Result<(), Error> {
+    if sigs.len() > pks.len() {
+        Err(ErrorKind::SigCountOverflow)?;
+    }
+    if m_threshold > sigs.len() {
```
quake: Need to filter none before checking threshold.

```suggestion
    if m_threshold > sigs.filter_map(|s| s).len() {
```

util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,147 @@
...
+                match SECP256K1.verify(&message, &sig, pk) {
+                    Ok(()) => Some(()),
+                    Err(err) => {
+                        debug!(target: "multisig", "verify secp256k1 sig error: {}", err);
+                        None
+                    }
+                }
+            })
+        })
+    .take(m_threshold)
```
doitian: Oops. I suggest remove that `trace!` line, format the code and add it back.

util/multisig/src/secp256k1.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+    if m_threshold > sigs.iter().filter_map(|s| *s).count() {
+        Err(ErrorKind::SigNotEnough)?;
+    }
+    let verified_sig_count = sigs
+        .iter()
+        .zip(pks.iter())
+        .filter_map(|(sig, pk)| {
+            sig.and_then(|sig| {
+                trace!(target: "multisig", "verify sig {:x?} with pubkey {:x?}", &sig.serialize_compact()[..], &pk.serialize()[..]);
```
doitian: Try to wrap this into multiple lines, maybe rustfmt will work again.

util/multisig/src/error.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Fail)]
+pub enum ErrorKind {
+    #[fail(display = "The count of sigs should less than pks.")]
+    SigCountOverflow,
+    #[fail(display = "The count of sigs less than threshold.")]
+    SigNotEnough,
+    #[fail(display = "Threshold not reach, sigs: {} required sigs: {}.", _0, _1)]
+    Threshold(usize, usize),
```
doitian: I personally dislike the tuple like structure when there are multiple fields, and the meaning of each field is not clear.

util/multisig/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "multisig"
+version = "0.1.0"
+license = "MIT"
+authors = ["Nervos Core Dev <dev@nervos.org>"]
+edition = "2018"
+
+[dependencies]
+secp256k1 = "0.12.2"
```
doitian: I lose the context, why we wrap the secp256k1 in `util/crypto` and why we don't use the wrapper here?

util/multisig/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Debug, Fail)]
+pub enum ErrorKind {
+    #[fail(display = "The count of sigs should less than pks.")]
+    SigCountOverflow,
+    #[fail(display = "The count of sigs less than threshold.")]
+    SigNotEnough,
+    #[fail(display = "Failed to meet threshold {:?}.", _0)]
+    Threshold(Threshold),
```
doitian: Enum supports struct like variant:

```suggestion
    Threshold {
        threshold: usize,
        pass_sigs: usize,
    },
```

jjyr: Ops didn't know this before

PR #895, opened 2019-05-24
fix: integration test send transaction capacity overflow

zhangsoledad:


test/src/specs/relay/transaction_relay.rs
```diff
@@ -61,12 +62,15 @@ impl Spec for TransactionRelayMultiple {
         });
 
         info!("Use generated block's cellbase as tx input");
+        let reward = block.transactions()[0].outputs()[0].capacity;
+        let txs_num = std::cmp::min(TXS_NUM as u64, reward.as_u64() / MIN_CAPACITY);
```
quake: need to change last assert code since we changed the total txs num.
```
for_each(|node| assert_eq!(node.get_tip_block().transactions().len(), TXS_NUM + 1));
```

zhangsoledad: :cry:

PR #897, opened 2019-05-24
feat: wrap the log macros to fix ill formed macros 

yangby-cryptape:
### Commits:

- chore: update the log filter template in configuration
- chore: apply new log macros to sync & relay
- chore: apply new log macros to util/* & ckb-bin
- chore: apply new log macros to verification & shared
- chore: apply new log macros to rpc & notify
- chore: apply new log macros to network
- chore: apply new log macros to miner
- chore: apply new log macros to chain
- **(important)** feat: wrap the log macros to fix ill formed macros

And, we have to update the log filters, add prefix `ckb-` for all our crates, since I use `target: env!("CARGO_PKG_NAME")`.


doitian:
Any plan to undraft? 

The team, @nervosnetwork/ckb-dev, Any option?

keroro520:
> Any plan to undraft?
> 
> The team, @nervosnetwork/ckb-dev, Any option?

Prefer to use `env!("CARGO_PKG_NAME").trim_left("ckb-")` as targets. And then we don't need to update the log filters.

yangby-cryptape:
Reply @keroro520:

- `env!("CARGO_PKG_NAME")` return a `const &'static str`, no more costs.
- `trim_start(..)` is not a `const` function.
  That means if the level of logs is enabled, it will take extra cost.
- `trim_left(..)` is deprecated since 1.33.0: superseded by `trim_start(..)`.

keroro520:
> trim_start(..) is not a const function.

Soga. Thanks for your reply.
Why not undraft this PR?

TheWaWaR:
https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#macros-using-local_inner_macros

PR #905, opened 2019-05-27
feat: add indexer related rpc

quake:
This PR added 5 indexer related [rpc](https://github.com/nervosnetwork/ckb/blob/e98046dac81e56d74d0f4c69035e8d0e08e290c0/rpc/README.md#indexer), most lines of code are unit / integration test and document, core logic is in [indexer/src/store.rs](https://github.com/nervosnetwork/ckb/blob/e98046dac81e56d74d0f4c69035e8d0e08e290c0/indexer/src/store.rs)


wallet/src/store.rs
```diff
@@ -0,0 +1,1196 @@
...
+    }
+
+    pub fn start<S: ToString>(self, thread_name: Option<S>, notify: &NotifyController) {
+        let mut thread_builder = thread::Builder::new();
+        if let Some(name) = thread_name {
+            thread_builder = thread_builder.name(name.to_string());
+        }
+
+        let new_tip_receiver = notify.subscribe_new_tip(WALLET_STORE_SUBSCRIBER);
+        thread_builder
```
xxuejie: Is there a thread pool we can borrow to help run the tasks here? I feel like a single thread might not be the best choice since it could be blocked by many things. Or at least leave a comment till the next time we are refactoring things.

wallet/src/store.rs
```diff
@@ -0,0 +1,1196 @@
...
+            .spawn(move || loop {
+                select! {
+                    recv(new_tip_receiver) -> msg => match msg {
+                        Ok(tip_changes) => self.update(&tip_changes.detached_blocks, &tip_changes.attached_blocks),
+                        _ => {
+                            error!(target: "wallet", "new_tip_receiver closed");
+                            break;
```
xxuejie: Correct me if I'm wrong: does this mean that at the first error, this thread would stop processing more tip updates? If so, I feel like this is too big an assumption.

wallet/src/store.rs
```diff
@@ -0,0 +1,1196 @@
...
+    }
+
+    pub fn sync_index_states(&self) {
+        info!(target: "wallet", "Start sync index states with chain store");
+        let mut lock_hash_index_states = self.get_lock_hash_index_states();
+        if lock_hash_index_states.is_empty() {
+            return;
+        }
+        // need to lock chain state, avoids inconsistent state in processing
+        let chain_state = self.shared.lock_chain_state();
```
xxuejie: Like the previous one, can we periodically release and re-acquire the lock to give other threads requesting the lock a chance to succeed?

wallet/src/store.rs
```diff
@@ -0,0 +1,1196 @@
...
+    }
+
+    fn remove_lock_hash(&self, lock_hash: &H256) {
+        self.commit_batch(|batch| {
+            let iter = self
+                .db
+                .iter(COLUMN_LOCK_HASH_LIVE_CELL, lock_hash.as_bytes())
+                .expect("wallet db iter should be ok");
+
+            iter.take_while(|(key, _)| key.starts_with(lock_hash.as_bytes()))
```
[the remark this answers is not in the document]
quake: we can't construct an `end_key` here for `delete_range`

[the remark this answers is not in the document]
xxuejie: All the keys you want to delete here starts with `lock_hash`, you can just treat `lock_hash` as an integer, add 1 to it, you will get the first key succeeding all the keys you need to delete here.

wallet/src/store.rs
quake: this error only occurs while channel closing, which means that main thread exits unexpectedly.

util/jsonrpc-types/src/wallet.rs
```diff
@@ -0,0 +1,31 @@
...
+
+// This is used as return value of get_transactions_by_lock_hash RPC
+#[derive(Serialize, Deserialize)]
+pub struct CellTransaction {
+    pub created_by: TransactionPoint,
+    pub consumed_by: Option<TransactionPoint>,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct TransactionPoint {
```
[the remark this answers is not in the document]
quake: we can't use `CellOutPoint` here, because of `index` here is a pointer to transaction outputs (create cell) or inputs (consume cell)

rpc/README.md
```diff
@@ -1074,3 +1080,230 @@ http://localhost:8114
...
+            "block_number": "1024",
+            "lock_hash": "0x9a9a6bdbc38d4905eace1822f85237e3a1e238bb3f277aa7b7c8903441123510"
+        }
+    ]
+}
+```
+
+### `get_transactions_by_lock_hash`
+
+Returns the transactions collection by the hash of lock script.
```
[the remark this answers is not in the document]
quake: there is a similar rpc `listtransactions` in bitcoin, in bitcoin, each transaction is categorized as `receive` or `send`, but using a paired `generated_by/consumed_by` is easier for api users.

util/jsonrpc-types/src/wallet.rs
```diff
@@ -0,0 +1,31 @@
...
+pub struct LiveCell {
+    pub created_by: TransactionPoint,
+    pub cell_output: CellOutput,
+}
+
+// This is used as return value of get_transactions_by_lock_hash RPC
+#[derive(Serialize, Deserialize)]
+pub struct CellTransaction {
+    pub created_by: TransactionPoint,
+    pub consumed_by: Option<TransactionPoint>,
```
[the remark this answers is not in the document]
quake: open for discussion

wallet/src/store.rs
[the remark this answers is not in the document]
quake: we need to delete `cell_out_point_lock_hash` which is a reverse index of lock hash, it can't utilize the `delete_range` function, to make it simple, I keep the same `iter/delete` pattern in `delete_lock_hash_transaction`.

quake:
@xxuejie most review issues have been resolved, others are open for discussion, please review again, thanks.

indexer/src/store.rs
```diff
@@ -0,0 +1,1241 @@
...
+/// | COLUMN_CELL_OUT_POINT_LOCK_HASH | CellOutPoint  | LockHashCellOutput       |
+/// +---------------------------------+---------------+--------------------------+
+
+const COLUMN_LOCK_HASH_INDEX_STATE: Col = 0;
+const COLUMN_LOCK_HASH_LIVE_CELL: Col = 1;
+const COLUMN_LOCK_HASH_TRANSACTION: Col = 2;
+const COLUMN_CELL_OUT_POINT_LOCK_HASH: Col = 3;
+
+pub trait IndexerStore: Sync + Send {
+    fn get_live_cells(&self, lock_hash: &H256, skip_num: usize, take_num: usize) -> Vec<LiveCell>;
```
jjyr: Need to support confirmation condition to make it work in the real world. http://chainquery.com/bitcoin-api/listunspent

quake: There is a `block_number` field in each live cell, api users can use this field and tip number to calculate the numbers of confirmation blocks.

db/src/lib.rs
```diff
@@ -33,6 +32,14 @@ pub trait KeyValueDB: Sync + Send {
     fn batch(&self) -> Result<Self::Batch>;
 }
 
+pub trait IterableKeyValueDB: KeyValueDB {
+    fn iter<'a>(
+        &'a self,
+        col: Col,
+        from_key: &'a [u8],
+    ) -> Result<Box<Iterator<Item = KeyValueIteratorItem> + 'a>>;
```
doitian: Why not associate type?

quake: I tried to use an associate type, however, it will introduce lifetime flag to the trait definition. I would like to keep simple trait design since we only have one implementation currently.

indexer/src/store.rs
```diff
@@ -0,0 +1,1240 @@
...
+        // attach blocks until reach tip or batch limit
+        let mut lock_hash_index_states = self.get_lock_hash_index_states();
+        let min_block_number: BlockNumber = lock_hash_index_states
+            .values()
+            .min_by_key(|index_state| index_state.block_number)
+            .expect("none empty index states")
+            .block_number;
+
+        // need to lock chain state, avoids inconsistent state in processing
+        let chain_state = self.shared.lock_chain_state();
```
jjyr: There is a way to avoid chain_state lock.
We can find the tip block first, then find it's ancestors until min_block_number.

jjyr: Since it is unnecessary to keep indexer 100% consistent with chain, I think it is worth to try to avoid the big `chain_state` lock, in case chain_state changed the indexer can resume the consistent in next sync.

indexer/src/store.rs
```diff
@@ -0,0 +1,1240 @@
...
+                index_lock_hashes.insert(lock_hash.to_owned());
+
+                let mut block = self
+                    .shared
+                    .store()
+                    .get_block(&index_state.block_hash)
+                    .expect("block exists");
+                // detach blocks until reach a block on main chain
+                self.commit_batch(|batch| {
+                    self.detach_block(batch, &index_lock_hashes, &block);
```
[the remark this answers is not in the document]
quake: in this loop, the list of `lock_hash_index_states` are lock hashes on fork chain, so detach all blocks until reach a block on main chain

indexer/src/store.rs
```diff
@@ -286,10 +286,12 @@ impl<CS: ChainStore + 'static> DefaultIndexerStore<CS> {
             .expect("none empty index states")
             .block_number;
 
-        // need to lock chain state, avoids inconsistent state in processing
-        let chain_state = self.shared.lock_chain_state();
+        let (tip_number, tip_hash) = {
+            let chain_state = self.shared.lock_chain_state();
```
jjyr: Consider using `store.get_tip_header` to completely avoid the lock?

quake: > Consider using `store.get_tip_header` to completely avoid the lock?

Resolved, please review again, thanks.

TheWaWaR:
Conflicted

doitian:
How to handle schema migration?

quake:
> How to handle schema migration? The main database stores the version into the database to check compatibility.

Currently, there are no version checking in the indexer store, I would like to refactor version and corruption checking of rocksdb to an utility function in future PR.

rpc/src/module/indexer.rs
```diff
@@ -0,0 +1,121 @@
...
+                (page.0 as usize).saturating_mul(per_page),
+                per_page,
+                reverse_order.unwrap_or_default(),
+            )
+            .into_iter()
+            .map(Into::into)
+            .collect())
+    }
+
+    fn index_lock_hash(
```
doitian: If I call this RPC repeatedly without removing the index first, the index may be in an invalid state.

For example:

- Assume that the current tip is 7. Call this RPC with `index_from` 5. Wait for the background thread to sync up to 7.
- Now tip becomes 10, and the background thread is still on 7.
- Call this RPC again with `index_from` 10.
- Background thread starts again, it will ignore blocks 8, 9

indexer/src/store.rs
```diff
@@ -0,0 +1,1298 @@
...
+                    batch.insert_lock_hash_index_state(lock_hash, &index_state);
+                });
+            });
+
+        // attach blocks until reach tip or batch limit
+        let mut lock_hash_index_states = self.get_lock_hash_index_states();
+        let min_block_number: BlockNumber = lock_hash_index_states
+            .values()
+            .min_by_key(|index_state| index_state.block_number)
+            .expect("none empty index states")
```
doitian: Should not `expect` here. It's in the multiple threads environment, line 274 cannot guarantee that the map is still non-empty here.

indexer/src/store.rs
```diff
@@ -0,0 +1,1298 @@
...
+                        .filter(|(_, index_state)| index_state.block_number < block_number)
+                        .map(|(lock_hash, _)| lock_hash)
+                        .cloned()
+                        .collect();
+                    let block = self
+                        .shared
+                        .get_ancestor(&tip_hash, block_number)
+                        .and_then(|header| self.shared.store().get_block(&header.hash()))
+                        .expect("block exists");
+                    self.attach_block(batch, &index_lock_hashes, &block);
```
doitian: Must check that the block is a child of the last indexed block. Because the index state may change between `detach` and `attach` when the thread is interrupted.

rpc/json/rpc.json
```diff
@@ -629,6 +629,155 @@
...
+            "1024"
+        ],
+        "result": {
+            "block_hash": "0xf35168c2e2e0c494ec97233091d5de51b7e5af7376bbc3d7572fc6438e2bb032",
+            "block_number": "1024",
+            "lock_hash": "0x9a9a6bdbc38d4905eace1822f85237e3a1e238bb3f277aa7b7c8903441123510"
+        },
+        "types": [
+            {
+                "index_from": "Create an index from starting block number, an optional parameter, null means starting from tip"
```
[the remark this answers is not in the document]
doitian: The doc must mention whether the block `index_from` is indexed.

indexer/src/store.rs
```diff
@@ -0,0 +1,1298 @@
...
+                .expect("tip header exists");
+            (tip_header.number(), tip_header.hash().to_owned())
+        };
+        self.commit_batch(|batch| {
+            (min_block_number + 1..=tip_number)
+                .take(BATCH_ATTACH_BLOCK_NUMS)
+                .for_each(|block_number| {
+                    let index_lock_hashes = lock_hash_index_states
+                        .iter()
+                        .filter(|(_, index_state)| index_state.block_number < block_number)
```
doitian: It seems we can never index the genesis block. Users can have cells in the genesis block.

test/src/rpc.rs
```diff
@@ -215,6 +216,72 @@ impl RpcClient {
             .call()
             .expect("rpc call process_block_without verify")
     }
+
+    pub fn get_live_cells_by_lock_hash(
```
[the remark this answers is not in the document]
quake: > Why add another suite of RPCs for test?

it's a helper fn for integration test.

rpc/src/module/indexer.rs
quake: it's rpc caller's responsibility to use an appropriate `index_from` value, indexer store won't handle this case.

indexer/src/store.rs
quake: won't fix this as same reason of https://github.com/nervosnetwork/ckb/pull/905#discussion_r294279649

rpc/json/rpc.json
```diff
@@ -629,6 +629,155 @@
...
+            },
+            {
+                "lock_hash": "Cell lock script hash"
+            }
+        ]
+    },
+    {
+        "description": "Get lock hash index states",
```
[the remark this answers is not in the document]
quake: will return empty array.

rpc/json/rpc.json
```diff
@@ -629,6 +629,155 @@
...
+            {
+                "per": "Page size, max value is 50"
+            },
+            {
+                "reverse_order": "Returns the live cells collection in reverse order, an optional parameter, default is false"
+            }
+        ]
+    },
+    {
+        "description": "Returns the transactions collection by the hash of lock script.",
```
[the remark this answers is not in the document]
quake: There are no error case for this rpc

PR #909, opened 2019-05-28
refactor: move best_known_headers/last_common_headers into peer_state

keroro520:
* `best_known_header` and `last_common_header` are both  state of peer. So I move them into `PeerState`. And add wrap functions for them, so that hides the locks outside.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -4,7 +4,6 @@ use crate::{BLOCK_DOWNLOAD_WINDOW, MAX_BLOCKS_IN_TRANSIT_PER_PEER, PER_FETCH_BLO
 use ckb_core::header::Header;
 use ckb_network::PeerIndex;
 use ckb_store::ChainStore;
-use ckb_util::try_option;
 use log::{debug, trace};
 use numext_fixed_hash::H256;
 use numext_fixed_uint::U256;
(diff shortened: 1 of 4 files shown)


sync/src/synchronizer/headers_process.rs
```diff
@@ -233,7 +232,7 @@ where
 
         if log_enabled!(target: "sync", log::Level::Debug) {
             let chain_state = self.synchronizer.shared.lock_chain_state();
-            let peer_state = self.synchronizer.peers.best_known_header(self.peer);
+            let peer_state = self.synchronizer.peers.get_best_known_header(self.peer);
```
TheWaWaR: variable name not match the value

TheWaWaR: variable name not match the value, better use `peer_best_known_header` or something close.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -48,38 +47,30 @@ where
...
     }
 
     pub fn last_common_header(&self, best: &HeaderView) -> Option<Header> {
-        let mut guard = self.synchronizer.peers.last_common_headers.write();
-
-        let last_common_header = try_option!(guard.get(&self.peer).cloned().or_else(|| {
-            if best.number() < self.tip_header.number() {
+        let last_common_header = {
+            if let Some(header) = self.synchronizer.peers().get_last_common_header(self.peer) {
+                Some(header.clone())
```
TheWaWaR: `header` already have ownership, no need `clone()`

PR #653, opened 2019-05-07
feat: add rpc experiment module

keroro520:
* Add rpc `dry_run_transaction`
* Add rpc `_compute_transaction_id`
* Enable Experiment moduel by default

PR #910, opened 2019-05-28
feat: network alert

jjyr:
Introduce: 
https://en.bitcoin.it/wiki/Alert_system

Purpose: 
Implement the alert system in CKB for urgent situation,
In CKB early stage we may meet the same crisis bugs that Bitcoin meet,
in urgent case, CKB core team can send an alert message across CKB P2P network,
the client will show the alert message, the other behaviors of CKB node will not change.

The Network Alert will be removed soon once the CKB network is considered mature.

Implement:
1. Add a new network protocol `alt`.
2. Add a default disabled RPC `send_alert`.

### DIFF
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -8,7 +8,7 @@ use ckb_miner::BlockAssembler;
 use ckb_network::{CKBProtocol, NetworkService, NetworkState};
 use ckb_network_alert::alert_relayer::AlertRelayer;
 use ckb_notify::NotifyService;
-use ckb_rpc::RpcServer;
+use ckb_rpc::{RpcServer, ServiceBuilder};
 use ckb_shared::shared::{Shared, SharedBuilder};
 use ckb_store::ChainStore;
 use ckb_sync::{NetTimeProtocol, NetworkProtocol, Relayer, SyncSharedState, Synchronizer};
@@ -116,16 +116,27 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
         .start(version, Some("NetworkService"))
         .expect("Start network service failed");
 
-    let rpc_server = RpcServer::new(
-        args.config.rpc,
-        network_controller,
-        shared,
-        synchronizer,
-        chain_controller,
-        block_assembler_controller,
-        alert_notifier,
-        alert_verifier,
-    );
+    let builder = ServiceBuilder::new(&args.config.rpc)
+        .enable_chain(shared.clone())
+        .enable_pool(shared.clone(), network_controller.clone())
+        .enable_miner(
+            shared.clone(),
+            network_controller.clone(),
+            chain_controller.clone(),
+            block_assembler_controller,
+        )
+        .enable_net(network_controller.clone())
+        .enable_stats(shared.clone(), synchronizer, alert_notifier.clone())
+        .enable_experiment(shared.clone())
+        .enable_integration_test(
+            shared.clone(),
+            network_controller.clone(),
+            chain_controller.clone(),
+        )
+        .enable_alert(alert_verifier, alert_notifier, network_controller);
+    let io_handler = builder.build();
+
+    let rpc_server = RpcServer::new(args.config.rpc, io_handler);
 
     wait_for_exit();
 
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -126,7 +126,7 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
             block_assembler_controller,
         )
         .enable_net(network_controller.clone())
-        .enable_stats(shared.clone(), synchronizer, alert_notifier.clone())
+        .enable_stats(shared.clone(), synchronizer, Arc::clone(&alert_notifier))
         .enable_experiment(shared.clone())
         .enable_integration_test(
             shared.clone(),
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -6,7 +6,7 @@ use ckb_db::RocksDB;
 use ckb_logger::info_target;
 use ckb_miner::BlockAssembler;
 use ckb_network::{CKBProtocol, NetworkService, NetworkState};
-use ckb_network_alert::{alert_relayer::AlertRelayer, config::Config as AlertConfig};
+use ckb_network_alert::alert_relayer::AlertRelayer;
 use ckb_notify::NotifyService;
 use ckb_rpc::RpcServer;
 use ckb_shared::shared::{Shared, SharedBuilder};
@@ -75,7 +75,8 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
         Arc::clone(synchronizer.peers()),
     );
     let net_timer = NetTimeProtocol::default();
-    let alert_relayer = AlertRelayer::new(version.to_string(), AlertConfig::default());
+    let alert_config = args.config.alert.unwrap_or_default();
+    let alert_relayer = AlertRelayer::new(version.to_string(), alert_config);
 
     let alert_notifier = alert_relayer.notifier();
     let alert_verifier = alert_relayer.verifier();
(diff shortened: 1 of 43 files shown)


quake:
need integration test for this new feature

rpc/src/module/alert.rs
```diff
@@ -0,0 +1,66 @@
...
+            Ok(()) => {
+                let fbb = &mut FlatBufferBuilder::new();
+                let message = AlertMessage::build_alert(fbb, &alert);
+                fbb.finish(message, None);
+                let data = fbb.finished_data().into();
+                if let Err(err) = self
+                    .network_controller
+                    .broadcast(NetworkProtocol::ALERT.into(), data)
+                {
+                    log::error!(target: "rpc", "Broadcast alert failed: {:?}", err);
```
keroro520: Should we return the p2p errors back to the client?

util/network-alert/src/alert_relayer.rs
```diff
@@ -0,0 +1,135 @@
...
+            known_lists: LruCache::new(KNOWN_LIST_SIZE),
+        }
+    }
+
+    pub fn notifier(&self) -> Arc<Mutex<Notifier>> {
+        Arc::clone(&self.notifier)
+    }
+
+    pub fn verifier(&self) -> Arc<Verifier> {
+        Arc::clone(&self.verifier)
```
keroro520: Avoid Arc clone implicitly.

util/network-alert/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+//! Network Alert will be removed soon once the CKB network is considered mature.
+//!
+pub mod alert_relayer;
+pub mod config;
+pub mod notifier;
+pub mod verifier;
+
+use std::time::Duration;
+
+pub(crate) const BAD_MESSAGE_BAN_TIME: Duration = Duration::from_secs(5 * 60);
```
keroro520: "malformed alert"?

util/network-alert/src/notifier.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn add(&mut self, alert: Arc<Alert>) {
+        if self.has_received(alert.id) {
+            return;
+        }
+        // checkout cancel_id
+        if alert.cancel > 0 {
+            self.cancel(alert.cancel);
+        }
+        // add to received alerts
+        self.received_alerts.insert(alert.id, Arc::clone(&alert));
```
keroro520: `if insert.is_some() return;` ?

keroro520: Can we merge `noticed_alerts` and `received_alerts`?

util/network-alert/src/alert_relayer.rs
jjyr: Why? This there some best practice recommendations?

util/network-alert/src/lib.rs
jjyr: Just use the same name `sync:: BAD_MESSAGE_BAN_TIME` is ok.
https://github.com/nervosnetwork/ckb/blob/develop/sync/src/lib.rs#L69

util/network-alert/src/notifier.rs
jjyr: There are for the different purpose, `noticed_alerts` store alerts that the current node should be noticed through RPC or other approaches. 

`received_alerts` store all alerts we received, even the current node ignores, we still store them for relay to other peers.

rpc/src/module/alert.rs
jjyr: Not sure, I' d instead send another PR to modify this and `send_transaction` RPC if needed.

util/network-alert/src/alert_relayer.rs
keroro520: Sorry that I don't know any best practices around it.
I just feel like this style since https://github.com/nervosnetwork/ckb/pull/653#discussion_r281888622

util/network-alert/src/notifier.rs
keroro520: Got it.

So what if we give the alert messages some identifiers, like

```
enum {
  Cancel(alert),
  Normal(alert),
  ...
}
```

I'm not sure which one is simpler.

util/network-alert/src/alert_relayer.rs
jjyr: Ok, make sense to me, I' ll update it.

core/src/alert.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.inner.max_version = max_version;
+        self
+    }
+
+    pub fn priority(mut self, priority: u32) -> Self {
+        self.inner.priority = priority;
+        self
+    }
+
+    pub fn signatures(mut self, signatures: Vec<Bytes>) -> Self {
```
doitian: I prefer the name `extend_signatures` to make the function behavior clear.

rpc/src/server.rs
```diff
@@ -1,32 +1,38 @@
...
 use jsonrpc_server_utils::cors::AccessControlAllowOrigin;
 use jsonrpc_server_utils::hosts::DomainsValidation;
+use std::sync::Arc;
 
 pub struct RpcServer {
     pub(crate) server: Server,
 }
 
 impl RpcServer {
+    #[allow(clippy::too_many_arguments)]
```
doitian: It's better to refactor this. When the alert is disabled, why do we still need to provide `alert_notifier` and `alert_verifier`. The clippy linters all have their reason, we should not just disable them.

doitian: We can split it into modules, with several `enable_xxx(xx_arg1, xx_arg2` methods, and use a `start` method to really start the service.

rpc/src/module/stats.rs
```diff
@@ -36,14 +40,22 @@ impl<CS: ChainStore + 'static> StatsRpc for StatsRpcImpl<CS> {
...
+            .join("\n");
 
         Ok(ChainInfo {
             chain,
             median_time: Timestamp(median_time),
             epoch: EpochNumber(epoch),
             difficulty,
             is_initial_block_download,
-            warnings: String::new(),
+            warnings,
```
doitian: I think we can just use `alerts` here.

core/src/alert.rs
jjyr: According to current CKB builder naming convention, use plural represents `extend`.

We should discuss the naming in other issues, this PR maintains the insistence of naming.

rpc/src/module/stats.rs
jjyr: Alert includes `signatures` and other fields which user does not care about, it's better to use another struct if we decide to expose more information.

But for now, I'm not sure what's field should be exposed other than a simple message.

jjyr: I' think we can return `id`, `received_at`, `notice_until` and `message` for now.

``` json
{
   id: 1,
   received_at: 1559641204,
   notice_until: 1559641204,
   message: "a alert message"
}
```

doitian: I mean the field name

ckb-bin/src/subcommand/run.rs
```diff
@@ -98,19 +104,39 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
...
+        )
+        .enable_net(network_controller.clone())
+        .enable_stats(shared.clone(), synchronizer, Arc::clone(&alert_notifier))
+        .enable_experiment(shared.clone())
+        .enable_integration_test(
+            shared.clone(),
+            network_controller.clone(),
+            chain_controller.clone(),
+        )
+        .enable_alert(alert_verifier, alert_notifier, network_controller);
```
doitian: Still has the problem I mentioned before. Is it by design that event RPC is disabled, we still need to relay alerts?

jjyr: Yes, the RPC is designed for administrators to send alerts; By default, every client needs to relayer alert.

util/network-alert/src/notifier.rs
```diff
@@ -0,0 +1,93 @@
...
+use lru_cache::LruCache;
+use std::sync::Arc;
+
+const CANCEL_FILTER_SIZE: usize = 128;
+
+pub struct Notifier {
+    /// cancelled alerts
+    cancel_filter: LruCache<u32, ()>,
+    /// alerts we received
+    received_alerts: FnvHashMap<u32, Arc<Alert>>,
```
quake: are `received_alerts` values same as `noticed_alerts`? why need to store duplicate data here?

keroro520: https://github.com/nervosnetwork/ckb/pull/910#discussion_r289673881

PR #913, opened 2019-05-29
New verification model

xxuejie:
Based on feedbacks gathered earlier, we are revising our verification
model with the following changes:

* When validating a transaction, CKB will grab all lock scripts from
  all inputs, and group them based on lock script hash. The script in
  each group will only be run once. The lock script itself will have
  to do the validation task for all inputs in the same group
* CKB will also grab all type scripts from inputs and outputs(notice
  different from previous version, the type scripts in inputs are
  included here as well), and group them based on type script hash as
  well. Each type script in each group will also be run once. The type
  script itself needs to handle the validation task within the group
* Syscalls are also revised to allow fetching all the
  inputs/outputs/witnesses within a single group.
* Input args is removed since the functionality can be replicated elsewhere


core/src/transaction.rs
```diff
@@ -204,41 +204,38 @@ impl OutPoint {
 pub struct CellInput {
     pub previous_output: OutPoint,
     pub since: u64,
-    // Depends on whether the operation is Transform or Destroy, this is the proof to transform
-    // lock or destroy lock.
-    pub args: Vec<Bytes>,
+    // Only used for cellbase input, normal input doesn't need this field
+    pub block_number: BlockNumber,
```
janx: Is this a temporary workaround?

script/src/syscalls/load_witness.rs
```diff
@@ -1,26 +1,37 @@
...
+        LoadWitness {
+            witnesses,
+            group_inputs,
+        }
     }
 
-    fn fetch_witness(&self, index: usize) -> Option<&Witness> {
-        self.witnesses.get(index).cloned()
+    fn fetch_witness(&self, source: Source, index: usize) -> Option<&Witness> {
+        match source {
```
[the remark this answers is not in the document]
xxuejie: Yes, I don't think attaching witnesses to inputs makes sense. What if there are more witnesses than inputs? What if outputs require witnesses as well?

script/src/syscalls/mod.rs
```diff
@@ -79,20 +79,60 @@ impl InputField {
...
+const SOURCE_GROUP_FLAG: u64 = 0x0100_0000_0000_0000;
+const SOURCE_GROUP_MASK: u64 = 0xFF00_0000_0000_0000;
+const SOURCE_ENTRY_MASK: u64 = 0x00FF_FFFF_FFFF_FFFF;
+
 #[derive(Debug, PartialEq, Clone, Copy, Eq)]
 enum Source {
-    Input = 1,
-    Output = 2,
-    Dep = 3,
+    Normal(SourceEntry),
```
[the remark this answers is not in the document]
xxuejie: `Source` is a name we kept using in RFCs, I'd argue if we are using tx or group context is just the details of a source.

script/src/verify.rs
```diff
@@ -98,12 +123,36 @@ impl<'a, CS: ChainStore> TransactionScriptsVerifier<'a, CS> {
...
+                }
+            }
+        }
+        for (i, output) in rtx.transaction.outputs().iter().enumerate() {
+            if let Some(t) = &output.type_ {
+                let type_group_entry = script_groups
+                    .entry(t.hash())
+                    .or_insert_with(|| ScriptGroup::new(&t));
+                type_group_entry.output_indices.push(i);
+            }
```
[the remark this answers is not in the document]
xxuejie: Oops this is a bug.

core/src/transaction.rs
xxuejie: Nope this is a solution I'm proposing here. Since args is absent we need a way to tell the difference between each cellbase input.

script/src/syscalls/load_cell.rs
```diff
@@ -27,28 +29,50 @@ impl<'a, CS: LazyLoadCellOutput + 'a> LoadCell<'a, CS> {
...
+            Source::Group(SourceEntry::Input) => self
+                .group_inputs
+                .get(index)
+                .ok_or(INDEX_OUT_OF_BOUND)
+                .and_then(|actual_index| {
+                    self.resolved_inputs
+                        .get(*actual_index)
+                        .ok_or(INDEX_OUT_OF_BOUND)
+                })
+                .and_then(|r| r.cell().ok_or(ITEM_MISSING)),
```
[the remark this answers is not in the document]
xxuejie: I actually thought about this, and I think one additional benefit here is that cycles can be saved from those syscalls on not-in-group inputs. In addition, I couldn't find a case where we need to map group indices to the original indices, so I'm opting for current solution now.

script/src/syscalls/load_witness.rs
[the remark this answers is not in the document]
janx: I don't understand your question, not sure if I miss something.

The mapping between witness and input is one-to-one, for every input there must be one and only one witness. In `load_input` syscall there're branches for both `Source::Normal(SourceEntry::Input)` and `Source::Group(SourceEntry::Input)`, so I'm wondering why there is only `Source::Group(SourceEntry::Input)` in `load_witness`.

core/src/transaction.rs
janx: We can use cellbase input's witness.

script/src/syscalls/load_cell.rs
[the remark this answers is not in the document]
janx: Good to me.

script/src/syscalls/load_witness.rs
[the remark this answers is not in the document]
xxuejie: Hmm, I actually have slightly different understanding here: to me, it's more of a convention, not a standard that each input has one and only one witness. The question is: our model is different from bitcoin, should we expand from this limitation? If a type script requires a signature, we can put it in one witness item as well. Any thoughts?

core/src/transaction.rs
xxuejie: The problem with witness, is that it won't be included when calculating TX hash. So if a miner creates 2 cellbase transactions within the epoch, the hash would be identical, hence there's no way to tell one from the other.

script/src/syscalls/load_witness.rs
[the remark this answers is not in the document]
janx: By 'witness' I actually mean 'witness slot'. The slot can be empty or carry arguments needed by some outputs, but the slot number is the same as input number in a transaction. With this constraint it's easy to get corresponding witness for an input (just use the same index) in normal case.

core/src/transaction.rs
janx: Another choice: can we require cellbase tx to have a `dep` refer to its own header?

xxuejie: Isn't that a circular dependency? A header would include transaction root, which needs the hashes of all transactions(including cellbase tx), and the hash of the cellbase tx cannot be calculated until we fill in the header hash in `dep`.

script/src/syscalls/load_witness.rs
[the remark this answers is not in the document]
doitian: That's a circular dependency. Block hash depends on cellbase hash, cellbase requires block hash

core/src/transaction.rs
doitian: That's a circular dependency. Block hash depends on cellbase hash, cellbase requires block hash

doitian: But we can let it depends on parent block header

xxuejie: What about use `block_hash` to reference parent block header? We can keep `cell` in `OutPoint` as null to represent a cellbase OutPoint, but `block_hash` can be used (it's designed to reference a block header anyway).

script/src/syscalls/load_witness.rs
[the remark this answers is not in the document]
xxuejie: Okay this makes sense to me now, a fix is added.

core/src/transaction.rs
jjyr: It still smells a bit strange, what about force the first output use block_number as data?

xxuejie: After some thought I do feel it's better if we leverage the data field of the first output in the cellbase transaction. We can add a verification rule so that the first 4 bytes of the first output's data should store the block number. This still leaves space if the miner wants to encode more message. On the other hand, this avoid one more quirky rule in OutPoint handling.

Thoughts? @doitian @janx

script/src/verify.rs
```diff
@@ -98,12 +124,37 @@ impl<'a, CS: ChainStore> TransactionScriptsVerifier<'a, CS> {
...
             }
+            // here we are only pre-processing the data, verify method validates
+            // each input has correct script setup.
+            if let Some(cell_meta) = resolved_input.cell.cell_meta() {
+                let output = store.lazy_load_cell_output(cell_meta);
+                let lock_group_entry = lock_groups
+                    .entry(output.lock.hash())
+                    .or_insert_with(|| ScriptGroup::new(&output.lock));
+                lock_group_entry.input_indices.push(i);
+                if let Some(t) = output.type_ {
```
jjyr: Why do we need to care about the type of previous output?

script/src/verify.rs
```diff
@@ -237,75 +264,63 @@ impl<'a, CS: ChainStore> TransactionScriptsVerifier<'a, CS> {
...
-            .get(&withdraw_header_hash)
-            .ok_or(ScriptError::InvalidDaoWithdrawHeader)?;
-
         let mut maximum_output_capacities = Capacity::zero();
-        for (input, resolved_input) in self.inputs.iter().zip(self.resolved_inputs.iter()) {
+        for input_index in &script_group.input_indices {
+            // Each DAO witness should contain 3 arguments in the following order:
+            // 0. pubkey(33 bytes) from witness
+            // 1. signature from witness
+            // 2. withdraw block hash(32 bytes) from input args
```
jjyr: this arg is also from witness now.

core/src/transaction.rs
janx: The first-class inflation will stop when block issuance is smaller than cellbase output capacity requirement. Use cellbase output to store extra data has impact on first-class inflation.

janx: Should be ok given our current second-class issuance setting.

xxuejie: Another side of the point is: if the issuance is not enough for a cell, there shouldn't be a cell base tx, correct?

script/src/verify.rs
xxuejie: In the new verification model, type scripts from both inputs and outputs will be run when validating a transaction.

core/src/transaction.rs
janx: I think we can just assume that won't happen to simplify the code.

xxuejie: This is now updated.

jjyr: Another approach (which I think is better) is to reuse `since` field to store block_number, and we can make sure it works without addition assumption, leave the first output data to other usages.

xxuejie: I'm not sure the logic of validating `since` alone is enough here. We will still need to verify if the `since` field stores the correct block number. Otherwise an attacker or a careless user might just skip filling in `since`, still resulting in cellbase transactions that are identical.

And if we will need the logic to verify block number anyway, I think putting it in data part is better for separation of concerns. In addition, it's not that the data field for this output is not usable, we are only reserving the first 8 bytes for block number. The miner is still free to put whatever data he/she wants after the initial 8 bytes.

janx: I actually think use `since` for block number is better, as `since` is designed for block number. Another not-so-important point to avoid use cellbase data here is it also increases occupied capacity which affects second-issuance to miner.

A rule to make sure cellbase input filled `since` with current block number is still needed.

jjyr: Yes, we will need the same logic to check the block number. 

The reason I like the `since` idea is we do not need to rely on output data in our consensus, the initial 8 bytes implies that the first output cannot be constrained under some special types, which limit the usage, for example, a potential usage, can a miner put outputs into NervosDAO directly?

Just some thoughts, I do not strongly insist on this.

jjyr: For note, put `block_number` to `since` is against the original `since` rule, the default `0x0000` flags require tip number at least equals to `since`, which is always fails because in our case the block number is `tip + 1`.

I think we can break the `since` rule for cell base tx, because in semantics `since` is meaningless for cell base tx.

xxuejie: I have mixed feelings on this: on the one hand, I actually agree that using since is better than output data part due to the output usage point mentioned above, on the other hand, I don't think it's ideal to break since rule just for cell base, we've had enough special rules, and adding one here is only gonna add to the complexity.

One thing I'm thinking is: can we expand since rule with another flag to require that the block number put in since field must equal current block number? This way we won't break since rule, we are just adding a feature here.

jjyr: My mistake, actually, it does not break the `since` rule, when validating a block we should use block_number as tip_number.

xxuejie: I've updated the code to keep block number in `since` field, I can confirm that original since rule works here.

zhangsoledad: upvote `since`, `cellbase` with block number is also helpful for reward tx proposer.

core/src/transaction.rs
```diff
@@ -204,41 +204,33 @@ impl OutPoint {
...
         } = self;
-        (previous_output, since, args)
+        (previous_output, since)
     }
 
     pub fn serialized_size(&self) -> usize {
-        self.previous_output.serialized_size()
-            + mem::size_of::<u64>()
-            + self.args.iter().map(Bytes::len).sum::<usize>()
+        self.previous_output.serialized_size() + mem::size_of::<u64>() + mem::size_of::<u64>()
```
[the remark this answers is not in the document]
xxuejie: Oops this is due to the original added `block_number`, changing this now

PR #914, opened 2019-05-29
refactor: add pow work struct

quake:
This PR added `Work` struct to pow module and minor refactoring, it avoids unnecessary calculation of `pow_hash` and `target_difficulty` in the inner loop, and it is required to implement the feature of miner status displaying.


quake:
Will submit a new PR to replace this one

PR #915, opened 2019-05-29
fix: Sync blocked by protected peer

TheWaWaR:


janx:
What is a protected peer?

TheWaWaR:
@janx 
For security reason (do not eviction outbound peer too often) we mark a fixed number(current=4) outbound peer as protected peer.

PR #919, opened 2019-05-30
feat: synchronizer and relayer share BlocksInflight

keroro520:
* test: Add test for losing BlockTXN messages

* fix: Remove relayer's compact_block_filter, filter duplicated compact blocks via BlocksInFlight
    Currently, we discard the new arriving compact blocks if we had already received the same one, via `RelayState::compact_block_filter`. The purpose is avoiding re-processing/re-requesting the same compact blocks. But the problem is, if we failed to reconstruct a compact block, we cannot process this compact block anymore, before receiving the corresponding transactions from peer1. That's terrible if peer1's inventory was lost on the way.

    Here I remove `RelayState::compact_block_filter`, and insteadly use `Peers::blocks_inflight` to avoid re-requesting to same peers.

* fix: Cache pending compact block at peer-level, but not global-level

### DIFF
--- a/sync/src/relayer/block_transactions_process.rs
+++ b/sync/src/relayer/block_transactions_process.rs
@@ -4,6 +4,7 @@ use ckb_network::{CKBProtocolContext, PeerIndex};
 use ckb_protocol::{cast, BlockTransactions, FlatbuffersVectorIterator};
 use ckb_store::ChainStore;
 use failure::Error as FailureError;
+use std::collections::hash_map::Entry;
 use std::convert::TryInto;
 use std::sync::Arc;
 
@@ -31,29 +32,31 @@ impl<'a, CS: ChainStore + 'static> BlockTransactionsProcess<'a, CS> {
 
     pub fn execute(self) -> Result<(), FailureError> {
         let block_hash = cast!(self.message.block_hash())?.try_into()?;
-        if let Some(compact_block) = self
-            .relayer
-            .state
-            .pending_compact_blocks
-            .lock()
-            .remove(&block_hash)
-        {
-            let transactions: Vec<Transaction> =
-                FlatbuffersVectorIterator::new(cast!(self.message.transactions())?)
-                    .map(TryInto::try_into)
-                    .collect::<Result<_, FailureError>>()?;
+        let pending_compact_blocks = &self.relayer.state.pending_compact_blocks;
+        if let Entry::Occupied(mut pending) = pending_compact_blocks.lock().entry(block_hash) {
+            let (compact_block, peers_set) = pending.get_mut();
+            if peers_set.remove(&self.peer) {
+                let transactions: Vec<Transaction> =
+                    FlatbuffersVectorIterator::new(cast!(self.message.transactions())?)
+                        .map(TryInto::try_into)
+                        .collect::<Result<_, FailureError>>()?;
 
-            let ret = {
-                let chain_state = self.relayer.shared.lock_chain_state();
-                self.relayer
-                    .reconstruct_block(&chain_state, &compact_block, transactions)
-            };
+                let ret = {
+                    let chain_state = self.relayer.shared.lock_chain_state();
+                    self.relayer
+                        .reconstruct_block(&chain_state, compact_block, transactions)
+                };
 
-            if let Ok(block) = ret {
-                self.relayer
-                    .accept_block(self.nc.as_ref(), self.peer, &Arc::new(block));
+                // TODO Add this (compact_block, peer) into RecentRejects if reconstruct_block failed?
+                // TODO Add this block into RecentRejects if accept_block failed?
+                if let Ok(block) = ret {
+                    pending.remove();
+                    self.relayer
+                        .accept_block(self.nc.as_ref(), self.peer, &Arc::new(block));
+                }
             }
         }
+
         Ok(())
     }
 }
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -43,18 +43,6 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
         let compact_block: CompactBlock = (*self.message).try_into()?;
         let block_hash = compact_block.header.hash().to_owned();
 
-        if self.relayer.state.already_known_compact_block(&block_hash) {
-            debug_target!(
-                crate::LOG_TARGET_RELAY,
-                "discarding already known compact block {:x}",
-                block_hash
-            );
-            return Ok(());
-        }
-        self.relayer
-            .state
-            .mark_as_known_compact_block(block_hash.clone());
-
         if let Some(parent_header_view) = self
             .relayer
             .shared
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -76,6 +76,19 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
             return Ok(());
         }
 
+        if let Some(flight) = self
+            .relayer
+            .peers
+            .blocks_inflight
+            .read()
+            .inflight_state_by_block(&block_hash)
+        {
+            if flight.peers.contains(&self.peer) {
+                debug!(target: "relay", "discarding already in-flight compact block {:x}", block_hash);
+                return Ok(());
+            }
+        }
+
         // The new arrived has greater difficulty than local best known chain
         let mut missing_indexes: Vec<usize> = Vec::new();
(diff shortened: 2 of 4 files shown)


test/src/specs/relay/compact_block.rs
```diff
@@ -152,13 +143,71 @@ impl CompactBlockBasic {
...
+            "Node0 should send GetBlockTransactions message for missing transactions",
+        );
+
+        // Submit the new block to node1. We expect node1 will relay the new block to node0.
+        node1.submit_block(&block);
+        assert!(
+            wait_until(10, || node0.get_tip_block() == node1.get_tip_block()),
+            "Node1 should relay the new block to node0, and node0 should accept it"
+        );
+    }
```
keroro520: Without this PR, the test will fail here, node0's tip is not equal to node1. since node0 will discard the arriving compact block relayed from node1.

sync/src/relayer/block_transactions_process.rs
```diff
@@ -31,29 +31,31 @@ impl<'a, CS: ChainStore + 'static> BlockTransactionsProcess<'a, CS> {
...
-            .state
-            .pending_compact_blocks
-            .lock()
-            .remove(&block_hash)
-        {
-            let transactions: Vec<Transaction> =
-                FlatbuffersVectorIterator::new(cast!(self.message.transactions())?)
-                    .map(TryInto::try_into)
-                    .collect::<Result<_, FailureError>>()?;
+        let mut pending_compact_blocks = self.relayer.state.pending_compact_blocks.lock();
```
quake: I prefer `entry` to `get_mut` here:
```
if let Entry::Occupied(pending) = pending_compact_blocks.entry(&block_hash) {
    let (compact_block, peers_set) = pending.get_mut();

    ...

    if let Ok(block) = ret {
      pending.remove();
      ...
    }
}
```

sync/src/relayer/compact_block_process.rs
```diff
@@ -43,22 +43,25 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
         let compact_block: CompactBlock = (*self.message).try_into()?;
         let block_hash = compact_block.header.hash().to_owned();
 
-        if self.relayer.state.already_known_compact_block(&block_hash) {
-            debug!(target: "relay", "discarding already known compact block {:x}", block_hash);
+        let parent = self
```
quake: prefer `match` to `unwrap` here:
```
match self.relayer.shared.get_header_view(compact_block.header.parent_hash()) {
    None => {...},
    Some(parent) =>  {...}
}
```

sync/src/relayer/compact_block_process.rs
```diff
@@ -69,35 +72,50 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
+                };
                 let resolver = HeaderResolverWrapper::new(
                     &compact_block.header,
                     self.relayer.shared.shared().to_owned(),
                 );
                 let header_verifier = HeaderVerifier::new(
                     CompactBlockMedianTimeView {
                         anchor_hash: compact_block.header.hash(),
-                        pending_compact_blocks: &pending_compact_blocks,
+                        fn_get_pending_header: Box::new(fn_get_pending_header),
```
quake: I don't understand why change to use closure, could you explain the detail issue here?

test/src/specs/relay/compact_block.rs
```diff
@@ -46,58 +48,50 @@ impl CompactBlockBasic {
...
+        let new_tx = node.new_transaction(node.get_tip_block().transactions()[0].hash().clone());
+        node.submit_block(
+            &node
                 .new_block_builder(None, None, None)
                 .proposal(new_tx.proposal_short_id())
                 .build(),
         );
         (0..3).for_each(|_| {
-            node0.generate_block();
+            node.generate_block();
```
quake:
```suggestion
            node.generate_blocks(3);
```

test/src/specs/relay/compact_block.rs
```diff
@@ -108,20 +102,17 @@ impl CompactBlockBasic {
...
```
quake:
```suggestion
            node.generate_blocks(3);
```

test/src/specs/relay/compact_block.rs
```diff
@@ -152,13 +143,71 @@ impl CompactBlockBasic {
...
+        node0.generate_block();
+        let new_tx = node0.new_transaction(node0.get_tip_block().transactions()[0].hash().clone());
+        node0.submit_block(
+            &node0
+                .new_block_builder(None, None, None)
+                .proposal(new_tx.proposal_short_id())
+                .build(),
+        );
+        // Proposal a tx, and grow up into proposal window
+        (0..6).for_each(|_| {
```
quake:
```suggestion
        node0.generate_blocks(6);
```

test/src/specs/relay/compact_block.rs
```diff
@@ -152,13 +143,71 @@ impl CompactBlockBasic {
...
+            node0.generate_block();
+        });
+
+        // Make node0 and node1 reach the same height
+        node1.generate_block();
+        node0.connect(node1);
+        assert!(wait_until(20, || node0.get_tip_block() == node1.get_tip_block()),);
```
quake: we can use the helper fn `waiting_for_sync`: 
```suggestion
    node0.waiting_for_sync(node1, 9, 20);
```

test/src/specs/relay/compact_block.rs
```diff
@@ -152,13 +143,71 @@ impl CompactBlockBasic {
...
+        let message = get_root::<RelayMessage>(&data).unwrap();
+        assert_eq!(
+            message.payload_type(),
+            RelayPayload::GetBlockTransactions,
+            "Node0 should send GetBlockTransactions message for missing transactions",
+        );
+
+        // Submit the new block to node1. We expect node1 will relay the new block to node0.
+        node1.submit_block(&block);
+        assert!(
```
quake: please use `waiting_for_sync` helper function.

sync/src/relayer/compact_block_process.rs
keroro520: Already check `parent.is_none()` below. So here `unwrap` is safe.

sync/src/relayer/compact_block_process.rs
keroro520: The structure of `pending_compact_blocks` has been changed, 
from 

```
Mutex<FnvHashMap<H256, CompactBlock>
```

to
 
```
Mutex<FnvHashMap<H256, (CompactBlock, FnvHashSet<PeerIndex>)>>
```

The `CompactBlockMedianTimeView` don't care the second element `FnvHashSet<PeerIndex>`. So I use closure to wrap it.

Another personal reason is, I think passing closure is more graceful, and I don't see any overhead here.

quake: `CompactBlockMedianTimeView`  is a private struct, I would like to keep it simple instead of reusable, how about using a type alias in this case? 

```
pub type PendingCompactBlocks = FnvHashMap<H256, (CompactBlock, FnvHashSet<PeerIndex>)>
```

```
struct RelayState {
    pending_compact_blocks: Mutex<PendingCompactBlocks>,
    ...
}

struct CompactBlockMedianTimeView {
    pending_compact_blocks: &'a PendingCompactBlocks,
    ...
}
```

test/src/specs/relay/compact_block.rs
```diff
@@ -152,13 +139,66 @@ impl CompactBlockBasic {
...
+        let message = get_root::<RelayMessage>(&data).unwrap();
+        assert_eq!(
+            message.payload_type(),
+            RelayPayload::GetBlockTransactions,
+            "Node0 should send GetBlockTransactions message for missing transactions",
+        );
+
+        // Submit the new block to node1. We expect node1 will relay the new block to node0.
+        node1.submit_block(&block);
+        node0.waiting_for_sync(node1, node0.get_tip_block().header().number(), 20);
```
quake: node1's tip number is higher than node0, should use node1 tip number as the target, this test will fail randomly otherwise.

```suggestion
        node0.waiting_for_sync(node1, node1.get_tip_block().header().number(), 20);
```

sync/src/relayer/compact_block_process.rs
keroro520: I do have already refactored it in other PR, but it has to wait for its prior PRs merged:
https://github.com/keroro520/ckb/pull/6
https://github.com/keroro520/ckb/pull/5
https://github.com/keroro520/ckb/pull/4

I want to leave it later.

doitian:

PR #922, opened 2019-05-31
 feat: proposer reward

zhangsoledad:
### consensus
1. earliest transaction proposer get 40% of the transaction fee as a reward.
2. block reward finalized after proposal window close.
3. enforce one-input one-output one-witness on cellbase.

### database
1. block_ext.txs_verified -> block_ext.verified
2. txs_fees persistence 

## Enhancement

1. lazy static always success script for test
2. introduce `AsCapacity` make `Capacity` interface more flexible


spec/src/lib.rs
```diff
@@ -207,6 +215,12 @@ impl ChainSpec {
             .set_max_block_cycles(self.params.max_block_cycles)
             .set_pow(self.pow.clone());
 
+        // if (consensus.cellbase_maturity() <= consensus.tx_proposal_window().start() + 1) {
+        //     return Err(SpecLoadError::consensus_params(
+        //         "cellbase_maturity MUST greater than tx_proposal_window incremented by one",
+        //     ));
+        // }
+
```
janx: Remove these comments?

sync/src/tests/synchronizer.rs
```diff
@@ -2,11 +2,12 @@ use crate::synchronizer::{BLOCK_FETCH_TOKEN, SEND_GET_HEADERS_TOKEN, TIMEOUT_EVI
 use crate::tests::TestNode;
 use crate::{Config, NetworkProtocol, SyncSharedState, Synchronizer};
 use ckb_chain::chain::ChainService;
-use ckb_chain_spec::consensus::Consensus;
+use ckb_chain_spec::{consensus::Consensus, Foundation};
 use ckb_core::block::BlockBuilder;
 use ckb_core::header::HeaderBuilder;
+// use ckb_core::script::Script;
```
yangby-cryptape: If u have to rebase, u can remove this.

spec/src/lib.rs
```diff
@@ -105,6 +112,10 @@ impl SpecLoadError {
     fn genesis_mismatch(expect: H256, actual: H256) -> Box<Self> {
         Box::new(SpecLoadError::GenesisMismatch { expect, actual })
     }
+
+    // fn consensus_params(reason: String) -> Box<Self> {
```
yangby-cryptape: Remove this.

spec/src/consensus.rs
```diff
@@ -44,6 +45,10 @@ impl ProposalWindow {
     pub fn start(&self) -> BlockNumber {
```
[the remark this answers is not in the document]
doitian: From proposing block's perspective, `block_number + self.0` is the `start` that its proposals become effective, and `block_number + self.1` is the `end` after which its proposals become out dated.

But from the current block's perspective, `block_number - self.1` is the `start` which proposals can commit, and `block_number - self.1` is the `end`.

The function name `start`, `end` and the response type `BlockNumber` also make me think they will return the block number of the start and end block, but indeed they return the offsets.

I'll start a new thread for this, lets just ignore it in the PR since it is not introduced here.

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Debug, PartialEq, Clone, Eq, Fail)]
+pub enum Error {
+    #[fail(display = "Can't resolve finalize target: {}", _0)]
+    Target(BlockNumber),
+    #[fail(display = "Can't parse Script from target witness: {:x}", _0)]
+    Script(H256),
+}
+
+pub struct RewardCalculator<'a, P> {
```
doitian: I think we can add some ascii chart here to describe how the reward is calculated.

doitian:
Plan to review `RewardCalculator` tomorrow morning. Need more focus.

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    ) -> Result<Capacity, FailureError> {
+        let mut target_proposals = self.get_proposal_ids_by_hash(target.hash());
+
+        let proposal_window = self.provider.consensus().tx_proposal_window();
+        let block_number = parent.number() + 1;
+        let store = self.provider.store();
+
+        let mut reward = Capacity::zero();
+        let commit_start = cmp::max(block_number.saturating_sub(proposal_window.length()), 2);
+        let proposal_start = cmp::max(commit_start.saturating_sub(proposal_window.start()), 1);
```
doitian: `competing_propposal_start` and `competing_proposal_table` are better names.

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+            .map(ProposalShortId::from_tx_hash)
+            .zip(
+                store
+                    .get_block_ext(index.hash())
+                    .expect("block body stored")
+                    .txs_fees
+                    .iter(),
+            )
+        {
+            if target_proposals.remove(&id) {
```
doitian: Suggest comment:

```
// target block is the earliest block with effective proposals for the parent block
```

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+                reward = reward.safe_add(tx_fee.safe_mul_ratio(4, 10)?)?;
+            }
+        }
+
+        index = store
```
doitian: Put it in the front of `while`, we can eliminate the duplicate index update.

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+                tx_fee.safe_mul_ratio(4, 10).and_then(|proposer| {
+                    tx_fee
+                        .safe_sub(proposer)
+                        .and_then(|miner| acc.safe_add(miner))
+                })
+            })
+            .map_err(Into::into)
+    }
+
+    pub fn proposal_reward(
```
doitian: Some notes for other reviewers:

The proposal window is `(2, 10)`.

- target block is `t`
- current block is `t + 11`
- parent block is `t + 10`
- `commit_start` is `t + 2`, which is the earliest block which can commit txs in target block's proposals.
- `proposal_start` is `t - 8`, which is the earliest block which proposals are competing with the target block.

util/reward-calculator/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        let proposal_window = self.provider.consensus().tx_proposal_window();
+        let block_number = parent.number() + 1;
+        let store = self.provider.store();
+
+        let mut reward = Capacity::zero();
+        let commit_start = cmp::max(block_number.saturating_sub(proposal_window.length()), 2);
+        let proposal_start = cmp::max(commit_start.saturating_sub(proposal_window.start()), 1);
+
+        let mut proposal_table = BTreeMap::new();
```
doitian: I propose a simpler algorithm below.

We only need to keep a single `HashSet`, which tracks the proposal ids that have already been proposed in previous blocks. This can avoid repeat 8 `flat_map`s.

We name the `HashSet` as `proposed`, which stores txs already proposed before the target block.

The loop starts from parent block, backward to `commit_start`. For each block `bn`:

- Find proposals of `bn - 10`, add the proposals to `proposed` unless `bn - 10 == target`.
- For each tx in block bn, if tx is not in `proposed` but in target's proposal, sum the proposal reward.

util/reward-calculator/src/lib.rs
doitian: Alternatively, the proposed ids can be removed from `target_proposals`, no need to maintain another hash set.

doitian:
The whole PR looks good to me, the only change I requested is:

> (4, 10) should be defined as consensus parameters.

The others are just refactoring suggestions, they have lower priority and we can improve them later.

util/reward-calculator/src/lib.rs
doitian: Alternatively, the proposed ids can be removed from target_proposals, no need to maintain another hash set.

spec/src/lib.rs
```diff
@@ -91,6 +97,7 @@ pub enum SpecLoadError {
     FileNotFound,
     ChainNameNotAllowed(String),
     GenesisMismatch { expect: H256, actual: H256 },
+    ConsensusParams(String),
```
keroro520: Seems never used `ConsensusParams`.

core/src/transaction.rs
```diff
@@ -490,8 +490,14 @@ impl Transaction {
         &self.witnesses
     }
 
+    // one-in one-out one-wit
     pub fn is_cellbase(&self) -> bool {
-        self.inputs.len() == 1 && self.inputs[0].previous_output.is_null()
```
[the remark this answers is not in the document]
zhangsoledad: We didn't provision the number of the output before, but now, miner creates `cellbase` not only for their own,  so, a new rule here, the cellbase must only have one output.
I moved other checks to verifier.

doitian:
[ios_scan_4250993727.pdf](https://github.com/nervosnetwork/ckb/files/3275267/ios_scan_4250993727.pdf)

This diagram demonstrates how the reward is finalized.

spec/src/consensus.rs
```diff
@@ -84,8 +84,8 @@ pub struct Consensus {
     // block version number supported
     pub max_block_proposals_limit: u64,
     pub genesis_epoch_ext: EpochExt,
-    // foundation lock
-    pub foundation: Foundation,
+    // genesis lock
```
doitian: better remove this comment

Issue #904, opened 2019-05-27
Optimize error message on rpc call `get_block_template` MethodNotFound


doitian:
Prompt user to enable Miner and setup `block_assembler` in `ckb.toml`

PR #926, opened 2019-05-31
feat: make a better error message for miner when method not found

yangby-cryptape:
For #904.


keroro520:
Should we exit the miner program after logging the warning message?

yangby-cryptape:
> Should we exit the miner program after logging the warning message?

You can just restart your `CKB server` then the `CKB miner` can works well again.

PR #928, opened 2019-05-31
chore: reload build info on commit

doitian:


build.rs
```diff
@@ -20,7 +20,7 @@ fn rerun_if_changed(path_str: &str) -> bool {
 
 fn main() {
     let files_stdout = std::process::Command::new("git")
-        .args(&["ls-tree", "--name-only", "HEAD"])
+        .args(&["ls-tree", "-r", "--name-only", "HEAD"])
```
yangby-cryptape: Is this `-r` dispensable?

doitian: The behavior is undefined when adding a directory to `build-if-changed`. Some system may report the directory has changed when any file has changed, some only when new files are added or existing files are deleted. So it better to list all the files instead.

PR #930, opened 2019-05-31
[HOLD] refactor: Flatten sync states into SyncSharedState

keroro520:
* refactor: Flatten relayer's state and synchronizer's state into SyncSharedState.
* refactor: Flatten peers' state into SyncSharedState.
* feat: Hide locks, avoid taking locks too long outside (Not done yet, will continue in future)
* fix: Clear related state when peer disconnects


keroro520:
I will split this PR into several smaller PRs later.

PR #931, opened 2019-06-01
feat: add a function to select all tx-hashes from storage for a block

yangby-cryptape:


driftluo:
Although the code is correct, I don't quite know what the role of this interface is.

yangby-cryptape:
> Although the code is correct, I don't quite know what the role of this interface is.

@zhangsoledad asked me how to get only transaction-hashes for a block. He want to calculate all short ids for transactions in a block.

And, some JSON-RPC protocols, such as `get_block*` can return only hashes (instead of full data) in default, as a optimization.

PR #932, opened 2019-06-01
[HOLD] feat: ban relaying invalid compact blocks

keroro520:
* feat: Ban relaying invalid compact blocks. Treat a compact block as invalid when `Relayer::accept_block` returns an error, and ban the corresponding sourced peer.

---

NOTE: Currently, when `Synchronizer::accept_block` returns an error, we just log a message but not ban the sourced peer. Maybe we should ban the sourced peer when the error is not "unknown parent".

### DIFF
--- a/sync/src/relayer/block_transactions_process.rs
+++ b/sync/src/relayer/block_transactions_process.rs
@@ -52,7 +52,7 @@ impl<'a, CS: ChainStore + 'static> BlockTransactionsProcess<'a, CS> {
                 if let Ok(block) = ret {
                     pending.remove();
                     self.relayer
-                        .accept_block(self.nc.as_ref(), self.peer, &Arc::new(block));
+                        .accept_block(self.nc.as_ref(), self.peer, &Arc::new(block))?;
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


doitian:
Should we ban the peer on any error?

keroro520:
> Should we ban the peer on any error?

There are 2 kinds of errors returned by `accept_block`: invalid block, and database errors.

I think we should ban invalid block, or involve `misbehavior` to lower the peer score;
As for database errors, we should just log it but not ban.
Any suggestions?

doitian:
> > Should we ban the peer on any error?
> 
> There are 2 kinds of errors returned by `accept_block`: invalid block, and database errors.
> 
> I think we should ban invalid block, or involve `misbehavior` to lower the peer score;
> As for database errors, we should just log it but not ban.
> Any suggestions?

lgtm

doitian:
Why there are two commits having the same commit message?

keroro520:
> Why there are two commits having the same commit message?

```
feat: Ban relaying invalid blocks 
test: Ban relaying invalid blocks
```

Add a test, and then fix the code.

PR #933, opened 2019-06-01
chore: group sentry events via fingerprint

doitian:


miner/src/client.rs
```diff
@@ -149,10 +149,6 @@ impl Client {
                 }
                 Err(e) => {
                     error!("rpc call submit_block error: {:?}", e);
-                    sentry::capture_message(
```
doitian: Duplicated of the errors reported in rpc miner.

yangby-cryptape:
Ref: [Customize Grouping with Fingerprints](https://docs.sentry.io/data-management/rollups/?platform=rust#custom-grouping).

PR #936, opened 2019-06-03
chore: remove some unsafe

u2:
Is the unsafe necessary?

I tried to add `#[cfg(test)]` for `build_unchecked` of `Block`, but the compiling failed.

### DIFF
--- a/store/src/flat_block_body.rs
+++ b/store/src/flat_block_body.rs
@@ -139,15 +139,13 @@ pub(crate) fn deserialize_transaction(
     let witnesses: Vec<Witness> = config.deserialize(&tx[header[5]..])?;
     let hash: H256 = config.deserialize(&tx[header[6]..])?;
     let witness_hash: H256 = config.deserialize(&tx[header[7]..])?;
-    unsafe {
-        Ok(TransactionBuilder::default()
-            .version(version)
-            .deps(deps)
-            .inputs(inputs)
-            .outputs(outputs)
-            .witnesses(witnesses)
-            .build_unchecked(hash, witness_hash))
-    }
+    Ok(TransactionBuilder::default()
+        .version(version)
+        .deps(deps)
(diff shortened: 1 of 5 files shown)


yangby-cryptape:
- If users use the `build(..)` to build those structs, the hashes will be correct.
- If users use the `build_unchecked(..)` to build those structs, the correctness of those hashes can not be guaranteed.

So, both @driftluo and @TheWaWaR suggested that add `unsafe` to let users pay attention to those methods.

In fact, we shouldn't use these methods in non-test code. We tried using `$[cfg(test)]`, but the `$[cfg(test)]` can't passed from crates to crates. (We add a `$[cfg(test)]` in crate A, and run `cargo test` in crate B, crate A couldn't know that it should be enable the `test` feature.)

store/src/flat_block_body.rs
```diff
@@ -137,17 +137,13 @@ pub(crate) fn deserialize_transaction(
...
-    unsafe {
-        Ok(TransactionBuilder::default()
-            .version(version)
-            .deps(deps)
-            .inputs(inputs)
-            .outputs(outputs)
-            .witnesses(witnesses)
-            .build_unchecked(hash, witness_hash))
-    }
+    Ok(TransactionBuilder::default()
```
yangby-cryptape: We wrote `hashes` into storage as caches. `build(..)` always computes those hashes again.

u2: updated.

doitian:
It is intended to mark those functions as unsafe.

u2:
> It is intended to mark those functions as unsafe.

The unsafe in Rust means that not enforcing the memory safety guarantees.

doitian:
It's the best way we know to prevent using the function by accident. It is not acceptable to simply remove the unsafe.

PR #938, opened 2019-06-03
refactor: remove low S check from util-crypto

jjyr:
Cause we already use witness, the low S rule for signature is unnecessary.

See https://github.com/bitcoin/bips/blob/master/bip-0062.mediawiki#new-rules

### DIFF
--- a/util/crypto/src/secp/signature.rs
+++ b/util/crypto/src/secp/signature.rs
@@ -50,15 +50,6 @@ impl Signature {
         Signature(sig)
     }
 
-    /// Check if this is a "low" signature.
-    pub fn is_low_s(&self) -> bool {
-        if let Ok(h_s) = H256::from_slice(self.s()) {
-            h_s <= HALF_N
-        } else {
-            false
-        }
-    }
-
     /// Check if each component of the signature is in range.
     pub fn is_valid(&self) -> bool {
         let h_r = match H256::from_slice(self.r()) {


util/crypto/src/secp/signature.rs
```diff
@@ -50,15 +50,6 @@ impl Signature {
         Signature(sig)
     }
 
-    /// Check if this is a "low" signature.
-    pub fn is_low_s(&self) -> bool {
```
janx: where is this function used?

util/crypto/src/secp/signature.rs
```diff
@@ -74,7 +65,7 @@ impl Signature {
                 return false;
             }
         };
-        self.v() <= 1 && h_r < N && h_r >= ONE && h_s < N && h_s >= ONE
+        self.v() < 4 && h_r < N && h_r >= ONE && h_s < N && h_s >= ONE
```
janx: why is this change needed?

util/crypto/src/secp/signature.rs
jjyr: Not used in current CKB, this PR intends to remove it to avoid confusion.

util/crypto/src/secp/signature.rs
jjyr: After checking the source code of secp256k1 and peter wullies's answer, it seems in practice the `v` is only considered either `0` or `1`.

I'll rollback this.

https://github.com/bitcoin-core/secp256k1/blob/544435fc90a5672d862e2a51f44c10251893b97d/src/ecdsa_impl.h#L298

https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v

jjyr: After checking the source code of secp256k1 and peter wullies's answer, it seems in practice the `v` is only considered either `0` or `1`.

I'll rollback this.

https://github.com/bitcoin-core/secp256k1/blob/544435fc90a5672d862e2a51f44c10251893b97d/src/ecdsa_impl.h#L298

https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v

Issue #900, opened 2019-05-25
Do not reuse the error `CapacityOverflow`


doitian:
The error `CapacityOverflow` means different things, it's hard to differentiate them:

- In a single output cell, the capacity is not enough to hold the cell serialized size.
- Numeric overflow when summing up inputs capacity or outputs capacity.

doitian:
We should leave `CapacityOverflow` for the numeric overflow error. Use `InsufficentCellCapacity` for cell occupied capacity verification.

PR #878, opened 2019-05-22
fix: calculate the current median time from tip

keroro520:
**BREAKING CHANGE**: Original implementation use `[Tip-BlockMedianCount .. Tip-1]` to calculate the current block median time. According to the notion of BlockMedianTime in [bip-0113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki#specification) , here change to use `[Tip-BlockMedianCount+1 .. Tip]` instead.

PR #940, opened 2019-06-03
[ᚬmaster] 🎉 Rc/v0.14.0

doitian:
This is the release for Rylai V3. The testnet will reset and deploy this version on Jun 15.

---

### Features

* #908: Peers handle disconnect (@keroro520)

    * Handle disconnected inside `Peers`
    * Rename `n_sync_started`/`n_protected_outbound_peers`
    * Remove imlicitly clone for `Peers`

* #891: Secp256k1 multisig (@jjyr)

    Used in alert system https://github.com/nervosnetwork/ckb/pull/887

* #845: Limit TXO set memory usage (@yangby-cryptape)

    This is a breaking change: b:database
    
    ### Commits
    
    - feat: add a function to traverse all keys for database
    
    - feat: save cell set into database (but not used)
    
      BREAKING CHANGE: Add new column to store cell set into database.
    
    - refactor: loading CellSet do not have to load all blocks
    
    - perf: delete transactions whose all cells are dead
    
    - fix: the order to apply cell set diff is very important
    
    ### Issue
    
    #228
    
    ### Notes
    
    - I insist on the opinion: save live cells is better than save transactions which include at least one live cell.
    
      We can use a only-key-and-no-value database (for example, Redis) to store live cells (`CellOutPoint` as key). 
    
      I tried, but it's too difficult to adapt to current logic, so I gave up.
    
    - I didn't change current design about how to use `CellSet`, I just made them be persisted.
    
    - **All transactions** which include at least one live cell, **are still kept in memory**.
    
      But **the transactions which don't have any live cells are removed from memory**.

* #874: Revise uncle rule (@zhangsoledad)

    This is a breaking change: b:consensus, b:database
    
    1. get rid uncle age limit
    2. try include disconnected block as uncle

* #920: Tweak consensus params (@zhangsoledad)

    This is a breaking change: b:consensus
    
    tweak block_cycles_limit and min_block_interval

* #897: Wrap the log macros to fix ill formed macros (@yangby-cryptape)

    ### Commits:
    
    - chore: update the log filter template in configuration
    - chore: apply new log macros to sync & relay
    - chore: apply new log macros to util/* & ckb-bin
    - chore: apply new log macros to verification & shared
    - chore: apply new log macros to rpc & notify
    - chore: apply new log macros to network
    - chore: apply new log macros to miner
    - chore: apply new log macros to chain
    - **(important)** feat: wrap the log macros to fix ill formed macros
    
    And, we have to update the log filters, add prefix `ckb-` for all our crates, since I use `target: env!("CARGO_PKG_NAME")`.

* #919: Synchronizer and relayer share BlocksInflight (@keroro520)

    * test: Add test for losing BlockTXN messages
    
    * fix: Remove relayer's compact_block_filter, filter duplicated compact blocks via BlocksInFlight
        Currently, we discard the new arriving compact blocks if we had already received the same one, via `RelayState::compact_block_filter`. The purpose is avoiding re-processing/re-requesting the same compact blocks. But the problem is, if we failed to reconstruct a compact block, we cannot process this compact block anymore, before receiving the corresponding transactions from peer1. That's terrible if peer1's inventory was lost on the way.
    
        Here I remove `RelayState::compact_block_filter`, and insteadly use `Peers::blocks_inflight` to avoid re-requesting to same peers.
    
    * fix: Cache pending compact block at peer-level, but not global-level

* #924: Add a transaction error `InsufficientCellCapacity` (@yangby-cryptape)

    For #900.

* #926: Make a better error message for miner when method not found (@yangby-cryptape)

    For #904.

### Bug Fixes

* #878: Calculate the current median time from tip (@keroro520)

    This is a breaking change: b:consensus
    
    **BREAKING CHANGE**: Original implementation use `[Tip-BlockMedianCount .. Tip-1]` to calculate the current block median time. According to the notion of BlockMedianTime in [bip-0113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki#specification) , here change to use `[Tip-BlockMedianCount+1 .. Tip]` instead.

* #915: Sync blocked by protected peer (@TheWaWaR)
* #906: Proposal table reload (@zhangsoledad)
* fix: Cuckoo cycle verification bug (@yangby-cryptape)

### Improvements

* #909: Move best_known_headers/last_common_headers into peer_state (@keroro520)

    * `best_known_header` and `last_common_header` are both  state of peer. So I move them into `PeerState`. And add wrap functions for them, so that hides the locks outside.

### Misc

* #913: New verification model (@xxuejie)

    This is a breaking change: b:consensus, b:database, b:p2p, b:rpc
    
    Based on feedbacks gathered earlier, we are revising our verification
    model with the following changes:
    
    * When validating a transaction, CKB will grab all lock scripts from
      all inputs, and group them based on lock script hash. The script in
      each group will only be run once. The lock script itself will have
      to do the validation task for all inputs in the same group
    * CKB will also grab all type scripts from inputs and outputs(notice
      different from previous version, the type scripts in inputs are
      included here as well), and group them based on type script hash as
      well. Each type script in each group will also be run once. The type
      script itself needs to handle the validation task within the group
    * Syscalls are also revised to allow fetching all the
      inputs/outputs/witnesses within a single group.
    * Input args is removed since the functionality can be replicated elsewhere


doitian:
Close because of the wrong source repo

PR #942, opened 2019-06-04
feat: add CI files for Windows environments

xxuejie:
This PR adds Windows CI configurations for both appveyor and Azure Pipeline, we can then decide which one to use(or if we want to keep both).


appveyor.yml
```diff
@@ -0,0 +1,21 @@
...
+  - cmd: rust-%RUST_VERSION%-%RUST_INSTALL_TRIPLE%.exe /VERYSILENT /NORESTART /DIR="%RUST_INSTALL_DIR%"
+  - cmd: CALL "%VCVARS%"
+  - cmd: SET PATH=%PATH%;%RUST_INSTALL_DIR%\bin
+  - rustc --version
+  - cargo --version
+
+build: false
+test_script:
+  - cmd: SET RUST_BACKTRACE=full
+  - cargo test
```
doitian:
```suggestion
  - cargo test --verbose --all
```

doitian:
Could you please check APPVEYOR_PULL_REQUEST_NUMBER and run different tasks?

- When it is set, run `cargo test --all`
- Otherwise, run integration tests.

xxuejie:
> Could you please check APPVEYOR_PULL_REQUEST_NUMBER and run different tasks?
> 
> * When it is set, run `cargo test --all`
> * Otherwise, run integration tests.

Yes I can definitely add those. Right now I'm still testing the pipeline so the build specification might be a bit rough, perhaps I should add a WIP flag :P

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+  - cp --version
+
+build: false
+test_script:
+  - cmd: SET RUST_BACKTRACE=full
+  # The plan we are using for appveyor now cannot finish unit tests
+  # and integration tests in one run, so we have to separate them to
+  # run unit in branches, and integration tests in PRs
+  - if NOT defined APPVEYOR_PULL_REQUEST_NUMBER make test
+  - if defined APPVEYOR_PULL_REQUEST_NUMBER make integration-windows
```
doitian: The condition is inversed. We only run unit tests for PR, and integration for branch.

ci/azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+    pool:
+      vmImage: 'VS2017-Win2016'
+    steps:
+      - template: windows-dependencies.yml
+        parameters:
+          rustup_toolchain: '1.34.2-x86_64-pc-windows-msvc'
+      - script: make test
+        displayName: Run unit tests
+  - job: IntegrationTest
+    condition: eq(variables['Build.Reason'], 'PullRequest')
```
doitian: The condition is inversed. We only run unit tests for PR, and integration for branch.

doitian: Better move the file into `devtools`. It seems Azure Pipeline allows us to choose the yml file path.

devtools/azure/pipelines.yml
```diff
@@ -0,0 +1,20 @@
+jobs:
+  - job: UnitTest
```
doitian: We can only run UnitTest if

- It is triggered by PR
- Or it is triggered for master branch

Here is the condition used in travis, FYI

```
if: 'tag IS NOT present AND (type == pull_request OR branch = master OR branch =~ /^test/)'
```

xxuejie: The logic has been fixed except for the tagging part, I couldn't find one variable containing git tag in Azure Pipeline conditions.

PR #953, opened 2019-06-05
[BREAKCHANGE] feat: display miner worker status

quake:
BREAKCHANGE: config file `ckb-miner.toml` changed

- feat: display miner worker status
- refactor: add miner worker trait, prepare for `Cuckoo` cpu and gpu miner plugin

tty


quake:
close via https://github.com/nervosnetwork/ckb/pull/961

PR #954, opened 2019-06-05
chore: use large timeout value for integration test

quake:
Change all `waiting_for_sync` timeout value to 60 seconds, resolve integration test fail issue on poor CI server.

### DIFF
--- a/test/src/net.rs
+++ b/test/src/net.rs
@@ -122,10 +122,11 @@ impl Net {
         });
     }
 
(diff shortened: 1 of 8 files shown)


yangby-cryptape:
In this version, some tests wait 60 secs for 5 height, and some tests wait 60 secs for 2 height, it seems not fair.

IMHO, a more flexible solution is: check if there is anything changed in a small interval.

Didn't reach the target? That's OK, if there is some changes in each interval, it means the program is still alive.

PR #961, opened 2019-06-05
[ᚬrc/v0.14.0] [BREAKCHANGE] feat: display miner worker status

quake:
BREAKCHANGE: config file `ckb-miner.toml` changed

- feat: display miner worker status
- refactor: add miner worker trait, prepare for `Cuckoo` cpu and gpu miner plugin

tty


miner/src/worker/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+pub fn start_worker(
+    pow: Arc<dyn PowEngine>,
+    config: &WorkerConfig,
+    seal_tx: Sender<(H256, Seal)>,
+) -> WorkerController {
+    let mp = MultiProgress::new();
+    let spinner_style = ProgressStyle::default_spinner()
```
doitian: How it resolves the conflicts with logger?

docs/ckb-core-dev.md
```diff
@@ -60,9 +60,13 @@ cycle_length = 12
...
 func = "Dummy"
+```
+
+and don't forget to modify `ckb-miner.toml` miner workers section:
 
-# Delay offset (in milliseconds)
-[pow.params.delay]
-type = "constant"
-value = 5000
+```
```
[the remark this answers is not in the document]
quake: it's related to dummy miner only, no need to set these value in chain config.

miner/src/worker/mod.rs
quake: progress bar log to stderr, no conflict with the logger.

docs/ckb-core-dev.md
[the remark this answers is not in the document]
doitian: Got it. My point is that it is easier to switch PoW if we can only edit it in one file. Maybe we can provide params for all PoW engines in `ckb-miner.yml`, and only the params for current PoW will be used.

miner/src/worker/mod.rs
doitian: Some may want to run miner as a daemon. The progress should be disabled if so. Let's add `StandardError=null` under `[Service]` in devtools/init/linux-systemd/ckb-miner.service

quake:
> Disable stderr in devtools/init/linux-systemd/ckb-miner.service

Done, please review again, thanks.

resource/ckb-miner.toml
```diff
@@ -32,11 +30,22 @@ dsn = "" # {{
...
 # _ => rpc_url = "http://127.0.0.1:{rpc_port}/"
 # }}
 block_on_submit = true
 
 # block template polling interval in milliseconds
 poll_interval = 1000
+
+[[miner.workers]]
+worker_type = "CuckooSimple"
+threads     = "1"
```
TheWaWaR: Why not integer?

TheWaWaR:
Maybe a config to turn off status output?

resource/ckb-miner.toml
```diff
@@ -32,11 +30,22 @@ dsn = "" # {{
...
 
-[miner]
+[miner.client]
 rpc_url = "http://127.0.0.1:8114/" # {{
 # _ => rpc_url = "http://127.0.0.1:{rpc_port}/"
 # }}
 block_on_submit = true
 
 # block template polling interval in milliseconds
 poll_interval = 1000
```
TheWaWaR: Maybe a config to turn off status output?

resource/ckb-miner.toml
quake: We need to support `bool` or `float` here, for example: 
```
[[miner.workers]]
worker_type = "CuckooCpuMean"
trim        = "true"

[[miner.workers]]
worker_type = "Dummy"
delay_type  = "Normal"
mean        =  "5000"
std_dev     =  "32.5"
```

resource/ckb-miner.toml
quake: > Maybe a config to turn off status output?

I'm considering adding more status to the console output, let me add this configuration parameter with more status in another PR.

docs/ckb-core-dev.md
```diff
@@ -60,9 +60,13 @@ cycle_length = 12
...
+[[miner.workers]]
+worker_type = "Dummy"
```
zhangsoledad: I think different types do not allow to exist at the same time, pow type should be global

yangby-cryptape:
Last week, I asked a question in chat: could the numbers in the result be equal the `max_edge`. When I was reviewing this PR, I felt confused again.

- In our implementation:

  We didn't allow that when generating the edges: https://github.com/nervosnetwork/ckb/blob/a1db895e57c6a3ced78b81d26af0c5815c26eab1/pow/src/cuckoo.rs#L273-L281

  But, we allowed that when doing verification: https://github.com/nervosnetwork/ckb/blob/a1db895e57c6a3ced78b81d26af0c5815c26eab1/pow/src/cuckoo.rs#L185-L187

- And, I thought the official implementation do the same logic, too.

docs/ckb-core-dev.md
quake: pow type is global, but the user may need to configure different worker type, for example, running 1 miner on CPU and two miners on different GPUs:

```
[[miner.workers]]
worker_type = "CuckooCpu"
threads     = "2"

[[miner.workers]]
worker_type = "CuckooGpuLean"
device_id   = "..."

[[miner.workers]]
worker_type = "CuckooGpuMean"
device_id   = "..."
```

resource/ckb-miner.toml
doitian: You can use `toml::Value` as the hashmap value.

doitian: Another choise: use internal tagged enum

```rust
#[derive(Serialize, Deserialize)]
#[serde(tag = "worker_type")]
enum WorkerConfig {
    CuckooCpuMean { trim: bool },
    Dummy { delay_type: String, mean: u32, std_dev: f64 },
}
```

docs/ckb-core-dev.md
zhangsoledad: I don't think such `Dummy` here is served for your purpose. It's different level

resource/ckb-miner.toml
quake: @doitian @TheWaWaR I pushed a new commit https://github.com/nervosnetwork/ckb/pull/961/commits/4059a26365b3cda571ebc7d5cfdae9ca62662d0c to resolve this issue, please review again, thanks

docs/ckb-core-dev.md
quake: @zhangsoledad I'm going to stick on this commit: https://github.com/nervosnetwork/ckb/commit/4059a26365b3cda571ebc7d5cfdae9ca62662d0c after fighting with `toml::Value`, please review again, thanks.

util/app-config/src/app_config.rs
```diff
@@ -401,45 +401,7 @@ mod tests {
                 miner_config.chain.spec,
                 PathBuf::from("specs/integration.toml")
             );
-            assert_eq!(miner_config.miner.rpc_url, "http://127.0.0.1:7000/");
-        }
-    }
-
-    #[cfg(all(unix, target_pointer_width = "64"))]
```
doitian: Wrong merge?

doitian: The test is added by @xxuejie when porting to Windows.

quake: sorry, resolved auto rebase error, please review again.

PR #966, opened 2019-06-06
[ᚬrc/v0.14.0] Backport windows support and sentry cleanup to v0.14.0

doitian:


miner/src/client.rs
```diff
@@ -149,10 +149,6 @@ impl Client {
                 }
                 Err(e) => {
                     error!("rpc call submit_block error: {:?}", e);
-                    sentry::capture_message(
```
xxuejie: Is this needed for this PR?

doitian: Yes, it can reduce noises in Sentry.

doitian: Not a part of Windows support, just port together.

doitian: Changed PR title

PR #967, opened 2019-06-06
feat: simplify git describe info in version

doitian:
The full `git describe` output easily cause confusion, this change will only
show the 7 hex digests commit sha and the dirty flag, such as

    ckb 0.15.0-pre (rylai-v4 d7d3baca-dirty 2019-06-06)

### DIFF
--- a/util/build-info/src/lib.rs
+++ b/util/build-info/src/lib.rs
@@ -60,7 +60,7 @@ impl std::fmt::Display for Version {
 
 pub fn get_commit_describe() -> Option<String> {
     std::process::Command::new("git")
-        .args(&["describe", "--dirty"])
+        .args(&["describe", "--dirty", "--always", "--exclude", "*"])
         .output()
         .ok()
         .and_then(|r| {


util/build-info/src/lib.rs
```diff
@@ -60,7 +60,7 @@ impl std::fmt::Display for Version {
 
 pub fn get_commit_describe() -> Option<String> {
     std::process::Command::new("git")
-        .args(&["describe", "--dirty"])
+        .args(&["describe", "--dirty", "--always", "--exclude", "*"])
```
TheWaWaR: Execute `git describe --dirty --always --exclude *` on ckb develop branch will report
> fatal: --dirty is incompatible with commit-ishes

doitian: try `git describe --dirty --always --exclude '*'`

Shell expands `*`

quake: the function name is `get_commit_describe', but it only returns the short hash id of last commit, how about changing the fn name and git command to:
```
git rev-parse --short HEAD
```

doitian: This does not check whether the repo is dirty

quake: I get it, thanks

rainchen: FYI, I preferred using  a build version as the build metadata, 
with the form "yyyymmddHHMMSS.commitshortid", and the full version number string will be: 
`${product_version}+${build_version}` 
for example: `v0.3.0+20190606201137.40564fe`

I know this is quite long but more friendly for human identification.
This form is validated for SemVer, see https://semver.org/#spec-item-10

here is how I accomplished  in other project https://github.com/cryptape/cita-monitor/pull/110/files

PR #970, opened 2019-06-08
perf(sync): Fix get ancestor performance issue

TheWaWaR:
- [x] Use skip list to search ancestor of a header
- [x] Adjust block fetch interval IBD/not-IBD
- [x] Add test for skip list

**NOTE**: The [implementation](https://github.com/bitcoin/bitcoin/blob/5d2ccf0ce9ca1571c650a69319fb9c1e0b626ecb/src/chain.cpp#L70-L120) and [test](https://github.com/bitcoin/bitcoin/blob/5d2ccf0ce9ca1571c650a69319fb9c1e0b626ecb/src/test/skiplist_tests.cpp#L13-L44) is almost copy from bitcoin.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -168,13 +168,17 @@ where
             let mut inflight = self.synchronizer.peers.blocks_inflight.write();
             let count = MAX_BLOCKS_IN_TRANSIT_PER_PEER
                 .saturating_sub(inflight.peer_inflight_count(&self.peer));
+            let max_height_header = self
+                .synchronizer
+                .shared
+                .get_ancestor(&best_known_header.hash(), max_height)?;
 
             while index_height < max_height && fetch.len() < count {
                 index_height += 1;
                 let to_fetch = self
                     .synchronizer
                     .shared
-                    .get_ancestor(&best_known_header.hash(), index_height)?;
+                    .get_ancestor(max_height_header.hash(), index_height)?;
                 let to_fetch_hash = to_fetch.hash();
 
                 let block_status = self.synchronizer.get_block_status(to_fetch_hash);
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -899,3 +899,82 @@ impl<CS: ChainStore> SyncSharedState<CS> {
         );
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use ckb_core::header::HeaderBuilder;
+    use rand::{thread_rng, Rng};
+
+    const SKIPLIST_LENGTH: u64 = 500_000;
+
+    #[test]
+    fn test_get_ancestor_use_skip_list() {
+        let mut header_map: HashMap<H256, HeaderView> = HashMap::default();
+        let mut hashes: BTreeMap<BlockNumber, H256> = BTreeMap::default();
+
+        let mut parent_hash = None;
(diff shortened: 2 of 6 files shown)


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -168,13 +168,17 @@ where
             let mut inflight = self.synchronizer.peers.blocks_inflight.write();
             let count = MAX_BLOCKS_IN_TRANSIT_PER_PEER
                 .saturating_sub(inflight.peer_inflight_count(&self.peer));
+            let max_height_header = self
```
doitian: Why this extra step is required?

sync/src/types.rs
```diff
@@ -507,6 +514,69 @@ impl HeaderView {
     pub fn into_inner(self) -> Header {
         self.inner
     }
+
+    pub fn build_skip<F>(&mut self, mut get_header_view: F)
+    where
+        F: FnMut(&H256) -> Option<HeaderView>,
```
doitian: Why need `FnMut`?

sync/src/types.rs
```diff
@@ -507,6 +514,69 @@ impl HeaderView {
...
+        }
+        let mut number_walk = current.number();
+        while number_walk > number {
+            let number_skip = get_skip_height(number_walk);
+            let number_skip_prev = get_skip_height(number_walk - 1);
+            match current.skip_hash {
+                Some(ref hash)
+                    if number_skip == number
+                        || (number_skip > number
+                            && !(number_skip_prev + 2 < number_skip
```
doitian: parent is better if `number_skip_prev < number_skip`, why need `+2` here?

sync/src/types.rs
TheWaWaR: Could be removed.

TheWaWaR: Could be just `Fn`

sync/src/synchronizer/block_fetcher.rs
TheWaWaR: A minor optimization. typical `get_ancestor` with 30W ancestors need about `60` steps, with this optimization will reduce to less than `(max_height - fixed_last_common_header.number())`

sync/src/types.rs
TheWaWaR: If we want test this method (measure steps for example) in the future, we need `FnMut`

sync/src/types.rs
TheWaWaR: `number_skip_prev` and `number_skip` are in different parity(even+odd or odd+even), when `number_skip_prev < number_skip` their distance can be 1, 3, 5...

I guess by current `get_skip_height`, neighbors' parents also neighbors is impossible, so their distance `>=3`

TheWaWaR: `number_skip_prev` and `number_skip` are in different parity (odd+even or even+odd), so their distance can be 1,3,5...

I guess by current `get_skip_height`, neighbor's parents also neighbor is impossible, so their distance `>=3`

TheWaWaR: Need a prove here?

doitian:
It's an important performance improvement, could you backport this PR to rc/v0.14.0? I think we need this PR in the next release.

TheWaWaR:
> It's an important performance improvement, could you backport this PR to rc/v0.14.0? I think we need this PR in the next release.

OK

PR #971, opened 2019-06-08
feat: abstract data loader layer to decouple ckb-script and ckb-store

jjyr:
### Intention

The needs to run `ckb-script` outside the ckb is wide, such as test contract scripts, verify tx in layer 2 program or in development tools. etc.

This PR tries to decouple `ckb-store` dependency from `ckb-script` via abstract a data loader trait, make `ckb-script` easier to be used standalone.

### Changes

1. abstract a trait `DataLoader`. `TransactionScriptVerifier` should only load data from this trait.
2. remove the `ckb-store` dependency from `ckb-script` and `dao-utils`


script/data-loader/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+use ckb_core::cell::CellMeta;
+use ckb_core::extras::BlockExt;
+use ckb_core::transaction::CellOutput;
+use numext_fixed_hash::H256;
+
+/// Script DataLoader
+/// abstract the data access layer
+pub trait DataLoader {
+    // load CellOutput
+    fn lazy_load_cell_output(&self, _cell: &CellMeta) -> CellOutput;
```
doitian: Why use underscore before parameter name?

TheWaWaR: If there is no default implementation, no need underscore.

util/dao/utils/src/lib.rs
```diff
@@ -13,8 +12,8 @@ use std::sync::Arc;
 // Those validation work is left to other verifiers to check.
 // TODO: revisit it here to see if we need to return correct error once we
 // manage to revise error handling in verification package
-pub fn calculate_transaction_fee<CS: ChainStore>(
-    store: Arc<CS>,
+pub fn calculate_transaction_fee<CS: DataLoader>(
+    store: &CS,
```
TheWaWaR: `CS` => `DL`

TheWaWaR:
Conflicted

PR #977, opened 2019-06-10
feat: Upgrade p2p

driftluo:
1. upgrade p2p dependence
2. support `upnp` optional

changelog: https://github.com/nervosnetwork/p2p/blob/master/CHANGELOG.md

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -53,6 +53,8 @@ ping_interval_secs = 120
 # 20 minutes
 ping_timeout_secs = 1200
 connect_outbound_interval_secs = 15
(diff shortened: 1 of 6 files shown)


resource/ckb.toml
```diff
@@ -53,6 +53,8 @@ ping_interval_secs = 120
 # 20 minutes
 ping_timeout_secs = 1200
 connect_outbound_interval_secs = 15
+# If true, try to register upnp
```
TheWaWaR: "If true" => "If set to true"

doitian:
Can you repost the changelog from 0.2.0 to 0.2.1 here?

PR #978, opened 2019-06-10
feat: use new identify protocol

driftluo:
BREAKING CHANGE: identify protocol break

The current identify protocol does not play a role in identifying the capabilities of both parties, and the message structure is not reasonable.

So, I rewrote it and added the capability ID and network ID.

Moreover, I changed the process of network connection from random protocol opening to the first one, which must be the identify protocol, and then decide whether to continue to open the remaining protocols.

This can slightly alleviate address pollution problems between different networks.


protocol/src/protocol_generated.rs
```diff
@@ -101,14 +102,14 @@ const ENUM_NAMES_SYNC_PAYLOAD:[&'static str; 9] = [
...
-  ENUM_NAMES_SYNC_PAYLOAD[index]
+  let index = e as u8;
+  ENUM_NAMES_SYNC_PAYLOAD[index as usize]
 }
 
 pub struct SyncPayloadUnionTableOffset {}
 #[allow(non_camel_case_types)]
 #[repr(u8)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
```
quake: Do you use a new version flatbuffer generator?

driftluo:
```
$ flatc --version
flatc version 1.11.0
```
It should have been upgraded some time ago, although I don’t remember when it was

protocol/src/protocol.fbs
```diff
@@ -256,3 +256,9 @@ table Alert {
     message: Bytes;
 }
 
+table Identify {
+  // network name
+  name: string;
```
u2: I don't think the `name: string` is a good identity for the network, its size is unlimited. What about `id: u64`.

And what about adding another field `version` used for forking.

network/src/network.rs
```diff
@@ -698,7 +703,11 @@ pub struct NetworkService {
 }
 
 impl NetworkService {
-    pub fn new(network_state: Arc<NetworkState>, protocols: Vec<CKBProtocol>) -> NetworkService {
+    pub fn new(
+        network_state: Arc<NetworkState>,
+        protocols: Vec<CKBProtocol>,
+        consensus: &Consensus,
+    ) -> NetworkService {
```
keroro520: IdentifyCallback only cares the service name, so passing `name: String` instead of `consensus: Consensus`, is enough.

keroro520: And can we try to reuse [`code_name`](https://github.com/nervosnetwork/ckb/blob/a5ce3dfe881dddf8a798db7f707ae0195c6e4ada/src/main.rs#L38) as service name here?

network/src/protocols/identify.rs
```diff
@@ -109,3 +152,102 @@ impl Callback for IdentifyCallback {
...
+            (Some(raw_name), flag) => {
+                if self.name != raw_name {
+                    debug!(
+                        "Not the same chain, self: {}, remote: {}",
+                        self.name, raw_name
+                    );
+                    return None;
+                }
+
+                if flag == 0 {
```
keroro520: What does `flag == 0` mean? Maybe add a comment?

network/src/network.rs
driftluo: Can't do this, now the name is `/chain_id/genesis_hash[..8]`.

`code_name` change does not represent `chan_id` and `genesis_hash` change

network/src/protocols/identify.rs
driftluo:
```
#[repr(u64)]
enum Flag {
    /// Support all protocol
    FullNode = 0x1,
}
```
zero is flatbuffer default value

network/src/network.rs
driftluo: > IdentifyCallback only cares the service name, so passing `name: String` instead of `consensus: Consensus`, is enough.

I am not sure if the data in genesis is still needed in the future. Passing the entire reference has no direct impact on the performance of the code.

keroro520: > > IdentifyCallback only cares the service name, so passing `name: String` instead of `consensus: Consensus`, is enough.
> 
> I am not sure if the data in genesis is still needed in the future. Passing the entire reference has no direct impact on the performance of the code.

Not about the performance.
It avoids introducing `ckb-chain-spec` as a dependency if only use the name as the parameter.

quake:
LGTM, we may refactor feeler protocol also after this PR merged.

PR #979, opened 2019-06-10
[ᚬpr-mirror/976] feat: wrap lock methods

keroro520:
* feat: Avoid locking a long time

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -977,4 +977,11 @@ impl<CS: ChainStore> SyncSharedState<CS> {
             locked.entry(id).or_default().insert(pi);
         }
     }
+
+    pub fn disconnected(&self, pi: PeerIndex) -> Option<PeerState> {
+        self.known_txs.lock().inner.remove(&pi);
+        self.known_blocks.lock().inner.remove(&pi);
+        self.inflight_blocks.write().remove_by_peer(&pi);
+        self.peers().disconnected(pi)
+    }
 }
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -32,6 +32,8 @@ use std::collections::{
     BTreeMap,
 };
 use std::fmt;
+use std::mem::swap;
+use std::ops::DerefMut;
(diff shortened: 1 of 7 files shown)


sync/src/types.rs
```diff
@@ -933,16 +921,60 @@ impl<CS: ChainStore> SyncSharedState<CS> {
 
     pub fn mark_as_known_tx(&self, hash: H256) {
         self.inflight_transactions().remove(&hash);
-        self.tx_filter().insert(hash, ());
+        self.tx_filter.lock().insert(hash, ());
     }
 
     pub fn already_known_tx(&self, hash: &H256) -> bool {
-        self.tx_filter().contains_key(hash)
+        self.tx_filter.lock().contains_key(hash)
```
driftluo: Use an inline function instead of a function call can reduce how much time of the lock holding? 

Maybe 1-10ms?

doitian:
I don't recommend to submit PR to a mirror branch. That branch is force-pushed whenever PR #976 has changed, and it is deleted when 976 is closed.

sync/src/types.rs
xxuejie: I don't really get this comment, care to explain?

keroro520: I just remove the `tx_filter()` method and call `tx_filter.lock()` directly. There are no differences.
So here will not reduce the lock holding time.

I am not sure I get your meaning.

driftluo: I seem to have misread the focus, seeing that the call becomes a function, shortening the time to acquire the lock, but in fact the focus of pr is not here.

keroro520:
> I don't recommend to submit PR to a mirror branch. That branch is force-pushed whenever PR #976 has changed, and it is deleted when 976 is closed.

Yeah. I separate #930 into two smaller PRs, #976 and #979. And #979 dependences on the previous one.

So I simply make #979 target to #976, so that the reviewers can review the two PRs respectively.

If I don't do this, it must wait until #976 been merged.

PR #983, opened 2019-06-11
[ᚬrc/v0.14.0] fix: uncle number should smaller than block

zhangsoledad:


doitian:
does it affect v0.14.0?

zhangsoledad:
> does it affect v0.14.0?

yes

doitian:
> > does it affect v0.14.0?
> 
> yes

You can rebase it onto rc/v0.14.0, we'll merge back rc to develop later.

PR #993, opened 2019-06-12
chore

u2:


network/src/peer_registry.rs
```diff
@@ -29,14 +30,12 @@ pub struct ConnectionStatus {
...
 
-fn sort_then_drop<T, F>(list: &mut Vec<T>, n: usize, compare: F)
+fn sort_then_truncate<T, F>(list: &mut Vec<T>, n: usize, compare: F)
 where
     F: FnMut(&T, &T) -> std::cmp::Ordering,
 {
     list.sort_by(compare);
-    if list.len() > n {
-        list.truncate(list.len() - n);
-    }
```
u2: Is it a mistake?

doitian: I think this function is intended to remove n elements from the list.

BTW, please do not mess up bug fixing with refactoring in one PR.

u2: OK, I dropped the associated commits.

u2: I see after rereading the code.

PR #994, opened 2019-06-12
feat: Wrap lock methods

keroro520:
* feat: Avoid locking a long time

---

#979 has closed by bot automatically and I cannot find the way to re-open it :(
So here re-submit #979 here.

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -977,4 +977,11 @@ impl<CS: ChainStore> SyncSharedState<CS> {
             locked.entry(id).or_default().insert(pi);
         }
     }
+
+    pub fn disconnected(&self, pi: PeerIndex) -> Option<PeerState> {
+        self.known_txs.lock().inner.remove(&pi);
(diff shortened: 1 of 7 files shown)


sync/src/types.rs
```diff
@@ -933,16 +921,67 @@ impl<CS: ChainStore> SyncSharedState<CS> {
...
+
+    pub fn clear_get_block_proposals(&self) -> FnvHashMap<ProposalShortId, FnvHashSet<PeerIndex>> {
+        let mut locked = self.pending_get_block_proposals.lock();
+        let old = locked.deref_mut();
+        let mut ret = FnvHashMap::default();
+        swap(old, &mut ret);
+        ret
+    }
+
+    pub fn insert_get_block_proposals(&self, ids: Vec<ProposalShortId>, pi: PeerIndex) {
```
TheWaWaR: I wonder why `pi: PeerIndex` is not the first argument.

keroro520: Placing `pi` in the first one is better! Let me change it right now.

PR #997, opened 2019-06-13
chore: improve performance of difficulty calculation

u2:


core/src/difficulty.rs
```diff
@@ -1,24 +1,25 @@
 use numext_fixed_hash::H256;
-use numext_fixed_uint::U256;
+use numext_fixed_uint::{U256, u256};
 
 const ONE: U256 = U256::one();
+// 1 << 255
+const T: U256 = u256!("57896044618658097711785492504343953926634992332820282019728792003956564819968");
```
keroro520: Add a small test to ensure the value of `T`? like `assert_eq!(T, one() << 255)`

And, I don't think `T` is a good name.

u2: updated.

PR #998, opened 2019-06-13
fix: fix confusing JsonBytes deserializing error message

### LINKED ISSUE #974: Confusing JsonBytes deserializing error message
In JsonBytes, It should report invalid length, instead of expecting 0x-prefixed hex string when the hex string contains odd digits.

FYI: https://github.com/cryptape/rust-numext/blob/61fc5e5dd9ef98288875932c396b6c81ed872c86/constructor/src/fixed_hash/core/extension/serde.rs#L74

driftluo:
fix #974 

1. Not 0x-prefixed hex string, error message:
```
Config Error: Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: "invalid value: string \"7f52f0fccdd1d11391c441adfb174f87bca612b0\", expected a 0x-prefixed hex string", key: ["block_assembler", "args"] } }
```

2. The illegal 0x-prefixed hex string
```
Config Error: Error { inner: ErrorInner { kind: Custom, line: None, col: 0, message: "invalid length 43, expected even length", key: ["block_assembler", "args"] } }
```

### DIFF
--- a/util/jsonrpc-types/src/bytes.rs
+++ b/util/jsonrpc-types/src/bytes.rs
@@ -51,9 +51,14 @@ impl<'b> serde::de::Visitor<'b> for BytesVisitor {
     where
         E: serde::de::Error,
     {
-        if v.len() < 2 || &v[0..2] != "0x" || v.len() & 1 != 0 {
+        if v.len() < 2 || &v[0..2] != "0x" {
             return Err(E::invalid_value(serde::de::Unexpected::Str(v), &self));
         }
+
+        if v.len() & 1 != 0 {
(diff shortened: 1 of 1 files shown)


quake:
could you please add test to cover `invalid prefix / invalid length / invalid char / etc` ?

driftluo:
> could you please add test to cover `invalid prefix / invalid length / invalid char / etc` ?

This test is very difficult to write, error returns cannot determine the type, can only compare errors through string comparison, at the same time, the error type and error of serde_json and `toml` are not the same, can only increase the dependency of `toml` to test

PR #999, opened 2019-06-13
[ᚬrc/v0.14.0] feat: avoid hard code code_hashes

doitian:
Generate the code hash constants in ckb-resource, and use the constants
instead of the hard code hash in different locations.


doitian:
@nervos-bot-user ci status e8b8e28d002bfaa3bb6e9f96d940aa8aa0f8d5e3
CI: success

Travis not triggered

PR #1000, opened 2019-06-13
feat: Allow miner add an arbitrary message into the cellbase

### LINKED ISSUE #888: Allow miner add an arbitrary message into the cellbase
Miner can fill any data into the cellbase output cell data. It can be a part of the block assembler config

driftluo:
Allow miner add an arbitrary message into the cellbase

closed #888

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -10,12 +10,13 @@ use ckb_core::transaction::{
 };
 use ckb_core::uncle::UncleBlock;
 use ckb_core::BlockNumber;
-use ckb_core::{Bytes, Cycle, Version};
+use ckb_core::{Cycle, Version};
 use ckb_logger::{error, info};
 use ckb_notify::NotifyController;
 use ckb_shared::{shared::Shared, tx_pool::ProposedEntry};
 use ckb_store::ChainStore;
 use ckb_traits::ChainProvider;
+use ckb_verification::TransactionError;
 use crossbeam_channel::{self, select, Receiver, Sender};
 use failure::Error as FailureError;
 use faketime::unix_time_as_millis;
@@ -407,10 +408,14 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
                 .consensus()
                 .genesis_epoch_ext()
                 .block_reward(candidate_number)?,
-            Bytes::default(),
+            self.config.data.clone().into_bytes(),
             self.shared.consensus().bootstrap_lock().clone(),
             None,
         );
+        // User-defined data has a risk of exceeding capacity
+        if output.is_lack_of_capacity()? {
+            return Err(TransactionError::InsufficientCellCapacity.into());
+        }
         Ok(TransactionBuilder::default()
             .input(input)
             .output(output)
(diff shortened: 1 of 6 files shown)


resource/ckb.toml
```diff
@@ -117,8 +117,16 @@ cell_output_cache_size  = 128
 #
 #     ckb cli secp256k1-lock <pubkey>
 #
+# Also, ckb allows the miner to add any data to the cellbase that he has dug out.
+# The data must be A 0x-prefixed hex string. **Please note** that the data cannot be
+# larger than the capacity value of the current cellbase. If it is larger,
```
u2: The normal users don't know the capacity value of the cellbase, this data size may be confused for them.

test/src/specs/mining/bootstrap.rs
```diff
@@ -65,6 +66,7 @@ impl Spec for BootstrapCellbase {
             config.block_assembler = Some(BlockAssemblerConfig {
                 code_hash: h256!("0xa2"),
                 args: vec![],
+                data: JsonBytes::default(),
```
quake: please use arbitrary data here and assert it in line#46

resource/ckb.toml
driftluo: 1. First of all, it is not a required field here. If you don’t understand it, you will not fill it out.
2. The capacity of cellbase is dynamically changed. It is impossible to determine the size of data by a static value.
3. I considered a lot of processing solutions, and finally decided to write the details in the comment, and verify the value when generating the block template to ensure that the user can find the filling error as soon as possible.

resource/ckb.toml
```diff
@@ -117,8 +117,16 @@ cell_output_cache_size  = 128
 #
 #     ckb cli secp256k1-lock <pubkey>
 #
+# Also, ckb allows the miner to add any data to the cellbase that he has dug out.
```
doitian:
```suggestion
# Also, ckb allows the miners to add any data to the cellbase that they have dug out.
```

resource/ckb.toml
doitian: I think it's better to truncate the data when it is too long. And add a warning here to tell users that do not rely on the data here, it may be truncated when the block reward capacity is not enough to hold the data.

u2: It should never be panic if the data size overflow sometimes, for the user view it's a bug.

driftluo: The current processing will not panic if overflow RPC will return an error when getting the block template.

driftluo: I changed the implementation to forced truncate

miner/src/block_assembler.rs
```diff
@@ -424,14 +429,41 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
...
+        let output = CellOutput::new(block_reward, data, target_lock, r#type);
+
         Ok(TransactionBuilder::default()
             .input(input)
             .output(output)
             .witness(witness)
             .build())
     }
 
+    fn custom_data(
```
doitian: It's better to construct a cell output first with empty data, then the max allowed data length is

    capacity - occupied_capacity

doitian: Better solution:

Construct the output first, truncate the tailing bytes if

```
occupied_capacity > capacity
```

Where `occupied_capacity - capacity` is the bytes to truncate.

If `occupied_capacity - capacity > data.length`, we should return an error.

PR #1004, opened 2019-06-14
chore(ci): force job Linters using its own cache

doitian:
This PR tries to fix the following error in CI:

    found possibly newer version of crate `*` which `*` depends on

Travis separate cache via OS, Language and Env. So the new env option added to
job Linters can force it to use its own cache.


doitian:
@nervos-bot-user give me five

PR #1011, opened 2019-06-14
fix: witnesses_root cal including cellbase

u2:
https://github.com/cryptape/ckb-internal/issues/1063


doitian:
Since cellbase witness is not used in v0.14.0 and early versions, we don't need to backport this. Am I right? @zhangsoledad

zhangsoledad:
> Since cellbase witness is not used in v0.14.0 and early versions, we don't need to backport this. Am I right? @zhangsoledad

yes

Issue #892, opened 2019-05-24
Should not use same struct for both storage and core logic in all cases


yangby-cryptape:
For example:

This is the definition of `CellMeta`:
https://github.com/nervosnetwork/ckb/blob/e09eca6ae2eba520ade4b4fc4e20538517c73148/core/src/cell.rs#L25-L33

The `CellMeta` is saved into storage **with three `Option` fields** (bring uncertainty into storage):
https://github.com/nervosnetwork/ckb/blob/e09eca6ae2eba520ade4b4fc4e20538517c73148/store/src/store.rs#L465-L475

Most fields in the `CellMeta` are **redundant**, and I've heard that some guys want to **add more fileds into this struct**.

This struct likes a virus, some methods have to set `None` when build it and `unwrap` before use it (or just skip handle the cases when `is_none()`).

This style makes it to be hard to implement a robust program.

doitian:
We even have a `#[serde(skip)]` field here. 😱

janx:
Can we list the features depending on this struct? Once we get those features done we can start refactoring.

PR #1014, opened 2019-06-14
fix: locate blocks by hash

keroro520:
When calculates block_median_time, we need to locate the specific blocks.
Using block_hash instead block_number to locate the specific blocks is more accurate.

* feat: Add `hash: H256` into `BlockInfo`
* fix: Locate blocks by block_hash when calculate block_median_time

-------

**BREAKING CHANGE**: The format of `TransactionMeta`  is changed, which is affected by `BlockInfo`

### DIFF
--- a/core/src/cell.rs
+++ b/core/src/cell.rs
@@ -15,11 +15,16 @@ use std::fmt;
 pub struct BlockInfo {
     pub number: BlockNumber,
     pub epoch: EpochNumber,
+    pub hash: H256,
 }
 
 impl BlockInfo {
-    pub fn new(number: BlockNumber, epoch: EpochNumber) -> Self {
-        BlockInfo { number, epoch }
+    pub fn new(number: BlockNumber, epoch: EpochNumber, hash: H256) -> Self {
+        BlockInfo {
+            number,
+            epoch,
+            hash,
+        }
     }
 }
 
@@ -374,6 +379,7 @@ impl<'a> CellProvider for BlockCellProvider<'a> {
                             block_info: Some(BlockInfo {
                                 number: self.block.header().number(),
                                 epoch: self.block.header().epoch(),
+                                hash: self.block.header().hash().to_owned(),
                             }),
                             cellbase: *i == 0,
                         })
@@ -691,6 +697,7 @@ mod tests {
             block_info: Some(BlockInfo {
                 number: 1,
                 epoch: 1,
+                hash: H256::zero(),
             }),
             capacity: cell_output.capacity,
             data_hash: Some(cell_output.data_hash()),
--- a/core/src/transaction_meta.rs
+++ b/core/src/transaction_meta.rs
@@ -1,5 +1,5 @@
-use crate::cell::BlockInfo;
 use bit_vec::BitVec;
+use numext_fixed_hash::H256;
 use serde_derive::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize)]
@@ -23,6 +23,7 @@ impl From<BitVecSerde> for BitVec {
 pub struct TransactionMeta {
     block_number: u64,
     epoch_number: u64,
+    block_hash: H256,
     cellbase: bool,
     /// each bits indicate if transaction has dead cells
     #[serde(with = "BitVecSerde")]
@@ -33,12 +34,14 @@ impl TransactionMeta {
     pub fn new(
         block_number: u64,
         epoch_number: u64,
+        block_hash: H256,
         outputs_count: usize,
         all_dead: bool,
     ) -> TransactionMeta {
         TransactionMeta {
             block_number,
             epoch_number,
+            block_hash,
             cellbase: false,
             dead_cell: BitVec::from_elem(outputs_count, all_dead),
         }
@@ -48,10 +51,17 @@ impl TransactionMeta {
     pub fn new_cellbase(
         block_number: u64,
         epoch_number: u64,
+        block_hash: H256,
         outputs_count: usize,
         all_dead: bool,
     ) -> Self {
-        let mut result = Self::new(block_number, epoch_number, outputs_count, all_dead);
+        let mut result = Self::new(
+            block_number,
+            epoch_number,
+            block_hash,
+            outputs_count,
+            all_dead,
+        );
         result.cellbase = true;
         result
     }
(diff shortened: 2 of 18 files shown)


yangby-cryptape:
Disapprove!

- As I said in [PR-892: Should not use same struct for both storage and core logic in all cases](https://github.com/nervosnetwork/ckb/issues/892), most fields in the `core::CellMeta` are redundant, we store those block info 4 times(`Block`, `Transaction` * 2, `Cell`).

- And, `CellMeta` has three optional fields, but they're NOT optional in store:
  - `cell_output` should always be `None`.
  - `block_info` should always be `Some(..)`.
  - `data_hash` should always be `Some(..)`.

  So, why store them as `Option`?

- We couldn't store all struts we defined.

  We can compose `ckb-core` structs from more than one `ckb-store` structs.

core/src/cell.rs
```diff
@@ -13,11 +13,16 @@ use std::fmt;
 pub struct BlockInfo {
     pub number: BlockNumber,
     pub epoch: EpochNumber,
+    pub parent: H256,
```
yangby-cryptape: I think this struct should not be named as `BlockInfo`.

keroro520:
> Disapprove!
> 
> * As I said in [PR-892: Should not use same struct for both storage and core logic in all cases](https://github.com/nervosnetwork/ckb/issues/892), most fields in the `core::CellMeta` are redundant, we store those block info 4 times(`Block`, `Transaction` * 2, `Cell`).
> * And, `CellMeta` has three optional fields, but they're NOT optional in store:
>   
>   * `cell_output` should always be `None`.
>   * `block_info` should always be `Some(..)`.
>   * `data_hash` should always be `Some(..)`.
>   
>   So, why store them as `Option`?
> * We couldn't store all struts we defined.
>   We can compose `ckb-core` structs from more than one `ckb-store` structs.

What this PR does is, adding a new field `parent: H256` into `BlockInfo`, for fixing the accurately locating block problem. So I think these opinions are out of this PR:

1. `BlockInfo` is stored redundantly.
2. Confused optional fields are defined inside `CellMeta`.
3. The service layer and database layer mix-use the same core-types.

I strongly agree with your opinions, we should refactor them later, but not in this PR.

yangby-cryptape:
I don't understand the `refactor` you said.

We can split that issue into two issues:

- The core type `CellMeta` is too big.
- How to fetch the core type `CellMeta`.

First issue need a refactor.
But second issue doesn't, you can just fetch the parent hash in another way, and do a composition.

store/src/store.rs
```diff
@@ -285,21 +285,16 @@ impl<T: KeyValueDB> ChainStore for ChainKVStore<T> {
 
         for tx in genesis.transactions() {
             let tx_meta;
+            let block_info = BlockInfo::new(
+                genesis.header().number(),
+                genesis.header().epoch(),
+                genesis.header().parent_hash().clone(),
+            );
             let ins = if tx.is_cellbase() {
```
jjyr: let(ins, tx_meta) = ...

doitian:
@keroro520  Any plan on this PR? What's the next step?

keroro520:
> @keroro520 Any plan on this PR? What's the next step?

I will talk to @yangby-cryptape about it. I am not sure which way is better, to acquire the corresponding block_hash or parent_hash based since_block_number.

keroro520:
> > @keroro520 Any plan on this PR? What's the next step?
> 
> I will talk to @yangby-cryptape about it. I am not sure which way is better, to acquire the corresponding block_hash or parent_hash based since_block_number.

Call for review.

core/src/transaction_meta.rs
```diff
@@ -97,12 +112,13 @@ impl TransactionMeta {
         }
     }
 
-    pub fn destruct(&self) -> (u64, u64, bool, Vec<u8>, usize) {
```
u2: This destruct is different from others. `destruct(&self)` vs `destruct(self)`, just a question.

shared/tx-pool-executor/src/tx_pool_executor.rs
```diff
@@ -109,8 +115,9 @@ impl<CS: ChainStore> TxPoolExecutor<CS> {
                 let verified_result = TransactionVerifier::new(
                     &tx,
                     &block_median_time_context,
-                    tip_number,
+                    parent_number + 1,
```
u2: This is a bug fix?

keroro520: Yes, it is a bugfix https://github.com/nervosnetwork/ckb/pull/1150

I should rebase #1150 after it merged.

core/src/transaction_meta.rs
keroro520: Confused +1 ... cc @yangby-cryptape ?

yangby-cryptape: Require a reference as input:

https://github.com/nervosnetwork/ckb/blob/858d623fc2c0472fd51b3d1a2181aef70859ee14/protos/src/convert/mod.rs#L53-L64

core/src/transaction_meta.rs
```diff
@@ -32,12 +34,14 @@ impl TransactionMeta {
     pub fn new(
         block_number: u64,
         epoch_number: u64,
+        block_hash: H256,
```
quake:
```suggestion
    pub fn new(block_info: BlockInfo, outputs_count: usize, all_dead: bool)
```

core/src/transaction_meta.rs
```diff
@@ -47,10 +51,17 @@ impl TransactionMeta {
     pub fn new_cellbase(
         block_number: u64,
         epoch_number: u64,
+        block_hash: H256,
```
quake:
```suggestion
    pub fn new_cellbase(block_info: BlockInfo, outputs_count: usize, all_dead: bool)
```

shared/src/cell_set.rs
```diff
@@ -13,7 +13,7 @@ pub struct CellSetDiff {
     pub old_inputs: FnvHashSet<OutPoint>,
     pub old_outputs: FnvHashSet<H256>,
     pub new_inputs: FnvHashSet<OutPoint>,
-    pub new_outputs: FnvHashMap<H256, (u64, u64, bool, usize)>,
+    pub new_outputs: FnvHashMap<H256, (u64, u64, H256, bool, usize)>,
```
quake:
```suggestion
    pub new_outputs: FnvHashMap<H256, (BlockInfo, bool, usize)>,
```

core/src/transaction_meta.rs
keroro520: Yes. I did, but then I deleted it after talking with @yangby-cryptape . @yangby-cryptape will refactor this part later. 
So this PR only focus its goal without any other refactorings.

PR #1016, opened 2019-06-14
chore(ci): skip integration in PR

doitian:
Integration tests will become slower and slower, we don't want to run all the
test suites.

Later we may pick random n test specs to run in PR.


TheWaWaR:
Conflicted

PR #1019, opened 2019-06-15
[ᚬrc/v0.14.0] fix: miner log wrong block hash

quake:


miner/src/miner.rs
```diff
@@ -88,25 +88,24 @@ impl Miner {
...
                 self.notify_workers(WorkerMessage::Stop);
+                let raw_header = work.block.header().raw().to_owned();
+                let block = BlockBuilder::from_block(work.block.clone())
+                    .header(raw_header.with_seal(seal))
+                    .build();
+
                 debug!(
-                    "found seal {:?} of block #{} {:x}",
-                    seal,
+                    "Found! #{} {:x}",
```
doitian: Use `{:#x}` to print hash. So does the line 118

quake: Resolved

PR #1028, opened 2019-06-15
fix: use Rust VM runner in bundled config file

doitian:
Some command still uses the bundled config file, for example, following
command will fail in Windows when `ckb.toml` not found in current directory:

    ckb cli hashes

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -78,7 +78,7 @@ max_cycles = 200_000_000_000
 max_verfify_cache_size = 100_000
 
 [script]
-runner = "Assembly" # {{
+runner = "Rust" # {{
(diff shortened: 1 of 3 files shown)


resource/ckb.toml
```diff
@@ -78,7 +78,7 @@ max_cycles = 200_000_000_000
 max_verfify_cache_size = 100_000
 
 [script]
-runner = "Assembly" # {{
+runner = "Rust" # {{
```
xxuejie: Won't this changes the default behavior on Linux/Mac to use Rust based VM?

doitian: Nope, we require exporting ckb.toml before running the node. The default value in the exported config file will be Assembly.

PR #1030, opened 2019-06-15
fix: uncle block serialized_size

u2:
I am just confused.
```
#[derive(Clone, Serialize, Deserialize, Eq, Debug)]
pub struct UncleBlock {
    pub header: Header,
    pub proposals: Vec<ProposalShortId>,
}

    pub fn serialized_size(&self, proof_size: usize) -> usize {
        Header::serialized_size(proof_size)
    }
```
Is it inconsistence?
I checked the paper:
```
The total size of the first four fields should be no larger than
the hard-coded block size limit. The uncle blocks’ proposal
zones do not count in the limit as they are usually already
synchronized. The number of txpids in a proposal zone also
has a hard-coded upper bound.
```

We should store the uncle proposals?

### DIFF
--- a/core/src/uncle.rs
+++ b/core/src/uncle.rs
@@ -49,6 +49,7 @@ impl UncleBlock {
 
     pub fn serialized_size(&self, proof_size: usize) -> usize {
         Header::serialized_size(proof_size)
+            + self.proposals.len() * ProposalShortId::serialized_size()
     }
 }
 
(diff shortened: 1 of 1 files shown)


doitian:
> The number of txpids in a proposal zone also has a hard-coded upper bound.

It's by design

u2:
> > The number of txpids in a proposal zone also has a hard-coded upper bound.
> 
> It's by design

Not this. The uncle proposals is included in serialize but not in serialize_size.

And I don't understand the design also,
> The uncle blocks’ proposal
zones do not count in the limit as they are usually already
synchronized.

PR #1034, opened 2019-06-16
[ᚬrc/v0.14.0] chore: use fixed interval for miner state output

quake:


miner/src/worker/cuckoo_simple.rs
```diff
@@ -162,15 +162,15 @@ impl Worker for CuckooSimple {
                     self.solve(&pow_hash, random());
```
doitian: Not related, but it seems we don't need to clone the hash in the line above. Just use `self.pow_hash.as_ref()`.

PR #1035, opened 2019-06-16
feat: ckb init allow set ba-data

driftluo:
1. `ckb init` allow set ba-data
2. Verify that the input is valid hex

### DIFF
--- a/ckb-bin/src/subcommand/init.rs
+++ b/ckb-bin/src/subcommand/init.rs
@@ -35,9 +35,11 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
             format!(
                 "[block_assembler]\n\
                  code_hash = \"{}\"\n\
-                 args = [ \"{}\" ]",
+                 args = [ \"{}\" ]\n\
+                 data = \"{}\"",
                 hash,
-                args.block_assembler_args.join("\", \"")
+                args.block_assembler_args.join("\", \""),
+                args.block_assembler_data.unwrap_or_default()
             )
         }
         None => {
--- a/ckb-bin/src/subcommand/init.rs
+++ b/ckb-bin/src/subcommand/init.rs
@@ -39,7 +39,8 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
                  data = \"{}\"",
                 hash,
                 args.block_assembler_args.join("\", \""),
-                args.block_assembler_data.unwrap_or_default()
+                args.block_assembler_data
+                    .unwrap_or_else(|| "0x".to_string())
             )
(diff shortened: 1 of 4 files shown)


ckb-bin/src/subcommand/init.rs
```diff
@@ -35,9 +35,11 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
             format!(
                 "[block_assembler]\n\
                  code_hash = \"{}\"\n\
-                 args = [ \"{}\" ]",
+                 args = [ \"{}\" ]\n\
+                 data = \"{}\"",
                 hash,
-                args.block_assembler_args.join("\", \"")
+                args.block_assembler_args.join("\", \""),
+                args.block_assembler_data.unwrap_or_default()
```
doitian: Default does not work, it is not a valid hex string. Either omit data or use "0x" as default.

util/app-config/src/cli.rs
```diff
@@ -266,3 +276,16 @@ fn init() -> App<'static, 'static> {
                 .hidden(true),
         )
 }
+
+fn is_hex(hex: &str) -> Result<(), String> {
```
doitian: 👍

util/app-config/src/cli.rs
```diff
@@ -238,6 +239,7 @@ fn init() -> App<'static, 'static> {
             Arg::with_name(ARG_BA_CODE_HASH)
                 .long(ARG_BA_CODE_HASH)
                 .value_name("code_hash")
+                .validator(|code_hash| is_hex(code_hash.as_str()))
```
quake: change the `is_hex` fn parameter type to `String` according to validator closure type:

```suggestion
                .validator(is_hex)
```

util/app-config/src/cli.rs
quake: need to check hex char range, e.g. `is_hex("0xZZZ")` should return error.

util/app-config/src/cli.rs
```diff
@@ -266,3 +276,25 @@ fn init() -> App<'static, 'static> {
...
         )
 }
+
+fn is_hex(hex: String) -> Result<(), String> {
+    let tmp = hex.as_bytes();
+    if tmp.len() < 2 {
+        Err("Must be a hexadecimal string".to_string())
+    } else if tmp.len() & 1 != 0 {
+        Err("Hexadecimal strings must be of even length".to_string())
+    } else if tmp[..2] == b"0x"[..] {
```
yangby-cryptape: How could the users know that they have to input a 0x-prefix string?

PR #1041, opened 2019-06-16
chore: delete useless cachedb

u2:
`CacheDB` is useless, is it worth to retain?


quake:
yes, we can delete `CacheDB` now.

PR #1046, opened 2019-06-17
chore: fix miner may mine on the stale block

u2:
1. sleep 100 milliseconds may be a bit longer.
2. If the new work is not reached in time for some reason, the miner would mine on the stale block (https://github.com/nervosnetwork/ckb/blob/develop/miner/src/miner.rs#L96-L112), it's a bug or feature? If it's a feature, I think we should remove the sleep when receiving the stop message, and keep the start flag is true.


quake:
`try_update_block_template` is a sync rpc call, won't triger stale blcok

PR #1049, opened 2019-06-17
chore: try_into to into

u2:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3cc77c2bf9357b5328b9627d0f2ff6ef

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -538,7 +538,6 @@ mod tests {
     use ckb_verification::{BlockVerifier, HeaderResolverWrapper, HeaderVerifier, Verifier};
     use jsonrpc_types::{BlockTemplate, CellbaseTemplate};
     use numext_fixed_hash::H256;
-    use std::convert::Into;
     use std::sync::Arc;
 
     fn start_chain(
@@ -614,10 +613,10 @@ mod tests {
             .parent_hash(parent_hash);
 
         let block = BlockBuilder::default()
-            .uncles(uncles.into_iter().map(Into::into).collect())
-            .transaction(cellbase.into())
-            .transactions(transactions.into_iter().map(Into::into).collect())
-            .proposals(proposals.into_iter().map(Into::into).collect())
+            .uncles(uncles)
+            .transaction(cellbase)
+            .transactions(transactions)
+            .proposals(proposals)
             .header_builder(header_builder)
             .build();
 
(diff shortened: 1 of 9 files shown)


miner/src/block_assembler.rs
```diff
@@ -538,7 +538,7 @@ mod tests {
     use ckb_verification::{BlockVerifier, HeaderResolverWrapper, HeaderVerifier, Verifier};
     use jsonrpc_types::{BlockTemplate, CellbaseTemplate};
     use numext_fixed_hash::H256;
-    use std::convert::TryInto;
+    use std::convert::Into;
```
yangby-cryptape: Unless we use `no_std`, this is unnecessary. (not a request change, just a comment)

miner/src/block_assembler.rs
```diff
@@ -614,28 +614,10 @@ mod tests {
...
-                    .unwrap(),
-            )
-            .proposals(
-                proposals
-                    .into_iter()
-                    .map(TryInto::try_into)
-                    .collect::<Result<_, _>>()
-                    .unwrap(),
-            )
+            .uncles(uncles.into_iter().map(Into::into).collect())
```
yangby-cryptape: (not a request change, just a comment)

```rust
impl BlockBuilder {
    fn uncles<T>(mut self, arg: Vec<T>) -> Self where T: Into<UncleBlock>;
    fn uncle<T>(mut self, arg: T) -> Self where T: Into<UncleBlock>;
}
```

TheWaWaR: I prefer (like std does):
``` rust
impl BlockBuilder {
    fn uncles<T: Into<UncleBlock>>(mut self, arg: Vec<T>) -> Self;
    fn uncle<T: Into<UncleBlock>>(mut self, arg: T) -> Self;
}
```

yangby-cryptape: > I prefer (like std does):
> 
> ... ...

How many or how long you will change to the `where` statement?

Try to run `rustfmt` for this:
```rust
impl BlockBuilder {
    fn new<H: Into<Header>, U: Into<UncleBlock>, T: Into<Transaction>, O: Into<Something>>(mut self, header: H, uncles: Vec<U>, transactions: Vec<T>, others: O) -> Self {}
}
```

So, why do not use a consistent syntax?

u2: update

PR #1051, opened 2019-06-17
refactor: get tip_header from store instead of from chain_state

jjyr:


sync/src/types.rs
```diff
@@ -727,15 +727,16 @@ impl<CS: ChainStore> SyncSharedState<CS> {
         self.shared.lock_txs_verify_cache()
     }
     pub fn tip_header(&self) -> Header {
-        self.shared.lock_chain_state().tip_header().to_owned()
```
doitian: Is it ok to bypass the lock?

zhangsoledad: It's ok.

PR #1055, opened 2019-06-18
feat: add an extra cost for dynamic length types

yangby-cryptape:
**BREAKING CHANGE**:
- The occupied capacity for dynamic length types is changed, and some data and hashes have to be updated.
- The serialized size for dynamic length types is changed.


doitian:
cc @ashchan @classicalliu

doitian:
> Please add a test case to show how many capacities an empty data cell with secp256k1 lock occupied?

Can ask @jjyr to provide an example cell with the recoverable signature.

doitian:
This change should also be applied `serialized_size`, but can in a new PR.

PR #1061, opened 2019-06-19
chore(ci): use release mode to run integration

doitian:


doitian:

u2:
Because storing the cache is very slow? CITA is release mode also.

doitian:
> Because storing the cache is very slow? CITA is release mode also.

Debug mode can share the cache with test. That why I also added a environment variable on the job so the integration test job now will use their own cache.

doitian:

doitian:

doitian:

doitian:

Makefile
```diff
@@ -25,16 +25,17 @@ setup-ckb-tools:
...
 integration: setup-ckb-test ## Run integration tests in "test" dir.
 	cargo build ${VERBOSE}
-	cd test && cargo run ../target/debug/ckb
+	cd test && cargo run ${VERBOSE} -- ../target/debug/ckb
 
 integration-windows:
 	cp -f Cargo.lock test/Cargo.lock
 	cargo build ${VERBOSE}
-	cd test && cargo run ../target/debug/ckb
+	mv target test/
```
keroro520: Why `mv` here?

doitian: ln is not available in Windows

doitian:
I have rolled back the Windows changes, which still run debug mode integration now. In the future, we may try to switch to release mode to see whether it is faster.

PR #1066, opened 2019-06-19
fix: Check new block based on orphan block pool

keroro520:
* `orphan_block_pool` use `ParentHash` rather than `BlockHash` as key. So when we want to check whether we have received the same block as the arrived one, based `orphan_block_pool`, we should check deeply.

### DIFF
--- a/sync/src/synchronizer/block_pool.rs
+++ b/sync/src/synchronizer/block_pool.rs
@@ -58,7 +58,9 @@ impl OrphanBlockPool {
     pub fn contains(&self, block: &Block) -> bool {
         self.blocks
             .read()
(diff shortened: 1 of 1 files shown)


sync/src/synchronizer/block_pool.rs
```diff
@@ -58,7 +58,9 @@ impl OrphanBlockPool {
     pub fn contains(&self, block: &Block) -> bool {
         self.blocks
             .read()
-            .contains_key(block.header().parent_hash())
+            .get(block.header().parent_hash())
+            .map(|blocks| blocks.contains(block))
```
u2: `unwrap_or(false)` ?

PR #1067, opened 2019-06-19
chore: use ckb-system-scripts crate

doitian:


doitian:
ping @nervosnetwork/ckb-code-review

PR #1068, opened 2019-06-19
chore: use python to format json

doitian:


devtools/doc/jsonfmt.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import json
+import fileinput
```
yangby-cryptape: This `import fileinput` is redundant.

devtools/doc/jsonfmt.py
```diff
@@ -0,0 +1,16 @@
...
+import sys
+from collections import OrderedDict
+
+for file in sys.argv[1:]:
+    with open(file) as fp:
+        loaded = json.load(fp, object_pairs_hook=OrderedDict)
+
+    with open(file, 'w') as fp:
+        for line in json.dumps(loaded, indent=4).splitlines():
+            print(line.rstrip(), file=fp)
```
TheWaWaR: Just curious, will `json.dums` output space at the end of the line?

doitian: Curious here too. It's a bug.

PR #1071, opened 2019-06-19
✋ feat: Implement canonical serializer

jjyr:
* Implement canonical serialize, see https://github.com/nervosnetwork/ckb/blob/a6f96037c088846f5166cd75938b3aba110a00c4/util/canonical-serializer/README.md for spec
* Use canonical serialize to compute hash of core types: `Transaction`, `Header`, `Alert`.

Benchmark for canonical serialize:

```
# serialize a 2-inputs-2-outputs tx, size after canonical serialize is 1422 bytes

canonical serialize txs/serialize txs
                        time:   [370.41 ns 371.34 ns 372.36 ns]
                        thrpt:  [2.6856 Melem/s 2.6930 Melem/s 2.6997 Melem/s]
                 change:
                        time:   [-1.4475% -0.5458% +0.4167%] (p = 0.24 > 0.05)
                        thrpt:  [-0.4150% +0.5488% +1.4688%]
                        No change in performance detected.
```


doitian:
Hold after v0.15.0 RC.

util/canonical-serializer/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+        ensure!(
+            v.len() == 32,
+            "serialize U256 length error expect 32, got {}",
+            v.len()
+        );
+        self.buf.write_all(v)?;
+        Ok(self)
+    }
+
+    pub fn encode_fix_length_bytes(&mut self, v: &[u8], len: usize) -> Result<&mut Self> {
```
yangby-cryptape: So, I have to call `v.len()` before call this method, then this method call `v.len()` again?

util/canonical-serializer/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+            v.len() == 32,
+            "serialize H256 length error expect 32, got {}",
+            v.len()
+        );
+        self.buf.write_all(v)?;
+        Ok(self)
+    }
+
+    pub fn encode_u256(&mut self, v: &[u8]) -> Result<&mut Self> {
+        ensure!(
```
yangby-cryptape: Why not just call `encode_fix_length_bytes(...)`?

util/canonical-serializer/src/lib.rs
jjyr: no, this function is only for the fixed length data, like a proposal_id, the length is always 10

yangby-cryptape: So, you should not test `v.len() == len`, just test `v.len() >= len` then `write_all(&v[0..len])`.

util/canonical-serializer/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+        s.serialize(self)?;
+        Ok(self)
+    }
+
+    pub fn encode_struct_ref<T: CanonicalSerialize>(&mut self, s: &T) -> Result<&mut Self> {
+        s.serialize(self)?;
+        Ok(self)
+    }
+
+    pub fn encode_option<T: CanonicalSerialize>(&mut self, item: Option<T>) -> Result<&mut Self> {
```
doitian: I don't think we need this interface. We may need `Option<&T>` in the feature when calculating the hash from flatbuffer or capnp buffer.

util/canonical-serializer/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+pub struct CanonicalSerializer<W> {
+    buf: W,
+}
+
+impl<W: Write> CanonicalSerializer<W> {
+    pub fn new(buf: W) -> Self {
+        CanonicalSerializer { buf }
+    }
+
+    pub fn encode_u8(&mut self, v: u8) -> Result<&mut Self> {
```
doitian: Is `serialize_xx` a better name?

util/canonical-serializer/src/lib.rs
jjyr: No, the purpose is you should not use a variable length data with this method.

In your case the length of v is longer than len, this is usually misused.

For example to encode an H256 `encode_fix_length_bytes(&h256, 32)`, the `len` variable should be a constant for a specific type data.

util/canonical-serializer/src/lib.rs
jjyr: We need this because of `OutPoint` fields are optional.

util/canonical-serializer/src/lib.rs
```diff
@@ -0,0 +1,220 @@
...
+            "serialize fix length bytes error expect {}, got {}",
+            len,
+            v.len()
+        );
+        self.buf.write_all(v)?;
+        Ok(self)
+    }
+
+    pub fn encode_bytes(&mut self, v: &[u8]) -> Result<&mut Self> {
+        self.encode_u32(v.len() as u32)?;
```
driftluo: may need panic when `v.len() > u32::max()`

util/canonical-serializer/src/lib.rs
yangby-cryptape: - If they input a fixed bytes, why they have to input the length?

- The input `len` is only for test, and it's **implicit**, if the users do the test before call this, this test will be **redundant**.

  And in most scenarios, they have to do the same test before call this.

  I have to call `writer.encode_fix_length_bytes(slice[..32], 32)`, if the fixed bytes was included in a larger bytes, then it will test `32 == 32` again, no meaningless.

- The input `len` doesn't bring any features.

- IMHO:

  - Don't input the `len`, force users who call it to do tests.

  - Or, just test `v.len() >= len` then `write_all(&v[0..len])`, let the `len` do something, make it be more flexible.

  - Or, `fn encode_fix_length_bytes(&mut self, v: T, len: usize) where T: AsRef<[u8]>;`.

    I think this is not suitable for this API.

jjyr: The third one makes sense. It is too easy to write errors without the assert.

util/canonical-serializer/src/lib.rs
doitian: `encode_option_ref` can cover `Option<OutPoint>`. If we want to keep this, I think we can add a guard `<T: CanonicalSerialize + Copy>`

util/canonical-serializer/src/lib.rs
doitian: Better returns an error. Add a method `encode_len(usize)` for this.

yangby-cryptape:
Suddenly, I realized that what's different between this and [`bincode`]?

- [`bincode`] is more easy to use (`derive`).
- [`bincode`] can both [`Serialization`] and [`Deserialization`].
- In fact, if we don't care about `serde`, the core part of [`bincode`] is very simple, and it has full test (last month, it was transferred to [`Servo`]).

We can fork [`bincode`] and remove the parts that we don't need, or rewrite a simple [`bincode`].

[`bincode`]: https://github.com/servo/bincode
[`Serialization`]: https://github.com/servo/bincode/blob/master/src/ser/mod.rs
[`Deserialization`]: https://github.com/servo/bincode/blob/master/src/de/read.rs
[`Servo`]: https://github.com/servo

jjyr:
> Suddenly, I realized that what's different between this and [`bincode`](https://github.com/servo/bincode)?
> 
> * [`bincode`](https://github.com/servo/bincode) is more easy to use (`derive`).
>   I thinks `derive` is better than the follow code:
>   https://github.com/nervosnetwork/ckb/blob/71e734d92561abf924382546561878dfe44f08d1/core/src/header.rs#L81-L97
> * [`bincode`](https://github.com/servo/bincode) can both [`Serialization`](https://github.com/servo/bincode/blob/master/src/ser/mod.rs) and [`Deserialization`](https://github.com/servo/bincode/blob/master/src/de/read.rs).
> * In fact, if we don't care about `serde`, the core part of [`bincode`](https://github.com/servo/bincode) is very simple, and it has full test (last month, it was transferred to [`Servo`](https://github.com/servo)).
> 
> We can fork [`bincode`](https://github.com/servo/bincode) and remove the parts that we don't need, or rewrite a simple [`bincode`](https://github.com/servo/bincode).

I don't think `bincode` is the key point here if the only thing you want is to implement `serde`.

The **key point** is that we need an easy-to-implement canonical serialization format to compute the determination hash. 

Current API is also reasonable, notice the order of fields to serialize is also a part of our consensus, if the order changed the hash result will be changed too. So explicit order is better than implicit order(which `serde` offers) in this case IMO.

yangby-cryptape:
> .. the order of fields to serialize ..

Yes, and it's why I think [`bincode`] is better.

In this PR, the order is **hard-coded manually**, and you **have to write it for all structs**. In [`bincode`], the order is same as the order in the origin struct.

The order (ensured by [`serde`]) and encoding (ensured by [`bincode`]) are split.
You can write tests for checking order and write tests for checking encoding, separately.

But, in this PR, they are mixed!
So, until we write tests for each structs, we CAN NOT ensure them works as expected.

[`serde`]: https://github.com/serde-rs/serde
[`bincode`]: https://github.com/servo/bincode

PR #1073, opened 2019-06-19
refactor: Define a general Filter struct

keroro520:
* refactor: Define a `Filter` struct

* ~~feat: Use LruCache to avoid memory leaking~~

    ~~If we use native HashMap/HashSet as cache/filter, we have to prevent memory overflow by timely check or else. Replacing HashMap/HashSet with LruCache help us limit the memory usage, and reduce works around pruning staled items.~~

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -26,12 +26,9 @@ use lru_cache::LruCache;
 use numext_fixed_hash::H256;
 use numext_fixed_uint::U256;
 use std::cmp;
-use std::collections::{
-    hash_map::{Entry, HashMap},
-    hash_set::HashSet,
-    BTreeMap,
-};
+use std::collections::{hash_map::HashMap, hash_set::HashSet, BTreeMap};
 use std::fmt;
+use std::hash::Hash;
 use std::mem::swap;
 use std::ops::DerefMut;
 use std::sync::atomic::AtomicUsize;
@@ -211,9 +208,30 @@ impl PeerState {
     }
 }
 
-#[derive(Clone, Default)]
+#[derive(Default)]
+pub struct Filter<T: Eq + Hash> {
+    inner: LruCache<T, ()>,
+}
+
+impl<T: Eq + Hash> Filter<T> {
+    pub fn new(size: usize) -> Self {
+        Self {
+            inner: LruCache::new(size),
+        }
+    }
+
+    pub fn contains(&self, item: &T) -> bool {
+        self.inner.contains_key(item)
+    }
+
+    pub fn insert(&mut self, item: T) -> bool {
+        self.inner.insert(item, ()).is_none()
+    }
+}
+
+#[derive(Default)]
 pub struct KnownFilter {
-    inner: FnvHashMap<PeerIndex, LruCache<H256, ()>>,
+    inner: FnvHashMap<PeerIndex, Filter<H256>>,
 }
 
 impl KnownFilter {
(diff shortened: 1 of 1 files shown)


xxuejie:
Have we benched this? My concern is that HashMap (well, std's one is slow but the argument holds) is an optimized architecture with performance guarantees, with newer version of Rust they will get faster with new implementations such as hashbrown.

On the other hand, it seems like we don't need strict LRU policy here, in that case adding an LruCache might bring in a lot of unneeded logic. I feel it's better if we just cap the size of the HashMap till we know for sure this won't bring any negative effects.

sync/src/relayer/block_transactions_process.rs
```diff
@@ -32,13 +31,13 @@ impl<'a, CS: ChainStore + 'static> BlockTransactionsProcess<'a, CS> {
...
     pub fn execute(self) -> Result<(), FailureError> {
         let block_hash = cast!(self.message.block_hash())?.try_into()?;
-        if let Entry::Occupied(mut pending) = self
+        let mut success = false;
+        if let Some((compact_block, peers_set)) = self
             .relayer
             .shared()
             .pending_compact_blocks()
-            .entry(block_hash)
+            .get_mut(&block_hash)
```
u2: entry is better as before

https://github.com/nervosnetwork/lru-cache/blob/master/src/lib.rs#L132

sync/src/types.rs
```diff
@@ -211,25 +211,45 @@ impl PeerState {
...
+
+impl<T: Eq + Hash> Filter<T> {
+    pub fn new(size: usize) -> Self {
+        Self {
+            inner: LruCache::new(size),
+        }
+    }
+
+    pub fn contains(&self, item: &T) -> bool {
+        self.inner.contains_key(item)
```
u2: https://github.com/contain-rs/lru-cache/blob/master/src/lib.rs#L100
https://github.com/contain-rs/lru-cache/blob/master/src/lib.rs#L148

I am not sure whether it should refresh order.

https://github.com/nervosnetwork/lru-cache/blob/master/src/lib.rs#L42

u2:
If we care about the performance, I think linkedHashMap or LinkedHashSet is enough.

u2:
Why I am authorized to close the PR?

keroro520:
> Have we benched this? My concern is that HashMap (well, std's one is slow but the argument holds) is an optimized architecture with performance guarantees, with newer version of Rust they will get faster with new implementations such as hashbrown.
> 
> On the other hand, it seems like we don't need strict LRU policy here, in that case adding an LruCache might bring in a lot of unneeded logic. I feel it's better if we just cap the size of the HashMap till we know for sure this won't bring any negative effects.

Thanks!

Let me describe the background of this PR. Recently I am considering "spam attacks", the malicious send lots of spam requests/blocks/transactions. These spams will eat lots of memory since we don't prune timely. So I replace `HashMap` with `LruCache`. With the help of the feature "limited capacity" of `LruCache`, we can prevent the above spam attacks.

I haven't bench `LruCache` vs `HashMap`, and may not bench it later since I am gonna revert the change "HashMap -> LruCache". 

After reading your comment, I review the two fields `pending_compact_blocks` and `header_map`, which is changed from `HashMap` to `LruCache` in this PR. And find that these two fields will never suffer from spam attacks since we always verify the header's PoW via `HeaderVerifier` before inserting into the cache.

So I am gonna revert [93a6def018cb7b7966c7df6063fcf625815578a3](https://github.com/nervosnetwork/ckb/pull/1073/commits/93a6def018cb7b7966c7df6063fcf625815578a3). Maybe we can keep the first commit and treat it as a refactoring PR?

cc @u2

u2:
lgtm

PR #1074, opened 2019-06-20
test: Add test for protocols

driftluo:
1. fix discovery miss impl on add_new_addr 
2. fix discovery quick exit on add_new_addrs 
3. add test for protocols

### DIFF
--- a/network/src/protocols/discovery.rs
+++ b/network/src/protocols/discovery.rs
@@ -267,9 +267,20 @@ pub struct DiscoveryAddressManager {
 }
 
 impl AddressManager for DiscoveryAddressManager {
-    fn add_new_addr(&mut self, _session_id: SessionId, _addr: Multiaddr) {}
+    fn add_new_addr(&mut self, session_id: SessionId, addr: Multiaddr) {
+        let event = DiscoveryEvent::AddNewAddrs {
+            session_id,
+            addrs: vec![addr],
+        };
+        if self.event_sender.unbounded_send(event).is_err() {
+            debug!("receiver maybe dropped! (DiscoveryAddressManager::add_new_addrs)");
+        }
+    }
 
     fn add_new_addrs(&mut self, session_id: SessionId, addrs: Vec<Multiaddr>) {
+        if addrs.is_empty() {
+            return;
+        }
         let event = DiscoveryEvent::AddNewAddrs { session_id, addrs };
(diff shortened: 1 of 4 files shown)


network/src/protocols/discovery.rs
```diff
@@ -267,9 +267,20 @@ pub struct DiscoveryAddressManager {
...
-    fn add_new_addr(&mut self, _session_id: SessionId, _addr: Multiaddr) {}
+    fn add_new_addr(&mut self, session_id: SessionId, addr: Multiaddr) {
+        let event = DiscoveryEvent::AddNewAddrs {
+            session_id,
+            addrs: vec![addr],
+        };
+        if self.event_sender.unbounded_send(event).is_err() {
+            debug!("receiver maybe dropped! (DiscoveryAddressManager::add_new_addrs)");
+        }
+    }
```
keroro520: Suggest reuse `add_new_addrs` :

```
fn add_new_addr(&mut self, session_id: SessionId, addr: Multiaddr) {
    self.add_new_addrs(session_id, vec![addr])
}
```

network/src/protocols/test.rs
```diff
@@ -0,0 +1,386 @@
...
+
+struct Node {
+    listen_addr: Multiaddr,
+    control: ServiceControl,
+    network_state: Arc<NetworkState>,
+}
+
+impl Node {
+    fn dial(&self, node: &Node, protocol: DialProtocol) {
+        let _ = self.control.dial(node.listen_addr.clone(), protocol);
```
keroro520: If we assume that dial will be ok, then can we use `unwrap` here?
If not, ignore my comment.
The same below.

network/src/protocols/test.rs
```diff
@@ -0,0 +1,386 @@
...
+    });
+
+    Node {
+        control,
+        listen_addr,
+        network_state,
+    }
+}
+
+fn assert_help(node: &Node, expect_num: usize, step_second: u64, max_round: u8) {
```
keroro520: https://github.com/nervosnetwork/ckb/blob/develop/test/src/utils.rs#L74

A similar and more general implementation.

driftluo: > https://github.com/nervosnetwork/ckb/blob/develop/test/src/utils.rs#L74
> 
> A similar and more general implementation.

but it can't reuse here, it on `test/` not util

PR #1075, opened 2019-06-20
chore: use rtx and &[]

u2:


core/src/cell.rs
```diff
@@ -229,7 +229,7 @@ impl HeaderStatus {
     }
 }
 
-#[derive(Debug)]
```
u2: Just for test usage.

PR #1077, opened 2019-06-20
fix: resolve `ChainSpec` script deserialize issue

quake:


doitian:

spec/src/lib.rs
```diff
@@ -237,23 +237,23 @@ impl GenesisCell {
     fn build_output(&self) -> Result<CellOutput, Box<dyn Error>> {
         let mut cell = CellOutput::default();
         cell.data = self.message.as_bytes().into();
-        cell.lock.clone_from(&self.lock);
```
u2: Is the `into` needed?

doitian: They have different types, one is core script, another is jsontypes script

PR #1081, opened 2019-06-21
More tests on ckb utils

jjyr:


util/network-alert/src/tests/test_notifier.rs
```diff
@@ -0,0 +1,88 @@
...
+use std::sync::Arc;
+
+#[test]
+fn test_notice_alerts_by_version() {
+    let mut notifier = Notifier::new("0.1.0".into());
+    let alert1 = Arc::new(
+        AlertBuilder::default()
+            .id(1)
+            .max_version(Some("0.2.0".into()))
+            .min_version(Some("0.1.0".into()))
```
keroro520: A question: max_version/min_version are declared as "String" type.  Is it ok that `"0.10.0" < "0.9.10"`?

PR #1085, opened 2019-06-21
chore: standardize crate name via adding "ckb-" prefix

yangby-cryptape:
### What does this PR do

- Standardize crate name via adding "ckb-" prefix.

- Add a test in `Makefile` to detect non-standard crate names.

### Why

- We have several very simple crate name, after #1083, the probability of the same log target will be increased (ref #873).

- Preparing for publishing to [crates.io](https://crates.io).


doitian:
Hold until #1084 is merged

PR #1088, opened 2019-06-21
feat: revise epoch rpc

zhangsoledad:


yangby-cryptape:
cc @ashchan @classicalliu

PR #1094, opened 2019-06-22
feat: Secondary miner issurance, split DAO as a separate contract

xxuejie:
TODO:
* [x] tests

~This PR is still in WIP progress, it's lacking fixes for the tests so CI is doomed to fail. however it serves 2 purposes:~

* ~It provides a base so we can start writing tests for the system script PR: https://github.com/nervosnetwork/ckb-system-scripts/pull/21~
* ~Since it's a relatively large PR, publishing it earlier means we can get review process started early as well.~

Now all the tests have been fixed.


db/src/rocksdb.rs
```diff
@@ -15,7 +15,7 @@ use std::sync::Arc;
 //      - If the data can be migrated manually: update "x.y1.z" to "x.y2.0".
 //      - If the data can not be migrated: update "x1.y.z" to "x2.0.0".
 pub(crate) const VERSION_KEY: &str = "db-version";
-pub(crate) const VERSION_VALUE: &str = "0.1400.0";
+pub(crate) const VERSION_VALUE: &str = "0.1500.0";
```
doitian:
```suggestion
pub(crate) const VERSION_VALUE: &str = "0.1401.0";
```

1500 is reserved for the 0.15.0 release.

xxuejie: This is a breaking change that needs SDK updates as well, I think we should leave it till the next version, correct?

resource/specs/dev.toml
```diff
@@ -6,6 +6,9 @@ parent_hash = "0x000000000000000000000000000000000000000000000000000000000000000
 timestamp = 0
 difficulty = "0x100"
 uncles_hash = "0x0000000000000000000000000000000000000000000000000000000000000000"
+# run `cargo run cli hashes -b` to get the genesis hash
+hash = "0xd65a017201803f37d87130a94f9e19abd23369f77ee76a369b4e5040c9a8f50d"
```
doitian: A rebase conflict. We have removed `hash` from `dev.toml`, since updating is tedious for development. `issued_cells` can be deleted as well.

store/src/store.rs
```diff
@@ -130,6 +128,8 @@ pub trait ChainStore: Sync + Send {
     fn is_uncle(&self, hash: &H256) -> bool;
     // Get cellbase by block hash
     fn get_cellbase(&self, hash: &H256) -> Option<Transaction>;
+    // Get the ancestor of a base block by a block number
+    fn get_ancestor(&self, base: &H256, number: BlockNumber) -> Option<Header>;
```
doitian: Is this a refactoring or a rebase conflict that we move this method from chain provider to store?

doitian:
I'll publish the secp256k1 update first, so it will not mess up in this PR.

util/dao/src/lib.rs
```diff
@@ -1,50 +1,310 @@
...
+            .get_transaction(&cell_out_point.tx_hash)
+            .ok_or(Error::InvalidOutPoint)?;
+        let output = tx
+            .outputs()
+            .get(cell_out_point.index as usize)
+            .ok_or(Error::InvalidOutPoint)?;
+        self.calculate_maximum_withdraw(&output, &block_hash, withdraw_header_hash)
+    }
+
+    pub fn transaction_fee(&self, rtx: &ResolvedTransaction) -> Result<Capacity, FailureError> {
```
doitian: This indeed will also verify DAO withdraw. It is a looser version of the dao.c script in the system cell.

util/dao/utils/src/lib.rs
```diff
@@ -1,81 +1,60 @@
...
-                .and_then(|y| {
-                    if x > y {
-                        x.safe_sub(y)
-                    } else {
-                        Ok(Capacity::zero())
-                    }
-                })
-                .ok()
-        })
+        .skip(1)
```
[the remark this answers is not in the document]
doitian: Or should we include the first genesis message cell as well?

store/src/store.rs
xxuejie: This is a refactoring, we need this method in DaoCalculator but due to dependency problem, DaoCalculator can only get hold of store, not ChainProvider, I checked the implementation of get_ancestor and noticed it only needs a store, hence I moved the method.

util/dao/src/lib.rs
```diff
@@ -1,48 +1,267 @@
...
+            * u128::from(withdraw_ar)
+            / u128::from(deposit_ar);
+        let withdraw_capacity =
+            Capacity::shannons(withdraw_counted_capacity as u64).safe_add(occupied_capacity)?;
+
+        Ok(withdraw_capacity)
+    }
 }
 
 fn calculate_g2(
```
doitian: Should we make it consistent with block reward? A.k.a, the reminder is split evenly to the first r blocks, where r is the remainder.

doitian: Since we cannot issue reward to genesis, we should exclude genesis from g2 accumulation. A.k.a, g2 of genesis block is 0.

- G2: Secondary Reward in an Epoch
- `r_i`: Secondary rate of block i

| Block | AR |
| ----- | ---- |
| Genesis | `1` |
| Block 1 | `1`  |
| Block 2 | `1 * (1 + r_1)` |
| Block 3 | `1 * (1 + r_1) + (1 + r_2)` |

util/dao/src/lib.rs
```diff
@@ -1,48 +1,267 @@
...
+            })?;
+
+        let (parent_ar, parent_c, parent_u) = extract_dao_data(parent.dao())?;
+
+        let parent_g2 = calculate_g2(
+            parent.number(),
+            &parent_epoch,
+            self.consensus.secondary_epoch_reward(),
+        )?;
+        let parent_g = if parent.number() > self.consensus.reserve_number() {
```
doitian: C contains secondary reward, according to the paper.

util/dao/src/lib.rs
```diff
@@ -1,48 +1,267 @@
...
+            .ok_or(Error::InvalidHeader)?;
+
+        let target_g2 = calculate_g2(
+            target.number(),
+            &target_epoch,
+            self.consensus.secondary_epoch_reward(),
+        )?;
+        let (_, _, target_parent_u) = extract_dao_data(target_parent.dao())?;
+        let (_, target_c, _) = extract_dao_data(target.dao())?;
+        let reward = u128::from(target_g2.as_u64()) * u128::from(target_parent_u.as_u64())
```
doitian: If we save `U_prev` in the block header, we don't need to get parent header to calculate the miner reward. And we also get following benefits:

- The `dao` in genesis block is determined, we don't have to calculate it on the fly.
- If miners want's to pick transactions based on the block assembler template, they don't need to update `dao` in the header, since it is affected by the transactions commit in the current block.

xxuejie: At first I thought this was a good idea, but after some thought I feel it's gonna be quite weird to save `U_prev` in the block header. The problem is that in order to calculate `U_prev`, we need to resolve the whole previous block(since we need to calculate sum of all input capacities consumed in previous block), which is gonna be quite strange. I feel like to validate a block, we should only rely on the resolved data of the current block, not previous block.

And yes miner needs to update `dao` when picking transactions in the header, notice they will also need to calculate very similar things such as fees they can get, and the process of calculating `U` in `dao` resembles fee calculation a lot, so I doubt it will bring much effort here.

xxuejie:
Right now RPC test still fails, will update the related hashes after #1119 is merged and we finished rebasing this PR.

util/dao/src/lib.rs
```diff
@@ -1,256 +1,660 @@
...
+                        output
+                            .occupied_capacity()
+                            .and_then(|c| tx_capacities.safe_add(c))
+                    })
+                    .and_then(|c| capacities.safe_add(c))
+            })?;
+
+        let (parent_ar, parent_c, parent_u) = extract_dao_data(parent.dao())?;
+
+        let parent_g2 = calculate_g2(
```
doitian: This block of code can be simplified by reusing the logic to calculate base reward:

- First get the reward target of the parent block.
- If target is Some, get its base reward, otherwise `unwrap_or_default`.

```
let g = consensus.finalize_target(parent.number()).map(|target|
  self.base_block_reward(&target)
).unwrap_or_default();
```

The method `base_block_reward` does not exist yet, but is a method I recommend to add. It just returns the sum of primary and secondary reward. I have searched the code base, we never only need one of them.

xxuejie: I'm not sure if we can simplify the code here: while we have `g = g1 + g2`, and `g1` is the primary block reward, `g2` is not the secondary block reward. It's only that part of `g2`, depending on the  utilization then, can be counted as secondary block reward.

So while we might argue if we should rename `secondary_block_reward` method to `secondary_miner_block_reward`, I don't think we can simplify the code here to simply use `base_block_reward`.

doitian: Good catch

doitian: G1 and g2 are inconsistent. G1 uses the reward target of the parent block, while g2 uses the parent block.

doitian:

xxuejie:
> The line 258 in `script/src/verify.rs` iterates inputs to check if the transaction is a valid dao withdraw transaction by `valid_dao_withdraw_transaction`, in which all inputs are iterated again, thus the time complexity is O(n^2) (n is input count). Is it necessary?

This is a necessary check, tho we can change it to run linear time.

janx:
Do we still need the special `ISSUING_DAO_HASH` to identify a dao withdraw transaction? Can we check input type script instead?

xxuejie:
> Do we still need the special `ISSUING_DAO_HASH` to identify a dao withdraw transaction? Can we check input type script instead?

Yes we need that to issue the additional tokens.

janx:
> > Do we still need the special `ISSUING_DAO_HASH` to identify a dao withdraw transaction? Can we check input type script instead?
> 
> Yes we need that to issue the additional tokens.

Could you elaborate?

xxuejie:
> > > Do we still need the special `ISSUING_DAO_HASH` to identify a dao withdraw transaction? Can we check input type script instead?
> > 
> > 
> > Yes we need that to issue the additional tokens.
> 
> Could you elaborate?

That's a mistake, second thought yes we can do that.

util/dao/src/lib.rs
xxuejie: This is fixed now.

xxuejie:
@doitian @janx Issuing DAO input has been removed.

verification/src/transaction_verifier.rs
```diff
@@ -272,12 +276,7 @@ impl<'a> CapacityVerifier<'a> {
         // withdraw transactions.
         // cellbase's outputs are verified by TransactionsVerifier#InvalidReward
```
janx: Comment needs update.

util/dao/src/lib.rs
```diff
@@ -117,8 +106,20 @@ impl<'a, CS: ChainStore> DaoCalculator<'a, CS, DataLoaderWrapper<CS>> {
...
                 .ok_or(Error::InvalidHeader)?;
-            self.primary_block_reward(&target)
-                .and_then(|c| c.safe_add(parent_g2).map_err(Into::into))?
+            let target_epoch = self
+                .store
+                .get_block_epoch_index(target.hash())
+                .and_then(|index| self.store.get_epoch_ext(&index))
+                .ok_or(Error::InvalidHeader)?;
+            let parent_g2 = calculate_g2(
+                parent.number(),
```
janx: Should be `target.number()`?

xxuejie: Ouch, fixing now.

xxuejie:

PR #1097, opened 2019-06-23
fix: ibd should remain false once returned false

quake:
This PR is trying to resolve ibd peer timeout slow switching issue for some special case, e.x. node resume after hibernation, it's a performance optimization also.

bitcoin reference: https://github.com/bitcoin/bitcoin/blob/0b68fca700713e8e4c843c982b6047dc04410bc0/src/validation.cpp#L1027


doitian:
What a surprise.

PR #1104, opened 2019-06-24
chore: random keypair

u2:


script/src/verify.rs
```diff
@@ -1643,13 +1514,17 @@ mod tests {
         let verifier = TransactionScriptsVerifier::new(&rtx, &data_loader, &config);
 
         assert!(verifier.verify(100_000_000).is_err());
+        // TODO: FIXME
+        // assert_eq!(
+        //     verifier.verify(100_000_000).err(),
+        //     Some(ScriptError::ValidationFailure(-7))
+        // );
```
u2: The error is changing every time running the test case, I will fix it later.

script/src/verify.rs
```diff
@@ -630,6 +631,35 @@ mod tests {
         ChainKVStore::new(MemoryKeyValueDB::open(COLUMNS as usize))
     }
 
+    fn random_keypair() -> (Privkey, Pubkey) {
+        let gen = Generator::new();
+        gen.random_keypair()
+    }
+
+    fn to_hex_pubkey(pubkey: Pubkey) -> Vec<u8> {
```
TheWaWaR: Better use `&Pubkey` as argument

script/src/verify.rs
```diff
@@ -630,6 +631,35 @@ mod tests {
...
+        let pubkey = pubkey.serialize();
+        let mut hex_pubkey = vec![0; pubkey.len() * 2];
+        hex_encode(&pubkey, &mut hex_pubkey).expect("hex pubkey");
+        hex_pubkey
+    }
+
+    fn to_hex_signature(signature: Signature) -> Vec<u8> {
```
TheWaWaR: Better use `&Signature`.

script/src/verify.rs
```diff
@@ -630,6 +631,35 @@ mod tests {
...
+        let signature_der = signature.serialize_der();
+        let mut hex_signature = vec![0; signature_der.len() * 2];
+        hex_encode(&signature_der, &mut hex_signature).expect("hex signature");
+        hex_signature
+    }
+
+    fn sign_args(args: &[Bytes], privkey: Privkey) -> Signature {
```
TheWaWaR: Better use `&Privkey`.

script/src/verify.rs
TheWaWaR: @xxuejie 
I think this maybe a serious problem. Can you explain this?

xxuejie: It shouldn't happen, what are the different errors getting here?

script/src/verify.rs
u2: updated.

script/src/verify.rs
xxuejie: Actually, we already moved those tests to a different repo: https://github.com/nervosnetwork/ckb-system-scripts/tree/master/tests. And those tests have been removed in an upcoming PR: https://github.com/nervosnetwork/ckb/pull/1094. So I think we should move the PR if needed.

u2: The error would be `Some(ScriptError::ValidationFailure(-7))` or `Some(ScriptError::ValidationFailure(-3))`.

I have not inspected carefully, maybe it's just the concurrent test problem. I would delete the second commit and consider another PR for https://github.com/nervosnetwork/ckb-system-scripts

PR #1105, opened 2019-06-24
feat: Add cli subcommands

TheWaWaR:
* Add a library `sdk` in `util` directory, this library is common code for client applications
* Extend current `ckb cli` subcommands
  - `account` for keystore management
  - `wallet` for transfer capacity, query live cells
  - `local` for local transaction debug
  - `rpc` for call rpc methods
* Move some code from `ckb-app-config` to `ckb-bin`


TheWaWaR:
We consider use a standalone repo for [ckb-cli](http://github.com/nervosnetwork/ckb-cli), we will distribute `ckb-cli` binary and `ckb` in one package for user to access.

doitian:
Let's just package the binary in release.

PR #1109, opened 2019-06-25
fix: mark failed dialing as feeler

jjyr:


TheWaWaR:
It's a dirty way to solve the problem (assume it can).

Better filter out target peers by `can_dial()` before `dial_peers()`

jjyr:
> It's a dirty way to solve the problem (assume it can).
> 
> Better filter out target peers by `can_dial()` before `dial_peers()`

What this PR does is to resolve an obvious bug. Filter address before `dial` is easy to miss the check, we have changed from that to the current implementation, why change it back is a `better way`?

TheWaWaR:
A more better way is to remember this and do not dial, not like this.

network/src/network.rs
```diff
@@ -387,15 +387,17 @@ impl NetworkState {
         self.dialing_addrs.write().remove(&peer_id);
     }
 
+    /// Dial
+    /// return value indicate the dialing is actually sent or dynied.
```
quake: typo
```suggestion
    /// return value indicates the dialing is actually sent or denied.
```

network/src/network.rs
```diff
@@ -387,15 +387,17 @@ impl NetworkState {
...
         &self,
         p2p_control: &ServiceControl,
         peer_id: &PeerId,
         mut addr: Multiaddr,
         target: DialProtocol,
-    ) {
+    ) -> bool {
         if !self.can_dial(peer_id, &addr) {
-            return;
+            return false;
```
quake:
```suggestion
        if self.can_dial(peer_id, &addr) {
            match Multihash::from_bytes(peer_id.as_bytes().to_vec()) {
                Ok(peer_id_hash) => {
                    addr.push(multiaddr::Protocol::P2p(peer_id_hash));
                    debug!("dialing {} with {:?}", addr, target);
                    if let Err(err) = p2p_control.dial(addr.clone(), target) {
                        debug!("dial failed: {:?}", err);
                    } else {
                        self.dialing_addrs
                            .write()
                            .insert(peer_id.to_owned(), Instant::now());
                        return true;
                    }
                }
                Err(err) => {
                    error!("failed to convert peer_id to addr: {}", err);
                }
            }
        }
        false
```

PR #1119, opened 2019-06-26
feat: Remove rules of special reserve blocks

doitian:
For block 1~11, the reward target is genesis block. Yes the reward for genesis is issued multiple times.

Genesis block must have the lock serialized in the cellbase witness, which is set to `bootstrap_lock`.


verification/src/tests/uncle_verifier.rs
```diff
@@ -128,7 +129,12 @@ fn prepare() -> (
         parent = new_block.header().to_owned();
     }
 
-    (shared, chain1, chain2)
+    // the second chain must have smaller hash
+    if chain1.last().unwrap().header().hash() < chain2.last().unwrap().header().hash() {
+        (shared, chain2, chain1)
```
u2: :+1:

doitian:
The most important change is in `Consensus::finalize_target`. Most of the changes are adding witness for the cellbase in the test.

verification/src/contextual_block_verifier.rs
```diff
@@ -232,33 +232,16 @@ where
...
+        let (target_lock, block_reward) = self
+            .provider
+            .finalize_block_reward(self.parent)
+            .map_err(|_| Error::CannotFetchBlockReward)?;
+        if cellbase.transaction.outputs_capacity()? != block_reward {
+            return Err(Error::Cellbase(CellbaseError::InvalidRewardAmount));
+        }
+        if cellbase.transaction.outputs()[0].lock != target_lock {
+            return Err(Error::Cellbase(CellbaseError::InvalidRewardTarget));
+        }
```
janx: What if the miner construct a cellbase with two outputs whose sum equals `block_reward` but use a different lock on second output?

doitian: There's a verification rule that cellbase can only have one input and one output.

zhangsoledad: pub fn is_cellbase(&self) -> bool {
        self.inputs.len() == 1
            && self.outputs.len() == 1
            && self.witnesses.len() == 1
            && self.inputs[0].previous_output.is_null()
    }

we enforce `cellbase` have only one output.

u2: https://github.com/nervosnetwork/ckb/blob/develop/verification/src/block_verifier.rs#L50

PR #1124, opened 2019-06-27
feat: Allow TransactionScriptsVerifier set custom debugger printer

TheWaWaR:
Current `Debugger` only log the output and can not detected by outside.

Here we pass a callback to process debug output.


TheWaWaR:
Current use case is `ckb-cli` need to print the output of `ckb_debug`.

xxuejie:
In that case I really suggest changing the code to allow plugging in a different debugger instance, the current solution is a little weird

TheWaWaR:
The difficulty is about lifetime + generic + associated type a parameter, I'll have a try.

xxuejie:
Why not just use a Box?

PR #1126, opened 2019-06-28
perf: skip stored block processing

quake:


keroro520:
> I think it is better to place the check of "is already stored" outside "insert_block", such as [`accept_block`](https://github.com/nervosnetwork/ckb/blob/develop/sync/src/synchronizer/mod.rs#L189-L190).
> And then we can re-use `get_block_status` to do this check.

https://github.com/nervosnetwork/ckb/pull/1127 makes synchronizer and relayer share the same implementation of `insert_new_block`/`accept_block`, so we can just add this check to `accept_block`.

quake:
> I think it is better to place the check of "is already stored" outside "insert_block", such as [`accept_block`](https://github.com/nervosnetwork/ckb/blob/develop/sync/src/synchronizer/mod.rs#L189-L190).
> And then we can re-use `get_block_status` to do this check.

we can't assume that the invoker will check the block existence before calling this function, e.x. importing data from json, we should skip already stored blocks:
 https://github.com/nervosnetwork/ckb/blob/develop/util/instrument/src/import.rs#L49, 

it's necessary to add this as final checking.

keroro520:
> > I think it is better to place the check of "is already stored" outside "insert_block", such as [`accept_block`](https://github.com/nervosnetwork/ckb/blob/develop/sync/src/synchronizer/mod.rs#L189-L190).
> > And then we can re-use `get_block_status` to do this check.
> 
> we can't assume that the invoker will check the block existence before calling this function, e.x. importing data from json, we should skip already stored blocks:
> https://github.com/nervosnetwork/ckb/blob/develop/util/instrument/src/import.rs#L49,
> 
> it's necessary to add this as final checking.

👌

PR #1127, opened 2019-06-28
fix: Process orphan blocks when their parents were relayed

keroro520:
* test: Add an integration-test `test_relay_parent_of_orphan_block`
* feat: Move `orphan_block_pool` into `SyncSharedState`
* fix: Process orphan blocks when their parents were relayed. If relayer does not process the orphan_block_pool when inserting a new relaying block, it may leave some orphan blocks that be unable to re-process any further. This change makes synchronizer and relayer share the same one implementation of `insert_new_block`, so the above bug been fixed.


sync/src/types.rs
```diff
@@ -1067,6 +1085,78 @@ impl<CS: ChainStore> SyncSharedState<CS> {
         self.inflight_blocks.write().remove_by_peer(&pi);
         self.peers().disconnected(pi)
     }
+
+    pub fn insert_new_block(&self, chain: &ChainController, pi: PeerIndex, block: Block) -> bool {
+        let known_parent = |block: &Block| {
+            self.store()
+                .get_block_header(block.header().parent_hash())
```
quake:
```suggestion
                .get_block_number(block.header().parent_hash())
```

test/src/specs/relay/compact_block.rs
```diff
@@ -193,6 +196,88 @@ impl CompactBlockBasic {
...
+    // Case: A <- B, A == B.parent
+    // 1. Sync B to node0. Node0 will put B into orphan_block_pool since B's parent unknown
+    // 2. Relay A to node0. Node0 will handle A, and by the way process B, which is in
+    // orphan_block_pool now
+    pub fn test_relay_parent_of_orphan_block(&self, net: &Net, node: &Node, peer_id: PeerIndex) {
+        node.generate_block();
+
+        // Proposal a tx, and grow up into proposal window
+        let new_tx = node.new_transaction_spend_tip_cellbase();
+        // node.generate_blocks(10);
```
quake: please clean up this line

sync/src/relayer/mod.rs
```diff
@@ -215,10 +215,11 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
...
     }
 
-    pub fn accept_block(&self, nc: &CKBProtocolContext, peer: PeerIndex, block: &Arc<Block>) {
-        let ret = self.chain.process_block(Arc::clone(&block), true);
-
-        if ret.is_ok() {
+    pub fn accept_block(&self, nc: &CKBProtocolContext, peer: PeerIndex, block: Block) {
+        if self
+            .shared()
+            .insert_new_block(&self.chain, peer, block.clone())
```
driftluo: This should be an Arc<block>.

sync/src/types.rs
```diff
@@ -43,6 +46,7 @@ const GET_HEADERS_CACHE_SIZE: usize = 10000;
 const GET_HEADERS_TIMEOUT: Duration = Duration::from_secs(15);
 const TX_FILTER_SIZE: usize = 50000;
 const TX_ASKED_SIZE: usize = TX_FILTER_SIZE;
+const ORPHAN_BLOCK_SIZE: usize = 1024;
```
driftluo: I am not sure if the configuration item should be removed here.

sync/src/relayer/mod.rs
keroro520: Updated. Change the parameters of `insert_new_block` to `block: Arc<Block>`

sync/src/types.rs
keroro520: If we want the sync-parameters configurable, then others like `TX_FILTER_SIZE`/`MAX_ASK_MAP_SIZE` should be put into `SyncConfig` too.

sync/src/types.rs
```diff
@@ -1067,6 +1085,82 @@ impl<CS: ChainStore> SyncSharedState<CS> {
...
+                debug!(
+                    "parent-unknown orphan block, block: {}, {:x}, parent: {:x}",
+                    block.header().number(),
+                    block.header().hash(),
+                    block.header().parent_hash(),
+                );
+                continue;
+            }
+
+            if let Err(err) = self.accept_block(chain, pi, block.clone()) {
```
quake: avoids clone here, and remove the beginning line of loop: `let block = Arc::new(block);`
, change `debug` to`warn` and only output block hash.

keroro520: 1. I will change the log as your comment.
2. Rust owner mechanism requires clone here if we wanna log `block_hash` within the error message. And this is Arc clone, it is cheap.

PR #1130, opened 2019-06-28
✋ chore: print diff using symbols in CI

doitian:


doitian:
See https://github.com/nervosnetwork/ckb/pull/1132

PR #1150, opened 2019-07-02
fix: correct block number from tx_pool_excutor

keroro520:
* fix: The block number of the coming block should be `tip_number + 1`

-------

NOTE: **This is a breaking change**

### DIFF
--- a/shared/tx-pool-executor/src/tx_pool_executor.rs
+++ b/shared/tx-pool-executor/src/tx_pool_executor.rs
@@ -57,11 +57,11 @@ impl<CS: ChainStore> TxPoolExecutor<CS> {
         }
         // resolve txs
         // early release the chain_state lock because tx verification is slow
-        let (resolved_txs, cached_txs, unresolvable_txs, consensus, tip_number, epoch_number) = {
+        let (resolved_txs, cached_txs, unresolvable_txs, consensus, block_number, epoch_number) = {
             let chain_state = self.shared.lock_chain_state();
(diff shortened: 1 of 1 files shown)


shared/tx-pool-executor/src/tx_pool_executor.rs
```diff
@@ -347,4 +348,54 @@ mod tests {
...
+
+    #[test]
+    fn test_verify_and_add_invalid_since_tx_to_pool() {
+        let (shared, always_success_out_point) = setup(10);
+        let last_block = shared
+            .store()
+            .get_block(&shared.lock_chain_state().tip_hash())
+            .unwrap();
+        let last_cellbase = last_block.transactions().first().unwrap();
+        let since_absolute_block_number = |block_number| block_number;
```
jjyr: why?

jjyr: why?

jjyr: Please remove this function.

keroro520: When we want to build a `since` which represents "since an absolute block number":

```
let since = 0b000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000 | block_number
```

But Clippy warns `0b000...000` is redundant. So I remove `0b000..000` but keep the function.

PR #1151, opened 2019-07-02
[ᚬrc/v0.15.z] feat: allow providing extra sentry info in config

doitian:
```toml
[sentry]
dsn = "https://48c6a88d92e246478e2d53b5917a887c@sentry.io/1422795"
org_ident = "ckb-dev"
# org_ident = "ckb-testnet"
# org_ident = "ckb-test"
org_contact = "test@example.org"
```


keroro520:
Maybe we can print a log to mention the users to fill `org_contact` when the node starts, just like `dsn` does.

PR #1164, opened 2019-07-03
[ᚬrc/v0.15.z] fix: Ban peer when validate received block failed

TheWaWaR:


TheWaWaR:
> I think we should not pass network context around, but pop up the error instead.

Agreed

sync/src/synchronizer/mod.rs
```diff
@@ -167,15 +167,15 @@ impl<CS: ChainStore> Synchronizer<CS> {
...
-            return;
+            return Ok(());
         }
 
         match self.shared().get_block_status(&block.header().hash()) {
             BlockStatus::VALID_MASK => {
-                self.insert_new_block(peer, block);
+                self.insert_new_block(peer, block)?;
             }
             status => {
```
keroro520: Why not return an error when `BlockStatus` is unexpected?

TheWaWaR: I'm not sure it's a case to ban this peer

keroro520: I am not sure, either. Keep as it is.

PR #1165, opened 2019-07-03
feat: Reference script code via dep cell's type hash

xxuejie:
This allows us to build a new paradigm that allows upgrading of
scripts without affecting lock/type hash.


xxuejie:

keroro520:
Oh, maybe we should add a `hash_type: 1` example into readme.

util/jsonrpc-types/src/blockchain.rs
```diff
@@ -14,29 +14,41 @@ use ckb_core::Capacity as CoreCapacity;
 use numext_fixed_hash::H256;
 use numext_fixed_uint::U256;
 use serde_derive::{Deserialize, Serialize};
+use std::convert::TryInto;
 
 #[derive(Clone, Default, Serialize, Deserialize, PartialEq, Eq, Hash, Debug)]
 pub struct Script {
     pub args: Vec<JsonBytes>,
     pub code_hash: H256,
+    pub hash_type: u8,
```
jjyr: Can we use an enum struct here, so the RPC request became `"hash_type": "data"` or `"hash_type": "type"`, invalid format will be rejected by RPC server.

xxuejie: I don't mind changing it to a string(tho I personally feel u8 is better since it's more consistent with the wire protocol). But please see [here](https://github.com/nervosnetwork/ckb/pull/1165/files#diff-f71eef27edcb99c70fe66cf82c33b18bR33) for reason why I didn't reject the invalid format. It's just due to right now we are all using plain `From` trait here, and adding this rejection would involve making a whole lot of change from `From` trait to `TryFrom` trait here. Since this PR is already huge, I prefer to leave it till another time when we really have the need.

jjyr: if `hash_type` is an unexpected enum string in JSON, `serde` will return an error, so we do not need to change `From` to `TryFrom`.

xxuejie: I didn't realize we could do it this way, already changed, thanks.

xxuejie:

PR #1168, opened 2019-07-03
refactor: Use BlockStatus to filter things

keroro520:
* refactor: Move `orphan_block_pool` into `SyncSharedState`
* feat: Rewrite `BlockStatus` and `block_status_map`
* perf: Use `BlockStatus` to avoid re-processing the same headers/blocks

### DIFF
--- a/sync/src/block_status.rs
+++ b/sync/src/block_status.rs
@@ -4,22 +4,23 @@ bitflags! {
     pub struct BlockStatus: u32 {
         const UNKNOWN                 =     0;
 
-        const HEADER_VERIFIED         =     1;
-        const BLOCK_RECEIVED          =     Self::HEADER_VERIFIED.bits | 2;
-        const BLOCK_STORED            =     Self::HEADER_VERIFIED.bits | Self::BLOCK_RECEIVED.bits | 16;
+        const HEADER_VALID            =     0b0000_0000_0000_0001;
+        const BLOCK_RECEIVED          =     Self::HEADER_VALID.bits | 0b0000_0000_0000_0010;
+        const BLOCK_STORED            =     Self::HEADER_VALID.bits | Self::BLOCK_RECEIVED.bits | 0b0000_0000_0000_1000;
 
-        const BLOCK_INVALID           =    32;
+        const BLOCK_INVALID           =     0b0010_0000_0000_0000;
     }
 }
 
 #[cfg(test)]
 mod tests {
     use crate::block_status::BlockStatus;
+    use std::collections::HashSet;
 
     fn all() -> Vec<BlockStatus> {
         vec![
             BlockStatus::UNKNOWN,
-            BlockStatus::HEADER_VERIFIED,
+            BlockStatus::HEADER_VALID,
             BlockStatus::BLOCK_RECEIVED,
             BlockStatus::BLOCK_STORED,
             BlockStatus::BLOCK_INVALID,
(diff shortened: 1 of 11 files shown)


sync/src/block_status.rs
```diff
@@ -0,0 +1,81 @@
+use bitflags::bitflags;
+
+bitflags! {
+    pub struct BlockStatus: u32 {
+        const UNKNOWN                 =     0;
+
+        const HEADER_VERIFIED         =     1;
+        const BLOCK_RECEIVED          =     Self::HEADER_VERIFIED.bits | 2;
+        const BLOCK_STORED            =     Self::HEADER_VERIFIED.bits | Self::BLOCK_RECEIVED.bits | 16;
```
TheWaWaR: Should clear old stored block in case of fork.

sync/src/relayer/compact_block_process.rs
```diff
@@ -43,6 +44,20 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
+        if status.contains(BlockStatus::BLOCK_STORED) {
+            return Ok(());
+        } else if status.contains(BlockStatus::BLOCK_INVALID) {
+            debug_target!(
+                crate::LOG_TARGET_RELAY,
+                "receive a compact block with invalid status, {:#x}, peer: {}",
+                block_hash,
+                self.peer
+            );
+            // TODO ban/punish this peer
```
TheWaWaR: Punish should happen when insert `BLOCK_INVALID`.

keroro520:

sync/src/relayer/compact_block_process.rs
keroro520: We should punish the peer in 2 scenarios:

  * when inserts `BLOCK_INVALID`
  * receive a block already marked as `BLOCK_INVALID` before

But I think "ban" is dangerous, we should do it in another PR, but not here. Maybe after refactoring the scoring mechanism. So I just leave a TODO.

sync/src/block_status.rs
keroro520: Discussed.

sync/src/block_status.rs
```diff
@@ -0,0 +1,88 @@
...
+        vec![
+            BlockStatus::UNKNOWN,
+            BlockStatus::HEADER_VALID,
+            BlockStatus::BLOCK_RECEIVED,
+            BlockStatus::BLOCK_STORED,
+            BlockStatus::BLOCK_INVALID,
+        ]
+    }
+
+    fn assert_(includes: Vec<BlockStatus>, target: BlockStatus) {
```
doitian:
```suggestion
    fn assert_contain(includes: Vec<BlockStatus>, target: BlockStatus) {
```

sync/src/synchronizer/headers_process.rs
```diff
@@ -393,49 +393,74 @@ where
 
     pub fn accept(&self) -> ValidationResult {
         let mut result = ValidationResult::default();
+
+        // FIXME If status == BLOCK_INVALID then return early. But which error
```
doitian: Use a separate error code such as DuplicatedInvalidBlock or save the invalid reason as well.

keroro520: I would like to persist a `recent_reject { hash => reason }` to record the invalid reasons.

PR #1183, opened 2019-07-04
[ᚬrc/v0.15.z] fix: ibd should remain false once returned false

quake:
Backports the following commits to rc/v0.15.z:
 - fix: ibd should remain false once returned false (#1097)


quake:
@nervos-bot-user give me five

PR #1190, opened 2019-07-04
[ᚬrc/v0.15.z] fix: fix sync logic

quake:
Backports the following commits to rc/v0.15.z:
 - fix: fix sync logic (#1103)


quake:
@nervos-bot-user give me five

PR #1191, opened 2019-07-05
fix: Use tip header to ignore compact block

quake:
Backports the following commits to develop:

fix: Use tip header to ignore compact block (#1176)


doitian:
duplicated of #1197

PR #1192, opened 2019-07-05
chore: better name

u2:


sync/src/relayer/tests/mod.rs
```diff
@@ -1,2 +1,2 @@
-mod compact_block_process;
 mod compact_block_verifier;
+mod reconstruct_block;
```
keroro520: I think the original file name is a little bit better. It is an unit-test file corresponding to `compact_block_process.rs`. Although it has only one test yet...

u2: It does not test compact block process at all.

keroro520: 👌

PR #1195, opened 2019-07-05
[ᚬrc/v0.15.z] hotfix rc0.15

zhangsoledad:
* fix fetch invalid block
* fix response invalid block
* fix repeat process block overwrite block ext

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -30,7 +30,7 @@ use std::sync::Arc;
 use std::{cmp, thread};
 use stop_handler::{SignalSender, StopHandler};
 
-type ProcessBlockRequest = Request<(Arc<Block>, bool), Result<(), FailureError>>;
+type ProcessBlockRequest = Request<(Arc<Block>, bool), Result<bool, FailureError>>;
 
 #[derive(Clone)]
 pub struct ChainController {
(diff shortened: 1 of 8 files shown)


chain/src/chain.rs
```diff
@@ -216,26 +220,36 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
...
         }
 
         let mut total_difficulty = U256::zero();
-
         let mut fork = ForkChanges::default();
         let mut chain_state = self.shared.lock_chain_state();
         let mut txs_verify_cache = self.shared.lock_txs_verify_cache();
 
+        // after acquire lock
+        if self.shared.store().block_exists(block.header().hash()) {
```
keroro520: We involve `insert_block` to update the database in a single one thread. So it is unnecessary to acquire the lock before checking. 

This thread does: `while chan.recv(block) { insert_block(block) }`

doitian:
```
error: using '.clone()' on a ref-counted pointer
  --> chain/src/tests/basic.rs:39:24
   |
39 |         .process_block(block.clone(), true)
   |                        ^^^^^^^^^^^^^ help: try this: `Arc::<ckb_core::block::Block>::clone(&block)`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_on_ref_ptr
```

PR #1202, opened 2019-07-08
fix: `--ba-code-hash` `--ba-arg` can be used together

doitian:
`--ba-data` requires any of them, so they are included in a `ArgGroup`. But
`ArgGroup` requires one and only one must be present by default, which is
disabled with `multiple(true)`


quake:
backport to 0.16?

doitian:
> backport to 0.16?

only affect develop

PR #1203, opened 2019-07-09
feat: Add bootnode mode

driftluo:
1. Add boot node mode to avoid the inability to serve due to inbound full
2. ~Make ibd time range configurable~


network/src/protocols/mod.rs
```diff
@@ -348,4 +352,17 @@ impl CKBProtocolContext for DefaultCKBProtocolContext {
     fn protocol_id(&self) -> ProtocolId {
         self.proto_id
     }
+
+    fn inbound_eviction_condition(&self) -> bool {
+        let status = self.network_state.connection_status();
```
quake: check bootnode mode first:

```suggestion
        if self.network_state.config.bootnode_mode {
           ...
        }
```

sync/src/synchronizer/mod.rs
```diff
@@ -287,6 +288,10 @@ impl<CS: ChainStore> Synchronizer<CS> {
                     }
                 }
             }
+            if inbound_eviction && !state.is_outbound && index & 0x1 != 0 {
```
quake:
```suggestion
            else if inbound_eviction && index & 0x1 != 0 {
```

network/src/network.rs
```diff
@@ -447,6 +447,41 @@ pub struct EventHandler {
...
 
+impl EventHandler {
+    fn inbound_eviction(&self, context: &mut ServiceContext) {
+        if self.network_state.config.bootnode_mode {
+            let status = self.network_state.connection_status();
+
+            if status.max_inbound
+                - status.unreserved_inbound
+                - self.network_state.config.reserved_peers.len() as u32
+                <= 10
```
keroro520: It assumes that `max_peers` is large(should be greater than 10 * 2).
What if I wanna set the `max_peers  = reserved_peers.len()`, in other words, I want to ensure that only connect to the reserved peers?

keroro520: Suggest change to avoid overflow panic:

```
 if status.max_inbound <=
                status.unreserved_inbound
                + self.network_state.config.reserved_peers.len() as u32
                + 10
```

driftluo: I found another overflow bug...

PR #1204, opened 2019-07-09
test: insert_new_block

keroro520:
* test: SyncSharedState insert_new_block
* integration-test: Relay cmpblock which difficulty lt shared best known
* fix: Correct return value of insert_new_block

### DIFF
--- a/sync/src/tests/sync_shared_state.rs
+++ b/sync/src/tests/sync_shared_state.rs
@@ -1,0 +1,125 @@
+use crate::tests::util::{build_chain, inherit_block};
+use crate::types::BlockStatus;
+use crate::SyncSharedState;
+use ckb_chain::chain::ChainService;
+use ckb_core::block::BlockBuilder;
+use ckb_core::header::HeaderBuilder;
+use ckb_core::Capacity;
+use ckb_db::MemoryKeyValueDB;
+use ckb_network::PeerIndex;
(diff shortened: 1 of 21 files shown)


sync/src/tests/sync_shared_state.rs
```diff
@@ -0,0 +1,125 @@
...
+        let invalid_cellbase = always_success_cellbase(tip_number, Capacity::zero());
+        let next_block = inherit_block(shared.shared(), tip_number)
+            .transaction(invalid_cellbase)
+            .build();
+        Arc::new(next_block)
+    };
+
+    assert!(shared
+        .insert_new_block(&chain, PeerIndex::new(1), Arc::clone(&invalid_block))
+        .is_err(),);
```
u2: `is_err` is obscure, assert the exact err type.

sync/src/types.rs
```diff
@@ -1148,18 +1149,19 @@ impl<CS: ChainStore> SyncSharedState<CS> {
             }
         }
 
-        Ok(true)
+        ret
```
u2: :+1:

sync/src/tests/sync_shared_state.rs
keroro520: It is `FailureError`. assert it is tedious...

PR #1210, opened 2019-07-10
chore: Remove useless known_blocks filter

keroro520:
Originally we use `known_blocks` to avoid re-relay the same blocks to the same peers.
But now our code ensures that, as for the same blocks, we accept and relay only once. So this block-filter is useless.


quake:
Conflicting... please rebase, thanks.

PR #1211, opened 2019-07-10
feature: ban the peer when receive misbehave compact-block, add test for compact block process

u2:


sync/src/relayer/compact_block_process.rs
```diff
@@ -23,6 +23,18 @@ pub struct CompactBlockProcess<'a, CS> {
...
+    UnknownParent,
+    NotBetter,
+    AlreadyInFlight,
+    AlreadyPending,
+    HeaderVerifyFailed,
+    InflightBlocksReachLimit,
+    SendMissingIndexes,
+    NoMissingIndexes,
+}
+
```
keroro520: 1. It is worth a more reasonable name. For example, `SendMissingIndexes` is not a "status", `HeaderVerifyFailed` is an error.
2. Many other places can share this notion. Maybe we should define it as a global type, and use it to help score-system.

sync/src/relayer/tests/helper.rs
```diff
@@ -0,0 +1,144 @@
+use crate::{Relayer, SyncSharedState};
```
keroro520: YFI, I am planning to move `sycn/src/relayer/tests` into `sync/src/tests/`, after https://github.com/nervosnetwork/ckb/pull/1204.

sync/src/relayer/compact_block_process.rs
u2: 1. Naming is hard.
2. Agree. There is too much `Result<(), E>`, it's not good to read, test and debug.

sync/src/relayer/tests/helper.rs
u2: And can reuse the test utils.

sync/src/relayer/compact_block_process.rs
u2: `Status` is the only name which I can think, any suggestion?

doitian: The problem here is that we misuse `Result` here, since we want to ignore some errors. It's better to categorize errors like HTTP, which 4xx are errors caused by clients, and 5xx are caused by the server.

sync/src/types.rs
```diff
@@ -1175,7 +1175,7 @@ mod tests {
     use ckb_core::header::HeaderBuilder;
     use rand::{thread_rng, Rng};
 
-    const SKIPLIST_LENGTH: u64 = 500_000;
```
doitian: This should be in a separate PR.

sync/src/relayer/compact_block_process.rs
u2: These are not errors.

u2:
@doitian , @keroro520 updated.

I consider is it necessary to cache the result after verifying the header?
https://github.com/nervosnetwork/ckb/pull/1211/files#diff-98eccb89a5611099d6adca819c9c8b08L100
https://github.com/nervosnetwork/ckb/pull/1211/files#diff-98eccb89a5611099d6adca819c9c8b08L156

u2:
@doitian , @keroro520  updated

sync/src/relayer/mod.rs
```diff
@@ -168,8 +168,10 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
         message: RelayMessage,
     ) {
         if let Err(err) = self.try_process(Arc::clone(&nc), peer, message) {
-            debug_target!(crate::LOG_TARGET_RELAY, "try_process error {}", err);
-            nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+            if let Ok(Error::Misbehavior(e)) = err.downcast() {
+                debug_target!(crate::LOG_TARGET_RELAY, "try_process error {}", e);
+                nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+            }
```
zhangsoledad: the original logical will ban malformed message, `Misbehavior` didn't cover this

u2: updated

sync/src/relayer/error.rs
```diff
@@ -2,6 +2,24 @@ use failure::Fail;
...
+    #[fail(display = "misbehavior error: {}", _0)]
+    Misbehavior(Misbehavior),
+    #[fail(display = "ignored error: {}", _0)]
+    Ignored(Ignored),
+}
+
+#[derive(Debug, Fail, Eq, PartialEq)]
+pub enum Internal {
+    #[fail(display = "InflightBlocksReachLimit")]
+    InflightBlocksReachLimit,
```
keroro520: It is better to change the category as below:

* move `InflightBlocksReachLimit` into `Ignore` 
* `MissingIndexesEmpty` is not an `Error`. So I don't think define `MissingIndexesEmpty` is a good idea. If we don't define it, then we can remove the definition of `Internal` (after moving `InflightBlocksReachLimit` into `Ignore` )
* rename `Ignore::BlockStored` -> `Ignore::AlreadyStored`

sync/src/relayer/mod.rs
```diff
@@ -168,8 +169,19 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
...
     ) {
         if let Err(err) = self.try_process(Arc::clone(&nc), peer, message) {
-            debug_target!(crate::LOG_TARGET_RELAY, "try_process error {}", err);
-            nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+            if let Some(&Error::Misbehavior(ref e)) = err.downcast_ref() {
+                debug_target!(crate::LOG_TARGET_RELAY, "try_process error {}", e);
+                nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+                return;
+            }
+            if let Some(&ProtocalError::Malformed) = err.downcast_ref() {
```
keroro520: Use `else if` or `match` statement.

sync/src/relayer/compact_block_process.rs
```diff
@@ -197,40 +210,44 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
-                fbb,
-                &block_hash,
-                &missing_indexes
-                    .into_iter()
-                    .map(|i| i as u32)
-                    .collect::<Vec<_>>(),
+        // The missing_indexes may never be empty
+        if missing_indexes.is_empty() {
+            return Err(Error::Internal(Internal::MissingIndexesEmpty).into());
+        }
```
keroro520: `if missing_indexes.is_empty()` means `Ok(())`, not an error.

sync/src/relayer/mod.rs
u2: If using `else if`, the Clippy would suggest it can be ignored.
We could no use `match` here, it could not derive the type, the downcast_ref return a generic type.

sync/src/relayer/compact_block_process.rs
u2: The `missing_indexes` should never be empty. Otherwise, there would be some logical error in `reconstruct_block`.

sync/src/relayer/error.rs
u2: > It is better to change the category as below:
> 
> * move `InflightBlocksReachLimit` into `Ignore`

I don't want to move it, because in this case, the node could not accept this compact block just because it's busy, the node does not ignore it deliberately.

> * `MissingIndexesEmpty` is not an `Error`. So I don't think define `MissingIndexesEmpty` is a good idea. If we don't define it, then we can remove the definition of `Internal` (after moving `InflightBlocksReachLimit` into `Ignore` )

ok

> * rename `Ignore::BlockStored` -> `Ignore::AlreadyStored`

ok

u2:
@doitian merge?

PR #1212, opened 2019-07-10
feat: partition nonce for miners who use multi-threads

yangby-cryptape:
Avoid same nonces for different threads.

### DIFF
--- a/miner/src/worker/mod.rs
+++ b/miner/src/worker/mod.rs
@@ -10,6 +10,8 @@ use cuckoo_simple::CuckooSimple;
 use dummy::Dummy;
 use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use numext_fixed_hash::H256;
+use rand::{random, Rng};
+use std::ops::Range;
 use std::sync::Arc;
 use std::thread;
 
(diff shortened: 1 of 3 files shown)


miner/src/worker/mod.rs
```diff
@@ -41,6 +43,17 @@ impl WorkerController {
     }
 }
 
+pub(crate) fn partition_nonce(id: u64, total: u64) -> Range<u64> {
+    let range = u64::max_value() / total;
```
doitian: name `span` is better

miner/src/worker/mod.rs
```diff
@@ -41,6 +43,17 @@ impl WorkerController {
...
+    let start = range * id;
+    let end = match id {
+        x if x < total - 1 => range * (x + 1),
```
doitian:
```suggestion
        x if x < total - 1 => start + range,
```

miner/src/worker/mod.rs
```diff
@@ -103,5 +118,11 @@ pub fn start_worker(
 }
 
 pub trait Worker {
-    fn run(&mut self, progress_bar: ProgressBar);
+    fn run(&mut self, nonce_range: Range<u64>, progress_bar: ProgressBar);
```
doitian: Alternative, we can just pass a random generator into it.

```suggestion
    fn run<G: FnMut() -> u64>(&mut self, rng: G, progress_bar: ProgressBar);
```

miner/src/worker/mod.rs
```diff
@@ -41,6 +43,23 @@ impl WorkerController {
     }
 }
 
+fn partition_nonce(id: u64, total: u64) -> Range<u64> {
```
quake: integer division is round down, we can skip overflow checking:

```suggestion
fn partition_nonce(id: u64, total: u64) -> Range<u64> {
    debug_assert!(id < total);
    let span = u64::max_value() / total;
    let start = span * id;
    let end = start + span;
    (start..end)
}
```

yangby-cryptape: There still are few nonces in `end..u64::max_value()`.

Maybe the lucky nonce is one of them.

quake: Get, thanks.

btw, why set the parameter `id` and `total` type to u64? doesn't it should be usize?

yangby-cryptape: The output `nonce range` should be `u64`.

I think, both `u64` and `usize` are OK.
If I set `id` and `total` to `usize`, I can do cast at begin of the function `partition_nonce(..)`.

PR #1213, opened 2019-07-10
feat: Remove block_number from BMT

keroro520:
* Remove `block_number` from API `BlockMedianTimeContext::block_median_time`. 

  Originally we pass `block_number` as a parameter to help calculate the span

  ```
  span_start = min(0, block_number-customed_span)
  span = [span_start, block_number]
  ```

  In fact, it is unnecessary to calculate the span if we can know whether we reach the genesis during traveling the blocks. We can get it with the help of `timestamp_and_parent`!
  So I reduce the `block_number` from `block_median_time`, to make the API more simplified.


quake:

keroro520:
> Give some explanation in PR description that why removing it.

Done. I added an explanation in the PR description. Please help review again.

PR #1215, opened 2019-07-10
feat: alert notify script

jjyr:


util/network-alert/src/alert_relayer.rs
```diff
@@ -33,9 +33,9 @@ pub struct AlertRelayer {
 }
 
 impl AlertRelayer {
-    pub fn new(client_version: String, config: Config) -> Self {
+    pub fn new(client_version: String, notifier_config: NotifierConfig, config: Config) -> Self {
```
keroro520: Why not merge them together?  Put `client_version` and `notify_script` into config when constructs. And `AlertRelayer` and `Notifier` could share the same config.

ckb-bin/src/subcommand/run.rs
```diff
@@ -67,8 +67,13 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
 
     let relayer = Relayer::new(chain_controller.clone(), sync_shared_state);
     let net_timer = NetTimeProtocol::default();
-    let alert_config = args.config.alert.unwrap_or_default();
-    let alert_relayer = AlertRelayer::new(version.to_string(), alert_config);
+    let alert_signature_config = args.config.alert_signature.unwrap_or_default();
```
doitian: `serde(default)` can be used if we always want something.

util/network-alert/src/notifier.rs
```diff
@@ -60,13 +64,30 @@ impl Notifier {
...
         self.noticed_alerts
             .sort_by_key(|a| std::u32::MAX - a.priority);
     }
 
+    fn notify(&self, alert: &Alert) {
+        warn!("receive a new alert: {}", alert.message);
+        if let Some(notify_script) = self.config.notify_script.as_ref() {
+            match Command::new(notify_script)
+                .args(&[alert.message.to_owned()])
+                .status()
```
doitian: The child process may never exit.

doitian: Not sure this still works https://github.com/alexcrichton/wait-timeout

jjyr: I think we should make it simple, user can set a timeout in his own customized script(IMO it's unnecessary in 99% cases) and decide how to exit child process which better than we kill the script.

PR #1220, opened 2019-07-12
feat: Network flood control

TheWaWaR:
Do not send blocks to peer when session send buffer is full.

### DIFF
--- a/network/src/config.rs
+++ b/network/src/config.rs
@@ -1,5 +1,7 @@
-use crate::errors::{ConfigError, Error};
-use crate::PeerId;
+use crate::{
+    errors::{ConfigError, Error},
+    PeerId, DEFAULT_SEND_BUFFER,
+};
 use ckb_logger::info;
 use p2p::{
     multiaddr::{Multiaddr, Protocol},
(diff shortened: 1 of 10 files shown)


network/src/config.rs
```diff
@@ -36,6 +38,9 @@ pub struct NetworkConfig {
     pub upnp: bool,
     #[serde(default)]
     pub bootnode_mode: bool,
+    // Max send buffer size
+    #[serde(default)]
```
doitian: Not necessary, since Option is allowed omit in the toml.

network/src/protocols/mod.rs
```diff
@@ -58,6 +58,7 @@ pub trait CKBProtocolContext: Send {
     fn connected_peers(&self) -> Vec<PeerIndex>;
     fn report_peer(&self, peer_index: PeerIndex, behaviour: Behaviour);
     fn ban_peer(&self, peer_index: PeerIndex, duration: Duration);
+    fn pause_send(&self) -> bool;
```
doitian: It looks like a method which performs an action to pause sending. `send_paused` is a better name.

TheWaWaR: `pause_send` is from bitcoin, but `send_paused` seems more readable.

TheWaWaR:
@doitian 
Can we merge this now?

PR #1221, opened 2019-07-12
test: Compare block template

keroro520:
https://github.com/nervosnetwork/ckb/pull/1214 resets `current_time` of block template. We should adjust the compare function.


quake:

PR #1227, opened 2019-07-14
fix: should check tx from pool when the short_id set is not empty

lerencao:
I believe it's a typo.


quake:
@lerencao thanks for your PR.
@keroro520 is this bug introduced by https://github.com/nervosnetwork/ckb/pull/614 ? Please help to review and add an integration test for this.

PR #1228, opened 2019-07-15
[ᚬrc/v0.16] fix(relayer): should check tx from tx_pool when the short_id set is not empty

doitian:
Backport #1227


doitian:

PR #1232, opened 2019-07-15
feat: IBD with whitelist

driftluo:
1. Enable IBD with whitelist peers
2. Rename reserved to whitelist

Breaking changes:

In ckb.toml:

```diff
-reserved_peers = []
-reserved_only = false
+whitelist_peers = []
+whitelist_only = false
```

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -70,6 +70,7 @@ pub struct ChainSyncState {
     pub sent_getheaders: bool,
     pub not_sync_until: Option<u64>,
     pub protect: bool,
+    pub is_reserved: bool
 }
 
 impl Default for ChainSyncState {
@@ -81,6 +82,7 @@ impl Default for ChainSyncState {
             sent_getheaders: false,
             not_sync_until: None,
             protect: false,
+            is_reserved: false
         }
     }
 }
@@ -122,8 +124,8 @@ impl PeerState {
     }
 
     pub fn can_sync(&self, now: u64, ibd: bool) -> bool {
-        // only sync with outbound peer in IBD
-        (self.is_outbound || !ibd)
+        // only sync with outbound/reserved peer in IBD
+        ((self.is_outbound || self.chain_sync.is_reserved) || !ibd)
             && !self.sync_started
             && self
                 .chain_sync
@@ -396,7 +398,7 @@ impl Peers {
         peer: PeerIndex,
         headers_sync_timeout: Option<u64>,
         protect: bool,
-        is_outbound: bool,
+        (is_outbound, is_reserved): (bool, bool),
     ) {
         self.state
             .write()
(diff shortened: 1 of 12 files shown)


sync/src/types.rs
```diff
@@ -396,18 +398,20 @@ impl Peers {
         peer: PeerIndex,
         headers_sync_timeout: Option<u64>,
         protect: bool,
-        is_outbound: bool,
+        (is_outbound, is_whitelist): (bool, bool),
```
jjyr: Why tuple? flat args is better.

jjyr: Why tuple? flat args is better.

driftluo: Did not see the obvious benefits and disadvantages, where is better?

quake:
I have a refactoring proposal for this PR, move `is_outbound`, `is_whitelist` and `protect_outbound` to a new `PeerFlags` struct, and change `PeerState` to:
```
  pub struct PeerState {
    peer_flags: PeerFlags
     ...
  }
```

The `Peers#on_connected` can be refactored to:
```
    pub fn on_connected(
        &self,
        peer: PeerIndex,
        peer_flags: PeerFlags
    ) {
```

btw, headers_sync_timeout is always `None` here, right?

driftluo:
> btw, headers_sync_timeout is always `None` here, right?

lots of unitest dependence `headers_sync_timeout` set on here, I really don't like these test code that invades business logic

sync/src/types.rs
```diff
@@ -122,8 +124,8 @@ impl PeerState {
     }
 
     pub fn can_sync(&self, now: u64, ibd: bool) -> bool {
-        // only sync with outbound peer in IBD
-        (self.is_outbound || !ibd)
+        // only sync with outbound/whitelist peer in IBD
+        ((self.is_outbound || self.chain_sync.is_whitelist) || !ibd)
```
keroro520: redundant brackets

driftluo: I think this bracket can help understand

sync/src/types.rs
driftluo: I don't think there is any essential difference between the two in terms of definition.

jjyr: We can randomly put two parameters into a tuple and there not any "essential difference", and the program still works, but why we do this? 

Why don't u put other parameters into a tuple but these two? Does it imply the inner link of these two? If it is what's the hidden reason? And if it is not, why don't we put other parameters into tuple `(protect, is_outbound, is_whitelist)`?

Unreasonable change is a disadvantage.

driftluo: Why is `protect` not put together, because protect is the self-judgment of the protocol layer, and the other two are the state of `nc`. 

I am going to write it according to the @quake  approach.

driftluo:
> Please add relative tests.
I don't think there is a need for this change to be tested separately.

sync/src/types.rs
```diff
@@ -80,17 +79,23 @@ impl Default for ChainSyncState {
...
             not_sync_until: None,
-            protect: false,
         }
     }
 }
 
+#[derive(Clone, Default, Debug, Copy)]
+pub struct PeerFlags {
+    pub is_outbound: bool,
+    pub protect: bool,
```
doitian:
```suggestion
    pub is_protected: bool,
```

sync/src/types.rs
```diff
@@ -50,7 +50,7 @@ const ORPHAN_BLOCK_SIZE: usize = 1024;
 
 // State used to enforce CHAIN_SYNC_TIMEOUT
 // Only in effect for outbound, non-manual connections, with
-// m_protect == false
+// m_protect == false and not whitelist
```
doitian: Suggest changes:

```
// Only in effect for connections that are outbound, non-manual, non-protected and non-whitelist.
```

PR #1235, opened 2019-07-15
chore: optional start some network service

quake:
optional start `OutboundPeerService` and `DnsSeedingService`, and fix random failure integration test `BlockSyncDuplicatedAndReconnect `

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -815,21 +815,24 @@ impl NetworkService {
             p2p_service.control().to_owned(),
             ping_receiver,
         );
-        let outbound_peer_service = OutboundPeerService::new(
-            Arc::clone(&network_state),
-            p2p_service.control().to_owned(),
-            Duration::from_secs(config.connect_outbound_interval_secs),
-        );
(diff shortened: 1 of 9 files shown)


quake:

network/src/network.rs
```diff
@@ -815,21 +815,24 @@ impl NetworkService {
...
-            network_state.config.dns_seeds.clone(),
-        );
-        let bg_services = vec![
+        let mut bg_services = vec![
             Box::new(ping_service.for_each(|_| Ok(()))) as Box<_>,
             Box::new(disc_service) as Box<_>,
-            Box::new(outbound_peer_service) as Box<_>,
-            Box::new(dns_seeding_service) as Box<_>,
         ];
+        if config.connect_outbound_interval_secs > 0 {
```
keroro520: We should add documentation or some comments about `connect_outbound_interval_secs == 0`

keroro520: Or wrap a function `config.enable_auto_connect_outbound() -> bool;`

quake: Updated, please review again, thanks.

PR #1236, opened 2019-07-15
feature: add BlockTransactions verifier

u2:


sync/src/relayer/block_transactions_verifier.rs
```diff
@@ -0,0 +1,41 @@
...
+        transactions: &[Transaction],
+    ) -> Result<(), Error> {
+        let indexes: FnvHashSet<u32> = FnvHashSet::from_iter(indexes.iter().cloned());
+
+        let short_ids: Vec<&ShortTransactionID> = block
+            .short_ids
+            .iter()
+            .enumerate()
+            .filter(|(i, _)| indexes.contains(&(*i as u32)))
+            .map(|(_, x)| x)
```
zhangsoledad: `filter_map`

doitian:
What about the window when the node restarted and lose all the maps and the peer send the get block transactions request after restarting.

doitian:
@nervos-bot dummy ci ee4ef60402a1073a8c6e056f95a0d06932ecb9a0

u2:
> What about the window when the node restarts and loses all the maps and then the peer sends the get block transactions request?

I don't get your question. The other peer sends the `block transaction` message after the node restart?

https://github.com/nervosnetwork/ckb/pull/1236/files#diff-b9f599c722bd1ef9c78d1cd7c2334409L39

It first tries to find the pending_compact_blocks. If it loses all the maps, it will fail at this step.

https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/compact_block_process.rs#L237

The peer sends `get_block_transactions` message only when it receives `compact block`. So if the node could find the matched pending_compact_block, it should always have the `pending_get_block_transaction`.

sync/src/relayer/compact_block_process.rs
```diff
@@ -234,21 +237,18 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
         if let Err(err) = self
             .nc
             .send_message_to(self.peer, fbb.finished_data().into())
         {
             ckb_logger::debug!("relayer send get_block_transactions error: {:?}", err);
         }
+
+        let mut pending_get_block_transactions =
+            self.relayer.shared().pending_get_block_transactions();
+        pending_get_block_transactions.insert((block_hash, self.peer), indexes);
```
keroro520: As for a CompactBlock with fresh transactions, here we record the CompactBlock inside `pending_compact_blocks`, and record the fresh transaction-ids inside `pending_get_block_transactions`. I think it is redundant and hard to keep consistency between 2 caches. 

So, 1. does it neccesary to cache `pending_get_block_transactions`? 2. If it is neccesary, can we merge `pending_get_block_transactions` and `pending_compact_blocks` into one?

u2: Thanks, updated, pls review again.

sync/src/relayer/block_transactions_process.rs
```diff
@@ -61,9 +65,10 @@ impl<'a, CS: ChainStore + 'static> BlockTransactionsProcess<'a, CS> {
                     self.relayer
                         .accept_block(self.nc.as_ref(), self.peer, block);
                 }
+                return Ok(());
             }
         }
 
-        Ok(())
+        Err(Error::Misbehavior(Misbehavior::UnexpectedBlockTransactions).into())
```
keroro520: 1. A send CompactBlock to B
2. B request for the corresponding GetBlockTransactions from A
3. B restart and lose its records.
4. A response B BlockTransactions. But A ban B unexpectly.

Am I getting it right?

sync/src/relayer/block_transactions_verifier.rs
```diff
@@ -0,0 +1,46 @@
...
+            .short_ids
+            .iter()
+            .enumerate()
+            .filter_map(|(i, x)| {
+                if indexes.contains(&(i as u32)) {
+                    Some(x)
+                } else {
+                    None
+                }
+            })
```
keroro520: 1. `indexes` -> `indices`
2. `let short_ids = indices.iter()....` is better.
3. `BlockTransactionsInvalid` -> `InvalidBlockTransactions`

sync/src/relayer/compact_block_process.rs
```diff
@@ -14,7 +14,7 @@ use ckb_traits::{BlockMedianTimeContext, ChainProvider};
 use ckb_verification::{HeaderResolver, HeaderVerifier, Verifier};
 use failure::Error as FailureError;
 use flatbuffers::FlatBufferBuilder;
-use fnv::FnvHashSet;
+use fnv::FnvHashMap;
```
keroro520: Use `HashMap`?

sync/src/relayer/compact_block_process.rs
```diff
@@ -177,21 +176,28 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
+
+                    // The missing_indexes should never be empty!
+                    if missing_indexes.is_empty() {
+                        debug_target!(
+                            crate::LOG_TARGET_RELAY,
+                            "Missing indexes should never be empty peer: {}, block: {}",
+                            self.peer,
+                            &block_hash,
+                        );
+                        return Ok(Status::MissingIndexesEmpty);
```
keroro520: `assert!(!missing_indexes.is_empty())` is better.

Makefile
```diff
@@ -90,7 +90,7 @@ security-audit: ## Use cargo-audit to audit Cargo.lock for crates with security
 ##@ Continuous Integration
 
 ci: ## Run recipes for CI.
-ci: check-cargotoml fmt check-dirty-doc clippy security-audit test
+ci: check-cargotoml fmt check-dirty-doc clippy test
```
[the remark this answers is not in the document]
u2: It was a mistake.

sync/src/relayer/block_transactions_process.rs
u2: Yes. 
Before there were two different caches, so it was ok.

sync/src/relayer/compact_block_process.rs
u2: Yes, we have talked about it before.

sync/src/relayer/compact_block_process.rs
u2: no

sync/src/relayer/block_transactions_verifier.rs
u2: 1. https://grammarist.com/usage/indexes-indices/
2. sorry, I did not get it.
3. updated

sync/src/relayer/block_transactions_process.rs
keroro520: I am not sure about it...

u2: anyway fixed it already.

sync/src/relayer/block_transactions_verifier.rs
keroro520:
```
        let short_ids: Vec<&ShortTransactionID> = indexes
            .iter()
            .filter_map(|index| {
                block.short_ids.get(*index as usize)
            })
            .collect();
```

doitian:
integration test `compact_block_relay_parent_of_orphan_block` failed

u2:
> integration test `compact_block_relay_parent_of_orphan_block` failed

Yes, already try to find the reason.

u2:
@keroro520, @quake There were some mistakes in `BlockTransactionVerifier` before. Rewrite it now,
please review. Thanks.

PR #1237, opened 2019-07-15
feat: cache BLOCK_INVALID/BLOCK_VALID status

keroro520:
* feat: Add `BLOCK_VALID` status.
* test: Add test to check the block status after switching forks

### DIFF
--- a/sync/src/block_status.rs
+++ b/sync/src/block_status.rs
@@ -4,12 +4,12 @@ bitflags! {
     pub struct BlockStatus: u32 {
         const UNKNOWN                 =     0;
 
-        const HEADER_VALID            =     0b0000_0000_0000_0001;
-        const BLOCK_RECEIVED          =     Self::HEADER_VALID.bits | 0b0000_0000_0000_0010;
-        const BLOCK_STORED            =     Self::HEADER_VALID.bits | Self::BLOCK_RECEIVED.bits | 0b0000_0000_0000_1000;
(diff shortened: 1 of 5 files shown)


sync/src/types.rs
```diff
@@ -145,7 +145,7 @@ impl PeerState {
     }
 
     // Not use yet
-    pub fn caught_up_sync(&mut self) {
+    pub fn caught_nup_sync(&mut self) {
```
u2: Typo?

keroro520: Thanks!

sync/src/block_status.rs
```diff
@@ -7,6 +7,7 @@ bitflags! {
         const HEADER_VALID            =     0b0000_0000_0000_0001;
         const BLOCK_RECEIVED          =     Self::HEADER_VALID.bits | 0b0000_0000_0000_0010;
         const BLOCK_STORED            =     Self::HEADER_VALID.bits | Self::BLOCK_RECEIVED.bits | 0b0000_0000_0000_1000;
+        const BLOCK_VALID             =     Self::HEADER_VALID.bits | Self::BLOCK_RECEIVED.bits | Self::BLOCK_STORED.bits | 0b0000_0000_0001_0000;
```
TheWaWaR: `(1 << 4)` instead of `0b0000_0000_0001_0000` would be more readable?

keroro520: Thanks! Updated as you suggested.

And I add another [commit](https://github.com/nervosnetwork/ckb/pull/1237/commits/b4a37519169a1b480bf527ccbf1fd0d9a05e7e4b), PTAL.

PR #1246, opened 2019-07-17
feat: Send a message to remote peer when disconnect

TheWaWaR:
* Add a new p2p protocol `StringMessageProtocol` for send simple string message
* Send a message to remote peer when disconnect

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -165,7 +165,7 @@ impl NetworkState {
         {
             info!("peer {:?} banned", peer_id);
             if let Some(session_id) = self.peer_registry.read().get_key_by_peer_id(peer_id) {
-                if let Err(err) = disconnect(p2p_control, session_id, Some("banned")) {
+                if let Err(err) = disconnect_with_message(p2p_control, session_id, "banned") {
                     debug!("Disconnect failed {:?}, error: {:?}", session_id, err);
                 }
             }
@@ -198,7 +198,9 @@ impl NetworkState {
         if let Some(peer) = peer_opt {
             self.peer_store.lock().ban_addr(&peer.address, duration);
             let message = format!("Ban for {} seconds", duration.as_secs());
-            if let Err(err) = disconnect(p2p_control, peer.session_id, Some(message.as_str())) {
+            if let Err(err) =
+                disconnect_with_message(p2p_control, peer.session_id, message.as_str())
+            {
                 debug!("Disconnect failed {:?}, error: {:?}", peer.session_id, err);
             }
         }
@@ -472,9 +474,11 @@ impl EventHandler {
                     .enumerate()
                 {
                     if index & 0x1 != 0 {
-                        if let Err(err) =
-                            disconnect(context.control(), peer, Some("bootnode random eviction"))
-                        {
+                        if let Err(err) = disconnect_with_message(
+                            context.control(),
+                            peer,
+                            "bootnode random eviction",
+                        ) {
                             debug!("Inbound eviction failed {:?}, error: {:?}", peer, err);
                             return;
                         }
@@ -516,7 +520,7 @@ impl ServiceHandle for EventHandler {
                 error,
             } => {
                 debug!("ProtocolError({}, {}) {}", id, proto_id, error);
-                if let Err(err) = disconnect(context.control(), id, Some("ProtocolError")) {
+                if let Err(err) = disconnect_with_message(context.control(), id, "ProtocolError") {
                     debug!("Disconnect failed {:?}, error {:?}", id, err);
                 }
             }
@@ -594,10 +598,10 @@ impl ServiceHandle for EventHandler {
                                 "evict peer (disonnect it), {} => {}",
                                 evicted_peer.session_id, evicted_peer.address,
                             );
-                            if let Err(err) = disconnect(
+                            if let Err(err) = disconnect_with_message(
                                 context.control(),
                                 evicted_peer.session_id,
-                                Some("accept peer eviction"),
+                                "accept peer eviction",
                             ) {
                                 debug!(
                                     "Disconnect failed {:?}, error: {:?}",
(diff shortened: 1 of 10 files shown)


u2:
What about declaring some global constants?

TheWaWaR:
> What about declaring some global constants for the messages.

I think it's not necessary, it's just for human read (debug purpose).

network/src/network.rs
```diff
@@ -1104,3 +1130,17 @@ impl Drop for NetworkController {
         self.stop.try_send();
     }
 }
+
+// Send a optional message before disconnect a peer
+pub(crate) fn disconnect(
```
quake: Suggest to move this fn to `ServiceControl` and rename it to `disconnect_with_message(&self, peer_index: SessionId, message: &str)`

network/src/protocols/disconnect_message.rs
```diff
@@ -0,0 +1,45 @@
...
+pub(crate) struct DisconnectMessageProtocol;
+
+impl ServiceProtocol for DisconnectMessageProtocol {
+    fn init(&mut self, _context: &mut ProtocolContext) {}
+
+    fn received(&mut self, context: ProtocolContextMutRef, data: bytes::Bytes) {
+        let session_id = context.session.id;
+        if let Ok(message) = String::from_utf8(data.to_vec()) {
+            debug!(
+                "Received string message from peer={}: {}",
```
quake:
```suggestion
                "Received disconnect message from peer={}: {}",
```

network/src/network.rs
TheWaWaR: For up level (ckb) maintainability `ServiceControl::disconnect_with_message`. Based on that idea, there are to solutions:
1. Setup a p2p internal protocol for receiving the message (Which is ugly)
2. p2p need a new basic mechanism to receive the message (In current case, I don't it's worth it)

How about this compromised way? :
1. `DisconnectMessageProtocol` still required
2. `ServiceControl::disconnect_with_message(&self, peer_index: SessionId, proto_id: ProtocolId, message: &str)`

network/src/network.rs
```diff
@@ -511,7 +516,7 @@ impl ServiceHandle for EventHandler {
                 error,
             } => {
                 debug!("ProtocolError({}, {}) {}", id, proto_id, error);
-                if let Err(err) = context.disconnect(id) {
+                if let Err(err) = disconnect(context.control(), id, Some("ProtocolError")) {
```
jjyr:
```suggestion
                if let Err(err) = disconnect(context.control(), id, Some(format!("protocol error {}", proto_id))) {
```

jjyr:
```suggestion
                if let Err(err) = disconnect(context.control(), id, Some(format!("protocol error {}", proto_id))) {
```

network/src/network.rs
```diff
@@ -605,7 +614,11 @@ impl ServiceHandle for EventHandler {
                                 "registry peer failed {:?} disconnect it, {} => {}",
                                 err, session_context.id, session_context.address,
                             );
-                            if let Err(err) = context.disconnect(session_context.id) {
+                            if let Err(err) = disconnect(
+                                context.control(),
+                                session_context.id,
+                                Some("accept peer error"),
```
jjyr:
```suggestion
                                Some("reject peer connection"),
```

network/src/network.rs
```diff
@@ -589,7 +594,11 @@ impl ServiceHandle for EventHandler {
                                 "evict peer (disonnect it), {} => {}",
                                 evicted_peer.session_id, evicted_peer.address,
                             );
-                            if let Err(err) = context.disconnect(evicted_peer.session_id) {
+                            if let Err(err) = disconnect(
+                                context.control(),
+                                evicted_peer.session_id,
+                                Some("accept peer eviction"),
```
jjyr:
```suggestion
                                Some("evict because accept new peer"),
```

network/src/network.rs
```diff
@@ -713,7 +726,11 @@ impl ServiceHandle for EventHandler {
                         .disconnecting_sessions
                         .write()
                         .insert(session_id);
-                    if let Err(err) = context.disconnect(session_id) {
+                    if let Err(err) = disconnect(
+                        context.control(),
+                        session_id,
+                        Some("already removed from registry"),
```
jjyr:
```suggestion
                        Some("peer already disconnect")
```

network/src/network.rs
```diff
@@ -992,7 +1018,7 @@ impl NetworkController {
             .read()
             .get_key_by_peer_id(peer_id)
         {
-            if let Err(err) = self.p2p_control.disconnect(session_id) {
+            if let Err(err) = disconnect(&self.p2p_control, session_id, Some("remove node")) {
```
jjyr:
```suggestion
            if let Err(err) = disconnect(&self.p2p_control, session_id, Some("disconnect manually")) {
```

sync/src/synchronizer/headers_process.rs
```diff
@@ -284,7 +284,10 @@ where
             && (is_outbound && !is_protected)
         {
             debug!("Disconnect peer({}) is unprotected outbound", self.peer);
-            if let Err(err) = self.nc.disconnect(self.peer) {
+            if let Err(err) = self.nc.disconnect(
+                self.peer,
+                Some("insufficient work unprotected outbound peer"),
```
jjyr:
```suggestion
                Some("useless outbound peer in IBD"),
```

network/src/network.rs
TheWaWaR: For debug purpose `removed from registry` is better.

network/src/protocols/mod.rs
```diff
@@ -337,8 +338,8 @@ impl CKBProtocolContext for DefaultCKBProtocolContext {
             .filter_broadcast(target, self.proto_id, data)?;
         Ok(())
     }
-    fn disconnect(&self, peer_index: PeerIndex) -> Result<(), Error> {
-        self.p2p_control.disconnect(peer_index)?;
+    fn disconnect(&self, peer_index: PeerIndex, message: Option<&str>) -> Result<(), Error> {
```
keroro520: It seems unnecessary to wrap an `Option` around the message.

keroro520: Yes, use empty `""` directly. It is the same as `unwrap_or_default("")`.

TheWaWaR: > Yes, use empty `""` directly. It is the same as `unwrap_or_default("")`.

OK for me. 
@jjyr 
What do you think?

jjyr: IMO we should always return a reason, but for now, it's OK to return an empty string.

PR #1248, opened 2019-07-17
chore: upgrade rustc to 1.36.0

driftluo:
Upgrade rustc to 1.36.0


doitian:
docker builder images upgraded

driftluo:
wait #1250

doitian:
Can we also remove the dependency hashdown?

driftluo:
> Can we also remove the dependency hashdown?

We can remove the `fnv` and `hashdown` dependencies, but this should not be made into one pr

u2:
> > Can we also remove the dependency hashdown?
> 
> We can remove the `fnv` and `hashdown` dependencies, but this should not be made into one pr

Agree to remove them both, fast enough.

PR #1249, opened 2019-07-17
refactor: apply new serialization to all crates

yangby-cryptape:
**READY for Review.**


TheWaWaR:
https://github.com/nervosnetwork/ckb/blob/1266224e82e602fd2a7f5ec293f0fc85f4a9e057/sync/src/relayer/mod.rs#L106-L108

https://github.com/nervosnetwork/ckb/blob/1266224e82e602fd2a7f5ec293f0fc85f4a9e057/sync/src/relayer/mod.rs#L404-L415

For backward compatibility requirement, when got an unexpected Union Item we should just ignore it.

yangby-cryptape:
Reply @doitian:

- [x] `Please add a section in `docs/ckb-core-dev.md` about how to install moleculec ...`
- [x] `Please add a comment at the beginning of the file, such as: ...`
- [x] `This rule is broken if someone add a new file into GEN_FILES. ...`
- [x] `Prefer fixed types before dynamic types.`
- [ ] `This field is 0 in most cases, should we use Uint64Opt here?`
      How about move this changes in another PR, after merge this?
- [ ] Upgrade `ckb-system-script`.

Spent three days, I have pass all tests in this repo in my local environment.
Just waiting for the PR in `ckb-system-script` repo, I will push the last commit.

Makefile
```diff
@@ -139,7 +133,7 @@ help:  ## Display help message.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 .PHONY: build prod prod-test prod-docker docker docker-publish
-.PHONY: gen gen-clean clean-node-files check-cfbc-version
+.PHONY: gen clean-node-files check-cfbc-version
```
u2: rm `check-cfbc-version`?

yangby-cryptape: yes, i will

indexer/src/store.rs
```diff
@@ -1056,199 +1183,247 @@ mod tests {
 
         store.sync_index_states();
 
-        let transactions = store.get_transactions(&script1.hash(), 0, 100, false);
+        println!("script1.lock_hash: {:x}", script1.calc_script_hash());
```
u2: `println!`

miner/src/block_assembler.rs
```diff
@@ -200,21 +203,19 @@ impl BlockAssembler {
     }
 
     fn transform_uncle(uncle: UncleBlock) -> UncleTemplate {
-        let UncleBlock { header, proposals } = uncle;
-
         UncleTemplate {
-            hash: header.hash().to_owned(),
+            hash: uncle.calc_header_hash(),
```
u2: The old implementation header.hash().to_owned()` is a cached hash, now need to calculate the hash every time?

doitian:
> * How about move this changes in another PR, after merge this?

OK with me

util/types/src/core/views.rs
```diff
@@ -0,0 +1,662 @@
...
+    define_block_unpacked_inner_getter!(number, BlockNumber);
+    define_block_unpacked_inner_getter!(difficulty, U256);
+    define_block_unpacked_inner_getter!(timestamp, u64);
+    define_block_unpacked_inner_getter!(parent_hash, H256);
+    define_block_unpacked_inner_getter!(transactions_root, H256);
+    define_block_unpacked_inner_getter!(witnesses_root, H256);
+    define_block_unpacked_inner_getter!(proposals_hash, H256);
+    define_block_unpacked_inner_getter!(uncles_hash, H256);
+    define_block_unpacked_inner_getter!(uncles_count, u32);
+    define_block_unpacked_inner_getter!(epoch, EpochNumber);
```
yangby-cryptape: @quake

Only `HeaderView`, `TransactionView` and `BlockView` have direct methods to access the inner data.
(And those data was unpacked, since those data were always required `unpack` version when used, if u want a packed version, I think u have to write like `block.data().xx().xx().xx()`).

When we remove `H256`, and only keep `Byte32`, most of these methods will return packed version data, too.

IMHO, it's unnecessarily for other structs. Too many methods is troublesome, too.

For example, a lot of `new` methods, no one can understand them, unless them jump to the definitions.
And we even have a `new` method has 8 inputs, it's not easy to write them in a right order, I think `builder` is better.

yangby-cryptape: I use macros because I was really tired to distinguish functions of those methods.

FOr example: a struct has eight fields, there are eight methods named as the field names: seven of them is `getter`, and only one has logical process and that one method return a processed data, not the original data. 
That sucks!

util/types/src/extension/shortcuts.rs
```diff
@@ -0,0 +1,226 @@
+use std::collections::HashSet;
```
yangby-cryptape: @quake I put some extra logic for packed data type here.

sync/src/relayer/compact_block_verifier.rs
```diff
@@ -16,34 +15,43 @@ impl CompactBlockVerifier {
...
         // Check indices order of prefilled transactions
-        let unordered = prefilled_transactions
-            .as_slice()
-            .windows(2)
-            .any(|pt| pt[0].index >= pt[1].index);
-        if unordered {
-            return Err(Error::Misbehavior(
-                Misbehavior::UnorderedPrefilledTransactions,
-            ));
+        for i in 0..(prefilled_transactions.len() - 1) {
```
u2: use `window` method as before

store/src/transaction.rs
```diff
@@ -79,135 +79,138 @@ impl StoreTransaction {
...
+        self.insert_raw(COLUMN_BLOCK_UNCLE, hash.as_slice(), uncles.as_slice())?;
+        self.insert_raw(
+            COLUMN_BLOCK_PROPOSAL_IDS,
+            hash.as_slice(),
+            proposals.as_slice(),
+        )?;
+        for (index, tx) in block.transactions().into_iter().enumerate() {
+            let key = packed::TransactionKey::new_builder()
+                .block_hash(hash.clone())
+                .index(index.pack())
```
yangby-cryptape: This `index` was defined as `BeUint32` in schema file, and I implement conversions between `u32.to_be_bytes()` and `BeUint32`. And the `TransactionKey` was defined as `struct` in schema file, so the `key.as_slice()` can use for prefix-search.

sync/src/synchronizer/mod.rs
```diff
@@ -827,13 +833,13 @@ mod tests {
...
-        let mut blocks: Vec<Block> = Vec::new();
+        let mut blocks: Vec<BlockView> = Vec::new();
         let mut parent = shared
             .store()
             .get_block_header(&shared.store().get_block_hash(0).unwrap())
             .unwrap();
         for i in 1..=block_number {
-            let parent_epoch = shared.get_block_epoch(&parent.hash()).unwrap();
+            let parent_epoch = shared.store().get_block_epoch(&parent.hash()).unwrap();
             let epoch = shared
```
u2: Please use `shared.get_block_epoch()`, there is a bit of nuance.

sync/src/synchronizer/mod.rs
```diff
@@ -789,13 +795,13 @@ mod tests {
...
         let peer: PeerIndex = 0.into();
 
-        let mut blocks: Vec<Block> = Vec::new();
+        let mut blocks: Vec<BlockView> = Vec::new();
         let mut parent = shared1
             .store()
             .get_block_header(&shared1.store().get_block_hash(0).unwrap())
             .unwrap();
         for i in 1..block_number {
-            let parent_epoch = shared1.get_block_epoch(&parent.hash()).unwrap();
```
u2: Please use `shared.get_block_epoch()`, there is a bit of nuance.

sync/src/synchronizer/mod.rs
```diff
@@ -707,7 +713,7 @@ mod tests {
         parent = blocks[150].header().to_owned();
         let fork = parent.number();
         for i in 1..=block_number {
-            let parent_epoch = shared2.get_block_epoch(&parent.hash()).unwrap();
+            let parent_epoch = shared2.store().get_block_epoch(&parent.hash()).unwrap();
```
u2: Please use `shared.get_block_epoch()`, there is a bit of nuance.

sync/src/synchronizer/mod.rs
```diff
@@ -685,10 +691,10 @@ mod tests {
         let (chain_controller2, shared2, _notify2) = start_chain(Some(consensus.clone()), None);
         let block_number = 200;
 
-        let mut blocks: Vec<Block> = Vec::new();
+        let mut blocks: Vec<BlockView> = Vec::new();
         let mut parent = consensus.genesis_block().header().to_owned();
         for i in 1..block_number {
-            let parent_epoch = shared1.get_block_epoch(&parent.hash()).unwrap();
+            let parent_epoch = shared1.store().get_block_epoch(&parent.hash()).unwrap();
```
u2: Please use `shared.get_block_epoch()`, there is a bit of nuance.

sync/src/synchronizer/mod.rs
```diff
@@ -592,7 +598,7 @@ mod tests {
             .store()
             .get_block_header(&shared.store().get_block_hash(number - 1).unwrap())
             .unwrap();
-        let parent_epoch = shared.get_block_epoch(&parent.hash()).unwrap();
+        let parent_epoch = shared.store().get_block_epoch(&parent.hash()).unwrap();
         let epoch = shared
```
u2: Please use `shared.get_block_epoch()`, there is a bit of nuance.

sync/src/tests/synchronizer.rs
```diff
@@ -110,50 +113,46 @@ fn setup_node(thread_name: &str, height: u64) -> (TestNode, Shared) {
         let number = block.header().number() + 1;
         let timestamp = block.header().timestamp() + 1;
 
-        let last_epoch = shared.get_block_epoch(&block.header().hash()).unwrap();
+        let last_epoch = shared
+            .store()
+            .get_block_epoch(&block.header().hash())
```
u2: `shared.get_block_epoch`

sync/src/tests/util.rs
```diff
@@ -45,44 +46,38 @@ pub fn generate_blocks(
     }
 }
 
-pub fn inherit_block(shared: &Shared, parent_hash: &H256) -> BlockBuilder {
+pub fn inherit_block(shared: &Shared, parent_hash: &Byte32) -> BlockBuilder {
     let parent = shared.store().get_block(parent_hash).unwrap();
-    let parent_epoch = shared.get_block_epoch(parent_hash).unwrap();
+    let parent_epoch = shared.store().get_block_epoch(parent_hash).unwrap();
```
u2: `shared.get_block_epoch`

sync/src/synchronizer/mod.rs
u2: https://github.com/nervosnetwork/ckb/pull/1424

sync/src/tests/util.rs
zhangsoledad: don't use `ChainProvider::get_block_epoch`, it will be deteted later

sync/src/tests/synchronizer.rs
zhangsoledad: don't use ChainProvider::get_block_epoch, it will be deleted later

sync/src/synchronizer/mod.rs
zhangsoledad: don't use ChainProvider::get_block_epoch, it will be deleted later

sync/src/synchronizer/mod.rs
zhangsoledad: don't use ChainProvider::get_block_epoch, it will be deleted later

sync/src/synchronizer/mod.rs
zhangsoledad: don't use ChainProvider::get_block_epoch, it will be deleted later

doitian:
bors r-

CI failed on Windows

PR #1250, opened 2019-07-18
test: Fix test

driftluo:
1. Fix feeler protocol test random behavior, sleep can't eliminate randomness
2. Comparison with HashSet

### DIFF
--- a/sync/src/tests/inflight_blocks.rs
+++ b/sync/src/tests/inflight_blocks.rs
@@ -1,6 +1,8 @@
 use crate::types::InflightBlocks;
 use crate::BLOCK_DOWNLOAD_TIMEOUT;
+use fnv::FnvHashSet;
 use numext_fixed_hash::{h256, H256};
+use std::iter::FromIterator;
 
 #[test]
 fn inflight_blocks_count() {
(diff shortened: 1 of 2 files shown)


sync/src/tests/inflight_blocks.rs
```diff
@@ -20,9 +21,11 @@ fn inflight_blocks_count() {
...
     assert_eq!(inflight_blocks.peer_inflight_count(2.into()), 1); // one block inflight
     assert_eq!(
-        inflight_blocks
-            .inflight_block_by_peer(1.into())
-            .map(|set| set.iter().collect()),
+        inflight_blocks.inflight_block_by_peer(1.into()).map(|set| {
+            let mut res = set.iter().collect::<Vec<&H256>>();
+            res.sort();
+            res
+        }),
```
u2: I think to compare the `HashSet` is more reasonable.

driftluo: `vec + sort` or `HashSet` both ok. But `HashSet` needs more code, so ...

u2: reasonable and less code

```rust
    use fnv::FnvHashSet;
    use std::iter::FromIterator;

    assert_eq!(
        inflight_blocks
            .inflight_state_by_block(&h256!("0x4"))
            .map(|state| state.peers.clone()),
        Some(FnvHashSet::from_iter(vec![(4.into()), (1.into())]))
    );
```

PR #1251, opened 2019-07-18
fix: there is an incorrect deserialization in indexer

yangby-cryptape:
### Bug

- Insert a `CellOutput`
  https://github.com/nervosnetwork/ckb/blob/bd7865a76e5b53608f698a5fbeafc1f797ceb69d/indexer/src/store.rs#L548-L555

- Select as `Option<CellOutput>`
  https://github.com/nervosnetwork/ckb/blob/bd7865a76e5b53608f698a5fbeafc1f797ceb69d/indexer/src/store.rs#L516-L522
  https://github.com/nervosnetwork/ckb/blob/bd7865a76e5b53608f698a5fbeafc1f797ceb69d/indexer/src/types.rs#L36-L41

### DIFF
--- a/indexer/src/store.rs
+++ b/indexer/src/store.rs
@@ -516,10 +516,10 @@ impl IndexerStoreBatch {
                 db.read(COLUMN_LOCK_HASH_LIVE_CELL, &lock_hash_index.to_vec())
                     .expect("indexer db read should be ok")
                     .map(|value| deserialize(&value).expect("deserialize CellOutput should be ok"))
-                    .map(|cell_output| LockHashCellOutput {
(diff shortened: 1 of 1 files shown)


quake:
Could you please add unit test for this PR? You can refer to `consume_txs_in_same_block`, create a new unit test `consume_txs_in_different_block`.

yangby-cryptape:
> Could you please add unit test for this PR? You can refer to `consume_txs_in_same_block`, create a new unit test `consume_txs_in_different_block`.

- If I wrote a test, I have to remove it in my next PR (use our own serialization).
- Here will be a compile-time error, after I finished the serialization.

PR #1252, opened 2019-07-18
feat: uncle descendant limit

zhangsoledad:
# break changes
Append uncle verification rule:  A block B1 is considered to be the uncle of another block B2, B1's parent is either B2's ancestor or embedded in B2 or its ancestors as an uncle

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -1,6 +1,6 @@
+use crate::candidate_uncles::CandidateUncles;
 use crate::config::BlockAssemblerConfig;
 use crate::error::Error;
-use ckb_core::block::Block;
 use ckb_core::cell::{resolve_transaction, OverlayCellProvider, TransactionsProvider};
 use ckb_core::extras::EpochExt;
 use ckb_core::header::Header;
@@ -34,7 +34,6 @@ use std::cmp;
 use std::sync::{atomic::AtomicU64, atomic::AtomicUsize, atomic::Ordering, Arc};
 use std::thread;
 
-const MAX_CANDIDATE_UNCLES: usize = 42;
 type BlockTemplateParams = (Option<u64>, Option<u64>, Option<Version>);
 type BlockTemplateResult = Result<BlockTemplate, FailureError>;
 const BLOCK_ASSEMBLER_SUBSCRIBER: &str = "block_assembler";
@@ -134,7 +133,7 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
         let new_uncle_receiver = notify.subscribe_new_uncle(BLOCK_ASSEMBLER_SUBSCRIBER);
         let thread = thread_builder
             .spawn(move || {
-                let mut candidate_uncles = LruCache::new(MAX_CANDIDATE_UNCLES);
+                let mut candidate_uncles = CandidateUncles::new();
                 loop {
                     select! {
                         recv(signal_receiver) -> _ => {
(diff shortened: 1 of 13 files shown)


zhangsoledad:

zhangsoledad:

zhangsoledad:

miner/src/candidate_uncles.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    pub fn insert(&mut self, uncle: Arc<UncleBlock>) -> bool {
+        let number = uncle.header().number();
+        if self.count >= MAX_CANDIDATE_UNCLES {
+            let first_key = *self.map.keys().next().expect("length checked");
+            if number > first_key {
+                if let Some(set) = self.map.get(&first_key) {
+                    self.count -= set.len();
+                }
+                self.map.remove(&first_key);
```
u2:
```
if let Some(set) = self.map.remove(&first_key) {
    self.count -= set.len();
}
```

verification/src/uncles_verifier.rs
```diff
@@ -93,6 +95,12 @@ where
                 return Err(Error::Uncles(UnclesError::InvalidNumber));
             }
 
+            if !(included.contains(uncle.header.parent_hash())
+                || self.provider.descendant(uncle.header.parent_hash()))
+            {
+                return Err(Error::Uncles(UnclesError::DescendantLimit));
+            }
```
u2: Should check the uncle parent's number is just one less than the uncle's number. Either in descendant or uncle verifier. Otherwise, the uncle's parent could be any history block with a not matched continuous number.

zhangsoledad: no it's not our rule

doitian:
@nervos-bot-user dummy ci 46067454a2a738375d47ced9ae61a09bb6983833

miner/src/block_assembler.rs
```diff
@@ -470,43 +468,48 @@ impl<CS: ChainStore + 'static> BlockAssembler<CS> {
...
         current_epoch_ext: &EpochExt,
-        candidate_uncles: &mut LruCache<H256, Arc<Block>>,
-    ) -> FnvHashSet<UncleBlock> {
+        candidate_uncles: &mut CandidateUncles,
+    ) -> Vec<UncleBlock> {
         let store = self.shared.store();
         let epoch_number = current_epoch_ext.number();
         let max_uncles_num = self.shared.consensus().max_uncles_num();
-        let mut uncles = FnvHashSet::with_capacity_and_hasher(max_uncles_num, Default::default());
+        let mut uncles: Vec<UncleBlock> = Vec::with_capacity(max_uncles_num);
```
jjyr: why not use HashSet?

zhangsoledad: Need a sorted map,  keep the highest candidate uncle

miner/src/candidate_uncles.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    pub fn values(&self) -> impl Iterator<Item = &Arc<UncleBlock>> {
+        self.map.values().flat_map(HashSet::iter)
+    }
+
+    pub fn remove(&mut self, uncle: &Arc<UncleBlock>) -> bool {
+        let number = uncle.header().number();
+
+        let mut is_empty = false;
```
quake:
```suggestion
        if let Entry::Occupied(mut entry) = self.map.entry(number) {
            let set = entry.get_mut();
            if set.remove(&uncle) {
              self.count -= 1;
              if set.is_empty() {
                entry.remove();
              }
              return true;
            }
        }
        false
```

PR #1256, opened 2019-07-18
chore: fix ci

u2:


u2:
@nervos-bot-user give me five? test..

u2:
The different python versions would generate different doc.

u2:
> > The different python versions would generate different doc.
> 
> It is not the actual reason.
> 
> The document format should change like:
> 
> ```
>         "types": [
>             {
>                 "address": "The IP/Subnet with an optional netmask (default is /32 = single IP)"
>             },
>             {
>                 "command": "`insert` to insert an IP/Subnet to the list, `delete` to delete an IP/Subnet from the list"
>             },
>             {
>                 "ban_time": "Time in milliseconds how long (or until when if [absolute] is set) the IP is banned, optional parameter, null means using the default time of 24h"
>             },
>             {
>                 "absolute": "If set, the `ban_time` must be an absolute timestamp in milliseconds since epoch, optional parameter"
>             },
>             {
>                 "reason": "Ban reason, optional parameter"
>             }
>         ]
> ```
> 
> And it would be good to add an assertion `assert(len(item) == 1)` here.

Even like this, different python versions lead different docs.

keroro520:
> > > The different python versions would generate different doc.
> > 
> > 
> > It is not the actual reason.
> > The document format should change like:
> > ```
> >         "types": [
> >             {
> >                 "address": "The IP/Subnet with an optional netmask (default is /32 = single IP)"
> >             },
> >             {
> >                 "command": "`insert` to insert an IP/Subnet to the list, `delete` to delete an IP/Subnet from the list"
> >             },
> >             {
> >                 "ban_time": "Time in milliseconds how long (or until when if [absolute] is set) the IP is banned, optional parameter, null means using the default time of 24h"
> >             },
> >             {
> >                 "absolute": "If set, the `ban_time` must be an absolute timestamp in milliseconds since epoch, optional parameter"
> >             },
> >             {
> >                 "reason": "Ban reason, optional parameter"
> >             }
> >         ]
> > ```
> > 
> > 
> > And it would be good to add an assertion `assert(len(item) == 1)` here.
> 
> Even like this, different python versions lead to different docs.

Why "different python versions lead to different docs" ???

u2:
I will crate a PR for this.

u2:
@keroro520 , thanks, please review.

PR #1259, opened 2019-07-19
chore: remove hashbrown

u2:
https://blog.rust-lang.org/2019/07/04/Rust-1.36.0.html#a-new-hashmapk,-v%3E-implementation


u2:
And fnv?

jjyr:
> And fnv?

IMO it should be removed too, I guess the performance improvement in our scenario is trivial https://cglab.ca/~abeinges/blah/hash-rs/

doitian:
@nervos-bot-user dummy ci 51985d998419905601c36edbce502097474f4694

PR #1267, opened 2019-07-20
fix: header verifier with wrong header resolver

u2:
https://github.com/nervosnetwork/ckb/blob/develop/verification/src/block_verifier.rs#L161-L173
https://github.com/nervosnetwork/ckb/blob/develop/sync/src/types.rs#L786
https://github.com/nervosnetwork/ckb/blob/develop/sync/src/types.rs#L811

Should find parent from header_map firstly, otherwise, it may not find the parent and lead to the UnknownParent.

todo:
- [ ] add test

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -584,7 +584,7 @@ mod tests {
         let block: BlockBuilder = block_template.into();
         let block = block.build();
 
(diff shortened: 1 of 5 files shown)


keroro520:

u2:
bors give me five

PR #1268, opened 2019-07-20
fix: shoud not reject not better compact block

u2:
https://github.com/cryptape/ckb-internal/issues/1150

### DIFF
--- a/sync/src/relayer/tests/compact_block_process.rs
+++ b/sync/src/relayer/tests/compact_block_process.rs
@@ -17,7 +17,6 @@ use crate::NetworkProtocol;
 use crate::MAX_PEERS_PER_BLOCK;
 use faketime::unix_time_as_millis;
 use fnv::FnvHashSet;
-use numext_fixed_uint::U256;
 use std::cell::RefCell;
 use std::convert::TryInto;
 use std::iter::FromIterator;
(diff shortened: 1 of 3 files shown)


sync/src/relayer/tests/compact_block_process.rs
```diff
@@ -144,50 +143,6 @@ fn test_unknow_parent() {
     );
 }
 
-#[test]
```
keroro520: I think this test case is still helpful.
So if we change to accept the not-better block, we should change the test case logic, but not remove it. Or add other tests instead.

keroro520:
Relayer should only accept blocks with a height greater than `tip - N`. Otherwise, the attackers may launch a spam attack by sending many blocks with low difficulty.

u2:
> Relayer should only accept blocks with a height greater than `tip - N`. Otherwise, the attackers may launch a spam attack by sending many blocks with low difficulty.

This another problem, create an issue?

sync/src/relayer/tests/compact_block_process.rs
u2: updated.

sync/src/relayer/tests/compact_block_process.rs
```diff
@@ -145,23 +144,23 @@ fn test_unknow_parent() {
...
+fn test_accept_not_a_better_block() {
     let (relayer, _) = build_chain(5);
-    let chain_state = relayer.shared.lock_chain_state();
-    let header = chain_state.tip_header();
+    let header = {
+        let chain_state = relayer.shared.lock_chain_state();
+        chain_state.tip_header().clone()
+    };
+
+    // The timestamp is random, so it may be not a better block.
```
keroro520: Can we give it less difficulty like [before](https://github.com/nervosnetwork/ckb/pull/1268/files#diff-0ae539c65d2a643736f1fdeb17c7cdb5L158) ?

u2: The less difficulty is an invalid header.

https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/compact_block_process.rs#L170

https://github.com/nervosnetwork/ckb/blob/develop/verification/src/header_verifier.rs#L149

Here it's a fork test(valid header but not a better block), not an invalid header test.

PR #1269, opened 2019-07-20
test: fix some commented code in tests

yangby-cryptape:
The commented code was broken, but it's required if we want to generate the `rpc.json` (for passing the rpc tests).

I have to spend a lot of time to understand the tests (breaking my original works).
So, we **SHOULD NOT** comment them again, keep them be available, so we can use it easily, next time.

### DIFF
--- a/rpc/src/test.rs
+++ b/rpc/src/test.rs
@@ -67,6 +67,7 @@ fn new_cellbase(number: BlockNumber, always_success_script: &Script) -> Transact
 
 fn setup_node(
     height: u64,
+    print_mode: bool,
 ) -> (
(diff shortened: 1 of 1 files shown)


quake:
is it duplicate with https://github.com/nervosnetwork/ckb/pull/1278 ?

yangby-cryptape:
> is it duplicate with #1278 ?

Merge or Close, both are OK.

I pushed this PR, because I want to pass CI for my new commits, and I spent a lot of time to understand the test and fix it.
The only way to get the correct hash is adding a `println!(..)`.

If #1278 did this, then this PR can be closed.

keroro520:
I think this PR can be closed. But before that, I need reviews

PR #1271, opened 2019-07-20
fix: ignoring get_block

u2:


keroro520:
It is not the real reason.
This PR would introduce another bug.

PR #1272, opened 2019-07-20
fix: Clean status of new inserted block

keroro520:
fix: Clear the newly inserted block from block_status_map.

  We don't know whether the actual block status is BLOCK_VALID or BLOCK_INVALID.
  So we just simply remove the corresponding in-memory block status,
  and the next time `get_block_status` would acquire the real-time
  status via fetching block_ext from the database.


keroro520:
@nervos-bot-user give me five

PR #1273, opened 2019-07-20
feat: tweak max block interval

zhangsoledad:


zhangsoledad:
https://github.com/nervosnetwork/ckb/pull/1274

PR #1278, opened 2019-07-20
test: Refactor rpc test

keroro520:
* Make it more easy to modified when APIs changed
* Generate document in JSON format when the test fails
* Print the all errors at the end

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -56,7 +56,7 @@ echo '{
     "jsonrpc": "2.0",
     "method": "get_block",
     "params": [
-        "0x99c7e95158b27cf165aea40f434773ddc5ffefa81df03806036c19af81563b38"
+        "0x81ca425c82c82dcadd289029aba4f573b1fea06d42ad95983f5b0de0f15a70d2"
     ]
 }' \
 | tr -d '\n' \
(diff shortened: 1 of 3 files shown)


yangby-cryptape:
I don't understand, you didn't change any `hash(..)` methods, why all hashes are changed.
And, why you updated the `dao` data? I don't understand change `"capacity": "50000000000000"` to `"capacity": "101930723083"` has any meanings in this PR.

If anyone change any code, just re-generate test cases, **how could us to distinguish the boundary of changes**.

I changed the `hash(..)` methods, so only block hash, uncle hash, transaction hash and witness hash are changed in my PR, it was enough to pass the tests.
So I can prove my PR didn't effect the irrelevant part. Maybe my `hash(..)` method was incorrect, but the effect has a explicit boundary.

keroro520:
> I don't understand, you didn't change any `hash(..)` methods, why all hashes are changed.

For example, I change `get_block_by_number(2)` to `get_block_by_number(1024)`, `get_block($block_hash(2))` to `get_block($block_hash(1024))`.
The reason is that the original `$block_hash(2)` was filled manually, and now the `$block_hash(1024)` can [be generated by program](https://github.com/nervosnetwork/ckb/pull/1278/files#diff-51a07bc2375b5d9aa3bc40ed4c6d667aR303). I cannot get the `$block_hash(2)` in the test.

---

> And, why you updated the `dao` data? I don't understand change `"capacity": "50000000000000"` to `"capacity": "101930723083"` has any meanings in this PR.

Yes, I shouldn't change the block reward in this PR. I will change it back later.

---

> If anyone change any code, just re-generate test cases, **how could us to distinguish the boundary of changes**.
> I changed the `hash(..)` methods, so only block hash, uncle hash, transaction hash and witness hash are changed in my PR, it was enough to pass the tests.
> So I can prove my PR didn't effect the irrelevant part. Maybe my `hash(..)` method was incorrect, but the effect has a explicit boundary.

I agree with you. In this PR, I change the target block to `1024` and some params, so the hashes changed. Later any changes for `hash(...)`, it will only affect the hashes.

keroro520:
>> And, why you updated the dao data? I don't understand change "capacity": "50000000000000" to "capacity": "101930723083" has any meanings in this PR.

> Yes, I shouldn't change the block reward in this PR. I will change it back later.

I have to say that the original `dao/capacity/...` is wrong, we give them fake but wrong numbers. The reason why we didn't notice this mistake before, is that we [insert blocks with `need_verification = false`](https://github.com/nervosnetwork/ckb/blob/709154450dd8383ae19c8a95ff0c99f6dcaf129d/rpc/src/test.rs#L136)

I fix this mistake in a new commit: [98d418](https://github.com/nervosnetwork/ckb/pull/1278/commits/98d418c82381247f3144cb99d5590a62fa2ce311)

doitian:
Please rebase

keroro520:
> Please rebase

Rebased. PTAL

doitian:
conflict again 😢

doitian:
@yangby-cryptape do you still have any concern?

PR #1280, opened 2019-07-21
refactor: explicit deny alert when version not match

jjyr:


util/network-alert/src/notifier.rs
```diff
@@ -44,23 +44,26 @@ impl Notifier {
...
-        {
-            return;
+            if let Some(ref max_v) = alert.max_version {
+                if Version::parse(&max_v)
+                    .map(|max_v| client_version > max_v)
+                    .unwrap_or(true)
+                {
+                    return;
+                }
+            }
```
keroro520: It's better to define a helper function, like `is_allowed_version` or else.

keroro520: Please add unit tests.

jjyr: the old test case is works well https://github.com/nervosnetwork/ckb/blob/develop/util/network-alert/src/tests/test_notifier.rs#L6

this PR just explicit the program's behaviour

u2:
bors ping

util/network-alert/src/notifier.rs
```diff
@@ -30,9 +30,32 @@ impl Notifier {
         }
     }
 
-    pub fn add(&mut self, alert: Arc<Alert>) {
+    fn is_accept_version(&self, alert: &Arc<Alert>) -> bool {
```
doitian: Syntax error

```suggestion
    fn is_version_effective(&self, alert: &Arc<Alert>) -> bool {
```

util/network-alert/src/notifier.rs
```diff
@@ -44,23 +67,10 @@ impl Notifier {
...
 
         // check conditions, figure out do we need to notice this alert
-        if alert
-            .min_version
-            .as_ref()
-            .map(|min_v| Version::parse(&self.client_version) < Version::parse(min_v))
-            == Some(true)
-        {
+        if !self.is_accept_version(&alert) {
             return;
```
doitian: Add a log here.

PR #1279, opened 2019-07-21
refactor: extract data field from CellOutput to Transaction

jjyr:
1. remove `data` field from `CellOutput`
2. add `data_hash` field to `CellOutput`
3. add `outputs_data: Vec<Bytes>` to `Transaction`, `hash(outputs_data[i]) == outputs[i].data_hash`

PR #1281, opened 2019-07-21
perf: use `CellMeta` cached data_hash value

quake:


quake:
This PR is stale after https://github.com/nervosnetwork/ckb/pull/1279

PR #1282, opened 2019-07-22
fix: byte index is not a char boundary for non-ASCII char

yangby-cryptape:


u2:
Add one test?

yangby-cryptape:
> Add one test?

DONE!

PR #1285, opened 2019-07-23
test: block proposal process test

u2:
test: assert request proposal for compact block process
test: block proposal process test


doitian:
Please fix the CI

PR #1286, opened 2019-07-23
refactor: use block_hash 

u2:
refactor: use block_hash instead of block_number in get_block_proposal message


TheWaWaR:
Where use the `GetBlockProposal::block_hash` field?

u2:
> block_hash

Now, it's useless as before. If we want to add more verifying or log, we should use the hash.

PR #1288, opened 2019-07-23
chore: ignore generated files in pre-commit hook

u2:


devtools/git/pre-commit
```diff
@@ -85,7 +100,7 @@ else
   FMTFAILED=0
   for file in $(git diff --name-only --cached);
   do
-    if [ -f "$file" ] && [ ${file: -3} == ".rs" ]; then
+    if [ -f "$file" ] && [ ${file: -3} == ".rs" ] && [ $(contains "${IGNORE[@]}" "$file") == "n" ] ; then
```
doitian: I recommend using `case` here to filter out files that should be ignored.

u2: updated

doitian:
ping @nervosnetwork/ckb-dev

PR #1297, opened 2019-07-24
feature: add get_block_finalized_reward_info rpc

u2:
Firstly fix test:  `cellbase` should have `witness`, otherwise the RPC test will panic.

https://github.com/nervosnetwork/ckb/pull/1297/commits/25445bc4ab1c35e160fefe2bacd89ada32fcda76

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -5,6 +5,7 @@
     *   [`get_block`](#get_block)
     *   [`get_block_by_number`](#get_block_by_number)
     *   [`get_block_hash`](#get_block_hash)
+    *   [`get_cellbase_output_capacity_details`](#get_cellbase_output_capacity_details)
     *   [`get_cells_by_lock_hash`](#get_cells_by_lock_hash)
     *   [`get_current_epoch`](#get_current_epoch)
     *   [`get_epoch_by_number`](#get_epoch_by_number)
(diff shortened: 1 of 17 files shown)


util/reward-calculator/src/lib.rs
```diff
@@ -33,7 +34,10 @@ impl<'a, P: ChainProvider> RewardCalculator<'a, P> {
 
     /// `RewardCalculator` is used to calculate block finalize target's reward according to the parent header.
     /// block reward consists of four parts: base block reward, tx fee, proposal reward, and secondary block reward.
-    pub fn block_reward(&self, parent: &Header) -> Result<(Script, Capacity), FailureError> {
+    pub fn block_reward(
+        &self,
+        parent: &Header,
+    ) -> Result<(Script, Capacity, BlockReward), FailureError> {
```
doitian: We can impl a method on `BlockReward` to return the total capacity, then we don't need to return the total capacity here.

u2: or add a new field `total`?

rpc/src/module/chain.rs
```diff
@@ -251,4 +254,17 @@ impl<CS: ChainStore + 'static> ChainRpc for ChainRpcImpl<CS> {
...
     }
+
+    fn get_block_finalized_reward_info(&self, hash: H256) -> Result<Option<BlockRewardView>> {
+        Ok(self
+            .shared
+            .store()
+            .get_block_header(&hash)
+            .and_then(|header| {
+                self.shared
+                    .finalize_block_reward(&header)
```
doitian: I'm afraid `finalize_block_reward` requires parent hash instead of the current block hash.

doitian: The reason is that the info is required when constructing the new block, which hash is not available yet.

doitian: Maybe we should rename `finalize_block_reward` to more verbose `finalize_block_reward_via_parent_hash`, and `RewardCalculator::block_reward` to `RewardCalculator::block_reward_via_parent_hash`.

rpc/json/rpc.json
```diff
@@ -666,6 +666,25 @@
             }
         ]
     },
+    {
+        "description": "Returns the information about block reward.",
```
doitian:
```suggestion
        "description": "Returns each component of the created CKB in this block's cellbase, which is issued to a block N - 1 - ProposalWindow.farthest, where this block's height is N",
```

doitian:
Suggest a new name `get_cellbase_output_capacity_details`

rpc/src/module/chain.rs
u2: `fn finalize_block_reward(&self, _parent: &Header)`  implicitly tell that.

rpc/json/rpc.json
u2: thanks

rpc/src/module/chain.rs
u2: updated, would find the parent first and then finalize_block_reward.

rpc/README.md
```diff
@@ -233,7 +246,45 @@ http://localhost:8114
...
+```json
+{
+    "id": 2,
+    "jsonrpc": "2.0",
+    "result": {
+        "primary": "100000000000",
+        "proposal_reward": "0",
+        "secondary": "3936000",
+        "total": "100003936000",
+        "tx_fee": "0"
```
keroro520: Why `proposal_reward` has a suffix but others not?

What about:

`primary_block_reward`
`secondary_block_reward`
`total_reward`
`transactions_fee`

u2: > Why `proposal_reward` has a suffix but others not?
> 
> What about:
> 
> `primary_block_reward`
> `secondary_block_reward`
> `total_reward`
> `transactions_fee`

https://github.com/cryptape/ckb-internal/issues/1206

PR #1300, opened 2019-07-25
chore: add a log on future task panic

driftluo:
add a log on future task panic

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -938,6 +938,7 @@ impl NetworkService {
                 let mut runtime = runtime::Builder::new()
                     .core_threads(num_threads)
                     .name_prefix("NetworkRuntime-")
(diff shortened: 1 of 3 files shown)


network/src/network.rs
```diff
@@ -938,6 +938,7 @@ impl NetworkService {
                 let mut runtime = runtime::Builder::new()
                     .core_threads(num_threads)
                     .name_prefix("NetworkRuntime-")
+                    .panic_handler(|_| warn!("One future task panic"))
```
doitian: See `util/logger/src/lib.rs` how to print the backtrace.

doitian: I think we can expose it as a global pub function. Given panic info, print the backtrace via log.

driftluo:
This implementation is of little significance, close it

PR #1308, opened 2019-07-26
refactor: using ProposalShortId in CompactBlock

quake:
This PR changes `CompactBlock#short_id` to `ProposalShortId`, and `reconstruct_block` will try to get tx from the entire tx pool instead of proposal tx pool only.

### DIFF
--- a/shared/src/tx_pool/pool.rs
+++ b/shared/src/tx_pool/pool.rs
@@ -212,6 +212,16 @@ impl TxPool {
         self.proposed.get_tx(id).cloned()
     }
 
+    pub fn get_tx_from_proposed_and_others(&self, id: &ProposalShortId) -> Option<Transaction> {
+        self.proposed
+            .get_tx(id)
+            .or_else(|| self.gap.get_tx(id))
+            .or_else(|| self.pending.get_tx(id))
+            .or_else(|| self.orphan.get_tx(id))
+            .or_else(|| self.conflict.get(id).map(|e| &e.transaction))
+            .cloned()
+    }
+
     pub(crate) fn remove_committed_txs_from_proposed<'a>(
         &mut self,
         txs: impl Iterator<Item = &'a Transaction>,
--- a/sync/src/relayer/compact_block.rs
+++ b/sync/src/relayer/compact_block.rs
@@ -7,14 +7,11 @@ use numext_fixed_hash::H256;
 use std::collections::HashSet;
 use std::convert::{TryFrom, TryInto};
 
-pub type ShortTransactionID = [u8; 6];
-
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct CompactBlock {
     pub header: Header,
     pub uncles: Vec<UncleBlock>,
-    pub nonce: u64,
-    pub short_ids: Vec<ShortTransactionID>,
+    pub short_ids: Vec<ProposalShortId>,
     pub prefilled_transactions: Vec<IndexTransaction>,
     pub proposals: Vec<ProposalShortId>,
 }
@@ -25,7 +22,6 @@ impl Default for CompactBlock {
         Self {
             header,
             uncles: Default::default(),
-            nonce: Default::default(),
             short_ids: Default::default(),
             prefilled_transactions: Default::default(),
             proposals: Default::default(),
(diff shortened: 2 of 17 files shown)


u2:
what about the performance?

sync/src/relayer/compact_block.rs
```diff
@@ -5,14 +5,11 @@ use ckb_protocol::{self, cast, FlatbuffersVectorIterator};
...
 
-pub type ShortTransactionID = [u8; 6];
-
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct CompactBlock {
     pub header: Header,
     pub uncles: Vec<UncleBlock>,
-    pub nonce: u64,
-    pub short_ids: Vec<ShortTransactionID>,
+    pub short_ids: Vec<ProposalShortId>,
```
driftluo: I think the name of the `ProposalShortId` can now cause ambiguity. I think it should be replaced by `TransactionShortId`.

u2: keep the name as before pls.

sync/src/relayer/mod.rs
```diff
@@ -286,15 +283,12 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
...
-        let mut short_ids_set: HashSet<&ShortTransactionID> =
-            compact_block.short_ids.iter().collect();
+        let mut short_ids_set: HashSet<&ProposalShortId> = compact_block.short_ids.iter().collect();
 
-        let mut txs_map: FnvHashMap<ShortTransactionID, Transaction> = transactions
+        let mut txs_map: FnvHashMap<ProposalShortId, Transaction> = transactions
             .into_iter()
             .filter_map(|tx| {
-                let short_id = short_transaction_id(key0, key1, &tx.witness_hash());
+                let short_id = tx.proposal_short_id();
```
u2: Rename the `proposal_short_id` to `short_id`?

sync/src/relayer/mod.rs
```diff
@@ -305,17 +299,11 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
...
-                let short_id = short_transaction_id(key0, key1, &entry.transaction.witness_hash());
-                if short_ids_set.remove(&short_id) {
-                    txs_map.insert(short_id, entry.transaction.clone());
-
-                    // Early exit here for performance
-                    if short_ids_set.is_empty() {
-                        break;
-                    }
+            short_ids_set.into_iter().for_each(|short_id| {
+                if let Some(tx) = chain_state.tx_pool().get_tx(&short_id) {
```
u2: Add a new method for it? Because of the fetching order.

https://github.com/nervosnetwork/ckb/blob/develop/shared/src/tx_pool/pool.rs#L192-L200

sync/src/relayer/tests/reconstruct_block.rs
```diff
@@ -75,28 +56,11 @@ fn test_reconstruct_block() {
...
+        let short_ids = short_transactions
             .iter()
-            .map(|tx| short_transaction_id(key0, key1, &tx.witness_hash()))
+            .map(|tx| tx.proposal_short_id())
             .collect();
         compact.short_ids = short_ids;
         compact.prefilled_transactions = prefilled;
-
-        // Split first 2 short transactions and move into pool. These pool transactions are not
-        // proposed, so it will not be acquired inside `reconstruct_block`
```
u2: keep this test case?

sync/src/relayer/compact_block.rs
quake: > I think the name of the `ProposalShortId` can now cause ambiguity. I think it should be replaced by `TransactionShortId`.

+1, I would like to do it in a sperate PR if this PR was accepted

sync/src/relayer/mod.rs
quake: Updated, please review again, thanks.

shared/src/tx_pool/pool.rs
```diff
@@ -212,6 +212,16 @@ impl TxPool {
...
+    pub fn get_tx_from_proposed_and_others(&self, id: &ProposalShortId) -> Option<Transaction> {
+        self.proposed
+            .get_tx(id)
+            .or_else(|| self.gap.get_tx(id))
+            .or_else(|| self.pending.get_tx(id))
+            .or_else(|| self.orphan.get_tx(id))
+            .or_else(|| self.conflict.get(id).map(|e| &e.transaction))
+            .cloned()
+    }
+
```
keroro520: What is the difference between `get_tx` and `get_tx_from_proposed_and_others`? If there is, please add some comments.

quake: I added this new fn according to review comment: https://github.com/nervosnetwork/ckb/pull/1308#discussion_r308047955

sync/src/relayer/mod.rs
keroro520: I am very confused about `get_tx` and `get_proposed_and_others`.
Please add some comments to explain the differences.
Why the fetching order matters and why we don't change the fetching order of `get_tx` and reuse it?

shared/src/tx_pool/pool.rs
doitian: It looks like they are the same.

sync/src/relayer/mod.rs
quake: Most txs in compact block will be in the proposal pool when we are trying to reconstruct, so add a new fn here to fetch tx from proposal pool first, then gap and pending pool.

PR #1310, opened 2019-07-27
fix: compact block median time

u2:


doitian:
Integration ran locally

doitian:
Merged

PR #1314, opened 2019-07-27
fix: randomly failed integration test valid since

jjyr:


test/src/specs/tx_pool/valid_since.rs
```diff
@@ -169,8 +169,7 @@ impl ValidSince {
 
         // Calculate current block median time
         let tip_number = node.rpc_client().get_tip_block_number();
-        let mut timestamps: Vec<u64> = (tip_number.saturating_sub(MEDIAN_TIME_BLOCK_COUNT + 1)
-            ..=tip_number)
+        let mut timestamps: Vec<u64> = ((tip_number - MEDIAN_TIME_BLOCK_COUNT + 1)..=tip_number)
```
doitian: Is it possible that `tip_number < MEDIAN_TIME_BLOCK_COUNT` here? Should we use the following expression?

```
(tip_number + 1).saturating_sub(MEDIAN_TIME_BLOCK_COUNT)
```

jjyr: It is not a random input test, it's ok to use minus.

PR #1318, opened 2019-07-30
fix: only accept blocks with a height greater than tip - N 

u2:


sync/src/relayer/compact_block_process.rs
```diff
@@ -57,6 +57,24 @@ impl<'a, CS: ChainStore + 'static> CompactBlockProcess<'a, CS> {
...
         let block_hash = compact_block.header.hash().to_owned();
 
+        // Only accept blocks with a height greater than tip - N
+        // where N is the current epoch length
+        let (lowest_number, tip) = {
+            let cs = self.relayer.shared.lock_chain_state();
+            let epoch_length = cs.current_epoch_ext().length();
+            let tip = cs.tip_header().clone();
+
+            if tip.number() > epoch_length {
```
doitian: It is simpler to use `saturating_sub`

sync/src/relayer/error.rs
```diff
@@ -42,4 +42,6 @@ pub enum Ignored {
     AlreadyInFlight,
     #[fail(display = "Already stored")]
     AlreadyStored,
+    #[fail(display = "Block is too early")]
+    TooEarlyBlock,
```
doitian: I think `TooOldBlock` is more accurate. Too early can be interpreted that the block is in the future, and comes too early.

sync/src/relayer/compact_block_process.rs
u2: thanks.

sync/src/relayer/error.rs
u2: fixed

PR #1322, opened 2019-07-31
fix: Reject hash type referencing via type with multiple matches

xxuejie:
Note this only applies to `Type` as a HashType, since if multiple
cells have the same data hash, the content must be the same.

### DIFF
--- a/script/src/verify.rs
+++ b/script/src/verify.rs
@@ -54,10 +54,7 @@ pub struct TransactionScriptsVerifier<'a, DL> {
     rtx: &'a ResolvedTransaction<'a>,
 
     binaries_by_data_hash: FnvHashMap<H256, Bytes>,
-    binaries_by_type_hash: FnvHashMap<H256, Bytes>,
-    // Used to denied the case when multiple matches are found for
-    // a type hash reference.
-    occurences_by_type_hash: FnvHashMap<H256, usize>,
+    binaries_by_type_hash: FnvHashMap<H256, (Bytes, bool)>,
     lock_groups: FnvHashMap<H256, ScriptGroup>,
     type_groups: FnvHashMap<H256, ScriptGroup>,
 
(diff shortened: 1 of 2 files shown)


script/src/verify.rs
```diff
@@ -740,6 +766,88 @@ mod tests {
...
+        };
+        let store = Arc::new(new_memory_store());
+        let data_loader = DataLoaderWrapper::new(store);
+
+        let config = ScriptConfig {
+            runner: Runner::default(),
+        };
+        let verifier = TransactionScriptsVerifier::new(&rtx, &data_loader, &config);
+
+        assert!(verifier.verify(100_000_000).is_err());
```
jjyr: It's better to use `assert_eq` to test the exact error type

script/src/verify.rs
```diff
@@ -93,12 +96,17 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
...
         let mut binaries_by_data_hash: FnvHashMap<H256, Bytes> = FnvHashMap::default();
         let mut binaries_by_type_hash: FnvHashMap<H256, Bytes> = FnvHashMap::default();
+        let mut occurences_by_type_hash: FnvHashMap<H256, usize> = FnvHashMap::default();
         for resolved_dep in resolved_deps {
             if let Some(cell_meta) = &resolved_dep.cell() {
                 let data = data_loader.load_cell_data(cell_meta).expect("cell data");
                 binaries_by_data_hash.insert(cell_meta.data_hash().to_owned(), data.to_owned());
                 if let Some(t) = &cell_meta.cell_output.type_ {
                     binaries_by_type_hash.insert(t.hash(), data.to_owned());
+                    occurences_by_type_hash
```
quake: Can we change the binaries_by_type_hash to `FnvHashMap<H256, (Bytes, bool)>`?
```
    binaries_by_type_hash.entry(t.hash())
        .entry(t.hash())
        .and_modify(|e| e.1 = true)
        .or_insert((data.to_owned(), false));
```

PR #1323, opened 2019-07-31
feature: send bulk of tx hashes

u2:
pending for benchmark


sync/src/relayer/mod.rs
```diff
@@ -419,20 +420,61 @@ impl<CS: ChainStore + 'static> Relayer<CS> {
...
+                for tx_hash in tx_hashes {
+                    let peers: Vec<PeerIndex> = nc
+                        .connected_peers()
+                        .into_iter()
+                        .filter(|target_peer| {
+                            known_txs.insert(*target_peer, tx_hash.clone())
+                                && (peer_index != *target_peer)
+                        })
+                        .take(MAX_RELAY_PEERS)
+                        .collect();
```
quake: no need to allocate a new vec here, we can use `for_each` after `take` directly

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    pub fn execute(self) -> Result<(), FailureError> {
+        let transaction_hashes: TransactionHashes = (*self.message).try_into()?;
+
+        let mut x_hashes = Vec::with_capacity(transaction_hashes.hashes.len());
+        {
+            let tx_filter = self.relayer.shared().tx_filter();
+            for tx_hash in transaction_hashes.hashes {
```
quake: Can we find a better name for `x_hashes`? naming is hard...
Should use `filter_map` here:

```suggestion
    let x_hashes: Vec<&H256> = transaction_hashes.hashes.iter().filter_map(|tx_hash| {
       ...
    }).collect();
```

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,106 @@
...
+                }
+            }
+        }
+
+        let mut knowned = Vec::with_capacity(x_hashes.len());
+        {
+            let state = self.relayer.shared.lock_chain_state();
+            let tx_pool = state.tx_pool();
+
+            for tx_hash in mem::replace(&mut x_hashes, vec![]) {
```
quake: Suggest to  use `retain` here:
```suggestion
            x_hashes.retain(|tx_hash| {
                ...
            });
```

sync/src/relayer/transactions_process.rs
```diff
@@ -0,0 +1,176 @@
...
+        TransactionsProcess {
+            message,
+            relayer,
+            nc,
+            peer,
+        }
+    }
+
+    pub fn execute(self) -> Result<(), FailureError> {
+        let relay_txes: RelayTransactions = (*self.message).try_into()?;
```
quake: Please use `txs` as `transaction`'s plural form abbreviation:
```suggestion
        let relay_txs: RelayTransactions = (*self.message).try_into()?;
```

sync/src/relayer/transactions_process.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let mut txes = Vec::with_capacity(relay_txes.transactions.len());
+
+        {
+            let tx_filter = self.relayer.shared().tx_filter();
+            for relay_tx in relay_txes.transactions {
```
quake: Please use `filter_map` and `collect`

sync/src/types.rs
```diff
@@ -1009,10 +1020,29 @@ impl<CS: ChainStore> SyncSharedState<CS> {
         self.tx_filter.lock().insert(hash);
     }
 
+    pub fn mark_as_known_txes(&self, hashes: Vec<H256>) {
```
quake: Please use `txs` as `transaction`'s plural form abbreviation:

```suggestion
    pub fn mark_as_known_txs(&self, hashes: Vec<H256>) {
```

sync/src/relayer/mod.rs
u2: updated, thanks.

sync/src/relayer/transaction_hashes_process.rs
u2: After changing to `filter_map`, Clippy suggested changing to `filter`.

thanks.

sync/src/relayer/transaction_hashes_process.rs
u2: thanks, updated.

sync/src/relayer/transactions_process.rs
u2: thanks.

sync/src/relayer/transactions_process.rs
u2: After changing to `fliter_map`, Clippy suggests changing to `filter`.

sync/src/types.rs
u2: thanks.

shared/src/tx_pool/pool.rs
```diff
@@ -158,6 +158,14 @@ impl TxPool {
             || self.orphan.contains_key(id)
     }
 
+    pub fn contains_tx(&self, id: &ProposalShortId) -> bool {
+        self.pending.contains_key(id)
+            || self.gap.contains_key(id)
+            || self.proposed.contains_key(id)
+            || self.orphan.contains_key(id)
+            || self.conflict.contains_key(id)
```
driftluo: I think the `conflict pool` should not be added to the judgment. The tx in the conflict pool may not be updated for a long time, there will be dirty data

sync/src/types.rs
```diff
@@ -1012,10 +1023,29 @@ impl<CS: ChainStore> SyncSharedState<CS> {
...
+            }
+        }
+
+        let mut tx_filter = self.tx_filter.lock();
+
+        for hash in hashes {
+            tx_filter.insert(hash);
+        }
+    }
+
```
keroro520: `mark_as_known_tx` can reuse `mark_as_known_txs`.

```
pub fn mark_as_known_tx(&self, hash: H256) {
    self.mark_as_known_txs(vec![hash])
}
```

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,111 @@
...
+        let transaction_hashes: TransactionHashes = (*self.message).try_into()?;
+
+        let mut transit_hashes: Vec<H256> = {
+            let tx_filter = self.relayer.shared().tx_filter();
+            transaction_hashes
+                .hashes
+                .into_iter()
+                .filter(|tx_hash| {
+                    if tx_filter.contains(&tx_hash) {
+                        debug_target!(
```
keroro520: Most of the `transaction_hash` is duplicated. So I would suggest removing this logging.

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,111 @@
...
+            .shared()
+            .peers()
+            .state
+            .write()
+            .get_mut(&self.peer)
+        {
+            let mut inflight_transactions = self.relayer.shared().inflight_transactions();
+
+            for tx_hash in transit_hashes {
+                debug_target!(
```
keroro520: Ditto

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,111 @@
...
+                    knowned.push(tx_hash.to_owned());
+                    false
+                } else {
+                    true
+                }
+            })
+        }
+
+        {
+            self.relayer.shared().mark_as_known_txs(knowned);
```
keroro520: I think it is unnecessary to mark known txs here. It should very rarely that "`tx_pool` has a tx but `tx_filter` not". So I would suggest removing this logic or add a log when `!knowned.is_empty()`

sync/src/relayer/transactions_process.rs
```diff
@@ -0,0 +1,180 @@
...
+            let tx_filter = self.relayer.shared().tx_filter();
+
+            relay_txs
+                .transactions
+                .into_iter()
+                .filter(|relay_tx| {
+                    let tx_hash = relay_tx.transaction.hash();
+                    if tx_filter.contains(tx_hash) {
+                        debug_target!(
+                            crate::LOG_TARGET_RELAY,
```
keroro520: Ditto

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -0,0 +1,111 @@
...
+                    tx_hash,
+                    self.peer,
+                );
+
+                let last_ask_timeout = inflight_transactions.get(&tx_hash).cloned();
+
+                if let Some(next_ask_timeout) =
+                    peer_state.add_ask_for_tx(tx_hash.clone(), last_ask_timeout)
+                {
+                    inflight_transactions.insert(tx_hash.clone(), next_ask_timeout);
```
keroro520: `tx_hash.clone()` -> `tx_hash`

sync/src/relayer/transaction_hashes_process.rs
u2: I am not sure, just keep the before logic.

sync/src/relayer/transaction_hashes_process.rs
u2: I am not sure, just keep the logic as before.

sync/src/types.rs
u2: ok

sync/src/relayer/transaction_hashes_process.rs
u2: would remove it later.

sync/src/relayer/transaction_hashes_process.rs
u2: agree, the filter is a false negative.

sync/src/relayer/transactions_process.rs
u2: Updated in the last commit, please review again, thanks.

PR #1324, opened 2019-07-31
chore: upgrade crossbeam channel

driftluo:
release note：
   -  Fix a bug in reference counting.
   - Optimize recv_timeout().
   - Add Select::remove().
   - Various small improvements, code cleanup, more tests.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -176,6 +176,8 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
         ChainService { shared, notify }
     }
 
+    // remove `allow` tag when https://github.com/crossbeam-rs/crossbeam/issues/404 is solved
+    #[allow(clippy::zero_ptr, clippy::drop_copy)]
     pub fn start<S: ToString>(mut self, thread_name: Option<S>) -> ChainController {
         let (signal_sender, signal_receiver) =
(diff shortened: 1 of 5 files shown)


chain/src/chain.rs
```diff
@@ -176,6 +176,7 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
         ChainService { shared, notify }
     }
 
+    #[allow(clippy::zero_ptr, clippy::drop_copy)]
```
doitian: Please add a comment why exclude these linter rules.

doitian: I think we should add a wrapper method to create the crossbeam channel temporarily, and add flags to that wrapper method.

quake:
reference issue: https://github.com/crossbeam-rs/crossbeam/issues/404

chain/src/chain.rs
driftluo: > I think we should add a wrapper method to create the crossbeam channel temporarily, and add flags to that wrapper method.

I think this behavior has no sense other than adding complexity. `select!` is a Macro with uncertain parameters, which means that if you want to wrap it, you also need a Macro with uncertain parameters, macro set macros, adding complexity, and almost no income

PR #1325, opened 2019-07-31
test: add DAO integration-tests

keroro520:
test: Add DAO integration-tests

- [x] DAO script

- [x] DAO deposit/withdraw flow

- [x] ~~DAO reward calculation. I didn't find a good way to test the reward calculation~~


keroro520:

doitian:
> @nervos-bot-user try integration

bors try

doitian:
bors try

test/integration.toml
```diff
@@ -24,7 +24,9 @@ hash_type = "Data"
 # the directory containing this config file.
 [genesis.system_cells]
 files = [
-  { file = "cells/always_success" }
```
doitian: Each integration spec can modify chainspec, maybe we don't need to include the two cells in every specs.

keroro520: Most of the test-cases don't care about the system cells.
I think that Including these two system cells is OK for integration-tests.

doitian:
Integration passed https://github.com/nervosnetwork/ckb/runs/182692185

test/src/specs/tx_pool/dao.rs
```diff
@@ -0,0 +1,403 @@
...
+                        cell_output.type_ = Some(deposit_dao_script());
+                        cell_output
+                    })
+                    .collect();
+                TransactionBuilder::from_transaction(transaction)
+                    .outputs_clear()
+                    .outputs(outputs)
+                    .build()
+            };
+            // TODO compare the reward
```
xxuejie: I did saw we introduced `calculate_maximum_withdraw` in this PR, so I'm curious what's preventing adding the actual logic here?

keroro520: Nothing prevents this logic here.
It is because of that I didn't think clearly about which aspects of reward I should test. So I didn't compare the reward in this PR and just leave a TODO.
Now it more clear. I will add several reward tests later.

test/src/specs/tx_pool/dao.rs
```diff
@@ -0,0 +1,403 @@
...
+            deposited = ensure_committed(node0, &transaction);
```
u2: Withdrawing the deposit many times is ok?

test/src/specs/tx_pool/dao.rs
```diff
@@ -0,0 +1,403 @@
...
+        let deposited = {
+            let transaction = deposit_dao_transaction(node0);
+            ensure_committed(node0, &transaction)
+        };
+        let transaction = withdraw_dao_transaction(node0, deposited.clone());
+        ensure_committed(node0, &transaction);
+    }
+}
+
+pub struct DepositDAOTwice;
```
u2: I think it's withdrawing many times, is not deposit DAO twice, the name is confusing.

test/src/specs/tx_pool/dao.rs
```diff
@@ -0,0 +1,403 @@
...
+                    .outputs(outputs)
+                    .build()
+            };
+            // TODO compare the reward
+            deposited = ensure_committed(node0, &transaction);
+        }
+    }
+}
+
+pub struct WithdrawDAOWithSinceMaturity;
```
u2: WithdrawDAOWithNotMaturitySince

test/src/specs/tx_pool/dao.rs
keroro520: No. It is not withdrawing multiple times.

I withdraw a DAO cell and then deposit it into DAO within the same one transaction. 

PS: `cell_output.type_ = Some(deposit_dao_script())` is depositing a cell.

test/src/specs/tx_pool/dao.rs
keroro520: OK. let's change it to `WithdrawAndDepositDAOWithinSameTx`.

test/src/specs/tx_pool/dao.rs
u2: Get it, thanks.

test/src/specs/tx_pool/dao.rs
```diff
@@ -0,0 +1,415 @@
...
+                        cell_output.type_ = Some(deposit_dao_script());
+                        cell_output
+                    })
+                    .collect();
+                TransactionBuilder::from_transaction(transaction)
+                    .outputs_clear()
+                    .outputs(outputs)
+                    .build()
+            };
+            // TODO compare the reward
```
u2: Add `node0.generate_blocks(20);` otherwise the tx is not  maturity.

keroro520: `ensure_committed` will do it.

PR #1331, opened 2019-08-02
feat: Refactor OutPoint and add dep group support 

TheWaWaR:
- [x] Change `struct OutPoint`
- [x] Add `enum CellDep`
- [x] Change `trait CellProvider` and `trait HeaderProvider`
- [x] Add dep group support
- [x] Fix tests for `chain` module
- [x] Add tests for dep group
- [x] Rebase to develop
- [ ] Fix `dao.c` contract (load header from dep)


core/src/cell.rs
```diff
@@ -489,52 +490,88 @@ impl BlockHeadersProvider {
...
+        let index_end = index_start + mem::size_of::<u32>();
+        let tx_hash =
+            H256::from_slice(&data[tx_hash_start..tx_hash_end]).expect("Invalid tx hash length");
+        let mut index_bytes = [0u8; mem::size_of::<u32>()];
+        index_bytes.copy_from_slice(&data[index_start..index_end]);
+        // Deserialize as little endian u32
+        let index = u32::from_le_bytes(index_bytes);
+
+        let sub_out_point = OutPoint::new(tx_hash, index);
+        if let Some(sub_cell_meta) = cell_resolver(&sub_out_point, false)? {
```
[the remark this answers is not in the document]
TheWaWaR: `cell_resolver` will handle this

core/src/cell.rs
```diff
@@ -546,93 +583,86 @@ pub fn resolve_transaction<'a, CP: CellProvider, HP: HeaderProvider>(
...
+    if !transaction.is_cellbase() {
+        for out_point in transaction.input_pts_iter() {
+            if !current_inputs.insert(out_point.to_owned()) {
                 return Err(UnresolvableError::Dead(out_point.to_owned()));
             }
-            (CellStatus::Unknown, _) => {
-                unknown_out_points.push(out_point.to_owned());
+            if let Some(cell_meta) = resolve_cell(out_point, false)? {
+                resolved_inputs.push(ResolvedInput::new(*cell_meta));
             }
```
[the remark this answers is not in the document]
TheWaWaR: It's for `Unknown` case, and the process should go on. I just keep the same logic as before.

core/src/cell.rs
```diff
@@ -546,93 +583,86 @@ pub fn resolve_transaction<'a, CP: CellProvider, HP: HeaderProvider>(
...
+            )),
+            HeaderStatus::Live(header) => Ok(header),
+        }
+    };
+    for dep in transaction.deps_iter() {
+        match dep {
+            CellDep::Cell(out_point) => {
+                if let Some(cell_meta) = resolve_cell(&out_point, false)? {
+                    resolved_deps.push(ResolvedDep::cell_only(*cell_meta));
+                }
```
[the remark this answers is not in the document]
TheWaWaR: Same as above.

core/src/transaction.rs
```diff
@@ -86,69 +83,88 @@ impl CellOutPoint {
...
-            + self
-                .block_hash
-                .as_ref()
-                .map(|_| H256::size_of())
-                .unwrap_or(0)
+    pub fn out_point(&self) -> Option<&OutPoint> {
+        match self {
+            CellDep::Cell(out_point) => Some(out_point),
+            CellDep::DepGroup(out_point) => Some(out_point),
+            _ => None,
```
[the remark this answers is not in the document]
TheWaWaR: Oops, forgot.

core/src/cell.rs
```diff
@@ -546,93 +583,86 @@ pub fn resolve_transaction<'a, CP: CellProvider, HP: HeaderProvider>(
...
             }
-            (CellStatus::Live(cell_meta), HeaderStatus::Unspecified) => {
-                resolved_deps.push(ResolvedOutPoint::cell_only(*cell_meta));
+            CellDep::CellWithHeader(out_point, block_hash) => {
+                let header = resolve_header(&block_hash, Some(&out_point))?;
+                if let Some(cell_meta) = resolve_cell(&out_point, false)? {
+                    resolved_deps.push(ResolvedDep::cell_and_header(*cell_meta, *header));
+                }
```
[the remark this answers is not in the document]
TheWaWaR: `resolve_cell` return `Result` which handle the `Dead` and `Unknown` cases, and we always need to handle the 2 cases.

shared/src/tx_pool/proposed.rs
```diff
@@ -160,7 +152,9 @@ impl ProposedPool {
                 }
 
                 for d in deps {
-                    self.edges.delete_value_in_deps(d, &id);
+                    if let Some(out_point) = d.out_point() {
+                        self.edges.delete_value_in_deps(out_point, &id);
+                    }
```
TheWaWaR: Should resolve the dep first, if its a dep group.

core/src/cell.rs
```diff
@@ -400,34 +406,30 @@ impl TransactionsProvider {
...
-                            CellMetaBuilder::from_cell_output((*cell).to_owned(), data.to_owned())
-                                .build(),
-                        )
-                    })
-                    .unwrap_or(CellStatus::Unknown),
-                None => CellStatus::Unknown,
-            }
-        } else {
-            CellStatus::Unspecified
+    fn cell(&self, out_point: &OutPoint, _with_data: bool) -> CellStatus {
```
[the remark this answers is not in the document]
TheWaWaR: If it set to true, it's not optional. If we load a cell meta with `with_data` argument but without the data then it's serious bug.

protocol/src/protocol.fbs
```diff
@@ -76,14 +76,19 @@ table Witness {
     data:           [Bytes];
 }
 
-table OutPoint {
+table CellDep {
+    tx_hash:        H256;
+    index:          uint32;
     block_hash:     H256;
+    dep_type:       uint8;
```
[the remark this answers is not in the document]
xxuejie: I have a note regarding this here: https://github.com/nervosnetwork/ckb/blob/59e7b47f5a1ff50b5a97a9d3aebe27d6681c58ab/core/src/script.rs#L11-L15

core/src/transaction.rs
```diff
@@ -84,71 +81,106 @@ impl CellOutPoint {
...
+    pub fn from_group_data(data: &[u8]) -> Result<OutPoint, usize> {
+        if data.len() != mem::size_of::<H256>() + mem::size_of::<u32>() {
+            Err(data.len())
+        } else {
+            let tx_hash_start = 0;
+            let tx_hash_end = tx_hash_start + mem::size_of::<H256>();
+            let index_start = tx_hash_end;
+            let index_end = index_start + mem::size_of::<u32>();
+            let tx_hash = H256::from_slice(&data[tx_hash_start..tx_hash_end])
+                .expect("Invalid tx hash length");
```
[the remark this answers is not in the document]
TheWaWaR: It will never panic, we already checked the length.

core/src/cell.rs
```diff
@@ -489,52 +490,90 @@ impl BlockHeadersProvider {
...
+    }
+
+    let dep_len = data.len() / OUT_POINT_LEN;
+    let mut out_points = Vec::with_capacity(dep_len);
+    for dep_idx in 0..dep_len {
+        let item_start = dep_idx * OUT_POINT_LEN;
+        let item_end = (dep_idx + 1) * OUT_POINT_LEN;
+        out_points.push(
+            OutPoint::from_group_data(&data[item_start..item_end])
+                .expect("parse group item data failed"),
```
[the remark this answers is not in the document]
TheWaWaR: It will not panic if the data length is correct, which we can sure it's right here.

core/src/transaction.rs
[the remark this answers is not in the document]
keroro520: I think `H256::from_slice` will return an error if the data contains unexpected characters. Am I understanding wrong?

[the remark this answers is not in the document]
TheWaWaR: Any data is valid, there is just one error, and that error is for invalid length. 
https://docs.rs/numext-fixed-hash/0.1.4/numext_fixed_hash/enum.FromSliceError.html

keroro520:
Previously we are allowed to specify `block_hash` in input-out-point, to ensure that the previous_output was contained in the given block.

I notice this PR remove `block_hash` from `OutPoint`, so that means we don't support this kind of inputs?

TheWaWaR:
> I notice this PR remove `block_hash` from `OutPoint`, so that means we don't support this kind of inputs?

Yes, we now should define this relationship in dep. `dao` will be a good use case.

util/jsonrpc-types/src/blockchain.rs
```diff
@@ -119,53 +119,62 @@ impl From<CellOutput> for CoreCellOutput {
...
         }
     }
 }
 
-#[derive(Clone, Default, Serialize, Deserialize, PartialEq, Eq, Hash, Debug)]
-pub struct OutPoint {
-    pub cell: Option<CellOutPoint>,
-    pub block_hash: Option<H256>,
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Debug)]
+#[serde(rename_all = "snake_case")]
```
doitian: It's better to use `Internally tagged` enum representation.

doitian: What does the JSON looks like of this type?

I recommend to use following structure:

```json
{ "type": "cell", "previous_output": { "index": "0", "tx_hash": "0x04e427..." } }
{ "type": "cell_with_header", "previous_output": { "index": "0", "tx_hash": "0x04e427..." }, "header_hash": "0xabcd..." }
{ "type": "dep_group", "previous_output": { "index": "0", "tx_hash": "0x04e427..." } }
{ "type": "header", "header_hash": "0xabcd..." }
```

A simple way to achieve this is to modify the enum structure to

```rust
#[serde(rename_all = "snake_case", tag = "type")]
pub enum CellDep {
  Cell { previous_output: OutPoint },
  CellWithHeader { previous_output: OutPoint, header_hash: H256 },
  DepGroup { previous_output: OutPoint },
  Header { header_hash: H256 },
}
```

Also it's better to include 4 different dep types in the RPC doc example response.

TheWaWaR: Since `ScriptHashType` is Camel-Case, can we keep the type field the default way?
``` javascript
{ "type": "Cell", "previous_output": { "index": "0", "tx_hash": "0x04e427..." } }
{ "type": "CellWithHeader", "previous_output": { "index": "0", "tx_hash": "0x04e427..." }, "header_hash": "0xabcd..." }
{ "type": "DepGroup", "previous_output": { "index": "0", "tx_hash": "0x04e427..." } }
{ "type": "Header", "header_hash": "0xabcd..." }
```

TheWaWaR: Or change `ScriptHashType` to `snake_case` ?

core/src/transaction.rs
```diff
@@ -84,71 +81,107 @@ impl CellOutPoint {
...
+            let mut index_bytes = [0u8; mem::size_of::<u32>()];
+            index_bytes.copy_from_slice(&data[index_start..index_end]);
```
u2: `index_bytes.copy_from_slice(&data[index_start..]);`

core/src/transaction.rs
```diff
@@ -84,71 +81,107 @@ impl CellOutPoint {
...
+            // Deserialize as little endian u32
+            let index = u32::from_le_bytes(index_bytes);
+            Ok(OutPoint::new(tx_hash, index))
+        }
+    }
+
+    /// Convert OutPoint to one dep group data item
+    pub fn to_group_data(&self) -> Vec<u8> {
+        [self.tx_hash.to_vec(), self.index.to_le_bytes().to_vec()].concat()
+    }
```
u2: Please add some tests for the from/to conversion.

TheWaWaR: Changed to CellKey's implementation (see the code above in the same file).

core/src/cell.rs
```diff
@@ -31,7 +32,7 @@ impl BlockInfo {
 pub struct CellMeta {
     #[serde(skip)]
     pub cell_output: CellOutput,
-    pub out_point: CellOutPoint,
+    pub out_point: OutPoint,
     pub block_info: Option<BlockInfo>,
     pub cellbase: bool,
```
u2: There is another struct `TransactionInfo` https://github.com/nervosnetwork/ckb/blob/develop/core/src/extras.rs#L19
It's similar to the `BlockInfo`, it seems that we can remove the `BlockInfo`, and we can remove the `pub cellbase: bool` as well, I think whether it's a `cellbase`  depends on the position in the Block. https://github.com/nervosnetwork/ckb/blob/develop/store/src/store.rs#L195

core/src/cell.rs
```diff
@@ -468,52 +470,90 @@ impl BlockHeadersProvider {
...
+    InvalidHeader(H256, Option<OutPoint>),
+    InvalidDepGroup(OutPoint),
     Dead(OutPoint),
     Unknown(Vec<OutPoint>),
     OutOfOrder(OutPoint),
 }
 
+pub fn parse_dep_group_data(data: Bytes) -> Result<Vec<OutPoint>, usize> {
+    // tx hash (32 bytes) + output index (4 bytes)
+    const OUT_POINT_LEN: usize = mem::size_of::<H256>() + mem::size_of::<u32>();
```
u2: put it in transaction.rs?

TheWaWaR:
Close this by: https://github.com/nervosnetwork/ckb/pull/1356

core/src/cell.rs
TheWaWaR: Maybe create a new issue for another PR.

u2: https://github.com/nervosnetwork/ckb/pull/1361

PR #1333, opened 2019-08-02
chore: quick exit on reconcile find an error

driftluo:
Once `found_error` has a value, the function `reconcile_main_chain` must return Error, which can be logged out early.

Take a closer look, maybe there is no need for a variable such as `found_error`

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -656,6 +656,7 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
                     }
                 } else {
                     *verified = Some(false);
+                    return Err(found_error.unwrap().into());
                 }
             } else {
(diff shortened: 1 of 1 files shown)


chain/src/chain.rs
```diff
@@ -656,6 +656,7 @@ impl<CS: ChainStore + 'static> ChainService<CS> {
                     }
                 } else {
                     *verified = Some(false);
+                    return Err(found_error.unwrap().into());
```
keroro520: We wanna mark all its descendants `block_ext.verified = Some(false)` when verification failed, so here should not return but continue to mark.

driftluo:
Turn off this pr first, the problem here is a bit wide, the main problem is not in this function

PR #1335, opened 2019-08-03
feat: tx_pool support sort by fee rate

jjyr:
Changes:

1. proposed pool and pending pool track txs in-pool ancestors and record the `ancestors_fee`, `ancestors_size`, `ancestors_cycles` for each tx.
2. proposed pool and pending pool are sort txs by tx's in-pool ancestors fee rate(aka ancestors score).
3. block_assembler package txs by the sort that tx_pool provides.

### DIFF
--- a/miner/src/block_assembler.rs
+++ b/miner/src/block_assembler.rs
@@ -9,7 +9,11 @@ use ckb_jsonrpc_types::{
 };
 use ckb_logger::{error, info};
 use ckb_notify::NotifyController;
-use ckb_shared::{shared::Shared, tx_pool::ProposedEntry};
+use ckb_shared::{
+    chain_state::ChainState,
+    shared::Shared,
+    tx_pool::{commit_txs_scanner::CommitTxsScanner, TxEntry},
+};
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_store::ChainStore;
 use ckb_traits::ChainProvider;
@@ -220,7 +224,7 @@ impl BlockAssembler {
     }
 
     fn transform_tx(
-        tx: &ProposedEntry,
+        tx: &TxEntry,
         required: bool,
         depends: Option<Vec<u32>>,
     ) -> TransactionTemplate {
@@ -351,8 +355,7 @@ impl BlockAssembler {
         let proposals = chain_state.get_proposals(proposals_limit as usize);
         let txs_size_limit =
             self.calculate_txs_size_limit(bytes_limit, cellbase.data(), &uncles, &proposals)?;
-
-        let (entries, size, cycles) = chain_state.get_proposed_txs(txs_size_limit, cycles_limit);
+        let (entries, size, cycles) = self.package_txs(&chain_state, txs_size_limit, cycles_limit);
         if !entries.is_empty() {
             info!(
                 "[get_block_template] candidate txs count: {}, size: {}/{}, cycles:{}/{}",
@@ -430,6 +433,16 @@ impl BlockAssembler {
         );
 
         Ok(template)
+    }
+
+    fn package_txs(
+        &self,
+        chain_state: &ChainState,
+        size_limit: usize,
+        cycles_limit: Cycle,
+    ) -> (Vec<TxEntry>, usize, Cycle) {
+        let tx_pool = chain_state.tx_pool();
+        CommitTxsScanner::new(tx_pool.proposed()).txs_to_commit(size_limit, cycles_limit)
     }
 
     /// Miner mined block H(c), the block reward will be finalized at H(c + w_far + 1).
@@ -563,11 +576,13 @@ mod tests {
             BlockBuilder, BlockNumber, BlockView, EpochExt, HeaderBuilder, HeaderView,
             TransactionBuilder, TransactionView,
         },
-        packed::{Block, CellInput, CellOutput},
+        packed::{Block, CellInput, CellOutput, CellOutputBuilder},
         H256,
     };
     use ckb_verification::{BlockVerifier, HeaderResolverWrapper, HeaderVerifier, Verifier};
     use std::sync::Arc;
+
+    const BASIC_BLOCK_SIZE: u64 = 706;
 
     fn start_chain(
         consensus: Option<Consensus>,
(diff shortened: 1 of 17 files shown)


miner/src/block_assembler.rs
```diff
@@ -413,6 +418,210 @@ impl BlockAssembler {
...
+                        iter,
+                        size_limit,
+                        cycles_limit,
+                        size,
+                        cycles,
+                        |_| false,
+                    );
+                });
+                // find tx with higher scores
+                let tx_entry = match (pool_tx.take(), modified_tx.take()) {
```
quake: Should implement trait `Ord` for `TxEntry` which delegate to `AncestorsScoreSortKey`, then use `Option#max`, ref: https://doc.rust-lang.org/std/option/enum.Option.html#impl-Ord, no need to use `take` and `replace`

```suggestion
                let tx_entry = [pool_tx, modified_tx].max();
```

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,210 @@ impl BlockAssembler {
...
+                    tx_entry.ancestors_count,
+                    ancestors.len(),
+                    "proposed tx pool ancestors count inconsistent"
+                );
+                // find all decendents then update and insert into modified
+                update_modified_entries(&tx_pool, &in_block_txs, &ancestors, &mut modified_entries);
+                // sort acestors by ancestors_count,
+                // if A is an ancestor of B, B.ancestors_count must large than A
+                let mut ancestors = ancestors.into_iter().collect::<Vec<_>>();
+                ancestors.sort_unstable_by(|a, b| a.ancestors_count.cmp(&b.ancestors_count));
```
quake:
```suggestion
                ancestors.sort_unstable_by_key(|a| a.ancestors_count);
```

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,210 @@ impl BlockAssembler {
...
+                // insert ancestors
+                for entry in ancestors {
+                    let entry_cycles = entry.cycles;
+                    let entry_size = entry.size;
+                    let short_id = entry.transaction.proposal_short_id();
+                    modified_entries.remove(&short_id);
+                    let is_inserted = in_block_txs.insert(short_id);
+                    assert!(is_inserted, "package duplicate txs");
```
quake: use debug_assert ?

shared/src/chain_state.rs
```diff
@@ -469,51 +476,68 @@ impl ChainState {
...
+    #[allow(clippy::too_many_arguments)]
+    fn send_tx_to_pool<F>(
         &self,
+        pool_name: &str,
         tx_pool: &mut TxPool,
         cycles: Option<Cycle>,
         size: usize,
         tx: Transaction,
-    ) -> Result<Cycle, PoolError> {
+        tx_resolved_result: Result<(Cycle, Capacity), PoolError>,
```
quake: I can't understand why we need `tx_resolved_result` here, what's the difference between fn param `cycles: Option<Cycle>` and `Result<Cycle>`, could you brief me?

shared/src/chain_state.rs
```diff
@@ -675,35 +839,43 @@ impl ChainState {
...
+            for id in pool.sorted_keys() {
+                if proposals.len() == proposals_limit {
+                    break;
+                } else if proposals.contains(&id) {
+                    // implies that ancestors are already in proposals
+                    continue;
+                }
+                let mut ancestors = pool.get_ancestors(&id);
+                ancestors.insert(*id);
+                let mut ancestors = ancestors.into_iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
jjyr: The `ancestors` is a `HashSet`, the second assign is to convert it to Vec, I can't figure out how to dismiss it and why it's bad.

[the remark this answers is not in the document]
quake: the position of `id` will be the last of vec if i understand correctly: 
```
let mut ancestors = ancestors.into_iter().collect::<Vec<_>>();
ancestors.sort;
ancestors.push(id);
```

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,210 @@ impl BlockAssembler {
...
+            modified_entries: &mut TxModifiedEntries,
+        ) {
+            for ptx in package_txs {
+                let ptx_id = ptx.transaction.proposal_short_id();
+                if in_block_txs.contains(&ptx_id) {
+                    continue;
+                }
+                let descendants = tx_pool.proposed().get_descendants(&ptx_id);
+                for id in descendants {
+                    let mut tx = match modified_entries.remove(&id) {
```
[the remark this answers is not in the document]
jjyr: `modified_entries` is a composite container, expose Entry will broken the internal consistent

miner/src/block_assembler.rs
jjyr: I'm not sure what happened when None cmp with a Some value.

shared/src/chain_state.rs
jjyr: see the following code, we need to put tx into different pools by error type.

miner/src/block_assembler.rs
quake: `Option` use `#[derive]` for ordering, doc of the sort order of derived ordering: https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable, it will return `Ordering::Less` when compare`None` with `Some`

quake:
hold for stress test

miner/src/block_assembler.rs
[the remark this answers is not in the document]
doitian: I think we should avoid using so many local functions here. In my opinion, we should save all the state required to fetch transactions from the pool in a single structure, and add methods to that structure. Some examples:

```
struct CommitTransactionsScanner<'a> {
    pool: &'a Pool,
    scanned: HashSet<ProposalShortId>,
    modified_entries: TxModifiedEntries,
    ...
}

impl CommitTransactionsScanner {
    fn update_modified_entries();
    fn transactions_to_commit() -> Vec<Transaction>;
}
```

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+                        Some(modified_tx) => modified_tx,
```
doitian: `unwrap_or_else` is clearer.

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+    fn package_txs(
+        &self,
+        chain_state: &ChainState,
+        size_limit: usize,
+        cycles_limit: Cycle,
+    ) -> (Vec<TxEntry>, usize, Cycle) {
+        /// update descendants txs into modified entries
+        fn update_modified_entries(
+            tx_pool: &TxPool,
+            in_block_txs: &HashSet<ProposalShortId>,
```
doitian: What is `in_block_txs` and what is `package_txs`. From this local function itself, I have no ideas. I have to search where this function is used.

doitian: `in_block_txs` is a bad name, the transactions selected are not in the block yet.

Suggest names:

- `in_block_txs`: `fetched_txs` transactions that already fetched
- `package_txs`: `new_fetched_txs` transactions that has just been fetched

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+        fn find_tx_entry<F: Fn(&TxEntry) -> bool>(
+            entry: &mut Option<TxEntry>,
+            iter: &mut dyn Iterator<Item = &TxEntry>,
+            size_limit: usize,
+            cycles_limit: Cycle,
+            size: usize,
+            cycles: Cycle,
+            check_tx_skip: F,
+        ) {
+            let check_limit_is_ok = |entry: &TxEntry| -> bool {
```
doitian: bad smell, this is a depth 3 local function. 😰

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+                            .expect("pool consistent"),
+                    };
+                    tx.sub_entry_weight(&ptx);
+                    modified_entries.insert(tx);
+                }
+            }
+        }
+
+        /// find a valid tx_entry
+        fn find_tx_entry<F: Fn(&TxEntry) -> bool>(
```
doitian: Really complex local function without comments about how it is used.

doitian: The code has many duplicated logic with `std::iter::Filter`，or `std::iter::Filter` + `std::iter::TakeWhilte`.

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+                        iter,
+                        size_limit,
+                        cycles_limit,
+                        size,
+                        cycles,
+                        |_| false,
+                    );
+                });
+                // take tx with higher scores
+                let tx_entry = match cmp::max(&mut pool_tx, &mut modified_tx).take() {
```
doitian: It's tricky to rely on that `Some > None`.

miner/src/block_assembler.rs
```diff
@@ -413,6 +418,192 @@ impl BlockAssembler {
...
+                    Some(entry) => entry,
+                    None => {
+                        // can't find any satisfied tx
+                        break;
+                    }
+                };
+                debug_assert!(!in_block_txs.contains(&tx_entry.transaction.proposal_short_id()));
```
doitian: If we do have a transaction that has already been proposed, we should skip the following code in the loop.

shared/src/chain_state.rs
```diff
@@ -469,51 +476,68 @@ impl ChainState {
...
+                    crate::LOG_TARGET_TX_POOL,
+                    "Failed to generate tx fee for {:x}, reason: {:?}",
+                    rtx.transaction.hash(),
+                    err
+                );
+                PoolError::TxFee
+            })
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
doitian: It seems the function does not use `self`, why it is declared in `ChainState`? It's also a bad smell that the function name is `send_tx_to_pool`, it still needs a closure to tell the function how to add a transaction.

doitian: I think the correct way to implement duplicated logic for different pool structures is defining some common operations as traits.

miner/src/block_assembler.rs
jjyr: In logic, duplicate txs should be filtered by the above lines.

miner/src/block_assembler.rs
jjyr: Have discussed with @quake .  The behaviour of derived order is documented https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable

Use order of option actually simplify our code, IMO we can rely on it unless we are worried the stdlib compatible may be broken by the rust team.

shared/src/chain_state.rs
jjyr: It's hard to abstract common trait on a single pool, for example, https://github.com/nervosnetwork/ckb/blob/9aa1890062b97818e7eda8c82e2db39887966043/shared/src/tx_pool/pool.rs#L116

This is only for reusing the error handle, since we force verify txs before it enters a pool. Need further refractory to simplify this.

miner/src/block_assembler.rs
jjyr: We need to maintain a failure counter in the loop, use iterator is hard to achieve this.

shared/src/tx_pool/commit_txs_scanner.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub fn new(proposed_pool: &'a ProposedPool) -> CommitTxsScanner<'a> {
+        CommitTxsScanner {
+            proposed_pool,
+            entries: Vec::new(),
+            modified_entries: TxModifiedEntries::default(),
+            fetched_txs: HashSet::default(),
+        }
+    }
+    /// find txs to commit, return TxEntry vector, total_size and total_cycles.
+    pub fn transactions_to_commit(
```
doitian: This function is still too complex, but I'm OK to approve this PR, since now it is in a single module and is easy to refactor later.

miner/src/block_assembler.rs
```diff
@@ -413,6 +416,16 @@ impl BlockAssembler {
...
     }
 
+    fn package_txs(
+        &self,
+        chain_state: &ChainState,
+        size_limit: usize,
+        cycles_limit: Cycle,
+    ) -> (Vec<TxEntry>, usize, Cycle) {
+        let tx_pool = chain_state.tx_pool();
+        CommitTxsScanner::new(tx_pool.proposed()).transactions_to_commit(size_limit, cycles_limit)
```
zhangsoledad: Not require, just comment, maybe we should uniform `txs` `transactions`, it's weird seeing they appear both.

shared/src/tx_pool/types.rs
```diff
@@ -106,69 +109,465 @@ impl DefectEntry {
...
+    pub fee: Capacity,
+    pub vbytes: u64,
+    pub id: ProposalShortId,
+    pub ancestors_fee: Capacity,
+    pub ancestors_vbytes: u64,
+}
+
+impl AncestorsScoreSortKey {
+    /// compare tx fee rate with ancestors fee rate and return the min one
+    fn min_fee_and_vbytes(&self) -> (Capacity, u64) {
```
u2: return weight is enough?

miner/src/block_assembler.rs
jjyr: Yeah, forgot to rename the function.

shared/src/tx_pool/types.rs
jjyr: It's for avoiding div.
```
fee / vbytes <=> fee2 / vbytes2
=>
fee * vbytes2 <=> fee2 * vbytes
```

shared/src/tx_pool/types.rs
```diff
@@ -106,69 +109,465 @@ impl DefectEntry {
...
+
+#[derive(Debug, Clone)]
+pub(crate) struct TxEntriesPool {
+    entries: HashMap<ProposalShortId, TxEntry>,
+    sorted_index: BTreeSet<AncestorsScoreSortKey>,
+    /// A map track transaction ancestors and descendants
+    links: HashMap<ProposalShortId, TxLink>,
+}
+
+impl Default for TxEntriesPool {
```
u2: derive default

shared/src/tx_pool/types.rs
u2: I see.

shared/src/tx_pool/pending.rs
```diff
@@ -82,3 +95,145 @@ impl CellProvider for PendingQueue {
...
+                    .build()
+            }))
+            .outputs_data((0..outputs_len).map(|_| Bytes::new()))
+            .build()
+    }
+
+    const MOCK_CYCLES: Cycle = 5_000_000;
+    const MOCK_SIZE: usize = 200;
+
+    #[test]
```
u2: It's better to add some detail description for the test case.

doitian:
bors r=quake,doitian

PR #1336, opened 2019-08-04
perf: refactoring block body store

quake:
This PR splits block body (transactions) into small value store and use rocksdb prefix seek API to improve the DB fetch performance.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -13,7 +13,7 @@ use std::sync::Arc;
 //      - If the data can be migrated manually: update "x.y1.z" to "x.y2.0".
 //      - If the data can not be migrated: update "x1.y.z" to "x2.0.0".
 pub(crate) const VERSION_KEY: &str = "db-version";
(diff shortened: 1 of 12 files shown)


jjyr:
typo `pref` -> `perf`

quake:
> typo `pref` -> `perf`

ops, updated, thanks

db/src/db.rs
```diff
@@ -13,7 +13,7 @@ use std::sync::Arc;
 //      - If the data can be migrated manually: update "x.y1.z" to "x.y2.0".
 //      - If the data can not be migrated: update "x1.y.z" to "x2.0.0".
 pub(crate) const VERSION_KEY: &str = "db-version";
-pub(crate) const VERSION_VALUE: &str = "0.1600.0";
+pub(crate) const VERSION_VALUE: &str = "0.1700.0";
```
doitian: Please bump to 1900

quake: Updated, please review again, thanks.

PR #1340, opened 2019-08-05
chore: Set long poll interval for dummy miner

keroro520:
Long `poll_interval` can prevent dummy miner re-poll frequently, so that avoid generating uncle blocks.

### DIFF
--- a/resource/ckb-miner.toml
+++ b/resource/ckb-miner.toml
@@ -40,7 +40,9 @@ rpc_url = "http://127.0.0.1:8114/" # {{
 block_on_submit = true
 
 # block template polling interval in milliseconds
-poll_interval = 1000
+poll_interval = 1000 # {{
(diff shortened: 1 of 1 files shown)


resource/ckb-miner.toml
```diff
@@ -40,7 +40,9 @@ rpc_url = "http://127.0.0.1:8114/" # {{
 block_on_submit = true
 
 # block template polling interval in milliseconds
-poll_interval = 1000
+poll_interval = 1000 # {{
+# dev => poll_interval = 1_000_000_000_000
```
u2: it's 1_000_000_000 seconds. It's a too long time.

keroro520: Yes. It means "never passively poll".
Miner will actively pull block-template.

u2: get it.

PR #1341, opened 2019-08-05
feat: Implement cell's type ID special script

xxuejie:


script/src/verify.rs
```diff
@@ -262,8 +276,13 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
 
         // Now run each script group
         for group in self.lock_groups.values().chain(self.type_groups.values()) {
-            let program = self.extract_script(&group.script)?;
-            let cycle = self.run(&program, &group, max_cycles).map_err(|e| {
+            let result = if group.script.code_hash == TYPE_ID_CODE_HASH {
```
doitian: should we check the hash type as well?

script/src/verify.rs
```diff
@@ -283,6 +304,62 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
...
+        max_cycles: Cycle,
+    ) -> Result<Cycle, ScriptError> {
+        if max_cycles < TYPE_ID_CYCLES {
+            return Err(ScriptError::ExceededMaximumCycles);
+        }
+        // TYPE_ID script should only accept one argument,
+        // which is the hash of all inputs when creating
+        // the cell.
+        if script_group.script.args.len() != 1 || script_group.script.args[0].len() != 32 {
+            return Err(ScriptError::ValidationFailure(-1));
```
keroro520: 1. Can we put `verify_type_id` to an independent module, like `TypeIdVerifier`, so that makes it looks more like an independent script?
2. Declare the exit-codes as constant variables.

script/src/verify.rs
```diff
@@ -262,8 +276,15 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
 
         // Now run each script group
         for group in self.lock_groups.values().chain(self.type_groups.values()) {
-            let program = self.extract_script(&group.script)?;
-            let cycle = self.run(&program, &group, max_cycles).map_err(|e| {
+            let result = if group.script.code_hash == TYPE_ID_CODE_HASH
+                && group.script.hash_type == ScriptHashType::Data
```
doitian: Since the code may change, maybe Type is better.

script/src/verify.rs
xxuejie: Done

doitian:
bors retry

PR #1344, opened 2019-08-06
[HOLD] feat: create ckb-error

keroro520:
* Create ckb-error and use it as the global system error type

----

NOTE:

- I will remove the `NOTE: the original name is *` lines later. These lines are used to help review.


error/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["Nervos Core Dev <dev@nervos.org>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+failure = "0.1.5"
+numext-fixed-hash = { version = "0.1", features = ["support_rand", "support_heapsize", "support_serde"] }
+numext-fixed-uint = { version = "0.1", features = ["support_rand", "support_heapsize", "support_serde"] }
+ckb-vm = { version = "0.15.1", features = ["asm"] }
```
u2: remove the `ckb-vm` dependency?

keroro520: Thanks~

error/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+impl PartialEq for Error {
+    fn eq(&self, other: &Self) -> bool {
+        if self.kind() != other.kind() {
+            return false;
+        }
+
+        match self.kind() {
+            ErrorKind::OutPoint => {
```
u2: what about `self.inner.get_context() == other.inner.get_context()`?

miner/src/block_assembler.rs
```diff
@@ -452,7 +452,7 @@ impl BlockAssembler {
         let occupied_capacity = output.occupied_capacity(Capacity::bytes(data.len())?)?;
 
         if reward < occupied_capacity {
-            return Err(TransactionError::InsufficientCellCapacity.into());
+            Err(TransactionError::OccupiedOverflowCapacity)?;
```
u2: the change from `InsufficientCellCapacity` to `OccupiedOverflowCapacity` is intended?

script/src/verify.rs
```diff
@@ -226,39 +227,31 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
...
+    pub fn verify(&self, max_cycles: Cycle) -> Result<Cycle, Error> {
         let mut cycles: Cycle = 0;
-        // Check if all inputs are resolved correctly
-        if self
-            .resolved_inputs()
-            .iter()
-            .any(|input| input.cell.is_none())
-        {
-            return Err(ScriptError::NoScript);
-        }
```
u2: Check if all inputs are resolved correctly

shared/src/chain_state.rs
```diff
@@ -354,11 +341,11 @@ impl ChainState {
 
     // Add a verified tx into pool
     // this method will handle fork related verifications to make sure we are safe during a fork
-    pub fn add_tx_to_pool(&self, tx: Transaction, cycles: Cycle) -> Result<Cycle, PoolError> {
+    pub fn add_tx_to_pool(&self, tx: Transaction, cycles: Cycle) -> Result<Cycle, Error> {
         let short_id = tx.proposal_short_id();
         let tx_size = tx.serialized_size();
         if self.reach_tx_pool_limit(tx_size, cycles) {
-            return Err(PoolError::LimitReached);
+            Err(InternalError::FullTransactionPool)?;
```
u2: TransactionPoolFull?

shared/src/chain_state.rs
```diff
@@ -536,14 +532,16 @@ impl ChainState {
...
                     // OutOfOrder should only appear in BlockCellProvider
-                    UnresolvableError::Empty
-                    | UnresolvableError::UnspecifiedInputCell(_)
-                    | UnresolvableError::InvalidHeader(_)
-                    | UnresolvableError::OutOfOrder(_) => {
+                    OutPointError::MissingInputCellAndHeader
+                    | OutPointError::MissingInputCell(_)
+                    | OutPointError::ExclusiveInputCell(_)
+                    | OutPointError::UnknownHeader(_)
+                    | OutPointError::OutOfOrder(_) => {
```
u2: I am not sure this part, it looks like this PR includes more changes than it expected.

shared/tx-pool-executor/src/tx_pool_executor.rs
```diff
@@ -59,7 +56,7 @@ impl TxPoolExecutor {
         let (
             resolved_txs,
             cached_txs,
-            unresolvable_txs,
+            mut unresolvable_txs,
```
u2: A new name for it or we can remove it?  https://github.com/nervosnetwork/ckb/pull/1344/files#diff-8f96d3a4c4df5653ffc1b92dff8a14adR72

shared/tx-pool-executor/src/tx_pool_executor.rs
```diff
@@ -100,8 +94,8 @@ impl TxPoolExecutor {
 
         // immediately return if resolved_txs is empty
         if resolved_txs.is_empty() && cached_txs.is_empty() {
-            let (_tx, err) = unresolvable_txs.get(0).expect("unresolved tx exists");
-            return Err(err.to_owned());
+            let (_tx, err) = unresolvable_txs.remove(0);
```
u2: the old one is better.

error/src/lib.rs
keroro520: `self.inner.get_context()` returns `kind: ErrorKind` only.

When compare `Error`, we need to compare both kinds and the actual errors. So here we compare `self.kind() == other.kind()` , and then downcast and compare the actual errors.

verification/src/block_verifier.rs
```diff
@@ -133,15 +133,15 @@ impl MerkleRootVerifier {
 
     pub fn verify(&self, block: &Block) -> Result<(), Error> {
         if block.header().transactions_root() != &block.cal_transactions_root() {
-            return Err(Error::CommitTransactionsRoot);
+            Err(BlockError::UnmatchedCommittedRoot)?;
```
u2: TransactionRoot?

miner/src/block_assembler.rs
keroro520: I have renamed `InsufficientCellCapacity` -> `OccupiedOverflowCapacity `
https://github.com/nervosnetwork/ckb/pull/1344/files#diff-9f6a2acd3643aa4c24fdbf8922dbe532R6

Please tell me if you have better names or error descriptions!

script/src/verify.rs
keroro520: This part will be removed in PR https://github.com/nervosnetwork/ckb/pull/1331/ . So I remove it here. I think it is OK.

https://github.com/nervosnetwork/ckb/pull/1331/files#diff-99bdd708d07843aa77a343e5e906e76fL254

shared/src/chain_state.rs
keroro520: I personally like `adj + noun` format... So most of the error names are in this format.

u2: ok

shared/tx-pool-executor/src/tx_pool_executor.rs
keroro520: Sorry, I don't get your point. Why rename or remove it `unresolvable_txs`?

shared/tx-pool-executor/src/tx_pool_executor.rs
keroro520: `unresolvable_txs.get(0).expect("unresolved tx exists")`

return a reference `&Error`, and clone below.
But the new error type `ckb_error::Error` is not clonable. So I have to use `Vec::remove` to extract the error entity.

verification/src/block_verifier.rs
keroro520: I am not sure about it. But I think it would be better to specify which type of transactions in error messages, proposal or committed (`DuplicatedCommittedTransactions/DuplicatedProposalTransactions`, `UnmatchedCommittedRoot/UnmatchedProposalRoot`)

shared/tx-pool-executor/src/tx_pool_executor.rs
u2: sorry, my mistake, forget it.

shared/tx-pool-executor/src/tx_pool_executor.rs
u2: what about deriving clone trait for `Error`?

keroro520: We have to derive Clone manually.
Updated in commit https://github.com/nervosnetwork/ckb/pull/1344/commits/750b0eb03618c918d2eb4dbdaf710d9f9a5b65a6 , PTAL

error/src/transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// The format of `transaction.since` is invalid
+    // NOTE: the original name is InvalidSince
+    #[fail(display = "Invalid Since format")]
+    InvalidSinceFormat,
+
+    /// The transaction is not mature which is required by `transaction.since`
+    // NOTE: the original name is Immature
+    #[fail(display = "Not mature cause of since condition")]
+    NotMatureSince,
```
quake: I prefer to follow the `adjective + noun` naming rule:

```suggestion
    ImmatureTransaction,
```

error/src/out_point.rs
```diff
@@ -0,0 +1,79 @@
+use failure::Fail;
+use numext_fixed_hash::H256;
+use std::fmt::{Debug, Formatter, Result as FmtResult};
+
+// NOTE: To avoid circle dependency, we re-define another `ckb_core::OutPoint`,
```
doitian: I thought that in the discussion, `ckb-error` does not define any context. Each module will add the context instead.

error/src/lib.rs
doitian: Why do we need to compare Error? I think we only need to compare ErrorKind

doitian:
> `ckb-error` should only define the error and error kind, the module should define various context, and the `From` trait to convert from module errors to ckb-error.

ErrorKind should only carry the necessary fields as well when we really want to compare the Error or we want to set up the error message with the fields.

doitian:
If we want to categorize errors, I think we should use a nested enum instead. The context is reserved for modules to put their own error type.

```rust
pub enum ErrorKind {
    InternalError(InternalErrorKind),
}

pub enum InternalErrorKind {
    DatabaseError,
}
```

keroro520:
>> ckb-error should only define the error and error kind, the module should define various context, and the From trait to convert from module errors to ckb-error.

> ErrorKind should only carry the necessary fields as well when we really want to compare the Error or we want to set up the error message with the fields.

> If we want to categorize errors, I think we should use a nested enum instead. The context is reserved for modules to put their own error type.
> 
> ```rust
> pub enum ErrorKind {
>     InternalError(InternalErrorKind),
> }
> 
> pub enum InternalErrorKind {
>     DatabaseError,
> }
> ```

Thanks for your suggestions! I think I get your point and it may be a better style of managing errors.

HOLD this PR, I will make a quick PoC for the suggestions above:

  * `ckb-error` only defines ErrorKind(nested ErrorKind)
  * Normal modules define their own Error types
  * Use `From` trait to convert the Error types to ErrorKind in the form of context

keroro520:
Close.
I will open another PR according to https://github.com/nervosnetwork/ckb/pull/1344#issuecomment-520107924

PR #1348, opened 2019-08-07
feat: 10x faster miner

kilb:


miner/build.rs
```diff
@@ -0,0 +1,12 @@
+extern crate cc;
+
+fn main() {
+    cc::Build::new()
+        .file("src/worker/cuckoo.c")
+        .include("src/worker")
+        .flag("-O3")
+        .flag("-lcrypto")
```
doitian: 1. Have you tested in Windows?
2. Need to update docs about this new dependency.

doitian: It's OK to drop Windows support for several versions, since this is not the final PoW function we will use.

miner/src/worker/cuckoo.c
```diff
@@ -0,0 +1,117 @@
...
+v0 ^= (nonce); \
+v2 ^= 0xff; \
+sip_round(); sip_round(); sip_round(); sip_round(); \
+h = ((v0 ^ v1 ^ v2  ^ v3) & mask) << 1; 
+
+
+int c_solve(uint32_t* prof, const uint8_t* mesg, const uint64_t max_edge, const uint32_t cycle_len) {
+	uint64_t graph_size = max_edge << 1;
+	uint64_t mask = max_edge - 1;
+    int graph[MAX_GN];
```
doitian: The indentation is not consistent in this file.

doitian:
I have tested the miner locally. It looks good to me. The GPS is about 1000 per core on a cloud instance.

miner/src/worker/cuckoo_simple.rs
```diff
@@ -58,96 +63,33 @@ impl CuckooSimple {
...
-                    let seal = Seal::new(nonce, proof_u8.into());
-                    debug!(
-                        "send new found seal, pow_hash {:x}, seal {:?}",
-                        pow_hash, seal
-                    );
-                    if let Err(err) = self.seal_tx.send((pow_hash.clone(), seal)) {
-                        error!("seal_tx send error {:?}", err);
-                    }
-                    self.seal_candidates_found += 1;
+        let result = blake2b_256(pow_message(pow_hash, nonce).as_ref());
```
quake:
```suggestion
        let input = blake2b_256(pow_message(pow_hash, nonce).as_ref());
```

miner/build.rs
```diff
@@ -0,0 +1,11 @@
+extern crate cc;
```
quake: we are using rust 2018 edition, extern crate is no longer needed
```suggestion
use cc::Build;
```

miner/src/worker/cuckoo_simple.rs
```diff
@@ -58,96 +63,33 @@ impl CuckooSimple {
...
+        unsafe {
```
quake: why `unsafe` here?

miner/Cargo.toml
```diff
@@ -40,3 +41,6 @@ ckb-chain-spec = { path = "../spec" }
 ckb-db = { path = "../db" }
 ckb-pow = { path = "../pow" }
 ckb-dao-utils = { path = "../util/dao/utils" }
+
+[build-dependencies]
+cc = "1.0"
```
u2: newline

miner/src/worker/cuckoo_simple.rs
u2: c_solve

doitian:
bors retry

PR #1349, opened 2019-08-07
fix: compact block short_id collides

u2:
nodes should attempt to reconstruct the full block by taking the prefilledtxn transactions
    from the original CompactBlock message and placing them in the marked positions,
    then for each short transaction ID from the original compact_block message, in order,
    find the corresponding transaction either from the BlockTransactions message or
    from other sources and place it in the first available position in the block
    then once the block has been reconstructed, it shall be processed as normal,
    keeping in mind that short_ids are expected to occasionally collide,
    and that nodes must not be penalized for such collisions, wherever they appear.

### DIFF
--- a/sync/src/relayer/block_transactions_process.rs
+++ b/sync/src/relayer/block_transactions_process.rs
@@ -1,11 +1,23 @@
 use crate::relayer::block_transactions_verifier::BlockTransactionsVerifier;
-use crate::relayer::Relayer;
+use crate::relayer::error::{Error, Misbehavior};
+use crate::relayer::{ReconstructionError, Relayer};
 use ckb_network::{CKBProtocolContext, PeerIndex};
 use ckb_types::{core, packed, prelude::*};
 use failure::Error as FailureError;
 use std::collections::hash_map::Entry;
 use std::sync::Arc;
 
+// Keeping in mind that short_ids are expected to occasionally collide.
+// On receiving block-transactions message,
+// while the reconstructed the block has a different transactions_root,
+// 1. If the BlockTransactions includes all the transactions matched short_ids in the compact block,
+// In this situation, the peer sends all the transactions by either prefilled or block-transactions,
+// no one transaction from the tx-pool or store,
+// the node should ban the peer but not mark the block invalid
+// because of the block hash may be wrong.
+// 2. If not all the transactions comes from the peer,
+// there may be short_id collision in transaction pool.
+// the node retreat to request all the short_ids from the peer.
 pub struct BlockTransactionsProcess<'a> {
     message: packed::BlockTransactionsReader<'a>,
     relayer: &'a Relayer,
@@ -13,6 +25,7 @@ pub struct BlockTransactionsProcess<'a> {
     peer: PeerIndex,
 }
 
+#[derive(Debug, Eq, PartialEq)]
 pub enum Status {
     // BlockTransactionsVerifier has checked it,
     // so shoud not reach here unless the peer loses the transactions in tx_pool
@@ -23,6 +36,8 @@ pub enum Status {
     // The peer may lose it's cache
     // or other peer makes some mistakes
     UnkownRequest,
+    // Maybe short_id collides, re-send get_block_transactions message
+    CollisionAndSendMissingIndexes,
 }
 
 impl<'a> BlockTransactionsProcess<'a> {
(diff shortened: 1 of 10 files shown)


sync/src/relayer/block_transactions_process.rs
```diff
@@ -1,21 +1,33 @@
...
+use flatbuffers::FlatBufferBuilder;
 use numext_fixed_hash::H256;
 use std::collections::hash_map::Entry;
 use std::convert::TryInto;
 use std::sync::Arc;
 
+// Keeping in mind that short transaction IDs are expected to occasionally collide.
+// On receiving block-transactions message, if reconstruct the block with a different transactions_root,
+// - if all the short_ids are included,
+// the node should ban the peer but not mark the block invalid
```
doitian: I think the node should not ban the peer, unless the downloaded block is really invalid.

doitian: Do you mean "if all the transactions are prefilled," here like the comment below?

u2: If the BlockTransactions includes all the transactions matched `short_ids` in the compact block. 
I need to update the comment to make it clear.

u2: update the comment, pls review again, thanks.

shared/src/tx_pool/pool.rs
```diff
@@ -136,6 +136,21 @@ impl TxPool {
         self.orphan.add_tx(cycles, size, tx, unknowns.into_iter())
     }
 
+    #[cfg(feature = "test")]
```
doitian: I think we don't need these dirty methods to insert transactions with fake short id. There's an easy way to test collisions.

- Generate a transaction T, and add that transaction to the proposed pool, as usual.
- Generate a block, which includes the transaction. Change the merkle root to other value.
- Send the block as compact block, which does not prefill T.

The test should work because from the peer's perspective, it cannot tell the differences between a collision and a real unmatched merkle root.

sync/src/relayer/block_transactions_process.rs
```diff
@@ -59,24 +82,67 @@ impl<'a> BlockTransactionsProcess<'a> {
...
-                    self.relayer
-                        .accept_block(self.nc.as_ref(), self.peer, block);
-                    return Ok(Status::Accept);
+                match ret {
+                    Ok(block) => {
+                        if block.header().transactions_root()
+                            != compact_block.header.transactions_root()
+                        {
+                            if compact_block.short_ids.len() == txs_len {
+                                self.relayer.shared().insert_block_status(
```
doitian: We can never assert the block is invalid only when the merkle root does not match.

sync/src/relayer/types.rs
```diff
@@ -84,6 +84,17 @@ impl CompactBlock {
         }
         block_short_ids
     }
+
+    pub(crate) fn txs_len(&self) -> usize {
+        self.prefilled_transactions.len() + self.short_ids.len()
+    }
+
+    pub(crate) fn prefilleds_indexes(&self) -> HashSet<usize> {
```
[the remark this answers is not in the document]
u2: thanks.

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -47,6 +47,10 @@ impl<'a> TransactionHashesProcess<'a> {
                 .retain(|tx_hash| !tx_pool.contains_tx(&ProposalShortId::from_tx_hash(&tx_hash)))
         }
 
+        if transit_hashes.is_empty() {
```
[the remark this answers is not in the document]
u2: thanks, will create a new PR for this.

shared/src/tx_pool/pool.rs
u2: thanks.

sync/src/relayer/block_transactions_process.rs
u2: a mistake. the code is inconsistent with the comment.

u2:
updated, the changes https://github.com/nervosnetwork/ckb/compare/f17095459809050881e0dbef9bfbe64fef647a72..71cbbe1fb03953439f243ed52306e8c25e2908f1

u2:
@quake https://github.com/nervosnetwork/ckb/compare/71cbbe1fb03953439f243ed52306e8c25e2908f1..4398d51d7f9e08b7693656d6eae0bca2215a742b

u2:
Since now we could not forge `hash` in the `header`, so we could not use a fake `transaction_root` in the collision test. Instead of that, forge a fake transaction with the same `proposal_short_id` but different `hash`. 

Check the commit. https://github.com/nervosnetwork/ckb/pull/1349/commits/7bbe5497767e2eb6b965ba8146229194b26050b4

@quake, @doitian Please review it again, thanks.

u2:
bors r=quake

PR #1354, opened 2019-08-08
propose the tx structure changes for discussion

doitian:


quake:
```rust
pub struct Transaction {
    version: Version,
    cell_deps: Vec<CellDep>,
    header_deps: Vec<H256>,
    inputs: Vec<CellInput>,
    data: Vec<Bytes>,
    outputs: Vec<CellOutput>,
    witnesses: Vec<Vec<Bytes>>,
}

pub struct OutPoint {
    tx_hash: H256,
    index: u32,
}

pub struct CellDep {
    previous_output: OutPoint,
    is_group: bool,
}

pub enum ScriptHashType {
    Data = 0,
    Type = 1,
}

 pub struct Script {
    args: Vec<Bytes>,
    code_hash: H256,
    hash_type: ScriptHashType,
}

 pub struct CellInput {
    previous_output: OutPoint,
    lock_script: Script,
    since: u64,
}

pub struct CellOutput {
    capacity: Capacity,
    lock_hash: H256,
    type_script: Option<Script>,
}
```

u2:
Where is the output data?

doitian:
> CellProperties -> Kernel {capacity, lock, type}, where lock and type are both `Script`.

So we are not going to introduce the feature which reveals the lock script only when the cell is used as an input?

janx:
> > CellProperties -> Kernel {capacity, lock, type}, where lock and type are both `Script`.
> 
> So we are not going to introduce the feature which reveals the lock script only when the cell is used as an input?

If there's no objection, yes. According to our discussion in channel, it seems unnecessary to do that if we're going to move data back into cell.

janx:
Is it better to replace `previous_output` with just `outpoint`?

doitian:
Relates to #1356

doitian:
Final version:

```rust
pub struct Transaction {
    version: Version,

    cell_deps: Vec<CellDep>,
    header_deps: Vec<H256>,

    // Each input has a corresponding Vec<Bytes> in witnesses
    inputs: Vec<Input>,
    witnesses: Vec<Vec<Bytes>>,

    // Each cell has a corresponding output and data in outputs and outputs_data respectively.
    outputs: Vec<Output>,
    outputs_data: Vec<Bytes>,
}

pub struct CellDep {
    out_point: OutPoint,
    is_group: bool,
}

pub struct OutPoint {
    tx_hash: H256,
    index: u32,
}

pub enum ScriptHashType {
    Data = 0,
    Type = 1,
}

pub struct Script {
    args: Vec<Bytes>,
    code_hash: H256,
    hash_type: ScriptHashType,
}

pub struct Input {
    out_point: OutPoint,
    since: u64,
}

pub struct Output {
    capacity: Capacity,
    lock: Script,
    type_: Option<Script>,
}
```

Transaction

PR #1355, opened 2019-08-08
[ᚬmaster] Rc/v0.18

doitian:
This is the version going to be deployed on this Sat.

### Features

* #1351: Difficulty adjustment rfc version (@zhangsoledad)
* #1358: Add 10x faster miner (@kilb)

### Bug Fixings

* #1267: Header verifier with wrong header resolver (@u2)


doitian:
@keroro520 we need to run some acceptance tests

PR #1356, opened 2019-08-08
refactor: Add CellDep and change OutPoint data structure

TheWaWaR:
1. Add a new `CellDep` data structure, and changed `Transaction`:
``` rust
// For support dep group
pub struct CellDep {
    out_point: OutPoint,
    is_dep_group: bool,
}

// deps changed to cell_deps and header_deps
pub struct Transaction {
    // ...
    cell_deps: Vec<CellDep>,
    header_deps: Vec<H256>,
    // ...
}
```

2. `OutPoint` data structure renamed from origin `CellOutPoint`.
3. `HeaderProvider` changed to `HeaderChecker` because header now lazy loaded.

4. `SourceEntry` rename `Dep` to `CellDep` and add a new variant `CellHeader`, you can load `CellHeader` from `deps.headers` by index in VM.
``` rust
enum SourceEntry {
    Input,
    Output,
    // Cell dep
    CellDep,
    // Header dep
    CellHeader,
}
```


protos/schemas/common.fbs
```diff
@@ -48,14 +49,18 @@ table Witness {
     data:           [Bytes];
 }
 
+table CellDep {
+    tx_hash:        Bytes32;
+    index:          uint32;
+    is_dep_group:   uint8;
+}
+
 table OutPoint {
```
quake: `OutPoint` is useless after we changed deps struct, please remove it from `common.fbs` and `protocol.fbs`

TheWaWaR: Oops, not realized that.

TheWaWaR: We can't remove it because it used in syscall to load out point (used by contract)

script/proptest-regressions/syscalls/mod.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
```
driftluo: remove  this test file

doitian:
The Dep in SourceEntry should rename to `DepCell` and Header should be named `DepHeader`

keroro520:
> Would be inconsistent with `cell_deps` and `header_deps`.

So ...... Why not rename to `dep_cells` and `dep_headers`?

TheWaWaR:
> So ...... Why not rename to `dep_cells` and `dep_headers`?

Because they are dependencies, and we just talked about it.

core/src/transaction.rs
```diff
@@ -46,34 +45,49 @@ impl AsRef<[u8]> for CellKey {
...
 }
 
-impl Default for CellOutPoint {
+impl Default for OutPoint {
     fn default() -> Self {
-        CellOutPoint {
+        OutPoint {
             tx_hash: H256::zero(),
-            index: u32::max_value(),
+            index: 0,
```
u2: `Self::null()`

indexer/src/types.rs
```diff
@@ -11,7 +11,7 @@ use std::convert::TryInto;
 pub struct LockHashIndex {
     pub lock_hash: H256,
     pub block_number: BlockNumber,
-    pub cell_out_point: CellOutPoint,
+    pub cell_out_point: OutPoint,
```
u2: `pub out_point: OutPoint`

protocol/src/protocol.fbs
```diff
@@ -76,14 +77,18 @@ table Witness {
     data:           [Bytes];
 }
 
+table CellDep {
+    tx_hash:        H256;
+    index:          uint32;
```
u2: out_point?

protocol/src/protocol.fbs
```diff
@@ -76,14 +77,18 @@ table Witness {
...
+    is_dep_group:   uint8;
```
u2: bool?

TheWaWaR: No bool in flatbuffers

protocol/src/protocol.fbs
TheWaWaR: Expanded, just like `CellInput`

script/src/syscalls/load_header.rs
```diff
@@ -2,48 +2,75 @@ use crate::syscalls::{
...
             Source::Transaction(SourceEntry::Input) => self
                 .resolved_inputs
                 .get(index)
                 .ok_or(INDEX_OUT_OF_BOUND)
-                .and_then(|r| r.header().ok_or(ITEM_MISSING)),
+                .and_then(|cell_meta| self.load_header(cell_meta).ok_or(ITEM_MISSING)),
             Source::Transaction(SourceEntry::Output) => Err(INDEX_OUT_OF_BOUND),
-            Source::Transaction(SourceEntry::Dep) => self
-                .resolved_deps
+            Source::Transaction(SourceEntry::CellDep) => self
```
u2: return INDEX_OUT_OF_BOUND directly.

indexer/src/types.rs
TheWaWaR: I think both make sense

core/src/transaction.rs
TheWaWaR: `null` is for `cellbase`, default is for test.

script/src/syscalls/load_header.rs
```diff
@@ -2,48 +2,75 @@ use crate::syscalls::{
...
+        if self.header_deps.iter().any(|hash| hash == block_hash) {
+            self.data_loader.get_header(block_hash)
+        } else {
+            None
+        }
+    }
+
+    fn fetch_header(&self, source: Source, index: usize) -> Result<Header, u8> {
         match source {
             Source::Transaction(SourceEntry::Input) => self
```
u2: return INDEX_OUT_OF_BOUND directly?

script/src/syscalls/load_header.rs
TheWaWaR: No, it's on purpose, otherwise `dao` contract will not work.

TheWaWaR: It means load the header where the out point located.

shared/src/cell_set.rs
```diff
@@ -108,65 +108,69 @@ impl CellSet {
...
-                                    true,
-                                )
-                            }
-                        });
-                        meta.unset_dead(cell_input.index as usize);
-                    }
+        for out_point in diff
+            .old_inputs
+            .iter()
+            .filter(|out_point| !out_point.is_null())
```
u2: why add this filter?

shared/src/cell_set.rs
```diff
@@ -108,65 +108,69 @@ impl CellSet {
...
-        for new_input in &diff.new_inputs {
-            if let Some(cell_input) = &new_input.cell {
-                if let Some(meta) = new.get_mut(&cell_input.tx_hash) {
-                    meta.set_dead(cell_input.index as usize);
-                    continue;
-                }
+        for out_point in diff
+            .new_inputs
+            .iter()
+            .filter(|out_point| !out_point.is_null())
```
u2: why add this filter?

shared/src/chain_state.rs
```diff
@@ -242,42 +242,40 @@ impl ChainState {
 
         let updated_old_inputs = old_inputs
             .into_iter()
+            .filter(|out_point| !out_point.is_null())
```
u2: why add this filter?

core/src/transaction.rs
u2: why change the index from max to 0?

TheWaWaR: 0 is a reasonable default value? And MAX is for `cellbase`

shared/src/cell_set.rs
TheWaWaR: Filter out cellbase.

shared/src/cell_set.rs
TheWaWaR: Filter out cellbase.

shared/src/chain_state.rs
TheWaWaR: Filter out cellbase.

doitian:
The PR description is out dated.

TheWaWaR:
> The PR description is out dated.

Updated

PR #1359, opened 2019-08-09
fix: Multiple type ID cell creation in single transaction

xxuejie:
While previous implementation allows cells to have different type IDs
when they are created in different transactions. One quirk of the
solution, is that all cells in a single transaction will have the same
type ID, since the type ID script arg only hashs all inputs in the
transaction. This commit fixes the problem by hashing the cell's own
index in outputs as well, this way different cells created together
can also have different type IDs.

Notice while we implement it in Rust, the logic here can also be
implemented via a script running in CKB VM

### DIFF
--- a/script/src/type_id.rs
+++ b/script/src/type_id.rs
@@ -48,38 +48,36 @@ impl<'a> TypeIdSystemScript<'a> {
 
         // If there's only one output cell with current
         // TYPE_ID script, we are creating such a cell,
-        // we also need to validate that the hash of all
-        // inputs match the first argument of the script.
+        // we also need to validate that the first argument matches
+        // the hash of following items concatenated:
+        // 1. Transaction hash of the first CellInput's OutPoint
+        // 2. Cell index of the first CellInput's OutPoint
+        // 3. Index of the first output cell in current script group.
         if self.script_group.input_indices.is_empty() {
             let mut blake2b = new_blake2b();
-            for input in self.rtx.transaction.inputs() {
-                // TODO: we use this weird way of hashing data to avoid
-                // dependency on flatbuffers for now. We should change
-                // this when we have a better serialization solution.
-                if let Some(cell) = &input.previous_output.cell {
-                    blake2b.update(b"cell");
-                    blake2b.update(cell.tx_hash.as_bytes());
-                    let mut buf = [0; 4];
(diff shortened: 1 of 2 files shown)


script/src/type_id.rs
```diff
@@ -72,6 +72,14 @@ impl<'a> TypeIdSystemScript<'a> {
                 LittleEndian::write_u64(&mut buf, input.since);
```
doitian: Indeed, we don't need `block_hash` and `since`, `outpoint` of inputs already provides the uniqueness. It is also enough to use the first input out point.

script/src/type_id.rs
```diff
@@ -72,6 +72,14 @@ impl<'a> TypeIdSystemScript<'a> {
...
                 blake2b.update(&buf[..]);
             }
+            // Hash output indices(actually there is only one index) for current
+            // script group as well, so we can generate cells with different types
+            // In one transaction
+            for index in &self.script_group.output_indices {
```
doitian: Use only the first index is enough.

So the digest can be simplified into `[first_input.previous_output.tx_hash, first_input.previous_output.index, first_output.index]`

xxuejie: Thanks the logic is updated now

xxuejie: Actually some more thought: can we just use `transaction hash` + `index of first output in the group`? It seems like that's also up to the task but more simpler without needing serializing outpoints

doitian: outpoint is required since outputs in the same transaction have the same previous transaction hash.

xxuejie: Yes but we can use transaction hash + first index of output in the group. This way outputs ini the same transaction will have different index of output, hence different hash. Transaction hash only replaces outpoint of first previous input. The benefit here is that it decouples the hash from the actual outpoint structure.

u2: It seems like `transaction hash + index of first output in the group` is enough.

xxuejie: Thinking again, transaction hash is not possible here. Since the type script included here will be included in transaction hash calculation, and to calculate type script argument, we will need transaction hash, another chicken-egg problem.

doitian:
bors r=doitian,u2

PR #1361, opened 2019-08-09
refactor: use `TransactionInfo` instead of `BlockInfo`

u2:
There is another struct TransactionInfo https://github.com/nervosnetwork/ckb/blob/develop/core/src/extras.rs#L19
It's similar to the BlockInfo, it seems that we can remove the BlockInfo.

`is_cellbase` should always depend on the position in the Block. https://github.com/nervosnetwork/ckb/blob/develop/store/src/store.rs#L195


core/src/cell.rs
```diff
@@ -46,8 +28,7 @@ pub struct CellMeta {
 pub struct CellMetaBuilder {
     cell_output: CellOutput,
     out_point: CellOutPoint,
-    block_info: Option<BlockInfo>,
-    cellbase: bool,
+    block_info: Option<TransactionInfo>,
```
quake: Should we rename the field name to `transaction_info` also?

u2: ok

PR #1366, opened 2019-08-09
[ᚬrc/v0.18] fix: orphan_rate_estimation panic at div zero

zhangsoledad:
backport  https://github.com/nervosnetwork/ckb/pull/1366


spec/src/consensus.rs
```diff
@@ -462,17 +462,17 @@ impl Consensus {
...
             } else {
-                let orphan_rate_estimation = RationalU256::one()
-                    / (((&last_orphan_rate + U256::one())
-                        * &epoch_duration_target_u256
-                        * &last_epoch_length_u256
-                        / (&last_orphan_rate * &last_epoch_duration * &next_epoch_length_u256))
-                        .saturating_sub_u256(U256::one()));
-
-                let denominator = if orphan_rate_estimation.is_zero() {
+                let orphan_rate_estimation_recip = ((&last_orphan_rate + U256::one())
```
doitian: Need a test case to cover it. But the test doesn't need to backport to 0.18, just add to the PR send to develop.

PR #1380, opened 2019-08-10
Twist transaction hashes log

u2:
1. return if no tx hashes missing in transaction_hashes_process
2. twist debug info


doitian:
bors r=driftluo,doitian

PR #1386, opened 2019-08-11
Create ckb-error and use it as the global system error type

keroro520:
* Create ckb-error and use it as the global system error type
* This PR is an update of #1344 based on https://github.com/nervosnetwork/ckb/pull/1344#issuecomment-520107924

* There are several error kinds, and the corresponding error types:

  * OutPoint: `util/types/src/core/error.rs`
  * Transaction: `verification/src/error.rs`
  * Script: `script/src/error.rs`
  * Header: `verification/src/error.rs`
  * Block: `verification/src/error.rs`
  * Internal: `error/src/internal.rs`
  * Dao: `util/dao/utils/src/error.rs`
  * Spec: `spec/src/error.rs`

This PR does not cover the error types in ckb-miner, ckb-rpc and ckb-sync. Perhaps do it later.

----

NOTE:

- I will remove the `NOTE: the original name is *` lines later. These lines are used to help review.


keroro520:
@u2 Thanks for your careful review. I have updated  according to the above suggestions, in the new commitment https://github.com/nervosnetwork/ckb/commit/029e9025779c86348ea890d522c26aeb1c3d9ed6

* Rename some fields
* Distinguish Pow and Epoch errors
* Explicitly list the malformed errors

verification/src/error.rs
```diff
@@ -96,119 +173,213 @@ pub enum CellbaseError {
...
         match self {
-            CapacityOverflow | DuplicateDeps | Empty | OutputsSumOverflow | InvalidScript
-            | ScriptFailure(_) | InvalidSignature | InvalidSince => true,
-            _ => false,
+            TransactionError::OutputOverflowCapacity
+            | TransactionError::DuplicatedDeps
+            | TransactionError::MissingInputsOrOutputs
+            | TransactionError::OccupiedOverflowCapacity
+            | TransactionError::InvalidSinceFormat
+            | TransactionError::TooLargeSize
```
u2: `OccupiedOverflowCapacity`, `UnmatchedOutputsDataLength`, `UnmatchedOutputsDataHashes` are `is_malformed_tx`, before they are not. But I think they should be.

But where is the `InvalidSignature` and `CapacityOverflow`?

error/src/internal.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Database exception
+    // NOTE: the original name is ckb_db::Error::DBError(String)
+    Database,
+
+    /// VM internal error
+    VM,
+
+    /// The transaction already exist in pool
+    DuplicatedPoolTransaction,
+
```
zhangsoledad: IMO, the subject comes before adjective will be better. such as 
CapacityArithmeticOverflow,
TransactionPoolFull,
PoolTransactionDuplicated...

script/src/error.rs
```diff
@@ -0,0 +1,28 @@
...
+#[derive(Fail, Debug, PartialEq, Eq, Clone)]
+pub enum ScriptError {
+    /// The field code_hash in script is invalid
+    #[fail(display = "InvalidCodeHash")]
+    InvalidCodeHash,
+
+    /// The script consumes too much cycles
+    // NOTE: the original name is ExceededMaximumCycles,
+    #[fail(display = "TooMuchCycles")]
+    TooMuchCycles,
```
zhangsoledad: `TooMuchCycles` is really a bad name,  How much is too much?
the original name is much better, why replace?

zhangsoledad:
I oppose all TooMuch TooMany-series rename

miner/src/block_assembler.rs
```diff
@@ -465,7 +465,7 @@ impl BlockAssembler {
         let occupied_capacity = output.occupied_capacity(Capacity::bytes(data.len())?)?;
 
         if reward < occupied_capacity {
-            return Err(TransactionError::InsufficientCellCapacity.into());
+            Err(TransactionError::OccupiedOverflowCapacity)?;
```
zhangsoledad: OccupiedOverflowCapacity is weird.

error/src/internal.rs
u2: We have discussed in the preview PR, but no conclusion.

keroro520:
@zhangsoledad @u2 @quake 

I would suggest changing the error names and descriptions back to the original. And then open another issue to discuss it. What are your opinions?

verification/src/error.rs
keroro520: `InvalidSignature` has been never used.

`CapacityOverflow` added.

u2:
> @zhangsoledad @u2 @quake
> 
> I would suggest changing the error names and descriptions back to the original. And then open another issue to discuss it. What are your opinions?

Agree. You can create some small PRs for it, it's better for review also.

keroro520:
> > @zhangsoledad @u2 @quake
> > I would suggest changing the error names and descriptions back to the original. And then open another issue to discuss it. What are your opinions?
> 
> Agree. You can create some small PRs for it, it's better for review also.

Updated: Change the error names back to the original.

Please help review again. Thanks~

zhangsoledad:
> @zhangsoledad @u2 @quake
> 
> I would suggest changing the error names and descriptions back to the original. And then open another issue to discuss it. What are your opinions?

yeah, that's good!

sync/src/types.rs
```diff
@@ -1208,12 +1207,12 @@ impl SyncSharedState {
...
+            self.peers()
+                .set_last_common_header(peer, block.header().clone());
         }
 
-        self.remove_header_view(&block.as_ref().hash());
-        self.peers()
-            .set_last_common_header(peer, block.header().clone());
-        ret
+        Ok(ret?)
     }
```
zhangsoledad: Oops :eyes:

keroro520: I write `Ok(ret?)` just for convenience to convert `Result<(), ckb_error::Error>` to `Result<(), FailureError>`.
If this is a bad style, I would like to change like before, maybe more explicitly.

verification/src/tests/uncle_verifier.rs
```diff
@@ -628,6 +613,6 @@ fn test_uncle_with_uncle_descendant() {
         let dummy_context = dummy_context(&shared);
         let uncle_verifier_context = UncleVerifierContext::new(&dummy_context, &epoch18);
         let verifier = UnclesVerifier::new(uncle_verifier_context, &new_block);
-        assert_eq!(verifier.verify(), Ok(()));
+        assert!(verifier.verify().is_ok());
```
zhangsoledad: I think `assert_eq!(verifier.verify(), Ok(()))` is more friendly, 
when assert failed, will got 
left: {error_kind} right: `Ok(())`

But `assert!(verifier.verify().is_ok())` only got 
left: false right: ture

Just comment here, not required, we shall improve it later for test friendly, someday.
`assert_eq!(ret, Ok(()), "testing sth got unexpected error {}", ret)`

sync/src/types.rs
zhangsoledad: I don't mean that. the original code makes me little confused.

```
if is_err {
   return 

} else {
   ~~~~;
}
~~~~;
~~~~;
```

zhangsoledad: I don't mean that. the original code makes me little confused.

```
if is_err {
   return err;
} else {
  ...
}
...
why else
```

shared/src/tx_pool_ext.rs
```diff
@@ -212,57 +208,71 @@ impl TxPool {
...
+                            // just be discarded.
+                            //
+                            // To avoid mis-discarding error types added in the future, please don't
+                            // use `Some(_)` as the match arm.
+                            //
+                            // OutOfOrder should only appear in BlockCellProvider
+                            OutPointError::InvalidHeaderDep(_)
+                            | OutPointError::InvalidDepGroup(_)
+                            | OutPointError::Empty(_)
+                            | OutPointError::UnspecifiedInputCell(_)
```
u2: When does it raise `UnspecifiedInputCell` and `UnknownHeader`?

keroro520: Oh, I didn't notice that they were removed from https://github.com/nervosnetwork/ckb/pull/1356

Thanks! I will remove the two errors.

verification/src/tests/uncle_verifier.rs
keroro520: Yes, I agree.

I just couldn't find a better way to fight against `rustc` about the compiling errors 😞 :

```
error[E0369]: binary operation `==` cannot be applied to type `std::result::Result<core::cell::ResolvedTransaction<'_>, ckb_error::Error>`
   --> util/types/src/core/cell.rs:707:9
    |
707 |         assert_eq!(result, Ok(()));
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |         |
    |         std::result::Result<core::cell::ResolvedTransaction<'_>, ckb_error::Error>
    |         std::result::Result<(), _>
    |
    = note: an implementation of `std::cmp::PartialEq` might be missing for `std::result::Result<core::cell::ResolvedTransaction<'_>, ckb_error::Error>`
    = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error: aborting due to previous error
```

u2:
bors r=quake,u2,zhangsoledad

u2:
bors r=quake,u2,zhangsoledad

zhangsoledad:
bors r=quake,u2,zhangsoledad

PR #1395, opened 2019-08-13
[HOLD] feat: min transaction fee fitler

jjyr:
This PR enable min fee rate feature to ckb.

This PR adds a `tx_pool.min_fee_rate` option in `ckb.toml`, the default value is `1_000` shannons for per kilobytes, txs that with a fee rate lower than this value will be ignored by the node.

The intention is two:
1. allow miners control which txs be mined by set tx's fee rate
2. to prevent zero-fee or extremely low fee spam txs occupy the network

However it's not a part of the consensus, a node can set the `tx_pool.min_fee_rate` to a lower value or `0`.

Changes:

* Cache tx fee in tx_verify_cache
* Add a new option to `ckb.toml`: tx_pool.min_fee_rate = 1_000 # shannons per kilobyte
* Return an error in RPC `send_transaction` when tx fee rate is lower than `min_fee_rate`
* txs propose && commit will ignores txs that fee rate lower than `min_fee_rate`
* txs will not relay by node if it's fee rate is lower than `min_fee_rate`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,13 +422,13 @@ dependencies = [
  "ckb-store 0.19.0-pre",
  "ckb-test-chain-utils 0.19.0-pre",
  "ckb-traits 0.19.0-pre",
+ "ckb-tx-cache 0.19.0-pre",
  "ckb-types 0.19.0-pre",
  "ckb-verification 0.19.0-pre",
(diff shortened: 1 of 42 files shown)


jjyr:
Send a new PR https://github.com/nervosnetwork/ckb/pull/1645

PR #1399, opened 2019-08-13
fix: check sync_started before handle InIBD

keroro520:
* fix: `ibd_process` will improperly substrate `n_sync_started` without checking whether the peer is `sync_started == true`, and then cause panic when disconnects peers.
* feat: Add `suspend_sync` function to help suspend the synchronization to the specified peers.
* feat: Assert `sync_started/n_protected_outbound_peers` overflow.


keroro520:
> Please review whether this PR has to backport.

No backport.

PR #1410, opened 2019-08-14
chore: remove unnecessary lock

driftluo:
1. Remove unnecessary lock
2. The discovery test of the occasional fails due to the timeout caused by resource competition, the test that can be passed on a more long time or run it on single


doitian:
Approvals threshold is set to 6 temporarily to prevent merging PRs before the serialization feature is merged.

PR #1413, opened 2019-08-14
fix: use remote peer observed address

jjyr:
Fix broken observe address feature.

Observe address is a feature to leverage remote peers to report external IPs for a node.

In p2p network, when a node establishes outbound connections, the remote peers can observe the external IP of the node, remote peers should report the external IP through identify protocol, to help the node find out it's IP, it's kinda like the DDNS.

We have this feature built-in the original design, see the [identity protocol](https://github.com/nervosnetwork/p2p/blob/master/protocols/identify/src/protocol.fbs#L11) for details.

However, after a few development cycles, the observe address feature is broken.

In the current version, a node will not discovery external IPs through identify protocol. call `local_node_info` [rpc](https://github.com/nervosnetwork/ckb/tree/develop/rpc#local_node_info) will only show listened `addresses` that from config file.

After apply this fix, the `local_node_info` will show additional new discovered external IPs.

Changes:

* Fix identify, skip an observed address if it's an unreachable IP.
* Save observed address in a queue `pending_observed_addrs`.
* Dial observed address(dial to self node) repeatedly in `outbound_service`.
* If dialing to self is succeeded, the node saves the address as listened addr or increases its score if already exists, the address will show in `local_node_info` RPC.

Additional: because of p2p library hard coded to filtered out local addresses for us, we can't write an integration test for now, I'd happy to add test after change the p2p library behavior.

### DIFF
--- a/network/src/errors.rs
+++ b/network/src/errors.rs
@@ -12,6 +12,8 @@ pub enum Error {
     Io(IoError),
     P2P(P2PError),
     DB(DBError),
+    Addr(String),
+    Dial(String),
     Shutdown,
 }
 
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -76,13 +76,16 @@ pub struct SessionInfo {
 pub struct NetworkState {
     pub(crate) peer_registry: RwLock<PeerRegistry>,
     pub(crate) peer_store: Mutex<Box<dyn PeerStore>>,
-    pub(crate) original_listened_addrs: RwLock<Vec<Multiaddr>>,
+    /// Node listened addresses
+    pub(crate) listened_addrs: RwLock<Vec<Multiaddr>>,
     dialing_addrs: RwLock<HashMap<PeerId, Instant>>,
 
     pub(crate) protocol_ids: RwLock<HashSet<ProtocolId>>,
-    listened_addrs: RwLock<HashMap<Multiaddr, u8>>,
+    /// Node public addresses,
+    /// includes manualy public addrs and remote peer observed addrs
+    public_addrs: RwLock<HashMap<Multiaddr, u8>>,
     pending_observed_addrs: RwLock<HashSet<Multiaddr>>,
-    // Send disconnect message but not disconnected yet
+    /// Send disconnect message but not disconnected yet
     disconnecting_sessions: RwLock<HashSet<SessionId>>,
     local_private_key: secio::SecioKeyPair,
     local_peer_id: PeerId,
@@ -94,7 +97,7 @@ impl NetworkState {
         config.create_dir_if_not_exists()?;
         let local_private_key = config.fetch_private_key()?;
         // set max score to public addresses
-        let listened_addrs: HashMap<Multiaddr, u8> = config
+        let public_addrs: HashMap<Multiaddr, u8> = config
             .listen_addresses
             .iter()
             .chain(config.public_addresses.iter())
@@ -127,8 +130,8 @@ impl NetworkState {
             config,
             peer_registry: RwLock::new(peer_registry),
             dialing_addrs: RwLock::new(HashMap::default()),
-            listened_addrs: RwLock::new(listened_addrs),
-            original_listened_addrs: RwLock::new(Vec::new()),
+            public_addrs: RwLock::new(public_addrs),
+            listened_addrs: RwLock::new(Vec::new()),
             pending_observed_addrs: RwLock::new(HashSet::default()),
             disconnecting_sessions: RwLock::new(HashSet::default()),
             local_private_key: local_private_key.clone(),
@@ -284,8 +287,8 @@ impl NetworkState {
         self.local_private_key().to_peer_id().to_base58()
     }
 
-    pub(crate) fn listened_addrs(&self, count: usize) -> Vec<(Multiaddr, u8)> {
-        self.listened_addrs
+    pub(crate) fn public_addrs(&self, count: usize) -> Vec<(Multiaddr, u8)> {
+        self.public_addrs
             .read()
             .iter()
             .take(count)
@@ -294,8 +297,8 @@ impl NetworkState {
     }
 
     pub(crate) fn vote_listened_addr(&self, addr: Multiaddr, votes: u8) {
-        let mut listened_addrs = self.listened_addrs.write();
-        let score = listened_addrs.entry(addr).or_default();
+        let mut public_addrs = self.public_addrs.write();
+        let score = public_addrs.entry(addr).or_default();
         *score = score.saturating_add(votes);
     }
 
@@ -303,16 +306,12 @@ impl NetworkState {
         self.peer_registry.read().connection_status()
     }
 
-    pub fn external_urls(&self, max_urls: usize) -> Vec<(String, u8)> {
-        let original_listened_addrs = self.original_listened_addrs.read();
-        self.listened_addrs(max_urls.saturating_sub(original_listened_addrs.len()))
+    pub fn public_urls(&self, max_urls: usize) -> Vec<(String, u8)> {
+        let listened_addrs = self.listened_addrs.read();
+        self.public_addrs(max_urls.saturating_sub(listened_addrs.len()))
             .into_iter()
-            .filter(|(addr, _)| !original_listened_addrs.contains(addr))
-            .chain(
-                original_listened_addrs
-                    .iter()
-                    .map(|addr| (addr.to_owned(), 1)),
-            )
+            .filter(|(addr, _)| !listened_addrs.contains(addr))
+            .chain(listened_addrs.iter().map(|addr| (addr.to_owned(), 1)))
             .map(|(addr, score)| (self.to_external_url(&addr), score))
             .collect()
     }
@@ -349,7 +348,7 @@ impl NetworkState {
             trace!("Do not dial self: {:?}, {}", peer_id, addr);
             return false;
         }
-        if self.listened_addrs.read().contains_key(&addr) {
+        if self.public_addrs.read().contains_key(&addr) {
             trace!(
                 "Do not dial listened address(self): {:?}, {}",
                 peer_id,
@@ -483,10 +482,10 @@ impl NetworkState {
     }
 
     pub fn add_observed_addrs(&self, iter: impl Iterator<Item = Multiaddr>) {
-        let mut listened_addrs = self.listened_addrs.write();
+        let mut public_addrs = self.public_addrs.write();
         let mut pending_observed_addrs = self.pending_observed_addrs.write();
         for addr in iter {
-            if let Some(score) = listened_addrs.get_mut(&addr) {
+            if let Some(score) = public_addrs.get_mut(&addr) {
                 *score = score.saturating_add(1);
                 trace!(
                     "increase score for exists observed addr: {:?} {}",
@@ -940,7 +939,7 @@ impl NetworkService {
                         self.network_state.to_external_url(&listen_address)
                     );
                     self.network_state
-                        .original_listened_addrs
+                        .listened_addrs
                         .write()
                         .push(listen_address.clone())
                 }
@@ -1049,8 +1048,8 @@ pub struct NetworkController {
 }
 
 impl NetworkController {
-    pub fn external_urls(&self, max_urls: usize) -> Vec<(String, u8)> {
-        self.network_state.external_urls(max_urls)
+    pub fn public_urls(&self, max_urls: usize) -> Vec<(String, u8)> {
+        self.network_state.public_urls(max_urls)
     }
 
     pub fn node_version(&self) -> &Version {
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -76,11 +76,12 @@ pub struct SessionInfo {
 pub struct NetworkState {
     pub(crate) peer_registry: RwLock<PeerRegistry>,
     pub(crate) peer_store: Mutex<Box<dyn PeerStore>>,
-    pub(crate) original_listened_addresses: RwLock<Vec<Multiaddr>>,
+    pub(crate) original_listened_addrs: RwLock<Vec<Multiaddr>>,
     dialing_addrs: RwLock<HashMap<PeerId, Instant>>,
 
     pub(crate) protocol_ids: RwLock<HashSet<ProtocolId>>,
-    listened_addresses: RwLock<HashMap<Multiaddr, u8>>,
+    listened_addrs: RwLock<HashMap<Multiaddr, u8>>,
+    pending_observed_addrs: RwLock<HashSet<Multiaddr>>,
     // Send disconnect message but not disconnected yet
     disconnecting_sessions: RwLock<HashSet<SessionId>>,
     local_private_key: secio::SecioKeyPair,
@@ -93,7 +94,7 @@ impl NetworkState {
         config.create_dir_if_not_exists()?;
         let local_private_key = config.fetch_private_key()?;
         // set max score to public addresses
-        let listened_addresses: HashMap<Multiaddr, u8> = config
+        let listened_addrs: HashMap<Multiaddr, u8> = config
             .listen_addresses
             .iter()
             .chain(config.public_addresses.iter())
@@ -126,8 +127,9 @@ impl NetworkState {
             config,
             peer_registry: RwLock::new(peer_registry),
             dialing_addrs: RwLock::new(HashMap::default()),
-            listened_addresses: RwLock::new(listened_addresses),
-            original_listened_addresses: RwLock::new(Vec::new()),
+            listened_addrs: RwLock::new(listened_addrs),
+            original_listened_addrs: RwLock::new(Vec::new()),
+            pending_observed_addrs: RwLock::new(HashSet::default()),
             disconnecting_sessions: RwLock::new(HashSet::default()),
             local_private_key: local_private_key.clone(),
             local_peer_id: local_private_key.to_public_key().peer_id(),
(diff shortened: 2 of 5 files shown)


network/src/network.rs
```diff
@@ -498,18 +547,15 @@ impl ServiceHandle for EventHandler {
...
                         .filter(|proto| match proto {
                             multiaddr::Protocol::P2p(_) => false,
                             _ => true,
                         })
                         .collect();
-                    self.network_state
-                        .listened_addresses
-                        .write()
-                        .insert(addr, std::u8::MAX);
+                    self.network_state.add_listened_addr(addr, 1);
```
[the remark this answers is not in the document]
jjyr: Because advise addr is scored by peers, each time the node receive a advise addr inc the addr score by 1.

network/src/services/outbound_peer.rs
```diff
@@ -106,15 +108,19 @@ impl Future for OutboundPeerService {
...
-                            // feeler peers
-                            self.dial_peers(true, FEELER_CONNECTION_COUNT);
+                        if !self.network_state.config.whitelist_only {
+                            if new_outbound > 0 {
+                                // dial peers
+                                self.dial_peers(false, new_outbound as u32);
+                            } else {
+                                // feeler peers
+                                self.dial_peers(true, FEELER_CONNECTION_COUNT);
+                            }
```
keroro520: Seems the previous code is buggy(didn't dial_whitelist when `whitelist_only == false`)?

driftluo: `try_dial_whitelist` in all time

keroro520: Oh, I see. jjyr has removed `self.try_dial_whitelist();` from `dial_peers` https://github.com/nervosnetwork/ckb/pull/1413/files#diff-bb4e2e506c663eed7307ac58d00d7071L62.

network/src/protocols/test.rs
```diff
@@ -212,19 +212,12 @@ fn net_service_start(name: String) -> Node {
     let peer_id = network_state.local_peer_id().clone();
 
     let mut listen_addr = p2p_service
-        .listen("/ip4/0.0.0.0/tcp/0".parse().unwrap())
+        .listen("/ip4/127.0.0.1/tcp/0".parse().unwrap())
```
driftluo: `127.0.0.1` will not trigger this discovery behavior, here must use `0.0.0.0`

https://github.com/nervosnetwork/p2p/blob/master/protocols/discovery/src/substream.rs#L381

jjyr: Unfortunately `0.0.0.0` is definitely also an unreachable "address".

Next step we need move `is_reachable` check from p2p lib to ckb, to allow us write tests.

driftluo: But `0.0.0.0` is not loopback，so it can pass the check，it will pass not loopback or can reachable

jjyr: I see, seems like a bug, why do we want to discovery a "not loopback and unreachable address"?

Also "0.0.0.0" is an "unusesal address", see https://en.wikipedia.org/wiki/0.0.0.0

In fact, even change listen_addr back to `0.0.0.0`, the IP of a peer the node read from socket is high probably still `127.0.0.1`

driftluo: Because the discovery of these two kind addresses is meaningless, it cannot be connected.

`0.0.0.0` is a special address, it will bind all the network cards of the device, and it is impossible to know exactly whether it is valid, so the verification can be let go.

`0.0.0.0`not equivalent to 127.0.0.1

driftluo: Maybe more reasonable conditions should be ` RawAddr::from(address).is_reachable() ||  !address.ip().is_loopback() `

driftluo: Maybe more reasonable conditions should be ` RawAddr::from(address).is_reachable() ||  !address.ip().is_loopback() `

jjyr: IMO the condition should be `RawAddr::from(address).is_reachable() && !address.ip().is_loopback()` or simply `RawAddr::from(address).is_reachable()`.

And this check should be removed from p2p library side, the p2p library can only focus on handle protocols, let ckb handle the logic.

driftluo: > And this check should be removed from p2p library side, the p2p library can only focus on handle protocols, let ckb handle the logic.

I think general inspections can be used in the library. 

> `RawAddr::from(address).is_reachable() && !address.ip().is_loopback()`

This will block `0.0.0.0`

jjyr: > > And this check should be removed from p2p library side, the p2p library can only focus on handle protocols, let ckb handle the logic.
> 
> I think general inspections can be used in the library.
> 
> > `RawAddr::from(address).is_reachable() && !address.ip().is_loopback()`
> 
> This will block `0.0.0.0`

You can't guarantee that other users of this library don't want to discovery internal IP addresses. And most important we can't write tests code to check discovery and identify ...

`RawAddr::from(address).is_reachable() && !address.ip().is_loopback()` is Ok.

network/src/network.rs
```diff
@@ -396,58 +407,96 @@ impl NetworkState {
...
+        let mut listened_addrs = self.listened_addrs.write();
+        let mut pending_advise_addrs = self.pending_advise_addrs.write();
+        for addr in iter {
+            if let Some(score) = listened_addrs.get_mut(&addr) {
+                *score = score.saturating_add(1);
+                trace!(
+                    "increase score for exists advise addr: {:?} {}",
+                    addr,
+                    score
+                );
```
TheWaWaR: Should remove address already in `pending_advise_addrs`

network/src/protocols/test.rs
driftluo: This is actually the built-in protocol behavior of discovery, which has two means:
1. Interchange stored discovered addresses
2. Timing interactively connecting addresses between nodes

This is the protocol behavior

network/src/network.rs
jjyr: There are exists verify little probabilities that when node receiving an advised addr, at the same time the node is dialing the same addr. I' ll handle this conflict in `ConnectToSelf` phase.

network/src/errors.rs
```diff
@@ -12,6 +12,7 @@ pub enum Error {
     Io(IoError),
     P2P(P2PError),
     DB(DBError),
+    Custom(String),
```
doitian: Can we give the error a name? `Custom` is generally used in a library that reserved for user-thrown errors, it makes no sense here.

network/src/network.rs
```diff
@@ -76,13 +76,14 @@ pub struct SessionInfo {
...
-    pub(crate) original_listened_addresses: RwLock<Vec<Multiaddr>>,
-    dialing_addrs: RwLock<FnvHashMap<PeerId, Instant>>,
+    pub(crate) original_listened_addrs: RwLock<Vec<Multiaddr>>,
+    dialing_addrs: RwLock<HashMap<PeerId, Instant>>,
 
-    pub(crate) protocol_ids: RwLock<FnvHashSet<ProtocolId>>,
-    listened_addresses: RwLock<FnvHashMap<Multiaddr, u8>>,
+    pub(crate) protocol_ids: RwLock<HashSet<ProtocolId>>,
+    listened_addrs: RwLock<HashMap<Multiaddr, u8>>,
+    pending_advise_addrs: RwLock<HashSet<Multiaddr>>,
```
doitian: What is an advise address? A weird term I have never seen before.

network/src/network.rs
```diff
@@ -76,13 +76,14 @@ pub struct SessionInfo {
 pub struct NetworkState {
     pub(crate) peer_registry: RwLock<PeerRegistry>,
     pub(crate) peer_store: Mutex<Box<dyn PeerStore>>,
-    pub(crate) original_listened_addresses: RwLock<Vec<Multiaddr>>,
-    dialing_addrs: RwLock<FnvHashMap<PeerId, Instant>>,
+    pub(crate) original_listened_addrs: RwLock<Vec<Multiaddr>>,
```
doitian: It wastes reviewers' time to mess up a PR with renaming refactoring.

jjyr:
> Please do not change the existing test behavior, write your own test based on the current fix.

The current test is based on a wrong behaviour, if you look carefully, the tests return two local IP addresses even we already add the `is_reachable` condition on the p2p side.

However, I'll leave this buggy test since you guys are happy with it.

network/src/network.rs
jjyr: I'll use term `observed address` - reference to external addresses that observed by remote peers.

network/src/network.rs
jjyr: Just `addresses` to `addrs`

jjyr:
The description is added.

doitian:
Can we let peers report each other's address in handshake?

driftluo:
> Can we let peers report each other's address in handshake?
we can't，because most nodes will listen on 0.0.0.0, it can't determine what address it is listening on.

jjyr:
> Can we let peers report each other's address in handshake?


I guess we can modify the p2p library to achieve this.

But in fact, we currently use `identify` as an extent version handshake, in the current implementation the ability flags and client versions are reported through `identify`, so exchange IP through `identify` is ok IMO.

network/src/peer_store/sqlite/peer_store.rs
```diff
@@ -370,8 +370,9 @@ impl PeerStore for SqlitePeerStore {
     }
 
     fn random_peers(&self, count: u32) -> Vec<PeerAddr> {
+        const EXTRA_PEERS: u32 = 2;
```
u2: What's the `EXTRA_PEERS`?

network/src/services/outbound_peer.rs
```diff
@@ -106,15 +109,19 @@ impl Future for OutboundPeerService {
...
                         }
+                        // keep whitelist peer on connected
+                        self.try_dial_whitelist();
+                        // try dial observedd addrs
```
u2: typo: observedd

network/src/peer_store/sqlite/peer_store.rs
jjyr: Removed, we do not actually need this now

doitian:
> > Can we let peers report each other's address in handshake?
> 
> we can't，because most nodes will listen on 0.0.0.0, it can't determine what address it is listening on.

I do not mean telling others what I'm listening on, but telling others what's my observed address of the peer on the other end of the connection.

driftluo:
> > > Can we let peers report each other's address in handshake?
> > 
> > 
> > we can't，because most nodes will listen on 0.0.0.0, it can't determine what address it is listening on.
> 
> I do not mean telling others what I'm listening on, but telling others what's my observed address of the peer on the other end of the connection.

identify do it on (handshake)[https://github.com/nervosnetwork/p2p/blob/master/protocols/identify/src/lib.rs#L251-L258], But I am confused about what this filter means. @TheWaWaR

jjyr:
> > > > Can we let peers report each other's address in handshake?
> > > 
> > > 
> > > we can't，because most nodes will listen on 0.0.0.0, it can't determine what address it is listening on.
> > 
> > 
> > I do not mean telling others what I'm listening on, but telling others what's my observed address of the peer on the other end of the connection.
> 
> identify do it on [handshake](https://github.com/nervosnetwork/p2p/blob/master/protocols/identify/src/lib.rs#L251-L258), But I am confused about what this filter means. @TheWaWaR

These lines remove `p2p` phase from multiaddr, to reduce the message size.

network/src/network.rs
```diff
@@ -282,26 +284,32 @@ impl NetworkState {
...
+    pub(crate) fn listened_addrs(&self, count: usize) -> Vec<(Multiaddr, u8)> {
+        self.listened_addrs
             .read()
             .iter()
             .take(count)
             .map(|(addr, score)| (addr.to_owned(), *score))
             .collect()
     }
 
+    pub(crate) fn score_listened_addr(&self, addr: Multiaddr, inc_score: u8) {
```
doitian: The function name `score_listened_addr` is confusing. I recommend `vote_listened_addr`, since votes are usually accumulated.

```suggestion
    pub(crate) fn vote_listened_addr(&self, addr: Multiaddr, votes: u8) {
```

network/src/network.rs
```diff
@@ -282,26 +284,32 @@ impl NetworkState {
...
+        let original_listened_addrs = self.original_listened_addrs.read();
+        self.listened_addrs(max_urls.saturating_sub(original_listened_addrs.len()))
             .into_iter()
-            .filter(|(addr, _)| !original_listened_addresses.contains(addr))
+            .filter(|(addr, _)| !original_listened_addrs.contains(addr))
             .chain(
-                original_listened_addresses
+                original_listened_addrs
                     .iter()
                     .map(|addr| (addr.to_owned(), 1)),
```
doitian: Create a constant for 1.

network/src/network.rs
```diff
@@ -125,13 +126,14 @@ impl NetworkState {
             peer_store,
             config,
             peer_registry: RwLock::new(peer_registry),
-            dialing_addrs: RwLock::new(FnvHashMap::default()),
-            listened_addresses: RwLock::new(listened_addresses),
-            original_listened_addresses: RwLock::new(Vec::new()),
-            disconnecting_sessions: RwLock::new(FnvHashSet::default()),
+            dialing_addrs: RwLock::new(HashMap::default()),
+            listened_addrs: RwLock::new(listened_addrs),
+            original_listened_addrs: RwLock::new(Vec::new()),
```
doitian: We don't need `original_listened_addrs` anymore. `NetworkService::start` can also use the vote interface and give the binded addr a very high score.

network/src/network.rs
```diff
@@ -282,26 +284,32 @@ impl NetworkState {
...
+    }
+
     pub(crate) fn connection_status(&self) -> ConnectionStatus {
         self.peer_registry.read().connection_status()
     }
 
     pub fn external_urls(&self, max_urls: usize) -> Vec<(String, u8)> {
-        let original_listened_addresses = self.original_listened_addresses.read();
-        self.listened_addresses(max_urls.saturating_sub(original_listened_addresses.len()))
+        let original_listened_addrs = self.original_listened_addrs.read();
```
doitian: We can remove `original_listened_addrs`.

network/src/network.rs
```diff
@@ -282,26 +284,32 @@ impl NetworkState {
...
+        self.listened_addrs(max_urls.saturating_sub(original_listened_addrs.len()))
             .into_iter()
-            .filter(|(addr, _)| !original_listened_addresses.contains(addr))
+            .filter(|(addr, _)| !original_listened_addrs.contains(addr))
```
doitian: It is slow when `original_listened_addrs` is large.

network/src/network.rs
```diff
@@ -331,12 +339,17 @@ impl NetworkState {
...
+        peer_id: &PeerId,
+        addr: &Multiaddr,
+        allow_dial_to_self: bool,
+    ) -> bool {
+        if !allow_dial_to_self && self.local_peer_id() == peer_id {
             trace!("Do not dial self: {:?}, {}", peer_id, addr);
             return false;
         }
-        if self.listened_addresses.read().contains_key(&addr) {
+        if self.listened_addrs.read().contains_key(&addr) {
```
doitian: Maybe it will have problems if we remove `original_listened_addrs` and just add the binded addr to `listened_addrs`.

network/src/network.rs
```diff
@@ -396,58 +409,96 @@ impl NetworkState {
...
+                addr,
+                DialProtocol::Single(IDENTIFY_PROTOCOL_ID.into()),
+                true,
+            ) {
+                debug!("try_dial_observed_addrs error {}", err);
+            }
+        }
+    }
+
+    pub fn new_observed_addrs(&self, iter: impl Iterator<Item = Multiaddr>) {
```
doitian: Prefer verbs for function names.

network/src/network.rs
jjyr: IMO `1` is a proper constant value here. Even we give a constant `ONE`, we should never give it value other than `1`, it is meaningless.

network/src/network.rs
jjyr: I don't think a user is possible to have hundreds to thoughts interfaces on a single machine. The `original listened addr` won't be large.

network/src/network.rs
jjyr: We use `u8` to record a score, if we remove `original_listened_addrs`, once node up long enough, we can't distinguish `original_listened_addrs` from `listened_addrs`.

IMO we should keep the distinguish, and keep the `original_listened_addrs`, it not slow down the code, and not confuse the code(if so we can chose a better name, the concept is simple).

In the future, we may introduce another vote or expired mechanism to `listened_addr` but not `original_listened_addrs`. Keep the simple concept will not complex our code IMO.

jjyr: Overload a single sctructure brings more confuse and complexity IMO.

doitian: I think it's better to rename these fields:

- `listened_addrs` to `observed_addrs`, `public_addrs` or `external_addrs`. Also, we use external in code, public in config file, maybe it's better to keep consistent on the terminology.
- `original_listened_addrs` to just `listened_addrs`.

jjyr: I'd like do the following changes:

* `listened_addrs` to `external_addrs`.
* `original_listened_addrs` to `listened_addrs`

`public_addrs` is used in the case that ckb can't find external IP through protocols, for example, the node is behandling a NAT or a reverse proxy, user can manually set public IP.

* `listened_addrs` - user config listened addresses.
* `public_addrs` - user config public addrs.
* `external_addrs` - addresses that discovered through identify + `public_addrs`

doitian:
bors r=keroro520,doitian,u2,driftluo

PR #1414, opened 2019-08-15
chore: remove block_hash from get_live_cell params

shaojunda:


u2:
bors r=doitian,u2,keroro520

PR #1415, opened 2019-08-15
feat: Ignore genesis cellbase maturity check

TheWaWaR:


verification/src/transaction_verifier.rs
```diff
@@ -221,14 +221,15 @@ impl<'a> MaturityVerifier<'a> {
...
-                        .as_ref()
-                        .expect("cell meta should have block number when transaction verify")
-                        .block_number
-                        + self.cellbase_maturity
+            let cell_block_number = meta
+                .transaction_info
+                .as_ref()
+                .expect("cell meta should have block number when transaction verify")
+                .block_number;
+            // Ignore genesis cellbase maturity check
```
jjyr:
```suggestion
            // Ignore genesis cells maturity check
```

TheWaWaR: `maturity` is a `cellbase` only property, why `cells`?

jjyr: Yes, use `cellbase` here is also makes sense.

verification/src/transaction_verifier.rs
```diff
@@ -225,14 +225,15 @@ impl<'a> MaturityVerifier<'a> {
...
-                    < meta
-                        .transaction_info
-                        .as_ref()
-                        .expect("cell meta should have block number when transaction verify")
-                        .block_number
-                        + self.cellbase_maturity
+            let cell_block_number = meta
+                .transaction_info
+                .as_ref()
+                .expect("cell meta should have block number when transaction verify")
```
doitian: `TransactionInfo` is missing if the cell is from a transaction in the pool. I dislike the smell of the old code, we should not rely that transaction info must exist when `is_cellbase` returns true. Let's be more verbose that:

- Add a method `is_cellbase` in `TransactionInfo`.
- `CellMeta` can call `TransactionInfo::is_cellbase` in its `is_cellbase`
- Use `meta.transaction_info.map(|info| info.is_cellbase() && info.block_number > 0 && ...) = Some(true)` here.

PR #1416, opened 2019-08-15
chore: refactor peer_store lock scope in accept_peer

u2:


network/src/network.rs
```diff
@@ -220,20 +220,21 @@ impl NetworkState {
...
             )
-        };
+        });
+
         if accept_peer_result.is_ok() {
-            peer_store.update_status(&peer_id, Status::Connected);
+            self.peer_store
+                .lock()
+                .as_mut()
+                .update_status(&peer_id, Status::Connected);
```
zhangsoledad: I'm not sure, do we need consistency here? 
@jjyr

jjyr: `update_status` is a relatively very fast operation, IMO the code may be slower because the 3 times lock acquisition, I do not see any benefit we can get here.

jjyr: Oh I see, `accept_peer` is also very fast, I' think we do not need to worry here.

jjyr: Can we close this PR if not strong reasons to merge? @u2

u2: Just be more readable than before. If there are other reasons for the objection, it's no problem to close it.

jjyr: The one strong reason to do not to change here is @zhangsoledad mentioned, in the current implementation the inconsistent maybe not cause big problems, but the change put a potential bug in the code, future refactoring may change the inner behavior and reactivate the bug unintentionally.

And IMO the old code is clear enough.

u2: ok, I will close it.

PR #1418, opened 2019-08-16
test: add randomness for uncle test

u2:
Add randomness for uncle block generation, make it would reconcile the main randomly.


doitian:

PR #1420, opened 2019-08-16
[ᚬrc/v0.18] fix: failed to sync for long forks

keroro520:
* fix: Two long forks(> 128) cannot synchronize with each other


doitian:
If it also affects develop, you have send a PR to develop as well.

PR #1423, opened 2019-08-19
feat: use eaglesong as new pow

quake:


miner/src/miner.rs
```diff
@@ -74,65 +70,64 @@ impl Miner {
...
-                    .header(header)
-                    .build()
-                    .into_view();
-                let block_hash: H256 = block.hash().unpack();
-                if self.stderr_is_tty {
-                    debug!("Found! #{} {:#x}", block.number(), block_hash);
-                } else {
-                    info!("Found! #{} {:#x}", block.number(), block_hash);
-                }
+            self.notify_workers(WorkerMessage::Stop);
```
u2: I think it's an optimization.

miner/Cargo.toml
```diff
@@ -30,7 +29,7 @@ ckb-verification = { path = "../verification" }
 indicatif = "0.11"
 console = "0.7.5"
 ckb-dao = { path = "../util/dao" }
-ckb-hash = { path = "../util/hash"}
+eaglesong = { git = "https://github.com/rink1969/eaglesong", version = "0.1" }
```
yangby-cryptape: I asked @rink1969 if he could publish it, and he did. So, we can use it from [crates.io](https://crates.io/crates/eaglesong).

doitian: published

miner/src/miner.rs
```diff
@@ -23,9 +19,9 @@ pub struct Miner {
     pub works: LruCache<H256, Work>,
     pub worker_controllers: Vec<WorkerController>,
     pub work_rx: Receiver<Work>,
-    pub seal_rx: Receiver<(H256, Seal)>,
+    pub nonce_rx: Receiver<(H256, u64)>,
     pub pb: ProgressBar,
-    pub seals_found: u64,
+    pub nonces_found: u64,
```
doitian: Is `nonces_found` the same as blocks found? I'm not sure which we should display in the progress, blocks found or nonces tried.

quake: it's same as blocks found

miner/src/worker/mod.rs
```diff
@@ -90,11 +90,11 @@ pub fn start_worker(
                 panic!("incompatible pow engine and worker type");
             }
         }
-        WorkerConfig::CuckooSimple(config) => {
-            if let Some(cuckoo_engine) = pow.as_any().downcast_ref::<CuckooEngine>() {
+        WorkerConfig::EaglesongSimple(config) => {
+            if let Some(_eaglesong_engine) = pow.as_any().downcast_ref::<EaglesongPowEngine>() {
```
doitian: If we don't need to extract the inner item, just use `is_some()`.

shared/src/config.rs
```diff
@@ -0,0 +1,52 @@
+use numext_fixed_hash::H256;
+use numext_fixed_uint::U256;
+use ckb_core::block::Block;
+use ckb_core::header::{BlockNumber, Header, RawHeader};
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Config {
```
doitian: Any reason we add this new file?

quake: ops, it's a rebase error...

PR #1432, opened 2019-08-21
fix: return error if pending_tx failed

u2:


shared/src/chain_state.rs
```diff
@@ -390,13 +390,14 @@ impl ChainState {
                             return Err(e);
                         }
                         tx_pool.update_statics_for_add_tx(tx_size, cycles);
-                    } else if self
-                        .pending_tx(&mut tx_pool, Some(cycles), tx_size, tx)
-                        .is_ok()
-                    {
+                        return Ok(cycles);
```
jjyr: Can we use `else if let Ok(cycles) {....}` to eliminate `return`? it's more readability

u2: It's on purpose. If using `else if`, there will be more nesting conditions, lead the code to complicated. Return result for every branch.

u2:
bors r=zhangsoledad,jjyr

u2:
fix the different_txs_with_same_input integration test in https://github.com/nervosnetwork/ckb/pull/1432/commits/08a2b18d325621596033a53ca94e6a2b3f38c562

test/src/specs/tx_pool/different_txs_with_same_input.rs
```diff
@@ -19,15 +20,10 @@ impl Spec for DifferentTxsWithSameInput {
...
-        let output = tx2_temp
-            .outputs()
-            .as_reader()
-            .get(0)
-            .unwrap()
-            .to_entity()
-            .as_builder()
-            .capacity(capacity_bytes!(40_000).pack())
+        let output = CellOutputBuilder::default()
+            .capacity(capacity_bytes!(100).pack())
```
u2: Before the output capacity was too high, the transaction returned error while RPC sending.

test/src/specs/tx_pool/different_txs_with_same_input.rs
```diff
@@ -37,7 +33,7 @@ impl Spec for DifferentTxsWithSameInput {
             .send_transaction(tx1.clone().data().into());
         node0
             .rpc_client()
-            .send_transaction(tx1.clone().data().into());
+            .send_transaction(tx2.clone().data().into());
```
u2: I think it was a mistake before.

u2:
bors r=jjyr,keroro520,zhangsoledad

PR #1384, opened 2019-08-11
feat: chain snapshot

zhangsoledad:
Feature: 
Introduce chain snapshot, which leverage rocksdb `snapshot` and `hamt` to achieve captures point-in-time view of the chain,  get rid of `chain_state` and global lock. get 3x improve when switch fork.

* https://github.com/facebook/rocksdb/wiki/Snapshot
* https://docs.rs/im/13.0.0/im/hashmap/struct.HashMap.html

PR #1441, opened 2019-08-21
chore: use `std::{HashMap, HashSet}` instead of `fnv`

yangby-cryptape:


u2:
bors r-

It will conflict with #1384, need to rebase after #1384 merged.

u2:
bors r=u2,zhangsoledad,driftluo

PR #1443, opened 2019-08-21
chore: lock exact ckb-system-scripts version

doitian:
When `ckb-releases` is used external, it may uses a newer version of
ckb-system-scripts if the caret requirement is used in Cargo.toml, since the
version is only locked in the Cargo.lock in this repository.

This change ensures that when this is used as a crate externally, it always
use the exact version of ckb-system-scripts.


u2:
bors r=xxuejie,u2

PR #1452, opened 2019-08-22
chore: Expand macros in calc_hash

keroro520:
I would like the frequently used APIs can be auto-completed by IDE. So I expand the macro `impl_calc_special_hash_for_entity` in `calc_hash.rs`

### DIFF
--- a/util/types/src/extension/calc_hash.rs
+++ b/util/types/src/extension/calc_hash.rs
@@ -21,20 +21,6 @@ where
     }
 }
 
-/*
- * Calculate special hash for packed bytes wrappers.
- */
-
-macro_rules! impl_calc_special_hash_for_entity {
-    ($entity:ident, $func_name:ident) => {
-        impl packed::$entity {
-            pub fn $func_name(&self) -> H256 {
-                self.as_reader().$func_name()
-            }
-        }
-    };
-}
-
 impl packed::CellOutput {
     pub fn calc_data_hash(data: &[u8]) -> H256 {
         if data.is_empty() {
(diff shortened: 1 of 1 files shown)


yangby-cryptape:
I have read all methods in `ckb-core`.

Almost all `getter` method return the field like it's function name, **but some are not, they have different logic, not a simple getter as expected**.

I accept all opinions why we shouldn't use macros.

**But, can you promise that no one will make mistakes and no one will break the rules those have no lints, again?**
An editor feature and correctness, which is important?

If I expand all macros, it also makes my workload looks more heavy.
But I can not promise I will never make mistakes when I wrote so many code, so I use macros.

Maybe you should write a better editor plugin.

doitian:
It seems we need voting here.

doitian:
Please add an emoji reaction here if you support to expand the macros in this PR.

doitian:
Please add an emoji reaction here if you reject the changes in this PR.

keroro520:
> I have read all methods in `ckb-core`.
> 
> Almost all `getter` method return the field like it's function name, **but some are not, they have different logic, not a simple getter as expected**.
> 
> I accept all opinions why we shouldn't use macros.
> 
> **But, can you promise that no one will make mistakes and no one will break the rules those have no lints, again?**
> An editor feature and correctness, which is important?
> 
> If I expand all macros, it also makes my workload looks more heavy.
> But I can not promise I will never make mistakes when I wrote so many code, so I use macros.
> 
> Maybe you should write a better editor plugin.

1. This PR expands `impl_calc_special_hash_for_entity ` inside `calcl_hash.rs`. So I think to discuss the other macros, like getter/setter, is out of the scope of this PR.

2. The reason I open this PR is that it took me some time to figure out "how can I get the lock-hash of a lock-script" when I try to upgrade `ckb-bench`. Compare to the time I had spent 
 to expand `impl_calc_special_hash_for_entity`, the time I spent to look for [`Script::calc_script_hash`](https://github.com/nervosnetwork/ckb/blob/develop/util/types/src/extension/calc_hash.rs#L53) is longer.

3. Now I learn I should always open cargo doc during coding. So it is OK for me that keep this macro.

doitian:
We'll just merge it when it succeeds to collect two approvals.

keroro520:
Close since it is unnecessary.

PR #1453, opened 2019-08-22
feat: Allow chain specs config dep groups

TheWaWaR:
- Allow chain specs config dep groups in `spec.toml`
- `docs/hashes.toml` with more information (`tx_hash`, `dep_groups`)


ckb-bin/src/subcommand/cli/hashes.rs
```diff
@@ -90,10 +116,22 @@ pub fn hashes<'m>(root_dir: PathBuf, matches: &ArgMatches<'m>) -> Result<(), Exi
...
         config.chain.spec.absolutize(&root_dir);
         let chain_spec = ChainSpec::load_from(&config.chain.spec).map_err(to_config_error)?;
-        specs.insert(chain_spec.name.clone(), chain_spec.try_into()?);
+        let spec_name = chain_spec.name.clone();
+        let spec_hashes: SpecHashes = chain_spec.try_into()?;
+        specs.push((spec_name, spec_hashes));
     }
 
-    print!("{}", toml::to_string(&specs).unwrap());
+    println!();
```
u2: It's intended?

docs/hashes.toml
```diff
@@ -1,72 +1,120 @@
...
 [[ckb_dev.system_cells]]
 path = "Bundled(specs/cells/secp256k1_ripemd160_sha256_sighash_all)"
+tx_hash = "0x57d4f9e30b71e949a126c4b9094cfecdb95b28993f638f74a9aba705c8a33389"
 index = 4
 code_hash = "0x1acf88382192fbf76fc5a457a11cc393a90d84a46c5e0d7ceebfd1d0dec871e4"
 
-[ckb_staging]
-genesis = "0xa0e8f608ea2cc11ce4eb583ed7ac2228b9dfbe71eadd960085f6c27858e2643f"
-cellbase = "0x480637c97e3b43aceb4e4d60b7fad293f60b7938054d9697192524b4e158395a"
+[[ckb_dev.dep_groups]]
```
doitian: I think it is better to add the info which system cells are included in this group.

resource/specs/staging.toml
```diff
@@ -57,6 +57,17 @@ lock.code_hash = "0x54811ce986d5c3e57eaafab22cdd080e32209e39590e204a99b32935f835
 lock.args = ["0x57ccb07be6875f61d93636b0ee11b675494627d2"]
 lock.hash_type = "Data"
 
+# Dep group cells
```
doitian: Put it after `[genesis.system_cells.lock]`.

spec/src/lib.rs
```diff
@@ -282,40 +281,33 @@ impl Genesis {
...
-        let outputs = vec![cell_blake160, cell_ripemd160];
-        let outputs_data = vec![data_blake160.pack(), data_ripemd160.pack()];
+        let (outputs, outputs_data): (Vec<_>, Vec<_>) = self
+            .dep_groups
+            .values()
+            .map(|files| {
+                let out_points: Vec<_> = files
+                    .iter()
+                    .map(|res| {
+                        let data: Bytes = res.get()?.into_owned().into();
```
doitian: I think it is a bit overhead to read the data to just check what's the position of it in the cellbase. We also can search the res in the system cells files list. Yes, it means the user must reference the same resource via the same path, but it is acceptable.

ckb-bin/src/subcommand/cli/hashes.rs
TheWaWaR: Yes

docs/hashes.toml
TheWaWaR: ok

resource/specs/staging.toml
TheWaWaR: Fine with me. My original thought is dep groups are in 2nd transaction.

spec/src/lib.rs
TheWaWaR: Or, we can cache the hash result?
`HashMap<Resource, H256>`

TheWaWaR:
bors cancel

doitian:
Merged in https://github.com/nervosnetwork/ckb/commit/644fdaa1a268be262576a97d1b0917c437e95c82

PR #1454, opened 2019-08-23
chore: change is_dep_group(bool) to dep_type(enum)

yangby-cryptape:
- [x] chore: change is_dep_group(bool) to dep_type(enum)

- [x] chore: change all enum to `snake_case` in jsonrpc

  - hash_type

  - dep_type

- [x] chore: check value for all enum from network

  We can trust the serialized data from storage and the structured data from jsonrpc.
  But we should not trust the serialized data from network.

  The current serialization solution doesn't handle this, now.

### DIFF
--- a/util/types/src/extension/check_data.rs
+++ b/util/types/src/extension/check_data.rs
@@ -60,15 +60,9 @@ impl<'r> packed::RawTransactionReader<'r> {
     }
 }
 
-impl<'r> packed::SlimTransactionReader<'r> {
-    pub fn check_data(&self) -> bool {
-        self.raw().check_data()
-    }
-}
-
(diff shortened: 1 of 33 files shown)


util/types/src/extension/check_data.rs
```diff
@@ -0,0 +1,228 @@
+use crate::{packed, prelude::*};
+
+/*
+ * Blockchain
+ */
+
+impl<'r> packed::ScriptHashTypeReader<'r> {
+    pub fn check_data(&self) -> bool {
+        self.as_slice()[0] <= 1
```
TheWaWaR: Comment for this

PR #1455, opened 2019-08-23
fix: remove the block_ext cache

u2:
If we ran the test case `test_switch_valid_fork` separately as `cargo test test_switch_valid_fork`, the test case would fail. But when we ran `cargo test`, it passed, it was weird.

In https://github.com/nervosnetwork/ckb/blob/ba72c852d2e70176dfff0b2b3992be79dd59ef41/sync/src/tests/sync_shared_state.rs#L204-L259, there was only one store instance. After diving into it, the problem is that before we called the `get_block_ext` and cached the result, but after the fork chain became to the main chain, we called the `get_block_ext` again, it would return the stale data.

The `block_ext` is different from other data in DB, we would change it in DB, so the cache may be stale. I have not found a good way to update/clean the stale keys, so just remove it.

I think we had better not to set the `cache_enable` to `false` in any test case, this may hide some underlying bugs. I will create another PR for it.

This test case https://github.com/nervosnetwork/ckb/blob/ba72c852d2e70176dfff0b2b3992be79dd59ef41/sync/src/tests/sync_shared_state.rs#L56 is different from `test_switch_valid_fork`, but I will also try to remove the `ckb_store::set_cache_enable(false);`.

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -104,7 +104,6 @@ header_cache_size          = 4096
 cell_data_cache_size       = 128
 block_proposals_cache_size = 30
 block_tx_hashes_cache_size = 20
-block_ext_cache_size       = 20
 block_uncles_cache_size    = 10
 cellbase_cache_size        = 20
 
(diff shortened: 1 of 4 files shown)


u2:
bors r=zhangsoledad,TheWaWaR

PR #1463, opened 2019-08-24
fix: Change default lock script

TheWaWaR:


ckb-bin/src/subcommand/init.rs
```diff
@@ -90,22 +90,29 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
     }
 
     let runner = Runner::default().to_string();
-    let default_hash = format!("{:#x}", CODE_HASH_SECP256K1_BLAKE160_SIGHASH_ALL);
+    let default_code_hash =
+        ChainSpec::load_from(&Resource::bundled(format!("specs/{}.toml", args.chain)))
+            .expect("Load chain spec fialed")
```
jjyr:
```suggestion
            .expect("Load chain spec failed")
```

ckb-bin/src/subcommand/init.rs
```diff
@@ -90,22 +90,29 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
...
+            Some(&default_code_hash_string)
         } else {
             None
         }
     });
 
     let block_assembler = match block_assembler_code_hash {
         Some(hash) => {
-            if default_hash != *hash {
+            if default_code_hash_string != *hash {
```
jjyr: We should also check `hash_type`

PR #1464, opened 2019-08-24
feat: Use secp256k1 referenced by hash type "type" as the default lock.

doitian:
Co-authored-by: Quake <quake@nervos.org>
Co-authored-by: Qian Linfeng <thewawar@cryptape.com>


doitian:
bors retry

PR #1474, opened 2019-08-26
feat(test): tail failed logs

keroro520:
* feat(test): Tail nodes' logs when fails

  We treat tests are failed when

  - tests panicked
  - any thread in nodes has panicked(via searching nodes' logs)

* chore(test): Refactor Net and Node


u2:
bros r=doitian,u2

u2:
bors r=doitian,u2

PR #1475, opened 2019-08-26
fix: should update cached missing_indexes when collision

u2:
If transaction collision happens during `BlockTransactionProcess`, the node should update the cached `missing_index`.


u2:
bor r=quake,zhangsoledad

u2:
bors r=quake,zhangsoledad

PR #1476, opened 2019-08-26
feat(test): Deref rpc client to node

keroro520:
feat(test): Deref node to rpc_client

### DIFF
--- a/test/src/net.rs
+++ b/test/src/net.rs
@@ -125,7 +125,7 @@ impl Net {
             }
         }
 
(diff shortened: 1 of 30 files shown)


doitian:
I think such usage of `Deref` is an anti-pattern.

keroro520:
> I think such usage of `Deref` is an anti-pattern.

Thanks for your reply. I read several posts about `Deref anti-pattern` after reading your reply. And then I agree with it now.

Close it. // Although `Deref` make APIs more flexible :(

PR #1478, opened 2019-08-26
fix: uncle header shouldn't be deleted when block detach from mainchain

zhangsoledad:


chain/src/tests/find_fork.rs
```diff
@@ -306,3 +307,106 @@ fn test_find_fork_case4() {
         HashSet::from_iter(fork.attached_blocks.iter().cloned())
     );
 }
+
+// this case is create for issuse from https://github.com/nervosnetwork/ckb/pull/1470
+#[test]
+fn fork_case5() {
```
u2: the test name is vague.

PR #1479, opened 2019-08-26
refactor(test): Move out utils functions

keroro520:
refactor(test): Move out utils functions

### DIFF
--- a/test/src/net.rs
+++ b/test/src/net.rs
@@ -1,16 +1,11 @@
 use crate::specs::TestProtocol;
-use crate::utils::wait_until;
 use crate::Node;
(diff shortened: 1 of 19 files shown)


doitian:
I don't see many benefits to move out the helper functions.

keroro520:
> I don't see many benefits to move out the helper functions.

To make the test framework cleaner.
`Net` and `Node` is the core type for the test framework, I think we should keep them clean. So I move out the helper functions(most of them were added carelessly) into `utils.rs`.

u2:
I think in net/node or in helper are both ok.

PR #1480, opened 2019-08-26
test: forks contain same transactions

keroro520:


keroro520:

PR #1486, opened 2019-08-27
refactor: use Byte32 to replace the majority of H256

yangby-cryptape:
### What does this PR do?

- [x] Merge `ckb-merkle-tree` into `ckb-types` (cyclic dependencies).
- [x] A new and simple `ckb-fixed-hash` to replace `numext-fixed-hash`.
- [x] Use `Byte32` to replace the majority of `H256`.

### Why don't remove all `H256`?

- Cyclic dependencies

  For examples:

  - I merge `ckb-merkle-tree` into `ckb-types`, because it's only a simple wrapper.
  - But, `ckb-crypto` has a lot of functions, I think it's not proper to merge into `ckb-types`.

- Constants

  `bytes::Bytes` is dynamically, but we use a lot of hashes constants, it can speed up and it looks more clearly.

  For examples:

  ```rust
  const N: H256 = h256!("0xffffffff_ffffffff_ffffffff_fffffffe_baaedce6_af48a03b_bfd25e8c_d0364141");
  ```

  So, I would like to keep those `macros`.

- JSON-RPC Serialization

  We define a lot of types for rpc, even `bytes::Bytes` support `serde`, we still define `JsonBytes` for `bytes::Bytes`, so I think we can keep `H256` as JSON-RPC types, too.

- Simplify this PR


chain/src/tests/basic.rs
```diff
@@ -142,7 +142,7 @@ fn test_transaction_spend_in_same_block() {
         assert_eq!(
             shared
                 .snapshot()
-                .cell(&OutPoint::new(hash.unpack(), 0), false),
+                .cell(&OutPoint::new(hash.to_owned().to_owned(), 0), false),
```
u2: `hash.to_owned().to_owned()`

yangby-cryptape: > `hash.to_owned().to_owned()`

It's `&&Byte32`.

u2:
Bench needed?

yangby-cryptape:
> Bench needed?

I don't think so.

doitian:
Correct me if I'm wrong.

This PR does two things:

- Replace H256 with Byte32 in most places
- Add a simplified version of fixed-hash to replace numext-fixed-hash

This PR does not include an attached version of H256, if we need an H256 from Byte32, we still need to clone the bytes.

yangby-cryptape:
> Correct me if I'm wrong.
> 
> ... ...

Yes, all are right.

> Add a simplified version of fixed-hash to replace numext-fixed-hash

- Still need structs to use as hashes, not in `ckb-types`.
  Or move some small crate (`ckb-crypto`) into `ckb-types` (cyclic dependencies)?

- I can't add methods / serialization to primitive types, **New Types are required**.
  For example, there still need a new type for serde.

> an attached version of H256

I don't get this. Maybe we can treat `Byte32Reader` as the attached version of `H256`?

doitian:
bors r=u2,doitian

doitian:
I think ckb-crypto is over-designed, we only need little features from it.

PR #1487, opened 2019-08-27
fix: Fix dao statistics in genesis block

### LINKED ISSUE #1436: The DAO in genesis header is incorrect after adding the dep group transaction
https://github.com/nervosnetwork/ckb/pull/1427#pullrequestreview-277552297

TheWaWaR:
Fix dao statistics missing second transaction in genesis block.

Closes #1436

### DIFF
--- a/util/dao/utils/src/lib.rs
+++ b/util/dao/utils/src/lib.rs
@@ -1,10 +1,11 @@
 use byteorder::{ByteOrder, LittleEndian};
 use ckb_types::{
     core::{Capacity, TransactionView},
-    packed::Byte32,
+    packed::{Byte32, OutPoint},
     prelude::*,
 };
 use failure::{Error as FailureError, Fail};
+use std::collections::HashSet;
 
 // This is multiplied by 10**16 to make sure we have enough precision.
 pub const DEFAULT_ACCUMULATED_RATE: u64 = 10_000_000_000_000_000;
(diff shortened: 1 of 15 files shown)


keroro520:
Please check this test case:
https://github.com/keroro520/ckb/blob/fix-dao-statistics/test/src/specs/tx_pool/dao.rs#L272-L309

This case will fail even with this bugfix PR.

util/dao/utils/src/lib.rs
```diff
@@ -25,36 +25,57 @@ pub enum Error {
...
-            capacity.safe_add(cap)
-        })?;
-    let u = genesis_cellbase_tx
-        .data()
-        .raw()
-        .outputs()
+pub fn genesis_dao_data(
+    txs: Vec<&TransactionView>,
+    // All input occupied capacities in txs
+    inputs: Vec<Capacity>,
```
doitian: It is simple to figure out which outputs have been spent in the genesis block. We just need to index all the inputs in the transaction view first and use that to filter out the outputs, like this:

```
let dead_cells = HashSet::new();
for tx in txs {
  for i in tx.inputs {
      dead_cells.insert(i.out_point)
  }
}
for tx in txs {
  for (i, o) in tx.outputs.enumerate() {
    if !dead_cells.contains((tx, i)) {
      count_c_and_u(o);
    }
  }
}
```

keroro520:
> Please check this test case:
> https://github.com/keroro520/ckb/blob/fix-dao-statistics/test/src/specs/tx_pool/dao.rs#L272-L309
> 
> This case will fail even with this bugfix PR.

Ignore it. The old test code has a mistake. I have fixed the test and it passes now!

doitian:
No need to backport

util/dao/utils/src/lib.rs
```diff
@@ -25,36 +27,55 @@ pub enum Error {
...
+        .flat_map(|tx| tx.inputs().into_iter().map(|input| input.previous_output()))
+        .collect::<HashSet<_>>();
+    let statistics_outputs = |tx: &TransactionView| -> Result<_, FailureError> {
+        let c = tx.data().raw().outputs().into_iter().try_fold(
+            Capacity::zero(),
+            |capacity, output| {
+                let cap: Capacity = output.capacity().unpack();
+                capacity.safe_add(cap)
+            },
+        )?;
```
keroro520: `c` should substrate `output.capacity()` if the output cell was dead, just like `u`.

keroro520: I have written a test for it on my branch. 
Please let me know after fixing this problem, or if I am wrong.

test: https://github.com/keroro520/ckb/blob/7a198fc7729bac9181348da9851d70696890804a/test/src/specs/reward/primary_reward.rs#L108-L159

TheWaWaR: I filtered out dead cells below (by not add it).
``` rust
.filter(|(index, _)| {
    !dead_cells.contains(&OutPoint::new(tx_hash.clone(), *index as u32))
})
```

doitian: we have to filter out dead cells in both c and u.

TheWaWaR: > we have to filter out dead cells in both c and u.

In common blocks we count every cell to `c`.
So, filter out dead cells in `c` is a genesis specific logic?

keroro520: In common blocks, we should filter out dead cells as well.

`c` is the total issued capacity, is the sum capacity of ***all the live cells***.

doitian: > > we have to filter out dead cells in both c and u.
> 
> In common blocks we count every cell to `c`.
> So, filter out dead cells in `c` is a genesis specific logic?

C should only count the outputs in the cellbase. And u should count occupied capacity in all live cells.

doitian: > `c` is the total issued capacity, is the sum capacity of _**all the live cells**_.

Unless no shannons were burned.

doitian:
bors r=keroro520,doitian

doitian:
ckb-failed on spec check_typical_2_in_2_out_tx

PR #1489, opened 2019-08-27
test: test dep groups in test_bundled_specs

yangby-cryptape:


spec/src/lib.rs
```diff
@@ -500,6 +509,51 @@ pub mod test {
                 assert_eq!(index_minus_one + 1, cell.index, "{}", bundled_spec_err);
                 assert_eq!(cell.data_hash, data_hash, "{}", bundled_spec_err);
                 assert_eq!(cell.type_hash, type_hash, "{}", bundled_spec_err);
+                system_cells.insert(cell.index, cell.path.as_str());
+            }
+
+            let cell2 = block.transaction(1).unwrap();
```
doitian: Cell means transaction output. The second transaction can be named as `dep_group_transaction`

yangby-cryptape: I updated the name, and added more comments (I think the code is hard to understand, even for me).

Since this PR is too simple, I rebase to latest commit and force push.

doitian:
bors r=TheWaWaR,doitian

PR #1491, opened 2019-08-27
feat: satoshi's gift

jjyr:
This PR allows a special mark in genesis block's cellbase tx, for short we call it `genesis tx` in the following.

`genesis tx` can use a special mark in output data, to enable `virtual occupied` feature, which means we store a special mark to data instead of actually store X bytes data, ckb will explain the special mark and act like there are X bytes are occupied in a cell.

Intention:
A satoshi's gift cell will place in CKB mainnet genesis block, and 60% of the cell will be fit in zero bytes. This PR is intended to implement satoshi's cell without actually write so many bytes.

Details:
The format of the special mark: `"occupied" | to_le_bytes(X)`.
This mark will only work for genesis tx's cells, any other cells store this format will be considered as normal data without any special meaning.

Major changes:
1. implement `virtual occupied` mark for genesis tx.
2. call occupied_capacity on `CellMeta` will return virtual occupied for special marked cells.
    1. This affects syscalls with OccupedCapacity field, to keep `DAO` contract works.
    2. This affects `dao` field in the block header, `virtual occupied` will also be count in used capacity in `dao` header field.
3. Allow `data` in spec file `issued_cells` section.
4. Add a sample satoshi's gift cell in `testnet.toml`

Minor changes(refactoring):
1. Delete `CellMeta` from `ckb.mol`, since we do not store this structure for a long time, this structure is reintroduced by a mistake in new serialize PR.
2. Rename `data_bytes` to `data_size`.


util/types/src/core/cell.rs
```diff
@@ -15,8 +15,12 @@ pub struct CellMeta {
     pub cell_output: CellOutput,
     pub out_point: OutPoint,
     pub transaction_info: Option<TransactionInfo>,
-    pub data_bytes: u64,
     /// In memory cell data and its hash
+    pub data_size: u64,
+    /// Data occupied capacity
+    /// may not equals to actually data size since virtual occupied
+    pub data_occupied_capacity: Capacity,
```
u2: I think only one field `occupied_capacity` is enough, it's `virtual_occupied` or data's size.

doitian: Is `data_size` enough? What will happen if `data_size` does not equal to the real data size?

jjyr: IMO we should keep both fields.

* `data_size` must be equals to real data size. we use `data_size` to check the validity of a syscall before actually load the `data` [details](https://github.com/jjyr/ckb/blob/satoshi-gift/script/src/syscalls/load_cell_data.rs#L91).
* `occupied_capacity` is the fields leave to us to hack the `occupied`.

We use the distinguished fields, to break the hidden-convention of `data_size` to `occupied capacity`.

It's clear enough, what's benefit can we get for using only one "overload" "hidden-convention" field?

jjyr:
https://github.com/nervosnetwork/ckb/pull/1576

PR #1500, opened 2019-08-28
chore: update block capacity

jjyr:
Increase the capacity of a block.

A block can contain 3875 typical 2 in 2 out txs.

### DIFF
--- a/resource/specs/testnet.toml
+++ b/resource/specs/testnet.toml
@@ -72,13 +72,5 @@ lock.code_hash = "0x68d5438ac952d2f584abf879527946a537e82c7f3c1cbf6d8ebf9767437d
 lock.args = ["0x57ccb07be6875f61d93636b0ee11b675494627d2"]
 lock.hash_type = "type"
 
-[params]
-epoch_reward = 1_250_000_00000000
-secondary_epoch_reward = 600_000_00000000
-# Cellbase outputs are "locked" and require 4 * MAX_EPOCH_LENGTH(1800) confirmations(approximately 16 hours)
-# before they mature sufficiently to be spendable,
-# This is to reduce the risk of later txs being reversed if a chain reorganization occurs.
-cellbase_maturity = 7_200
-
 [pow]
 func = "Eaglesong"
--- a/resource/specs/testnet.toml
+++ b/resource/specs/testnet.toml
@@ -75,7 +75,6 @@ lock.hash_type = "type"
 [params]
 epoch_reward = 1_250_000_00000000
 secondary_epoch_reward = 600_000_00000000
-max_block_cycles = 10_000_000_000
 # Cellbase outputs are "locked" and require 4 * MAX_EPOCH_LENGTH(1800) confirmations(approximately 16 hours)
 # before they mature sufficiently to be spendable,
 # This is to reduce the risk of later txs being reversed if a chain reorganization occurs.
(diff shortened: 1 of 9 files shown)


resource/specs/testnet.toml
```diff
@@ -75,7 +75,6 @@ lock.hash_type = "type"
 [params]
 epoch_reward = 1_250_000_00000000
 secondary_epoch_reward = 600_000_00000000
-max_block_cycles = 10_000_000_000
```
doitian: Leave a comment here, that testnet uses the default `MAX_BLOCK_CYCLES` in spec/src/consensus.rs

spec/src/consensus.rs
```diff
@@ -46,8 +48,8 @@ pub(crate) const MIN_EPOCH_LENGTH: u64 = EPOCH_DURATION_TARGET / MAX_BLOCK_INTER
 // can divide DEFAULT_EPOCH_REWARD and can be divided by ORPHAN_RATE_TARGET_RECIP.
 pub(crate) const GENESIS_EPOCH_LENGTH: u64 = 1_000;
 
-pub(crate) const MAX_BLOCK_BYTES: u64 = 2_000_000; // 2mb
-pub(crate) const MAX_BLOCK_CYCLES: u64 = TWO_IN_TWO_OUT_CYCLES * 200 * 8;
+pub(crate) const MAX_BLOCK_BYTES: u64 = TWO_IN_TWO_OUT_BYTES * TWO_IN_TWO_OUT_COUNT;
+pub(crate) const MAX_BLOCK_CYCLES: u64 = TWO_IN_TWO_OUT_CYCLES * TWO_IN_TWO_OUT_COUNT;
```
doitian: Please go through these constants, make them private if possible.

spec/src/lib.rs
```diff
@@ -52,6 +52,7 @@ pub struct ChainSpec {
 pub struct Params {
     pub epoch_reward: Capacity,
     pub secondary_epoch_reward: Capacity,
+    #[serde(default)]
```
quake: Maybe not related to this PR, I think that we should move `default` to struct `Params` level, testnet/mainnet use the default value including `max_block_size`, other specs ( dev / integration) can use the different configurable values.

jjyr: I think maybe we should just remove `params` fields from `testnet.toml`, since actually we do not use this config. @doitian

jjyr: Apply the change, also remove the `params` from `testnet.toml`.

resource/specs/testnet.toml
jjyr: I think @quake 's suggestion is good, and we should remove the entire `params` section from `testnet.toml` until we actually need to config these fields with non-default values.

test/src/specs/tx_pool/send_secp_tx.rs
```diff
@@ -9,13 +9,16 @@ use ckb_resource::CODE_HASH_SECP256K1_BLAKE160_SIGHASH_ALL;
...
-    packed::{CellDep, CellInput, CellOutput, OutPoint, Script},
+    core::{capacity_bytes, Capacity, Cycle, DepType, ScriptHashType, TransactionBuilder},
+    packed::{CellDep, CellInput, CellOutput, OutPoint, Script, Witness},
     prelude::*,
     H256,
 };
 use log::info;
 
+const TX_2_IN_2_OUT_SIZE: usize = 673;
+const TX_2_IN_2_OUT_CYCLES: Cycle = 13_290_092;
```
TheWaWaR: Should mention where those values come from.

test/src/specs/tx_pool/send_secp_tx.rs
```diff
@@ -101,34 +104,166 @@ impl Spec for SendSecpTxUseDepGroup {
...
+
+impl Default for CheckTypical2In2OutTx {
+    fn default() -> Self {
+        Self::new(42)
+    }
+}
+
+impl CheckTypical2In2OutTx {
+    pub fn new(seed: u64) -> Self {
+        let mut generator = Generator::non_crypto_safe_prng(seed);
```
TheWaWaR: Why certainty is needed for private key?

PR #1502, opened 2019-08-28
Fix invalid usage of vector builder method.

doitian:
In block builder, `transactions` extends transactions. In the test we really
want to replace all the transactions.


yangby-cryptape:
bors r=u2,keroro520,yangby-cryptape

u2:
bors r-

Code owner review required

PR #1508, opened 2019-08-29
feat: Add load_header_by_field syscall for fetching epoch data

xxuejie:
Note this only adds epoch related data as header fields for now, later we will revise all syscalls to make sure the missing parts are fill in if needed(such as other header fields)

### DIFF
--- a/script/src/syscalls/load_header.rs
+++ b/script/src/syscalls/load_header.rs
@@ -1,6 +1,7 @@
 use crate::syscalls::{
-    utils::store_data, Source, SourceEntry, INDEX_OUT_OF_BOUND, ITEM_MISSING,
-    LOAD_HEADER_SYSCALL_NUMBER, SUCCESS,
+    utils::{store_data, store_u64},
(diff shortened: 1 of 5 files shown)


script/src/syscalls/load_header.rs
```diff
@@ -93,6 +94,38 @@ impl<'a, DL: DataLoader + 'a> LoadHeader<'a, DL> {
...
+
+    fn load_by_field<Mac: SupportMachine>(
+        &self,
+        machine: &mut Mac,
+        header: &HeaderView,
+    ) -> Result<(u8, u64), VMError> {
+        let field = HeaderField::parse_from_u64(machine.registers()[A5].to_u64())?;
+        let epoch = match self.data_loader.get_block_epoch(&header.hash()) {
+            Some(epoch) => epoch,
+            None => return Ok((ITEM_MISSING, 8)),
```
jjyr: Use `std::mem::size_of<u64>()`, it makes the code easier to read.

xxuejie: Ooops this is a mistake, on an item-missing case the second value should be 0, fixing now.

PR #1510, opened 2019-08-30
feat: Check system cells lock script when build genesis block

TheWaWaR:
Besides genesis message cell's lock script, all other cells' lock scripts must exists in genesis block.

### DIFF
--- a/spec/src/lib.rs
+++ b/spec/src/lib.rs
@@ -28,7 +28,7 @@ use ckb_types::{
     H256, U256,
 };
 use serde_derive::{Deserialize, Serialize};
-use std::collections::BTreeMap;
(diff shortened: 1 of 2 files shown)


spec/src/lib.rs
```diff
@@ -200,6 +200,54 @@ impl Genesis {
             .transaction(cellbase_transaction)
             .transaction(dep_group_transaction)
             .build();
+
+        // Check lock scripts
```
u2: It's better to put this part in a single function.

TheWaWaR: make sense

u2:
https://travis-ci.com/nervosnetwork/ckb/jobs/231168775#L981

The integration test failed

PR #1394, opened 2019-08-13
chroe(rpc): keep rpc doc endpoints order same as rpc.json

zjhmale:
Closes #1229

PR #1514, opened 2019-08-31
fix: Sort json rpc methods in README doc

TheWaWaR:


u2:
close it? https://github.com/nervosnetwork/ckb/pull/1394

PR #1515, opened 2019-09-01
refactor: change enum from `[byte; 1]` to `byte`

quake:
This PR changes enum types in ckb.mol (`ScriptHashType` and `DepType`) from `[byte; 1]` to `byte`, it simplifies builder API and avoids some unnecessary conversions.

### DIFF
--- a/benches/benches/benchmarks/util.rs
+++ b/benches/benches/benchmarks/util.rs
@@ -230,7 +230,7 @@ lazy_static! {
         let script = Script::new_builder()
             .code_hash(CellOutput::calc_data_hash(&data))
             .args(Bytes::from(PUBKEY_HASH.as_bytes()).pack())
-            .hash_type(ScriptHashType::Data.pack())
+            .hash_type(ScriptHashType::Data.into())
             .build();
 
         (cell, data, script)
--- a/script/src/verify.rs
+++ b/script/src/verify.rs
@@ -32,6 +32,7 @@ use ckb_vm::{
 };
 use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::convert::TryFrom;
 
 // A script group is defined as scripts that share the same hash.
 // A script group will only be executed once per transaction, the
@@ -250,7 +251,7 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
 
     // Extracts actual script binary either in dep cells.
     fn extract_script(&self, script: &'a Script) -> Result<Bytes, Error> {
-        match script.hash_type().unpack() {
+        match ScriptHashType::try_from(script.hash_type()).expect("checked data") {
             ScriptHashType::Data => {
                 if let Some(data) = self.binaries_by_data_hash.get(&script.code_hash()) {
                     Ok(data.to_owned())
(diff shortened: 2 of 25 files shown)


util/types/src/conversion/blockchain.rs
```diff
@@ -74,39 +74,6 @@ impl<'r> Unpack<[u8; 10]> for packed::ProposalShortIdReader<'r> {
...
-
-impl<'r> Unpack<core::DepType> for packed::DepTypeReader<'r> {
-    fn unpack(&self) -> core::DepType {
-        use std::convert::TryFrom;
-        core::DepType::try_from(self.as_slice()[0]).expect("internal error: fail to unpack DepType")
-    }
-}
-
-impl_conversion_for_entity_unpack!(core::DepType, DepType);
-
```
keroro520: It's better to keep the conversion even if change `[byte; 1]` to `byte`, to avoid verbosely converting manually like `core::ScriptHashType::try_from(input.hash_type()).expect("checked data").into()` above

benches/benches/benchmarks/util.rs
```diff
@@ -213,7 +213,7 @@ lazy_static! {
         let script = Script::new_builder()
             .code_hash(CellOutput::calc_data_hash(&data))
             .args(vec![Bytes::from(PUBKEY_HASH.as_bytes()).pack()].pack())
-            .hash_type(ScriptHashType::Data.pack())
+            .hash_type(ScriptHashType::Data as u8)
```
yangby-cryptape: Expose interfaces is better than expose inner structures.

We can put `as u8` in `pack()` and let `pack()` be `inline`.

TheWaWaR: Agreed, we can unify the interface.

quake: I prefer `std::convert::Into` to `Pack` trait,  updated code, please review again, thanks

script/src/verify.rs
```diff
@@ -272,7 +273,7 @@ impl<'a, DL: DataLoader> TransactionScriptsVerifier<'a, DL> {
         // Now run each script group
         for group in self.lock_groups.values().chain(self.type_groups.values()) {
             let result = if group.script.code_hash() == TYPE_ID_CODE_HASH.pack()
-                && group.script.hash_type().unpack() == ScriptHashType::Type
+                && group.script.hash_type() == Into::<u8>::into(ScriptHashType::Type)
```
u2: what about `ScriptHashType::Data.into()`?

quake: rust compiler is not smart enough for this operation, have to provide a specific type here.

benches/benches/benchmarks/util.rs
yangby-cryptape: 1. I can **NOT** change `Into` but I can change `Pack`.
2. When put `JsonScriptHashType` and `ScriptHashType` together, it's a bit difficult to distinguish between them (both have `Into`).
3. `hash_type` is `u8`, it's an implementation detail, it's an optimization.
   In a complete logic, it's should be `molecule::Byte(bytes::Bytes)` (`it.as_slice().len() == 1`).
   So, make it be special, when I update internal implementation, I have to update all things here.

PR #1517, opened 2019-09-02
Add stratum protocal documents

yulidai:


quake:
I think that you should open this PR in https://github.com/nervosnetwork/rfcs repo.

yulidai:
Yes, you are right

PR #1519, opened 2019-09-02
fix: indexer store should lock the state when syncing data

quake:
This PR fixed an accidental bug of incorrect indexer store state when syncing indexer state and deindex an index in the same time.

### DIFF
--- a/indexer/src/store.rs
+++ b/indexer/src/store.rs
@@ -12,6 +12,7 @@ use ckb_types::{
     packed::{self, Byte32, CellOutput, OutPoint},
     prelude::*,
 };
+use ckb_util::Mutex;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


indexer/src/store.rs
```diff
@@ -369,6 +374,7 @@ impl DefaultIndexerStore {
                 })
         });
 
+        drop(sync_lock);
```
driftluo: lock will automatically unlock after leaving the scope, the `drop` is not necessary

doitian:
bors r=u2,doitian

indexer/src/store.rs
quake: > lock will automatically unlock after leaving the scope, the `drop` is not necessary

I want to release lock asap, so calling the `drop` explicitly in all code blocks.

PR #1521, opened 2019-09-03
test: Node should fetch blocks from all peers with better chains

xxuejie:
Whether peer's chain is known best's ancestor is irrelevant here.

### DIFF
--- a/test/src/specs/sync/block_sync.rs
+++ b/test/src/specs/sync/block_sync.rs
@@ -284,6 +284,57 @@ impl Spec for BlockSyncOrphanBlocks {
     }
 }
 
+pub struct BlockSyncNonAncestorBestBlocks;
+
+impl Spec for BlockSyncNonAncestorBestBlocks {
+    crate::name!("block_sync_non_ancestor_best_blocks");
+
+    crate::setup!(
+        num_nodes: 2,
+        connect_all: false,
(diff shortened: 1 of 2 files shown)


test/src/specs/sync/block_sync.rs
```diff
@@ -284,6 +284,57 @@ impl Spec for BlockSyncOrphanBlocks {
...
+        // the same difficulty, but block a has a lower block hash. So
+        // later when we sync the header of block a with node0, node0's
+        // global shared best header will be updated, but the tip will stay
+        // unchanged. Then we can connect node0 with node1, node1 will provide
+        // a better chain that is not the known best's ancestor.
+        let (a, b) = (
+            node0.new_block(None, None, None),
+            node1.new_block(None, None, None),
+        );
+        let (a, b) = if a.hash() < b.hash() { (a, b) } else { (b, a) };
```
u2: It's better to use `is_better_than` for further changing. We may change to the first-received policy.

xxuejie: `is_better_than` is buried deep in the HeaderView struct defined in sync package. I'm not sure if it's good strategy to expose that. Besides, it might actually justify the purpose that when we change first-received policy, integration tests need to be revisited to make sure things are still correct.

PR #1525, opened 2019-09-03
chore: let sentry capture panics when failed to verify the data from the storage

yangby-cryptape:
This week, sentry reported few panics that the data from the storage can not deserialize.
After reviewing the code, I think it should be impossible.
I added the hex string of the bytes to sentry messages. So, next time, we can figure it out.

### DIFF
--- a/util/types/src/prelude.rs
+++ b/util/types/src/prelude.rs
@@ -1,4 +1,7 @@
-pub use molecule::prelude::{Builder, Entity, Reader};
+pub use molecule::{
+    faster_hex::hex_string,
+    prelude::{Builder, Entity, Reader},
+};
 
 // An alias for unwrap / expect.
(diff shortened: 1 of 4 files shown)


util/types/src/prelude.rs
```diff
@@ -15,7 +18,41 @@ impl<T> ShouldBeOk<T> for Option<T> {
...
+                                 reason: {}; \
+                                 data: 0x{}.",
+                                err,
+                                hex_string(slice).unwrap()
+                            ),
+                            Level::Info,
+                        )
+                    },
+                );
+                panic!("verify slice should be ok, but {}", err);
```
doitian: `panic!` is enough, we don't need to send the sentry message manually.

yangby-cryptape: DONE!

PR #1527, opened 2019-09-03
feat: Return cell data in get_live_cell rpc

TheWaWaR:
BREAKING CHANGE: `get_live_cell` add `with_data` argument and result structure changed.


rpc/src/module/chain.rs
```diff
@@ -44,7 +44,7 @@ pub trait ChainRpc {
     ) -> Result<Vec<CellOutputWithOutPoint>>;
 
     #[rpc(name = "get_live_cell")]
-    fn get_live_cell(&self, _out_point: OutPoint) -> Result<CellWithStatus>;
+    fn get_live_cell(&self, _out_point: OutPoint, _with_data: bool) -> Result<CellWithStatus>;
```
quake: change `with_data` to optional param:

```suggestion
    fn get_live_cell(&self, _out_point: OutPoint, _with_data: Option<bool>) -> Result<CellWithStatus>;
```

quake:
> BREAKING CHANGE: `get_live_cell` add `with_data` argument and result structure changed.

we can use the optional param, it will keep rpc compatibility.

rpc/json/rpc.json
```diff
@@ -242,24 +242,28 @@
         ]
     },
     {
-        "description": "Returns the information about a cell by out_point if it is live.",
+        "description": "Returns the information about a cell by out_point if it is live. If given second with_data argument, will return cell data and data_hash if it is live",
```
quake: where is the `data_hash` , I can't find it in result example.

TheWaWaR: The `data` field is optional.
``` rust
pub struct CellInfo {
    pub output: CellOutput,
    pub data: Option<(JsonBytes, H256)>,
}
```

rpc/src/module/chain.rs
TheWaWaR: I think it's better explicit here, if we add another more argument in the future, it will make user confused.

rpc/json/rpc.json
quake: I don't think that we need return datahash in rpc response, one data content field is enough

TheWaWaR: It's just like the `hash` filed in `Transaction` struct, user do not need to calculate the hash themselves. If we attend to remove data_hash in `get_cell_data`, then I think remove hash here is reasonable.
https://github.com/nervosnetwork/ckb/blob/develop/rpc/json/rpc.json#L617

TheWaWaR:
> > BREAKING CHANGE: `get_live_cell` add `with_data` argument and result structure changed.
> 
> we can use the optional param, it will keep rpc compatibility.

`get_live_cell` is not compatible with old version any more, because the return data structure has changed.

rpc/json/rpc.json
quake: Get, let's keep hash here. Please change tuple to named field struct for consistency:
```json
"data": {
    "content":  "0x...",
    "hash": "0x..."
}
```

TheWaWaR:
> According to the code implementation, if we do not set the second argument `with_data`, it will return the wrong argument number, it's inconsistent with the RPC document.

Oops, you are right.

rpc/json/rpc.json
TheWaWaR: ok

u2:
bors r=quake,u2,keroro520

PR #1529, opened 2019-09-03
docs: Add calculate_dao_maximum_withdraw example

keroro520:
docs: Add calculate_dao_maximum_withdraw example

I mark this example `skip: true` because it is difficult to construct its test.


u2:
bors r=doitian,u2

doitian:
bors retry

PR #1533, opened 2019-09-04
fix: Consensus constructor should init epoch reward from parameter.

doitian:
The constructor accepts the parameter `epoch_reward` but it is not used to
initialize the field `epoch_reward`.

When a chain spec does not use the default epoch reward, it will still
reverted to the default epoch reward starting from the second epoch.


doitian:
bors retry

doitian:
bors r=quake,u2

doitian:
bors retry

PR #1534, opened 2019-09-04
fix: Update secp type script hash in genesis

doitian:


u2:
bors r=quake,u2

doitian:
bors retry

doitian:
bors retry

PR #1535, opened 2019-09-04
docs: Group methods of same module together

ashchan:
Do not let a module appear in multiple places.

Update: doc tool was updated to only group RPC methods by module name in the generated doc. We keep the `rpc.json` intact as the unit tests depend on that file's content order to run successfully.

### DIFF
--- a/devtools/doc/rpc.py
+++ b/devtools/doc/rpc.py
@@ -8,6 +8,9 @@ def newline(n):
         print("")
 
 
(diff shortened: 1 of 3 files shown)


ashchan:
> I think we should modify the `devtools/doc/rpc.py`, so that the generated document keeps grouped even if the `rpc.json` is out of order.

Agree. Will do that.

PR #1542, opened 2019-09-04
chore: replace constructor of EpochExt by builder

yangby-cryptape:
- The constructor is too complex, it's difficult to review when we just see the invocation of `new(...)`
- The field `block_reward` is `base_block_reward` in actual, so rename it to the corrected name.

### DIFF
--- a/util/types/src/core/extras.rs
+++ b/util/types/src/core/extras.rs
@@ -55,7 +55,7 @@ impl TransactionInfo {
 #[derive(Clone, Eq, PartialEq, Debug, Default)]
 pub struct EpochExt {
     pub(crate) number: EpochNumber,
-    pub(crate) block_reward: Capacity,
+    pub(crate) base_block_reward: Capacity,
     pub(crate) remainder_reward: Capacity,
     pub(crate) previous_epoch_hash_rate: U256,
     pub(crate) last_block_hash_in_previous_epoch: packed::Byte32,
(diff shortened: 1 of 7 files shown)


util/types/src/core/extras.rs
```diff
@@ -105,80 +143,91 @@ impl EpochExt {
...
+            self.base_block_reward
+                .safe_add(Capacity::one())
+                .map_err(Into::into)
+        } else {
+            Ok(self.base_block_reward)
+        }
     }
+}
+
+impl EpochExtBuilder {
```
u2: I think the builder is redundant

yangby-cryptape: I didn't get your point.

IMHO, I think we should remove all setters from `EpochExt`.

yangby-cryptape: See this:
https://github.com/nervosnetwork/ckb/blob/87c6912deb4427f8ac32cfe54d5b1bea3ca44da6/spec/src/consensus.rs#L211-L218

If we calling `set_genesis_block` before calling `set_genesis_epoch_ext`, a bug was created.

https://github.com/nervosnetwork/ckb/blob/ecd88dcffe7c18c7112d2d4f1cce71560ba2618d/spec/src/lib.rs#L189-L190

Two independent methods have to be used in a specific order.

I think we can remove all `setters` from `EpochExt` to avoid that. (We can do it after `ConsensusBuilder`, the `ConsensusBuilder` is on the task list.)
If u want to update a `EpochExt`, u can do `ext.into_builder().set_xxx().build()`.

doitian:
I also touched this in the PR to half primary reward. The issue is the `set_length` method, which really changes the epoch reward.

https://github.com/nervosnetwork/ckb/pull/1537/files#diff-6ded9a204ae8172088a45152773c9df6R106

yangby-cryptape:
> I also touched this in the PR to half primary reward. The issue is the `set_length` method, which really changes the epoch reward.
> 
> https://github.com/nervosnetwork/ckb/pull/1537/files#diff-6ded9a204ae8172088a45152773c9df6R106

So, I think, we should (base on your PR):
- remove `base_block_reward` and `remainder_reward` from `EpochExtBuilder`
- remove all `setters` from `EpochExt`
- If anyone want to change fields of `EpochExt`, just rebuild it from `EpochExtBuilder`.

u2:
bors r=doitian,u2

PR #1550, opened 2019-09-05
feat: Headers can only be referenced after maturity period

xxuejie:


zhangsoledad:
bors r=u2,keroro520,zhangsoledad

PR #1551, opened 2019-09-05
feat: cellbase output data must be empty

driftluo:
Because the ckb reward is delayed, the ownership of the cellbase of the current block is not the miner who digs out the block, so cellbase's output data must be disabled.


u2:
bors r=u2,zhangsoledad

PR #1557, opened 2019-09-06
test: add some tests back: lost some tests when replace `ckb-core` by `ckb-types`

yangby-cryptape:


yangby-cryptape:
bors r=zhangsoledad,quake

PR #1558, opened 2019-09-06
refactor: Expose method to invoke a single script on a transaction

xxuejie:
This can be helpful in CKB script debugger's development


zhangsoledad:

PR #1559, opened 2019-09-06
fix: block serialized size should not include uncles proposals serialized size

yangby-cryptape:


yangby-cryptape:
bors r=zhangsoledad,quake

PR #1563, opened 2019-09-07
chore: adjust MEDIAN_TIME_BLOCK_COUNT to 37

u2:


test/src/utils.rs
```diff
@@ -16,7 +16,7 @@ use std::thread::sleep;
 use std::time::{Duration, Instant};
 use tempfile::tempdir;
 
-pub const MEDIAN_TIME_BLOCK_COUNT: u64 = 11;
+pub const MEDIAN_TIME_BLOCK_COUNT: u64 = 37;
```
yangby-cryptape: It's weird, there are few const variables that are defined more than once, and they have to be equal. Why?

yangby-cryptape:
> In fact, the present integration tests using a self-defined [`MEDIAN_TIME_BLOCK_COUNT = 11`](https://github.com/nervosnetwork/ckb/blob/develop/test/src/utils.rs#L19). Yes, it is a bit dirty.
> 
> These test cases in `test/src/specs/tx_pool/valid_since.rs` should be failed if this PR merged.
> So you have to replace `MEDIAN_TIME_BLOCK_COUNT` with `node.consensus().median_time_block_count()` around the integration-tests.

Maybe we can do this in another PR.

And,  `Consensus` in `Node` is `Option` :cry: :
https://github.com/nervosnetwork/ckb/blob/409d3fe5072206df8d2c1356a1d6e26a89f82d4b/test/src/node.rs#L31

u2:
> In fact, the present integration tests using a self-defined [`MEDIAN_TIME_BLOCK_COUNT = 11`](https://github.com/nervosnetwork/ckb/blob/develop/test/src/utils.rs#L19). Yes, it is a bit dirty.
> 
> These test cases in `test/src/specs/tx_pool/valid_since.rs` should be failed if this PR merged.
> So you have to replace `MEDIAN_TIME_BLOCK_COUNT` with `node.consensus().median_time_block_count()` around the integration-tests.

It's not related to this PR, but I create another commit for convenience.

u2:
bors r=jjyr,quake,keroro520

PR #1564, opened 2019-09-07
chore: fix rpc print_document

u2:
1. Before in `print_document`, the `params_of` would request the RPC again and the `send_transaction` would return `duplicated`.
2. fix indents in the `rpc.expect.json`.


yangby-cryptape:
A bit weird: why the order of fields in `rpc.json` are changed?

u2:
> A bit weird: why the order of fields in `rpc.json` are changed?

I have noticed this, but no idea.

PR #1567, opened 2019-09-08
tx_pool service

zhangsoledad:
#### Main:
independent tx-pool service. eliminate global thread lock.
####  Side Changes: 
* change ResolveTransaction to owned, since we apply new memory types https://github.com/nervosnetwork/ckb/pull/1249, we can clone tx without having to worry about performance, we need to eliminate lifetime for using it in Future.
* move snapshot to util due to cycle dependence
* BlockAssembler is now part of tx-pool


tx-pool/src/service.rs
```diff
@@ -0,0 +1,640 @@
...
+        sender
+            .try_send(Message::NotifyTxs(notify))
+            .map_err(Into::into)
+    }
+
+    pub fn get_tx_pool_info(
+        &self,
+    ) -> Result<Box<dyn Future<Item = TxPoolInfo, Error = oneshot::Canceled>>, FailureError> {
+        let mut sender = self.sender.clone();
+        let (responder, response) = oneshot::channel();
```
driftluo: I personally think that it may need to be changed to a synchronized channel, which is crossbeam channel.

Because after `std::future`, future must depend on the runtime, there is no wait for the synchronous blocking execution method, and from the current code, there is no need for an asynchronous channel here, so I think the synchronization channel should be used here.

tx-pool/src/service.rs
```diff
@@ -0,0 +1,640 @@
...
+        let mut sender = self.sender.clone();
+        let notify = Notify::notify((txs, callback));
+        sender
+            .try_send(Message::NotifyTxs(notify))
+            .map_err(Into::into)
+    }
+
+    pub fn get_tx_pool_info(
+        &self,
+    ) -> Result<Box<dyn Future<Item = TxPoolInfo, Error = oneshot::Canceled>>, FailureError> {
```
driftluo: Also because of `std::future`, I think this return value should not be a future. 

I think this will be better compatible with later modifications:
```
Future.then(|res| responder.send(res))
```

tx-pool/src/service.rs
```diff
@@ -0,0 +1,620 @@
...
+            .spawn(move || {
+                let server = receiver
+                    .for_each(move |message| {
+                        let service_clone = service.clone();
+                        tokio::spawn(service_clone.process(message))
+                    })
+                    .select2(signal_receiver)
+                    .map(|_| ())
+                    .map_err(|_| ());
+                tokio::run(server);
```
driftluo: Do we need to share a runtime with the network, or do we have one each like this?

zhangsoledad: We have discussed it before, network uses own standalone runtime, other services share global one.
But now, we only have one service use Tokio, Default global runtime is fine for now. Thoughts?

driftluo: ok

sync/src/relayer/mod.rs
```diff
@@ -140,17 +134,23 @@ impl Relayer {
...
+            let fresh_proposals_filter = tx_pool.fresh_proposals_filter(proposals);
+
+            if let Err(ref err) = fresh_proposals_filter {
+                debug_target!(
+                    crate::LOG_TARGET_RELAY,
+                    "relayer tx_pool_controller send fresh_proposals_filter error {:?}",
+                    err,
+                );
+            };
+            fresh_proposals_filter.unwrap()
```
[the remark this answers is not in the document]
zhangsoledad: nice catch

tx-pool/src/component/entry.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl Ord for TxEntry {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.as_sorted_key().cmp(&other.as_sorted_key())
+    }
+}
+
+// A template data struct used to store modified entries when package txs
+pub struct TxModifiedEntries {
```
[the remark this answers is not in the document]
zhangsoledad: I guess the comments has a typo, It's a temporary data struct? 
It's not used to holding entries, not like other containers,  more likely an Iterator,  perform operations on the entries.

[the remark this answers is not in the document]
jjyr: This container is used to shadow expired entries. however, it's hard to say where to put it, both way I can accept.

chain/src/chain.rs
```diff
@@ -375,16 +365,22 @@ impl ChainService {
...
-                fork.detached_proposal_id().iter(),
-                &mut txs_verify_cache,
+            if let Err(e) = self.shared.tx_pool_controller().update_tx_pool_for_reorg(
+                fork.detached_blocks().clone(),
+                fork.attached_blocks().clone(),
+                fork.detached_proposal_id().clone(),
                 new_snapshot,
-            );
+            ) {
+                error!("notify update_tx_pool_for_reorg error {}", e);
```
jjyr: Can we return errors or just use `.expect`? IMO the code should not continue to run after this, it will cause inconsistent silently.

sync/src/relayer/transaction_hashes_process.rs
```diff
@@ -37,13 +38,28 @@ impl<'a> TransactionHashesProcess<'a> {
...
+                .map_err(|e| {
+                    debug_target!(
+                        crate::LOG_TARGET_RELAY,
+                        "[TransactionHashesProcess] request fresh_proposals_filter error {:?}",
+                        e
+                    );
+                    e
+                })?;
+            fresh_ids
+                .iter()
```
jjyr: into_iter

chain/src/chain.rs
zhangsoledad: The current implementation, notify error can be tolerated. 
TxPool can work asynchronously with chain. 
This error only occurs in high pressure, the channel is full that can't send messages anymore, in this situation, I'm afraid such retry does not any help, so I think just log is fine.
Also, we don't have to interrupt the program immediately, it won't break anything, chain data persistence already finished, maybe we can do more such persistent TxPool, someday.

quake:
bors r=driftluo,u2,jjyr,quake

PR #1569, opened 2019-09-09
refactor: rewrite peer store

jjyr:
Major changes:
1. remove the dependencies of SQLite.
2. separate the concepts `peer_id` and `addr`, only persistent `addr`.
3. remove trait `PeerStore`
4. peer store do not track connected peers' addrs anymore, RPC get_peers use addresses from peer registry.


network/src/network.rs
```diff
@@ -1024,6 +1035,15 @@ impl NetworkService {
                     warn!("send shutdown message to p2p error: {:?}", err);
                 }
 
+                debug!("Dumping peer store data ...");
```
quake: Please implement the `Drop` trait for `DumpPeerStoreService` and invoke the `dump_peer_store` in the drop fn.

network/src/peer_store/peer_store_impl.rs
```diff
@@ -0,0 +1,244 @@
...
+    ) -> Result<()> {
+        let network = ip_to_network(addr.extract_ip_addr()?.ip);
+        self.ban_network(network, timeout_ms, ban_reason)
+    }
+
+    pub(crate) fn ban_network(
+        &mut self,
+        network: IpNetwork,
+        timeout_ms: u64,
+        ban_reason: String,
```
quake: Suggest to use `&str`
```suggestion
        ban_reason: &str,
```

network/src/peer_store/peer_store_db.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn dump<P: AsRef<Path>>(&self, path: P) -> Result<(), Error> {
+        let addrs: Vec<_> = self.addrs_iter().collect();
+        if let Some(dir) = path.as_ref().parent() {
+            create_dir_all(dir)?;
+        }
+        let file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .append(true)
```
quake: it will create duplicate address
```suggestion
            .append(false)
```

network/src/peer_store/peer_store_db.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn dump<P: AsRef<Path>>(&self, path: P) -> Result<(), Error> {
+        let banned_addrs = self.get_banned_addrs();
+        if let Some(dir) = path.as_ref().parent() {
+            create_dir_all(dir)?;
+        }
+        let file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .append(true)
```
quake:
```suggestion
            .append(false)
```

network/src/network.rs
jjyr: Nope, it's bad to do this in Drop,  it's implicit and tricky and hard to handle error.

network/src/peer_store/peer_store_impl.rs
jjyr: `String` is better, we can't use `format!` with `&str`, it's limit the reason report.

TheWaWaR: You can `format!("{} xxx", yyy).as_str()`. But I think it's better use `String` here, since we take the ownership to `BannedAddr`.

quake: I have no strong opinions one way or the other, only concerned with consistency `NetworkState#ban` 
```
pub fn ban(&self, address: IpNetwork, ban_until: u64, ban_reason: &str)
```

network/src/network.rs
jjyr: Update, dump peer store when `DumpPeerStoreService` receives an exit signal.

network/src/peer_store/addr_manager.rs
```diff
@@ -0,0 +1,96 @@
+use crate::peer_store::types::{AddrInfo, IpPort};
+use rand::Rng;
+use std::collections::{HashMap, HashSet};
+
+#[derive(Default)]
+pub struct AddrManager {
```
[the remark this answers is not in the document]
jjyr: IMO our code is not such complex, I'd rather do not to introduce another unnecessary dependency.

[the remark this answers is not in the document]
jjyr: And the intention here is to arrange id for each `AddrInfo`, so we can use multiple way's to accces it, such as `random_ids`. 
It's not just for the sequence access, when we delete a elem from `IndexMap` we need to depent the internal implement to keep the `random_ids` consistent with `IndexMap` internal.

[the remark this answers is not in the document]
quake: You can access `IndexMap` via index, the original requirement of `&mut self` in a getter fn `get_xxx` is weird.  The `get_random` function can be simplified to:
```rust
pub fn get_random<F>(&self, count: usize, filter: F) -> Vec<AddrInfo> {
    let mut indexes = (0..addres.len()).collect();
    thread_rng().shuffle(&mut indexes);
    indexes.iter().for_each(|i| {
        let (ip_port, addr_info) = self.addres.get_index(i);
        ...
    })
}
```

btw, there is a bug in `get_random` fn, result limit `count` param is ignored, it will return all addresses, please add an unit test and fix it.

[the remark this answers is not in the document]
jjyr: Fix the `count` and `serde`.

This change may reduce a few lines of code, but it makes code slower. The current implementation only exchange `count` number elements, but with `shuffle` we need to exchange `addrs.len()` number elements. 

I do not think the `get` with `&mut self` is unacceptable, in `LRUCache` we already use it. I'am ok to change the name `get_xxx` to `fetch_xxx`.

network/src/peer_store/peer_store_db.rs
```diff
@@ -0,0 +1,95 @@
...
+        if let Some(dir) = path.as_ref().parent() {
+            create_dir_all(dir)?;
+        }
+        let file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .append(false)
+            .open(path)?;
+        serde_json::to_writer(file, &banned_addrs).map_err(|err| PeerStoreError::Serde(err).into())
+    }
```
TheWaWaR: https://github.com/bitcoin/bitcoin/blob/9debfd0e28413df1ba669f614aa44bb652fedb32/src/addrdb.cpp#L36-L73

PR #1571, opened 2019-09-10
feature: is_better_chain uses first-received policy

u2:


keroro520:
I think it's tests-required.

u2:
> I think it's tests-required.

The `block_sync_forks` tests it.

keroro520:
> > I think it's tests-required.
> 
> The `block_sync_forks` tests it.

`block_sync_forks` only test sync protocol. I mean we should test relay protocol and miner submission as well(or even more complicated cases).

u2:
> > > I think it's tests-required.
> > 
> > 
> > The `block_sync_forks` tests it.
> 
> `block_sync_forks` only test sync protocol. I mean we should test relay protocol and miner submission as well(or even more complicated cases).

 There are two places to test:
1. block sync
2. block processing, no matter where it comes from, whether from miner or other peers.

And the `compact_block` reconstructing is not caring about the better chain policy.

u2:
@keroro520 updated, please review.

quake:
bors r=keroro520,quake

PR #1572, opened 2019-09-10
test: sleep less time in test

u2:
Add comment to declare the purpose of the `sleep` and ajust it to a smaller value.

### DIFF
--- a/test/src/specs/sync/chain_forks.rs
+++ b/test/src/specs/sync/chain_forks.rs
@@ -509,7 +509,8 @@ impl Spec for ForksContainSameTransactions {
 
         // Build `chain1`, contain the target `transaction`, with length = 61
         {
-            sleep(Duration::from_secs(2));
+            // `sleep` to make sure that the chain1[2] != chain2[2]
+            sleep(Duration::from_millis(1));
             node1.generate_blocks(30);
             node1.submit_transaction(&transaction);
             node1.generate_blocks(30);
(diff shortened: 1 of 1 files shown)


test/src/specs/sync/chain_forks.rs
```diff
@@ -509,18 +507,32 @@ impl Spec for ForksContainSameTransactions {
 
         // Build `chain1`, contain the target `transaction`, with length = 61
         {
-            sleep(Duration::from_secs(2));
```
keroro520: `sleep(2s)` makes the timestamp of blocks are different, to avoid `chain1 == chain0`. I agree that `sleep` is tricky. We can change the `block.nonce` or `block.timestamp` manually to archive the intention.

test/src/specs/sync/chain_forks.rs
```diff
@@ -509,18 +507,32 @@ impl Spec for ForksContainSameTransactions {
...
             node2.generate_blocks(50);
         }
 
+        let rpc_client0 = node0.rpc_client();
+        let rpc_client1 = node1.rpc_client();
+        let rpc_client2 = node2.rpc_client();
+
+        let finish_generating = wait_until(10, || {
```
keroro520: It is nonsense to check height here.
The nodes of cause are on the target height now.

test/src/specs/sync/chain_forks.rs
u2: Why it is 2 seconds?

u2: `chain1` and `chain2` have different height and the `transaction` is in the different height block, they are different.

keroro520: I change my previous description: to ensure that `chain1[1] != chain2[1]`.

u2: updated.

quake:
bors r=keroro520,quake

PR #1573, opened 2019-09-10
feature: ignore compact-block which is higher than tip + N

u2:


sync/src/relayer/compact_block_process.rs
```diff
@@ -64,17 +64,23 @@ impl<'a> CompactBlockProcess<'a> {
...
+        // where N is the current epoch length
         if lowest_number > header.number() {
             return Err(Error::Ignored(Ignored::TooOldBlock).into());
         }
 
+        // If the CompactBlock height is greater than tip + 2.
+        // In high probability the node is syncing.
+        if header.number() > tip.number() + 2 {
+            return Err(Error::Ignored(Ignored::TooHighBlock).into());
+        }
```
TheWaWaR: We should log `Error::Ignored` in `try_process()` too.

u2: updated

sync/src/relayer/compact_block_process.rs
```diff
@@ -64,17 +64,23 @@ impl<'a> CompactBlockProcess<'a> {
...
         let epoch_length = snapshot.epoch_ext().length();
         let lowest_number = tip.number().saturating_sub(epoch_length);
 
+        // Only accept blocks with a height greater than tip - N
+        // where N is the current epoch length
         if lowest_number > header.number() {
             return Err(Error::Ignored(Ignored::TooOldBlock).into());
         }
 
+        // If the CompactBlock height is greater than tip + 2.
```
doitian: Need to fallback to sync with the peer, a.k.a, send `GetHeaders` to peer.

PR #1574, opened 2019-09-10
feat: Keep only one version of VM by removing ScriptConfig

xxuejie:
NOTE: one side effect here, is that CKB will only be able to run on
64-bit Windows, Linux and maxOS platforms with this change. We can
discuss if this will bring future problems. One alternative solution
here, is that we can still remove ScriptConfig and leave only one VM
here, then use conditional compilation to revert to Rust based VMs on
32-bit platforms, or ARM processors. This way on each platform, we are
still only supporting just one VM for simplicity


xxuejie:
bors try

zhangsoledad:
bors r=quake,u2,TheWaWaR,yangby-cryptape

PR #1576, opened 2019-09-10
feat: satoshi's gift

jjyr:
This PR allows a special cell "satoshi's gift" in the genesis block. When the mainnet launching, this cell will issue 1/4 of total genesis capacity, and 60% capacity of the cell will be calculated as occupied, this affects the Nervos DAO contract interests.

Satoshi's gift cell, as the name, the lock script of this cell verifies an `H160(pubkey)` that satoshi used in Bitcoin's genesis, satoshi can use the private key to sign a tx to spent the cell on CKB.

Major changes:
* Add a new option `SatoshiGift` on genesis configuration.
* Nervos dao contract calculate the satoshi's gift cell's occupied.
* issue a satoshi's cell in `testnet.toml`

### DIFF
--- a/resource/specs/testnet.toml
+++ b/resource/specs/testnet.toml
@@ -8,7 +8,7 @@ difficulty = "0x11940000000"
 uncles_hash = "0x0000000000000000000000000000000000000000000000000000000000000000"
 nonce = 0
 # run `cargo run cli hashes -b` to get the genesis hash
-hash = "0x4193e67e1a8ff5d3a5ade1e1a1deb1c7308d0221588efd743c69c1be2d01bab1"
+hash = "0x7c69755329171b3f01b4c81000904738d214b22ce85993614822c81e3fec33f7"
 
 [genesis.genesis_cell]
 message = "rylai-v10"
@@ -71,6 +71,13 @@ capacity = 50_000_000_00000000
 lock.code_hash = "0x1892ea40d82b53c678ff88312450bbb17e164d7a3e0a90941aa58839f56f8df2"
 lock.args = ["0x57ccb07be6875f61d93636b0ee11b675494627d2"]
 lock.hash_type = "type"
+# Satoshi's gift, a cell send to satoshi's lock,
+# 60% capacity of this cell is occupied.
+[[genesis.issued_cells]]
+capacity = 50_000_000_00000000
+lock.code_hash = "0x88fce31aa68db1bb2d5bde7a731b7e6b2429e7aa6fdf60203f397eb8e2147794"
+lock.args = ["0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18"]
+lock.hash_type = "data"
 
 [pow]
 func = "Eaglesong"
--- a/resource/specs/testnet.toml
+++ b/resource/specs/testnet.toml
@@ -8,7 +8,7 @@ difficulty = "0x11940000000"
 uncles_hash = "0x0000000000000000000000000000000000000000000000000000000000000000"
 nonce = 0
 # run `cargo run cli hashes -b` to get the genesis hash
-hash = "0x7c69755329171b3f01b4c81000904738d214b22ce85993614822c81e3fec33f7"
+hash = "0x7feb7839f6dc69867e7afd143aec916165b13bc4b2b9d614d075a5e0086861c2"
 
 [genesis.genesis_cell]
 message = "rylai-v10"
@@ -75,9 +75,9 @@ lock.hash_type = "type"
 # 60% capacity of this cell is occupied.
 [[genesis.issued_cells]]
 capacity = 50_000_000_00000000
-lock.code_hash = "0x88fce31aa68db1bb2d5bde7a731b7e6b2429e7aa6fdf60203f397eb8e2147794"
+lock.code_hash = "0x9cc3121726ed40c0ece7b43f7f28af712c7f13101f78699c5668341a52ddf280"
 lock.args = ["0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18"]
-lock.hash_type = "data"
+lock.hash_type = "type"
 
 [pow]
 func = "Eaglesong"
--- a/spec/src/consensus.rs
+++ b/spec/src/consensus.rs
@@ -8,10 +8,10 @@ use ckb_types::{
         capacity_bytes, BlockBuilder, BlockNumber, BlockView, Capacity, Cycle, EpochExt,
         HeaderView, Ratio, TransactionBuilder, Version,
     },
-    h256,
+    h160,
     packed::{Byte32, CellInput, Script},
     prelude::*,
-    u256, H256, U256,
+    u256, H160, H256, U256,
 };
 use std::cmp;
 use std::sync::Arc;
@@ -59,9 +59,8 @@ pub(crate) const MAX_BLOCK_CYCLES: u64 = TWO_IN_TWO_OUT_CYCLES * TWO_IN_TWO_OUT_
 const MAX_BLOCK_PROPOSALS_LIMIT: u64 = 3_000;
 const PROPOSER_REWARD_RATIO: Ratio = Ratio(4, 10);
 
-// A special lock for satoshi, use satoshi's private key can unlock this.
-pub(crate) const SATOSHI_LOCK_HASH: H256 =
-    h256!("0x1d11151e87cf8f77d2130e7ee8ca66be197b310ed39cc74c40a5df4933ca8a75");
+// Satoshi's pubkey hash in Bitcoin genesis.
+pub(crate) const SATOSHI_PUBKEY_HASH: H160 = h160!("0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18");
 // Ratio of satoshi cell occupied of capacity,
 // only affects genesis cellbase's satoshi lock cells.
 pub(crate) const SATOSHI_CELL_OCCUPIED_RATIO: Ratio = Ratio(6, 10);
@@ -120,7 +119,7 @@ pub struct ConsensusBuilder {
     block_version: Version,
     max_block_proposals_limit: u64,
     genesis_epoch_ext: EpochExt,
-    satoshi_lock_hash: Byte32,
+    satoshi_pubkey_hash: H160,
     satoshi_cell_occupied_ratio: Ratio,
 }
 
@@ -135,7 +134,7 @@ impl Default for ConsensusBuilder {
             .build();
         let dao = genesis_dao_data_with_satoshi_gift(
             vec![&cellbase],
-            &SATOSHI_LOCK_HASH,
+            &SATOSHI_PUBKEY_HASH,
             SATOSHI_CELL_OCCUPIED_RATIO,
         )
         .unwrap();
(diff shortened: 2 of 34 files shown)


spec/src/consensus.rs
```diff
@@ -260,6 +280,16 @@ impl Consensus {
         self
     }
 
+    pub fn set_satoshi_lock_hash(mut self, lock_hash: Byte32) -> Self {
```
doitian: Please do not add `set_` methods to Consensus any longer.

doitian: Prefer builder

util/types/src/core/mod.rs
```diff
@@ -19,7 +19,7 @@ pub use reward::BlockReward;
 pub use transaction_meta::{TransactionMeta, TransactionMetaBuilder};
 pub use views::{BlockView, HeaderView, TransactionView, UncleBlockVecView, UncleBlockView};
 
-pub use ckb_occupied_capacity::{capacity_bytes, Capacity, Ratio};
+pub use ckb_occupied_capacity::{self, capacity_bytes, Capacity, Ratio};
```
TheWaWaR: If you want to use `CapacitResult` just reexport it alone?

spec/src/consensus.rs
```diff
@@ -556,13 +665,35 @@ fn u256_low_u64(u: U256) -> u64 {
...
+        let consensus = ConsensusBuilder::new(genesis, capacity_bytes!(100)).build();
         assert_eq!(capacity_bytes!(100), consensus.epoch_reward);
     }
+
+    #[test]
+    fn test_satoshi_lock_hash() {
+        let script_code =
+            Resource::bundled("specs/cells/secp256k1_ripemd160_sha256_sighash_all".to_string())
+                .get()
+                .expect("Load secp script data failed");
```
TheWaWaR: This hash is stored in `ckb_resource`.

spec/src/consensus.rs
```diff
@@ -556,13 +665,35 @@ fn u256_low_u64(u: U256) -> u64 {
...
     use super::*;
-    use ckb_types::core::{BlockBuilder, TransactionBuilder};
+    use ckb_types::{
+        core::{BlockBuilder, ScriptHashType, TransactionBuilder},
+        h160,
+        packed::{CellOutput, Script},
+        H160,
+    };
+
+    // Satoshi's lock in Bitcoin genesis.
```
TheWaWaR:
```suggestion
    // Satoshi's pubkey hash in Bitcoin genesis.
```

spec/src/consensus.rs
```diff
@@ -58,6 +59,13 @@ pub(crate) const MAX_BLOCK_CYCLES: u64 = TWO_IN_TWO_OUT_CYCLES * TWO_IN_TWO_OUT_
 const MAX_BLOCK_PROPOSALS_LIMIT: u64 = 3_000;
 const PROPOSER_REWARD_RATIO: Ratio = Ratio(4, 10);
 
+// A special lock for satoshi, use satoshi's private key can unlock this.
+pub(crate) const SATOSHI_LOCK_HASH: H256 =
+    h256!("0x1d11151e87cf8f77d2130e7ee8ca66be197b310ed39cc74c40a5df4933ca8a75");
```
TheWaWaR: Better calculate from pubkey hash `0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18` in a `lazy_static!`. If the serialization changed, it's invalid.

jjyr: This test guarantee the lock hash https://github.com/nervosnetwork/ckb/pull/1576/files/e49e1dec8527c7e6248aac2a035e327905b0628f#diff-ca7d4f96a4f6eae78978a2e2622a7c38R687

TheWaWaR: But, if the serialization changed, you should manually update here. It's inconvenient and not straightforward.

jjyr: This value should seem as a constant, if serialization is changed, all the on-chain consensus are broken, we only need to manually handle this before mainnet launch.

resource/specs/testnet.toml
```diff
@@ -71,6 +71,13 @@ capacity = 50_000_000_00000000
 lock.code_hash = "0x1892ea40d82b53c678ff88312450bbb17e164d7a3e0a90941aa58839f56f8df2"
 lock.args = ["0x57ccb07be6875f61d93636b0ee11b675494627d2"]
 lock.hash_type = "type"
+# Satoshi's gift, a cell send to satoshi's lock,
+# 60% capacity of this cell is occupied.
+[[genesis.issued_cells]]
+capacity = 50_000_000_00000000
+lock.code_hash = "0x88fce31aa68db1bb2d5bde7a731b7e6b2429e7aa6fdf60203f397eb8e2147794"
+lock.args = ["0x62e907b15cbf27d5425399ebf6f0fb50ebb88f18"]
+lock.hash_type = "data"
```
TheWaWaR:
```
code_hash = "0x9cc3121726ed40c0ece7b43f7f28af712c7f13101f78699c5668341a52ddf280"
hash_type = "type"
```

spec/src/consensus.rs
```diff
@@ -556,13 +665,32 @@ fn u256_low_u64(u: U256) -> u64 {
...
         assert_eq!(capacity_bytes!(100), consensus.epoch_reward);
     }
+
+    #[test]
+    fn test_satoshi_lock_hash() {
+        let lock = Script::new_builder()
+            .code_hash(CODE_HASH_SECP256K1_RIPEMD160_SHA256_SIGHASH_ALL.pack())
+            .hash_type(ScriptHashType::Data.pack())
+            .args(vec![SATOSHI_H160.0.pack()].pack())
+            .build();
```
TheWaWaR: This system script have type script (type id), should use `ScriptHashType::Type` instead.

resource/specs/testnet.toml
TheWaWaR: Also add this to `dev.toml`, `staging.toml`

spec/src/lib.rs
```diff
@@ -249,7 +275,10 @@ impl Genesis {
             .transactions()
             .into_iter()
             .flat_map(|tx| tx.outputs().into_iter().map(|output| output.lock()))
-            .filter(|lock_script| lock_script != &genesis_cell_lock)
+            .filter(|lock_script| {
+                lock_script != &genesis_cell_lock
+                    && lock_script.calc_script_hash() != self.satoshi_gift.satoshi_lock_hash.pack()
```
TheWaWaR: Should not exclude `satoshi_lock_hash`

util/dao/utils/src/lib.rs
```diff
@@ -22,11 +23,19 @@ pub const DAO_VERSION: u8 = 1;
 pub const DAO_SIZE: usize = 32;
 
 pub fn genesis_dao_data(txs: Vec<&TransactionView>) -> Result<Byte32, Error> {
+    genesis_dao_data_with_satoshi_gift(txs, &H256([0u8; 32]), Ratio(1, 1))
+}
```
TheWaWaR: Should tell people it's only for test purpose.

spec/src/lib.rs
jjyr: Ok, I misunderstand the original check.

util/dao/utils/src/lib.rs
jjyr: IMO it's ok for this, logically the satoshi gift is not neccesary for the DAO, maybe in the future we need a testnet do not have a satoshi gift cell.

Since we have integration test , it's hard to misuse this method.

spec/src/consensus.rs
```diff
@@ -195,15 +199,64 @@ impl Consensus {
...
+        }
+    }
+
+    pub fn id(mut self, id: String) -> Self {
         self.id = id;
         self
     }
 
-    pub fn set_genesis_block(mut self, genesis_block: BlockView) -> Self {
+    pub fn genesis_block(mut self, genesis_block: BlockView) -> Self {
```
yangby-cryptape: I think this builder didn't fix the issue that why we need a builder, we still use setters with side effects.

```rust
pub fn genesis_block(mut self, genesis_block: BlockView) -> Self {
    // ... omit ...
    self.genesis_epoch_ext
        .set_difficulty(genesis_block.difficulty());	            .set_difficulty(genesis_block.difficulty());
    self.genesis_hash = genesis_block.hash();	        self.genesis_hash = genesis_block.hash();
    // ... omit ...
}
```

Anyway, we can do this optimization later.

spec/src/consensus.rs
```diff
@@ -93,59 +100,56 @@ impl ProposalWindow {
...
-        Consensus::new(genesis_block, DEFAULT_EPOCH_REWARD)
+        ConsensusBuilder::new(genesis_block, DEFAULT_EPOCH_REWARD)
     }
 }
 
-#[allow(clippy::op_ref)]
-impl Consensus {
-    pub fn new(genesis_block: BlockView, epoch_reward: Capacity) -> Consensus {
+impl ConsensusBuilder {
+    pub fn new(genesis_block: BlockView, epoch_reward: Capacity) -> Self {
```
yangby-cryptape: I think this `new(..)` method has side effects.

Remove Side Effects:
- Please use `ConsensusBuilder::default().set(..)....set(..).build(..)` to rewrite it, if any parameters we have to set, then these parameters should be the arguments of `new(..)`.

util/dao/utils/src/lib.rs
yangby-cryptape: I really hate it: `H160([0u8; 20])`. No body like `Shotgun Surgery`! Please call APIs!

**Avoid using constructor methods for those structs that has a lot of features!**

test/src/specs/dao/satoshi_dao_occupied.rs
```diff
@@ -0,0 +1,224 @@
...
+        assert!(
+            is_committed(&tx_status),
+            "ensure_committed failed {:#x}",
+            tx_hash
+        );
+    }
+
+    fn modify_chain_spec(&self) -> Box<dyn Fn(&mut ChainSpec) -> ()> {
+        Box::new(|spec_config| {
+            let satoshi_cell = issue_satoshi_cell(H160([0u8; 20]));
```
yangby-cryptape: `H160([0u8; 20])` Call APIs is better.

test/src/specs/dao/satoshi_dao_occupied.rs
```diff
@@ -0,0 +1,224 @@
...
+        .build()
+        .calc_script_hash()
+        .unpack()
+}
+
+fn pubkey_hash160(data: &[u8]) -> H160 {
+    fn ripemd160(data: &[u8]) -> H160 {
+        use ripemd160::{Digest, Ripemd160};
+        let digest = Ripemd160::digest(data);
+        H160(digest.into())
```
yangby-cryptape: Call APIs is better.

test/src/specs/dao/satoshi_dao_occupied.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    fn sha256(data: &[u8]) -> H256 {
+        use sha2::{Digest, Sha256};
+        let digest: [u8; 32] = Sha256::digest(data).into();
+        H256(digest)
```
yangby-cryptape: Call APIs is better.

spec/src/consensus.rs
jjyr: I don't think the `initialize` is `side effects`.

yangby-cryptape: May be I did not say clearly, or I used inappropriate words.

I mean: `T::new(inputs)` should equal to `T::default().update(inputs)`, not `T::default().update(inputs).update_some_extras_things()`, so the function signature can show what does the function do, clearly.

jjyr: If new is just 1 to 1 equals to the builder, why we need the `new` function? This function is of course surely to put some extra initialize logic.

spec/src/consensus.rs
jjyr: I am curious how can a setter can without side effects?

PR #1584, opened 2019-09-11
feat: Hexilize jsonrpc numbers

keroro520:
* feat: Return numbers in heximal format(without leading zeros)
* feat: Allow accept numbers in heximal or decimal format


util/jsonrpc-types/src/uint64.rs
```diff
@@ -0,0 +1,136 @@
+use ckb_types::prelude::{Pack, Unpack};
+use ckb_types::{core, packed};
+use serde::de::{Error, Visitor};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
+
+#[derive(Clone, Default, PartialEq, Eq, Hash, Debug)]
+pub struct Uint64(u64);
```
zhangsoledad: Why not derive `Copy`？

keroro520: Oh, I didn't think about it. I will update it later.

util/jsonrpc-types/src/uint64.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        write!(formatter, "a hex-encoded or decimal uint64")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        match value {
+            _ if value.starts_with("0x") => u64::from_str_radix(&value[2..], 16)
```
yangby-cryptape: I'm not sure about this:

```rust
fn main () {
    let x1 = u32::from_str_radix("0", 16).unwrap();
    let x2 = u32::from_str_radix("00", 16).unwrap();
    let x3 = u32::from_str_radix("000", 16).unwrap();
    let x4 = u32::from_str_radix("0000", 16).unwrap();
    assert_eq!(x1, x2);
    assert_eq!(x1, x3);
    assert_eq!(x1, x4);
}
```

Should we remove this ambiguity?

ashchan:
> I asked @ashchan , we both think a strict format is better.
> 
> Please remove the compatibility: `feat: Allow accept numbers in heximal or decimal format`.
> 
> My **personal** opinion (may be not right):
> 
> * A kernel system should keep simple and strict, no ambiguity is allowed.
> * So we can easy to test it, and keep the correctness easily.

My two cents for this. Force using hex string as number for both RPC params and outputs would free us a lot of trouble in the future (especially documentation and support).

keroro520:
> > I asked @ashchan , we both think a strict format is better.
> > Please remove the compatibility: `feat: Allow accept numbers in heximal or decimal format`.
> > My **personal** opinion (may be not right):
> > 
> > * A kernel system should keep simple and strict, no ambiguity is allowed.
> > * So we can easy to test it, and keep the correctness easily.
> 
> My two cents for this. Force using hex string as number for both RPC params and outputs would free us a lot of trouble in the future (especially documentation and support).

Thanks for the feedback. I will change it to only allow heximal numbers now.

util/jsonrpc-types/src/uint64.rs
keroro520: I have no strong opinions for it. 

@ashchan Any suggestions?

yangby-cryptape: The current/present situation:
- Return numbers in heximal format(**without** leading zeros)
- U32/U64: Accept numbers in heximal format(**with** any number leading zeros)
- U256:  Accept numbers in heximal format(**without** leading zeros, expected `0x0`)

keroro520: Agree!
So, I should make a more strict limitation on json-numbers: no leading zeros, no mather inputs or outputs, expect `0x0`.

PR #1585, opened 2019-09-12
feat: Include fractions in epoch number representations

xxuejie:
This change introduce fractions in 2 places where epoch numbers might
be used:

* The epoch field in the header
* Cell input's since part when using epoch values

Here we use a rational number format to represent epoch number.
The lower 54 bits of the epoch number field are split into 3
parts(listed in the order from higher bits to lower bits):

* The highest 16 bits represent the epoch length
* The next 16 bits represent the current block index in the epoch
* The lowest 24 bits represent the current epoch number

Different from previous solution, with a rational number format the
precision can be guaranteed.

Assuming we are at block number 11555, epoch 50, and epoch 50 starts
from block 11000, has a length of 1000. The epoch number for this
particular block will then be 9326559282, which is calculated
in the following way:

```
50 | ((11555 - 11000) << 24) | (1000 << 16)
```

This way, we can alleviate the trouble that CKB uses variable epoch
length. The actual contract or dapps can just read this epoch number
with fractions and perform numerical operations as normal.

### DIFF
--- a/util/types/src/core/extras.rs
+++ b/util/types/src/core/extras.rs
@@ -178,14 +178,13 @@ impl EpochExt {
         }
     }
 
-    pub fn number_with_fraction(&self, number: BlockNumber) -> EpochNumberWithFraction {
+    pub fn number_with_fraction(&self, number: BlockNumber) -> DetailedEpochNumber {
         debug_assert!(
             number >= self.start_number() && number < self.start_number() + self.length()
         );
-        let fraction = EpochNumberWithFraction::FRACTION_MAXIMUM_VALUE
-            * (number - self.start_number())
+        let fraction = DetailedEpochNumber::FRACTION_MAXIMUM_VALUE * (number - self.start_number())
             / self.length();
-        EpochNumberWithFraction::new(self.number(), fraction)
+        DetailedEpochNumber::new(self.number(), fraction, self.length())
     }
 }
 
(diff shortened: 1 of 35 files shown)


util/types/src/core/extras.rs
```diff
@@ -231,3 +244,44 @@ impl EpochExtBuilder {
...
+    pub const FRACTION_MASK: u64 = (Self::FRACTION_MAXIMUM_VALUE - 1);
+
+    pub fn new(number: u64, fraction: u64) -> EpochNumberWithFraction {
+        debug_assert!(number < Self::NUMBER_MAXIMUM_VALUE);
+        debug_assert!(fraction < Self::FRACTION_MAXIMUM_VALUE);
+        EpochNumberWithFraction((number << Self::FRACTION_BITS) | fraction)
+    }
+
+    pub fn number(&self) -> u64 {
+        (self.0 >> 16) & ((1u64 << 40) - 1)
```
yangby-cryptape:
```suggestion
        (self.0 >> Self::FRACTION_BITS) & (Self::NUMBER_MASK)
```

util/types/src/core/extras.rs
```diff
@@ -231,3 +244,44 @@ impl EpochExtBuilder {
...
+    }
+
+    pub fn fraction(&self) -> u64 {
+        u64::from(self.0 as u16)
```
yangby-cryptape:
```suggestion
        self.0 & Self::FRACTION_MASK
```

util/types/src/core/extras.rs
```diff
@@ -231,3 +243,71 @@ impl EpochExtBuilder {
...
+        debug_assert!(fraction < Self::FRACTION_MAXIMUM_VALUE);
+        debug_assert!(length < Self::LENGTH_MAXIMUM_VALUE);
+        DetailedEpochNumber(
+            (length << Self::NUMBER_WITH_FRACTION_BITS)
+                | (number << Self::FRACTION_BITS)
+                | fraction,
+        )
+    }
+
+    pub fn number(&self) -> u64 {
```
jjyr: The `DetailedEpochNumber ` and `EpochNumber` both exist in the current implementation, are we still use `EpochNumber` to indicate the `number` part? if not the `primitive ::EpochNumber` alias should be removed.

verification/src/tests/transaction_verifier.rs
```diff
@@ -372,12 +390,59 @@ fn test_invalid_since_verify() {
...
+fn test_invalid_zero_length_since() {
+    // use remain flags
+    let tx = create_tx_with_lock(0xa000_0000_0000_0000);
+    let rtx = create_resolve_tx_with_transaction_info(
+        &tx,
+        MockMedianTime::get_transaction_info(1, EpochNumberWithFraction::new(0, 0, 10), 1),
+    );
+
+    let median_time_context = MockMedianTime::new(vec![0; 11]);
+    assert!(verify_since(&rtx, &median_time_context, 5, 1).is_ok(),);
```
u2: Is it a valid or invalid test since? For the test name `test_invalid_zero_length_since`, it's an invalid since, but `verify_since` is `is_ok()`.

u2:
> The lower 54 bits of the epoch number field are split into 3

56 bits?

> 50 | ((11555 - 11000) << 24) | (1000 << 16)

```
50 | ((11555 - 11000) << 24) | (1000 << 40)?
```

PR #1586, opened 2019-09-12
feat: script package build adjustment

xxuejie:
This commit adds 2 changes:

* It adds a feature to allow disabling logging, hence saving us a
  lot of depdendencies
* It adds a feature to allow compiling without assembly interpreter
  mode. In CKB we won't allow this since we want only one interpreter,
  but when integrating with alternative projects, such as web
  debugger, this can be quite helpful


zhangsoledad:
bors r=yangby-cryptape,quake

PR #1587, opened 2019-09-12
refactor: Remove global store cache

zhangsoledad:


store/src/db.rs
```diff
@@ -41,8 +49,14 @@ impl<'a> ChainStore<'a> for ChainDB {
 }
 
 impl ChainDB {
-    pub fn new(db: RocksDB) -> Self {
-        ChainDB { db }
+    pub fn new(db: RocksDB, config: Option<StoreConfig>) -> Self {
+        let cache = config
+            .map(StoreCache::from_config)
+            .unwrap_or_else(StoreCache::default);
```
jjyr: Use `StoreConfig::Default` as second function is more intuition, I thought None is to disable cache until I see the internal.

store/src/store.rs
```diff
@@ -46,19 +44,19 @@ pub trait ChainStore<'a>: Send + Sync {
...
     fn get_block_header(&'a self, hash: &packed::Byte32) -> Option<HeaderView> {
-        let cache_enable = cache_enable();
-        if cache_enable {
-            if let Some(header) = HEADER_CACHE.lock().get_refresh(hash) {
+        if let Some(cache) = self.cache() {
+            if let Some(header) = cache.headers.lock().get_refresh(hash) {
                 return Some(header.clone());
             }
-        }
+        };
```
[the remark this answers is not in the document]
zhangsoledad: No, please review more attentive, it's no `if` ,it's `if let` , without `;`  semicolon, it compile?

store/src/db.rs
zhangsoledad: No !!!

None is not representing disable the cache now, one of the purposes of the refactor is fixing disable cache manually like before. before we use a global static cache, it will obstruct some tests, have to disable it in some case, but disable cache will cover some internal error.

`StoreConfig` is optional originally, users don't need to config it in the config file.  
I don't see any non-intuition here.

And `None` representing disable cache, I say No for this.

store/src/store.rs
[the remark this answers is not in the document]
keroro520: It can compile without `;`

store/src/db.rs
jjyr: I am not suggest to support disable the cache, but to use `StoreConfig::default` to replace the `None` as the second param.

The invoker can create a DB as:
`ChainDB::new(RocksDB::open_tmp(COLUMNS), Default::default())`
or
`ChainDB::new(RocksDB::open_tmp(COLUMNS), StoreConfig::default())`

This is clearer than using `None` to represent default.

PR #1592, opened 2019-09-17
feat: rpc strict JSON parsing

TheWaWaR:


TheWaWaR:
https://github.com/cryptape/ckb-internal/issues/1400

PR #1518, opened 2019-09-02
feat: Add chain_root to block header.

jjyr:
This PR intent to add a new field `chain_root` to block header, to help build flyclient protocol based light client.

`chain_root` is a merkle root of all previous headers in the chain, for the genesis block `chain_root` is `0x00..00`.

This PR implement MMR to calculate `chain_root` for each header. We push `block hash` and `difficulty` of a header into MMR, both information are digested by `chain_root`, so we can easily verify the total difficulty of a block in flyclient protocol.

Changes:
1. Implement `ckb-merkle-mountain-range`.
2. Add a new core type [HeaderDigest](https://github.com/nervosnetwork/ckb/pull/1518/files#diff-e1868d1b0796d594be368230eed82bf5), used to digest header hash and total difficulty in MMR.
3. Add a new field `chain_root` on block header.

### Chain benchmark

Choose always success benchmark, to eliminate the verification effects.

Before this PR.

```
Benchmarking always_success main_branch/100: Collecting 10 samples in estimated 36.260 s (55 ite                                                                                                always_success main_branch/100                        
                        time:   [286.90 ms 288.89 ms 290.05 ms]
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low mild
  1 (10.00%) high severe
Benchmarking always_success main_branch/200: Collecting 10 samples in estimated 53.182 s (55 ite                                                                                                always_success main_branch/200                        
                        time:   [554.96 ms 567.52 ms 583.15 ms]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild
Benchmarking always_success main_branch/500: Collecting 10 samples in estimated 124.45 s (55 ite                                                                                                always_success main_branch/500                        
                        time:   [1.4034 s 1.4787 s 1.5330 s]
Benchmarking always_success main_branch/1000: Collecting 10 samples in estimated 204.72 s (55 it                                                                                                always_success main_branch/1000                        
                        time:   [2.5919 s 2.6406 s 2.6785 s]

Benchmarking always_success side_branch/100: Collecting 10 samples in estimated 25.505 s (55 ite                                                                                                always_success side_branch/100                        
                        time:   [2.3724 ms 2.3911 ms 2.4114 ms]
Benchmarking always_success side_branch/200: Collecting 10 samples in estimated 27.941 s (55 ite                                                                                                always_success side_branch/200                        
                        time:   [3.9726 ms 4.1627 ms 4.3823 ms]
[...]

PR #1598, opened 2019-09-19
Revert chain_root field

jjyr:
Need more research on light client protocol, revert the `chain_root` for now, we can make a hard-fork to add it back after we decide the light client protocol.

revert https://github.com/nervosnetwork/ckb/pull/1518


u2:
bors r=doitian,u2

doitian:
@jjyr please resolve conflicts

jjyr:
bors r=doitian,u2

u2:
bors r=doitian,u2

jjyr:
bors r=doitian,u2

jjyr:
bors r=doitian,u2

jjyr:
bors retry

jjyr:
bors r=doitian,u2

zmitton:
Im wondering why this got reverted. Its literally the only consensus changes needed to open up the world of Fly Proofs. 

I imagine the light client protocol will continue to evolve for quite some time after this field of possibilities starts getting explored.

janx:
We did plenty of research then but we got no time to figure out how to handle difficulty adjustment verification (hint: it's more complicated than Bitcoin) as the mainnet is going to launch in Nov. Consider header synchronization is fast in the early stage anyway, we decided not to merge this but leave incorporation of flyclient (or better solutions) to future hardforks.

jjyr:
The details of the light client protocol haven't been fully discussed; the chain_root field could be very different depending on the specific design, this field should be introduced via a hard fork after fully discuss a light protocol design.

zmitton:
One of the problems with "shoe-horning" it in later (even as hard fork) is that a FlyProof will never be able to validate the work back to genesis.

For some background, Im a fan of FlyClient, have implemented MMRs in [JS](https://github.com/zmitton/merkle-mountain-range) and [Golang](https://github.com/zmitton/go-merklemountainrange), and would like to see the technology finally brought live (somewhere or another).

I worked past a lot of potential issues in my head, but am not too familiar with Nervos specifics. Would one of you be willing to chat with me for a few minutes about the chain_root field and difficulty adjustment issues?

zmitton:
Or is there a resource you can point me to? The FlyClient paper specifically tackles difficulty adjustment. 

As for the chain_root question, I'm not sure what the problem could be. If you implement the mmr_root as the hash of concatenated peaks, it can be calculated functionally (on-the-fly) in log(n)/2 time (and not saved anywhere). This allows the entire MMR to be append only. Not sure if that helps 🤷

janx:
We like FlyClient too :smile: 

The difficulty adjustment in CKB depends on the number of uncle blocks in an epoch. You can find the details in [Dynamic Difficulty Adjustment Mechanism section of rfc#0020](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#Dynamic-Difficulty-Adjustment-Mechanism) (better to read the whole document).

zmitton:
I think as uncles get committed they would just add to the total difficulty at that point. The merkle-sum tree would work as usual. As long as uncles are included in the header, a flyProof would just have to include any referenced uncles.

janx:
There're two problems to solve when we apply FlyClient to CKB:

1. FlyClient assumes fixed epoch length while NC-Max (the consensus algorithm used in CKB) has variable epoch length;
2. FlyClient assumes original NC style difficulty adjustment, which depends only on timestamps of the first block of each epoch, while difficulty adjustment in NC-Max depends on the number of uncles in the last epoch.

I think we need to improve the Variable Difficulty MMR design in FlyClient paper to accommodate NC-Max, probably adding more auxiliary fields to MMR interval/leaf nodes. A Merkle proof for uncle inclusion of sampled nodes is not enough, because at epoch transition the verifier needs to verify the difficulty adjustment is correct, which means verifying the number of uncles in the last epoch is correct. So certain epoch information should be encoded in MMR nodes.

PR #1599, opened 2019-09-19
feat: Adjust NervosDAO stats calculation logic

xxuejie:


spec/src/consensus.rs
```diff
@@ -132,24 +132,87 @@ impl Default for ConsensusBuilder {
...
+    // Genesis block has special issurance rule, hence we only add up
+    // issurance from block 1 up to the length of proposal window here.
+    let initial_primary_issurance = genesis_epoch_ext
+        .block_reward(0)
+        .and_then(|c| {
+            c.safe_mul(tx_proposal_window.farthest())
+                .map_err(Into::into)
+        })
+        .expect("initial primary issurance calculation overflow!");
+    let initial_secondary_issurance = genesis_epoch_ext
```
doitian: Block 1 ~ 11 only creates cellbase outputs for primary issuance, we don't need to add `10*g2` here

spec/src/consensus.rs
```diff
@@ -132,24 +132,87 @@ impl Default for ConsensusBuilder {
...
+    txs: Vec<&TransactionView>,
+    satoshi_pubkey_hash: &H160,
+    satoshi_cell_occupied_ratio: Ratio,
+    genesis_epoch_ext: &EpochExt,
+    tx_proposal_window: ProposalWindow,
+    secondary_epoch_reward: Capacity,
+) -> Byte32 {
+    // Genesis block has special issurance rule, hence we only add up
+    // issurance from block 1 up to the length of proposal window here.
+    let initial_primary_issurance = genesis_epoch_ext
```
doitian: typo: issuance

util/dao/src/lib.rs
```diff
@@ -99,61 +96,58 @@ impl<'a, CS: ChainStore<'a>> DaoCalculator<'a, CS, DataLoaderWrapper<'a, CS>> {
...
-        let current_ar = u128::from(parent_ar)
-            * u128::from((parent_c.safe_add(parent_g2)?).as_u64())
-            / (max(u128::from(parent_c.as_u64()), 1));
+        let current_c = parent_c.safe_add(current_g)?;
         let current_u = parent_u
             .safe_add(added_occupied_capacities)
             .and_then(|u| u.safe_sub(freed_occupied_capacities))?;
-
-        Ok(pack_dao_data(current_ar as u64, current_c, current_u))
+        let current_s = parent_s
```
doitian: It's better to add first since the interest now includes the current block.

spec/src/consensus.rs
doitian: Since we're going to remove it soon, it's OK to just keep it.

xxuejie: Sorry, not sure I get it: what are we going to remove soon?

util/dao/utils/src/lib.rs
```diff
@@ -19,18 +19,22 @@ pub use crate::error::DaoError;
...
-
-pub const DAO_SIZE: usize = 32;
-
 pub fn genesis_dao_data(txs: Vec<&TransactionView>) -> Result<Byte32, Error> {
-    genesis_dao_data_with_satoshi_gift(txs, &H160([0u8; 20]), Ratio(1, 1))
+    genesis_dao_data_with_satoshi_gift(
+        txs,
+        &H160([0u8; 20]),
+        Ratio(1, 1),
+        capacity_bytes!(1000),
```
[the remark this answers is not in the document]
xxuejie: This is just some dummy values used only in tests, normal code path won't use it. Even tho we fixed it to the correct value, the reward might change, making it invalid again

spec/src/consensus.rs
doitian: Remove the extra 10 blocks rewards.

PR #1600, opened 2019-09-19
chore: System cells force consistent locations

TheWaWaR:


spec/src/lib.rs
```diff
@@ -298,6 +316,23 @@ impl Genesis {
...
+            if data_hashes.get(&hash.pack()) != Some(&(tx_index, output_index)) {
+                return Err(format!(
+                    "Invalid output data for tx-index: {}, output-index: {}, expected data hash: {:x}",
+                    tx_index, output_index,
+                    hash,
+                ));
+            }
+            Ok(())
+        };
+        check_data_hash((0, 1), &CODE_HASH_SECP256K1_BLAKE160_SIGHASH_ALL)?;
```
keroro520: Why not define constants to represent the index of the certain system cells?

TheWaWaR: Good catch

PR #1602, opened 2019-09-19
feat: use zero lock system cells

driftluo:
use zero lock system cells


u2:
bors r=zhangsoledad,u2

PR #1603, opened 2019-09-19
fix: template timestamp should greater than tip timestamp

zhangsoledad:


doitian:
No need to backport

doitian:
bors r=quake,u2,keroro520

PR #1604, opened 2019-09-19
doc: fix the missing parameter of get_live_cell

Keith-CY:


doitian:
This file is generated from JSON

TheWaWaR:
https://github.com/nervosnetwork/ckb/blob/da8897dbc8382293bdf8fadea380a0b79c1efa92/rpc/json/rpc.json#L279

You should change the doc here, then `make ci` to verify if it's correct.

Keith-CY:
> This file is generated from JSON. Please update `rpc/json/rpc.json` and regenerate the file using `make gen-rpc-doc`

The fact is that the interface in `json` is correct while the `doc` is not.

Keith-CY:
> > This file is generated from JSON. Please update `rpc/json/rpc.json` and regenerate the file using `make gen-rpc-doc`
> 
> The fact is that the interface in `json` is correct while the `doc` is not.

I've found the right field in the `json` file

rpc/json/rpc.json
```diff
@@ -272,7 +272,8 @@
         },
         "types": [
             {
-                "out_point": "OutPoint object {\"tx_hash\": <tx_hash>, \"index\": <index>}."
+                "out_point": "OutPoint object {\"tx_hash\": <tx_hash>, \"index\": <index>}.",
+                "with_data": "Boolean"
```
keroro520: Should change to:

```
            {
                "out_point": "OutPoint object {\"tx_hash\": <tx_hash>, \"index\": <index>}."
            },
            {
                "with_data": "Boolean"
            }
```

Please take a look [here](https://github.com/nervosnetwork/ckb/blob/9bbd2110be80347a1db6ac8deb161a12a3b40200/devtools/doc/rpc.py#L31). This assertion will prevent your present unexpected format.
It is better to print the failed reason and what is the expected format. Could you help us to improve it?

quake:
bors r=keroro520,u2,quake

PR #1607, opened 2019-09-19
feat(test): Print panicked logs when node panicked

keroro520:
* feat(test): Print panicked logs when node panicked
* chore: tweak based Clippy warning `the function has a cognitive complexity`


quake:
bors r=u2,quake

PR #1609, opened 2019-09-19
fix: Use dao type script hash in dao transaction

TheWaWaR:
Dependent on this PR https://github.com/nervosnetwork/ckb/pull/1600

### DIFF
--- a/spec/src/consensus.rs
+++ b/spec/src/consensus.rs
@@ -1,3 +1,7 @@
+use crate::{
+    OUTPUT_INDEX_DAO, OUTPUT_INDEX_SECP256K1_BLAKE160_SIGHASH_ALL,
+    OUTPUT_INDEX_SECP256K1_RIPEMD160_SHA256_SIGHASH_ALL,
+};
 use ckb_dao_utils::genesis_dao_data_with_satoshi_gift;
 use ckb_pow::{Pow, PowEngine};
 use ckb_rational::RationalU256;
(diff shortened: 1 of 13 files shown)


test/src/node.rs
```diff
@@ -330,7 +330,7 @@ impl Node {
         since: u64,
         capacity: Capacity,
     ) -> TransactionView {
-        let always_success_out_point = OutPoint::new(self.genesis_cellbase_hash.clone(), 1);
+        let always_success_out_point = OutPoint::new(self.genesis_cellbase_hash.clone(), 5);
```
u2: Use `SYSTEM_CELL_ALWAYS_SUCCESS_INDEX` instead of `5`

test/src/node.rs
```diff
@@ -384,7 +388,7 @@ impl Node {
         self.always_success_code_hash = CellOutput::calc_data_hash(
             &consensus.genesis_block().transactions()[0]
                 .outputs_data()
-                .get(1)
+                .get(5)
```
u2: SYSTEM_CELL_ALWAYS_SUCCESS_INDEX

u2:
bors r=quake,u2

spec/src/consensus.rs
```diff
@@ -227,8 +231,27 @@ impl ConsensusBuilder {
...
+
+        let genesis_cellbase = genesis_block
+            .transaction(0)
+            .expect("Genesis must have cellbase");
+        let get_type_hash = |output_index: u64| {
+            genesis_cellbase
+                .output(output_index as usize)
+                .and_then(|output| output.type_().to_opt())
+                .map(|type_script| type_script.calc_script_hash())
+                .unwrap_or_default()
```
keroro520: Better `expect` here but not return a wrong hash value.

keroro520:
@TheWaWaR resolve conflict

PR #1613, opened 2019-09-20
fix: use serialized_size while calculate_txs_size_limit

u2:
Now for the blank block, the `serialized_size` and `as_slice().len()` are the same, but I think we should use `serialized_size`.


doitian:
@yangby-cryptape Please help to review this PR.

yangby-cryptape:
Rebase is required, since #1621 was merged.

u2:
> Rebase is required, since #1621 was merged.

rebased

u2:
bors r=quake,yangby-cryptape

PR #1614, opened 2019-09-21
Fix tx fee calculation overflow

jjyr:


jjyr:
> Please use u128 and add unit test for this bug

No reason to use u128, what if u128 still overflow? use u256?

quake:
bors r=u2,quake

jjyr:
bors r=u2,quake

PR #1617, opened 2019-09-23
feat: setup issuance schedule

doitian:
BREAKING CHANGE: primary/secondary epoch reward has changed


zhangsoledad:
bors r=zhangsoledad,xxuejie

PR #1619, opened 2019-09-23
fix: peer store persistent

jjyr:
This error cause peer store persistent error:
1. peer store will not persistent at exit(will still persistent to file every 1 hour when the node is running).
2. peer store persistent failed if `tmp dir` and `ckb dir` use a different file system.

A node will connect to bootnodes on start if the above error happened, other functions are not affected.

* Fix dump error when tmp files use a different file system [detail](https://stackoverflow.com/questions/24209886/invalid-cross-device-link-error-with-boost-filesystem?answertab=votes#tab-top).
* Fix peer store do not dump data on exit.

### DIFF
--- a/network/src/config.rs
+++ b/network/src/config.rs
@@ -64,7 +64,7 @@ impl NetworkConfig {
 
     pub fn peer_store_path(&self) -> PathBuf {
         let mut path = self.path.clone();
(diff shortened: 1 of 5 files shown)


quake:
bors r=quake,u2

PR #1620, opened 2019-09-23
doc: Note generating via `make gen-rpc-doc`

keroro520:
doc: Note generating via `make gen-rpc-doc`


u2:
bors r=doitian,u2,yangby-cryptape

doitian:
bors r=doitian,u2,yangby-cryptape

PR #1622, opened 2019-09-23
fix: default executor misused

zhangsoledad:


quake:
bors r=quake,driftluo

PR #1627, opened 2019-09-24
chore: manage internal error

keroro520:
* chore: Use `failure::Context` to manage internal error
* chore: Remove derivation from Fail for error_kinds

NOTE: The previous implementation shows incomplete error messages. Here is an example:

```
Before: Internal(Database)

Now: Internal(Database(failed to open the database: Invalid argument: You have to open all column families. Column families not opened: 12))
```


doitian:
bors r=zhangsoledad,doitian

PR #1628, opened 2019-09-24
doc: Correct result of get_block_template

keroro520:
* The result of `get_block_template` in RPC document is wrong


zhangsoledad:
bors r=zhangsoledad,quake

PR #1629, opened 2019-09-24
fix: fix peer store fetch_random return empty

jjyr:


quake:
bors r=zhangsoledad,driftluo

jjyr:
bors r=zhangsoledad,driftluo

PR #1632, opened 2019-09-24
refactor: change args and witness to single bytes

quake:
1. Change args and witness from `Vec<Bytes>` to `Bytes`
2. Add `load_script` system call


script/src/verify.rs
```diff
@@ -591,23 +597,23 @@ mod tests {
 
         // Not enough cycles
         assert_error_eq(
-            verifier.verify(100).unwrap_err(),
+            verifier.verify(11).unwrap_err(),
```
keroro520: FYI, we should define these cycles as meaningful constant variables later.

script/src/verify.rs
```diff
@@ -761,97 +767,21 @@ mod tests {
         );
     }
 
-    #[test]
```
keroro520: Why remove it?

quake: This unit test case is trying to run a script with low cycle, it's covered in `check_signature` line 600

script/src/verify.rs
quake: Updated, please review again, thanks.

quake:
bors r=doitian,xxuejie,keroro520

PR #1642, opened 2019-09-25
test: Pack uncles during switching epoch and reorg

keroro520:
* test: A uncle inherited from a fork-block in side fork is invalid
* test: A uncle inherited from a fork-uncle in side fork is invalid
* test: Miner should not add uncles into the epoch starting
* test: Different forks contain a same uncle

### DIFF
--- a/test/src/specs/mining/uncle.rs
+++ b/test/src/specs/mining/uncle.rs
@@ -19,7 +19,7 @@ impl Spec for UncleInheritFromForkBlock {
     // Case: A uncle inherited from a fork-block in side fork is invalid, because that breaks
     //       the uncle rule "B1's parent is either B2's ancestor or embedded in B2 or its ancestors
     //       as an uncle"
-    fn run(&self, net: Net) {
+    fn run(&self, net: &mut Net) {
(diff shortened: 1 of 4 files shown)


test/src/specs/mining/uncle.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        info!("(4) Add all the fork-blocks as uncle blocks into the chain. And now re-submit block with `uncle_child` should be success");
+        until_no_uncles_left(target_node);
+        let block = target_node
+            .new_block_builder(None, None, None)
+            .set_uncles(vec![uncle_child.as_uncle()])
+            .build();
+        target_node.submit_block(&block.data());
+    }
+}
```
u2: It's not easy to get all the things of the above two test specs, what about adding an ASCII graph.

keroro520: Many of integration-tests try to simulate complicated scenarios.
Yes, I will consider adding more descriptions to complicated cases. And ASCII graph is a good idea.

keroro520:
local CI passed.

PR #1643, opened 2019-09-25
compress header

zhangsoledad:
* remove uncles_count
* merge witnesses_root & transaction_root
* replace difficulty with compact_target


util/types/src/utilities/difficulty.rs
```diff
@@ -30,9 +30,104 @@ pub fn difficulty_to_target(difficulty: &U256) -> Byte32 {
...
 }
 
+/**
+* The "compact" format is a representation of a whole
+* number N using an unsigned 32bit number similar to a
+* floating point format.
+* The most significant 8 bits are the unsigned exponent of base 256.
+* This exponent can be thought of as "number of bytes of N".
+* The lower 23 bits are the mantissa.
+* Bit number 24 (0x800000) represents the sign of N.
```
doitian: We can use mantissa as an unsigned number.

util/types/src/utilities/difficulty.rs
```diff
@@ -30,9 +30,104 @@ pub fn difficulty_to_target(difficulty: &U256) -> Byte32 {
...
+* N = (-1^sign) * mantissa * 256^(exponent-3)
+*/
+fn bits(target: &U256) -> u32 {
+    let bytes = target.to_le_bytes();
+    for (pos, chunk) in bytes.chunks(4).enumerate().rev() {
+        let pn = u32::from_le_bytes(chunk.try_into().expect("bound checked"));
+        if pn > 0 {
+            for bits in (0..31).rev() {
```
doitian: There's a `leading_zeros` function.

benches/benches/benchmarks/util.rs
```diff
@@ -60,7 +60,7 @@ pub fn new_always_success_chain(txs_size: usize, chains_num: usize) -> Chains {
 
     let genesis_block = BlockBuilder::default()
         .dao(dao)
-        .difficulty(U256::from(1000u64).pack())
+        .compact_target(0x1f41_8937u32.pack()) // diff 1000
```
doitian: I think it's better to use a method like `diff_to_compact_target` to avoid the magic number and the comment.

util/types/src/utilities/difficulty.rs
```diff
@@ -1,58 +1,208 @@
...
-        boundary
-            .into_big_endian(&mut inner)
-            .expect("convert from U256 to Byte32");
-        Byte32::new_unchecked(inner.into())
+        ret = mantissa.clone();
+        ret <<= 8 * (exponent - 3);
+    }
+
+    let overflow = !mantissa.is_zero()
+        && ((exponent > 34) // > (256 ** 31)
```
u2: If `exponent > 32`, it must be overflow.

```rust
        let compact = target_to_compact(u256!(
            "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
        ));
        println!("{:#x}", compact);
        let (_, overflow) = compact_to_target(compact);
        assert_eq!(overflow, false, "should not overflow");
```

This test case should no fail.

verification/src/contextual_block_verifier.rs
```diff
@@ -471,21 +513,41 @@ impl<'a, CS: ChainStore<'a>> ContextualBlockVerifier<'a, CS> {
             prepare_epoch_ext(&self.context, &parent)?
         };
```
jjyr: Do we need to verify `header number == parent number + 1` in ContextialBlockVerifier?

zhangsoledad: No
HeaderVerifier -> BlockVerifier(non-contextial) -> ContextialBlockVerifier 
this order must be guarantee.

util/types/src/utilities/difficulty.rs
```diff
@@ -1,58 +1,226 @@
...
+            || (mantissa > U256::from(0xffu32) && exponent > 33)
+            || (mantissa > U256::from(0xffffu32) && exponent > 32));
+    (ret, overflow)
+}
```
u2: I am not sure whether we should add some validity check for the compact in `compact_to_target`, or we have already checked it at somewhere.

```rust
    fn test_compact_to_target2() {
        let c1 = 0x21000001;
        let (x, overflow) = compact_to_target(c1);
        println!("{:#x}", x);
        assert_eq!(overflow, false, "should not overflow");

        let c = target_to_compact(x);
        println!("{:#x}", c);
        assert_eq!(c, c1);
    }
```

This test case will fail because even the compact is not overflowed, this compact is not a rightful number which could convert from the target.

zhangsoledad: target is verified by consensus rule,  
so there is no correctness issue here.

this question depends on how you define overflow.
if consider compact_to_target individually, treat compact as any 32-bit integer, so the limit is 256U::max
if consider precision and target_to_compact, 0x21.. should not appear.

zhangsoledad:
@u2 your opinion is more reasonable.
`compact_to_target`, not `compact_encode`,  it's not a general encode. 
I apply it after thinking

zhangsoledad:
bors r=u2,doitian

zhangsoledad:
https://github.com/nervosnetwork/ckb-system-scripts/pull/50

zhangsoledad:
bors retry

zhangsoledad:
bors r=u2,doitian

PR #1644, opened 2019-09-25
Fix duplicate p2p phase in get_peers

jjyr:


quake:
bors r=quake,driftluo

PR #1645, opened 2019-09-25
Feat min transaction fee filter

jjyr:
This PR enable min fee rate filter to ckb.

This PR adds a `tx_pool.min_fee_rate` option in `ckb.toml`, the default value is `1_000` shannons/KB, txs that with a fee rate lower than this will be ignored by the node.

The intention of the feature:
1. allow miners control which txs be mined by set tx's min fee rate
2. to prevent zero-fee or extremely low fee spam txs occupy the network

The min fee rate is not a part of the consensus, a node can set the `tx_pool.min_fee_rate` to a lower value than `1000` or even `0`.

Changes:

* Cache tx fee
* Add a new option to `ckb.toml`: tx_pool.min_fee_rate = 1_000 # shannons/KB
* Return error in RPC `send_transaction` when tx fee rate is lower than `min_fee_rate`
* txs propose and commit will ignores txs that fee rate lower than `min_fee_rate`
* txs will not relay by node if it's fee rate is lower than `min_fee_rate`


test/src/specs/relay/transaction_relay_low_fee_rate.rs
```diff
@@ -0,0 +1,96 @@
...
+            .broadcast_transaction(tx_low_fee.data().into(), cycles)
+            .unwrap();
+
+        info!("Waiting for relay");
+        let rpc_client = node0.rpc_client();
+        let ret = wait_until(10, || rpc_client.get_transaction(hash.pack()).is_some());
+        assert!(ret, "Transaction should be broadcast to node0");
+
+        let rpc_client = node2.rpc_client();
+        let ret = wait_until(1, || rpc_client.get_transaction(hash.pack()).is_some());
```
u2: Two questions:
- why only to wait for 1 second here?
- why the tx should not be relayed to node2?

test/src/specs/tx_pool/send_low_fee_rate_tx.rs
```diff
@@ -0,0 +1,85 @@
...
+            .build();
+        let tx_high_fee = tx_high_fee
+            .data()
+            .as_advanced_builder()
+            .set_outputs(vec![])
+            .output(output)
+            .build();
+        node0
+            .rpc_client()
+            .send_transaction(tx_high_fee.data().into());
```
u2: Add an `assert` for the high fee tx?

u2:
One question: If the miner updates the min-fee-rate config while node running, the miner needs to restart the node?

jjyr:
> One question: If the miner updates the min-fee-rate config while node running, the miner needs to restart the node?

The only way to change the `min-fee-rate` is through `ckb.toml`, so it is.

test/src/specs/relay/transaction_relay_low_fee_rate.rs
jjyr: * This case is a failure case, so wait for more seconds just increase the waiting time.
* tx fee is lower than min-fee-rate

test/src/specs/tx_pool/send_low_fee_rate_tx.rs
jjyr: `send_transaction ` panic on error, so we do not need `assert`.

test/src/specs/relay/transaction_relay_low_fee_rate.rs
u2: The three nodes have the same `min_fee_rate` configuration, but the tx could be relayed to node0.

jjyr: `broadcast_transaction ` is not `relay protocol`, it's an interface for integration test.

rpc/src/module/pool.rs
```diff
@@ -48,7 +59,13 @@ impl PoolRpc for PoolRpcImpl {
...
 
         match submit_txs.unwrap() {
-            Ok(_) => {
+            Ok(submit_result) => {
+                let cache_entry = submit_result.get(0).expect("exists");
+                let min_fee = self.min_fee_rate.fee(tx.serialized_size());
+                // reject txs which fee lower than min fee rate
+                if cache_entry.fee < min_fee {
+                    return Err(RPCError::custom(RPCError::Invalid, format!("transaction fee rate lower than min_fee_rate: {} shannons/KB. hint: at least pay {} shannons fee", self.min_fee_rate, min_fee)));
+                }
```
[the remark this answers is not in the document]
zhangsoledad: Additional, `add_tx_to_pool` is no longer a public interface. All public interface is listed under process path belong with service, we just remain some legacy code.

verification/src/transaction_verifier.rs
```diff
@@ -96,10 +99,11 @@ where
                 epoch_number_with_fraction,
                 parent_hash,
             ),
+            fee_calculator: FeeCalculator::new(rtx, &consensus, &chain_store),
         }
     }
 
-    pub fn verify(&self, max_cycles: Cycle) -> Result<Cycle, Error> {
+    pub fn verify(&self, max_cycles: Cycle) -> Result<CacheEntry, Error> {
```
zhangsoledad: https://github.com/nervosnetwork/ckb/blob/develop/chain/src/chain.rs#L631-L636
develop branch move out calculate fee from `Verifier`,
you need to merge this.

verification/src/contextual_block_verifier.rs
```diff
@@ -550,19 +554,13 @@ impl<'a, CS: ChainStore<'a>> ContextualBlockVerifier<'a, CS> {
...
-                parent_hash,
-                resolved,
-            )
-            .verify(txs_verify_cache, executor)?
-        } else {
-            0
-        };
-
-        Ok(cycles)
+        BlockTxsVerifier::new(
```
quake: rebase error? I think that `switch` is introduced in https://github.com/nervosnetwork/ckb/pull/1643

jjyr: No, since we put the tx fee in the verifier cache, the `disable_txs` is no-sense, the `dao` will not calculate correctly if we `disable_txs`. furthermore I don't see this flag enabled anywhere.

jjyr: IMO we can add it back when we really need the flag.

PR #1649, opened 2019-09-26
test: add whitelist test

driftluo:
add whitelist test

### DIFF
--- a/test/src/specs/sync/ibd_process.rs
+++ b/test/src/specs/sync/ibd_process.rs
@@ -123,7 +123,8 @@ impl Spec for IBDProcessWithWhiteList {
         node0.connect(node3);
         node0.connect(node4);
 
-        // will never connect
+        // will never connect, protect node default is 4, see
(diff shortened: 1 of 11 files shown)


test/src/specs/sync/ibd_process.rs
```diff
@@ -81,3 +81,101 @@ impl Spec for IBDProcess {
...
+        let node4 = &net.nodes[4];
+        let node5 = &net.nodes[5];
+        let node6 = &net.nodes[6];
+
+        node0.connect(node1);
+        node0.connect(node2);
+        node0.connect(node3);
+        node0.connect(node4);
+
+        // will never connect
```
jjyr: I am confused by this comment, why it will never connect?

driftluo: Because it is not protected/whitelist node and it is on ibd mode, so, the outbound session will be disconnected by node0

jjyr: I see, it's better to use the explicit constant and comments https://github.com/nervosnetwork/ckb/blob/da8897dbc8382293bdf8fadea380a0b79c1efa92/sync/src/lib.rs#L57 to offload the mind.

PR #1652, opened 2019-09-26
chore: remove dup deps in Cargo.toml

u2:
`dev-dependencies` does not need to keep the crates in `dependencies`.

### DIFF
--- a/pow/Cargo.toml
+++ b/pow/Cargo.toml
@@ -12,6 +12,3 @@ ckb-hash = { path = "../util/hash"}
 serde = "1.0"
 serde_derive = "1.0"
 eaglesong = "0.1"
-
-[dev-dependencies]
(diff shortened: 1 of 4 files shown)


u2:
https://github.com/est31/cargo-udeps/issues/27

https://github.com/est31/cargo-udeps/issues/28

pow/Cargo.toml
```diff
@@ -12,6 +12,3 @@ ckb-hash = { path = "../util/hash"}
 serde = "1.0"
 serde_derive = "1.0"
 eaglesong = "0.1"
-
-[dev-dependencies]
-ckb-hash = { path = "../util/hash" }
```
yangby-cryptape: You should keep this line, and remove the line 11 in this file.

PR #1657, opened 2019-09-27
chore: update block assembler template note

shaojunda:


zhangsoledad:
bors r=zhangsoledad,quake

PR #1660, opened 2019-09-28
fix: json type number must use hex string

driftluo:
json type number must use hex string


zhangsoledad:
bors r=zhangsoledad,quake

PR #1664, opened 2019-09-29
chore: update get_block_template rpc doc

shaojunda:
replace difficulty with compact_target


u2:
bors r=quake,u2,keroro520,yangby-cryptape

PR #1666, opened 2019-09-29
feat: expand nonce to 128-bit

zhangsoledad:
### breaking change
* expand nonce to 128-bit
* change pow_message from [nonce + pow_hash] to [pow_hash + nonce ]


zhangsoledad:
https://github.com/nervosnetwork/ckb-system-scripts/pull/52/files

util/jsonrpc-types/src/uint128.rs
```diff
@@ -0,0 +1,158 @@
...
+        write!(formatter, "a hex-encoded, 0x-prefixed uint128")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        if !value.starts_with("0x") {
+            return Err(Error::custom(format!(
+                "Invalid uint64 {}: without `0x` prefix",
```
u2: invalid u128?

util/jsonrpc-types/src/uint128.rs
```diff
@@ -0,0 +1,158 @@
...
+                value
+            )));
+        }
+
+        let number = u128::from_str_radix(&value[2..], 16)
+            .map(Uint128)
+            .map_err(|e| Error::custom(format!("Invalid uint64 {}: {}", value, e)))?;
```
u2: invalid u128?

util/jsonrpc-types/src/uint128.rs
```diff
@@ -0,0 +1,158 @@
...
+        if number.to_string() != value {
+            return Err(Error::custom(format!(
+                "Invalid uint64 {}: with redundant leading zeros, expected: {}",
```
u2: invalid u128?

spec/src/lib.rs
```diff
@@ -99,7 +99,7 @@ pub struct Genesis {
     pub compact_target: u32,
     pub uncles_hash: H256,
     pub hash: Option<H256>,
-    pub nonce: u64,
+    pub nonce: u64, // u128 is not supported by https://github.com/alexcrichton/toml-rs
```
driftluo: We can use `U128` on `numext-fixed-uint` instead, it supports `serde` and accurately express u128

sync/src/synchronizer/mod.rs
```diff
@@ -675,11 +675,11 @@ mod tests {
         let num = 200;
```
driftluo: let num = 200u128;

zhangsoledad: this doesn't help, 
in those test will also take the `num` as `BlockNumber`,  cast can't be avoided.
```
fn insert_block(
        chain_controller: &ChainController,
        shared: &Shared,
        nonce: u128,
        number: BlockNumber,
) 
```

PR #1671, opened 2019-09-30
test: generated blocks in the connected state

driftluo:
generated blocks in the connected state

### DIFF
--- a/test/src/specs/sync/ibd_process.rs
+++ b/test/src/specs/sync/ibd_process.rs
@@ -138,7 +138,9 @@ impl Spec for IBDProcessWithWhiteList {
             panic!("refuse to connect fail");
         }
 
-        node6.disconnect(node0);
+        // After the whitelist is disconnected, it will always try to reconnect.
+        // In order to ensure that the node6 has already generated two blocks when reconnecting,
+        // it must be in the connected state, and then disconnected.
         node6.generate_blocks(2);
 
         let generate_res = wait_until(10, || net.nodes[6].get_tip_block_number() == 2);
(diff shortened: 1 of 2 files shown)


test/src/specs/sync/ibd_process.rs
```diff
@@ -147,7 +147,16 @@ impl Spec for IBDProcessWithWhiteList {
             panic!("node6 can't generate blocks to 2");
         }
 
-        node0.connect_uncheck(node6);
+        let is_disconnect_already = wait_until(10, || {
```
keroro520: Why this checking matter? Better adds some comments to explain the reason.

keroro520: Sorry that I still don't understand. What if they don't disconnect each other?

driftluo: If it is not possible to confirm that both parties have been disconnected, Node0 and node6 will not synchronize during the test time, because node0 has confirmed that node6 is in the ibd state.

quake:
bors r=u2,keroro520,quake

u2:
bors r=u2,keroro520,quake

driftluo:
There is a neglected problem here, causing random errors in the test.

**Whitelist will auto connect by network**, if miss this information in the test, there will be various strange problems.

driftluo:
bors r=u2,keroro520,quake

driftluo:
> > ckb-failed on spec whitelist_on_session_limit
> 
> https://travis-ci.org/keroro520/ckb/jobs/596414021
> 
> I guess that the above failure is caused by the same reason as this PR. Please check again.

The cause of the problem is not consistent, it is an asynchronous timing problem, I will fix it....

PR #1674, opened 2019-10-02
fix: Allow put a message in cellbase witness

TheWaWaR:


jjyr:
https://github.com/nervosnetwork/ckb-system-scripts/pull/54 introduce WitnessArgs `lock/type/extra`.

Can you rename the `message` to `extra` to keep consistency, basically the two fields describe the samething.

zhangsoledad:
bors r=quake,xxuejie

PR #1678, opened 2019-10-04
[ᚬrc/v0.22] fix: get_block_transactions_process should fill missing uncle in resp…

zhangsoledad:
fix: get_block_transactions_process should fill missing uncle in response

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -226,8 +226,6 @@ impl<'a> CompactBlockProcess<'a> {
                 }
             }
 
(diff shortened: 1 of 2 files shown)


sync/src/relayer/compact_block_process.rs
```diff
@@ -226,8 +226,6 @@ impl<'a> CompactBlockProcess<'a> {
                 }
             }
 
-            assert!(!missing_transactions.is_empty() || !missing_uncles.is_empty());
```
doitian: What happens if code runs here and both `missing_transactions` and `missing_uncles` are empty?

PR #1680, opened 2019-10-04
fix: get_block_transactions_process

zhangsoledad:


sync/src/relayer/compact_block_process.rs
```diff
@@ -226,8 +226,6 @@ impl<'a> CompactBlockProcess<'a> {
                 }
             }
 
-            assert!(!missing_transactions.is_empty() || !missing_uncles.is_empty());
```
u2: If there are no missing transactions and missing uncles, it should return `Ok(Status::AcceptBlock);` https://github.com/nervosnetwork/ckb/pull/1680/files#diff-98eccb89a5611099d6adca819c9c8b08L202

zhangsoledad:
```

u2: If this is a collision, the `missing_transactions` is not empty.

quake:
bors r=u2,quake

zhangsoledad:
bors retry

PR #1682, opened 2019-10-04
feat: Allow setting the spec file in ckb init.

doitian:
Add a new option `ckb init --import-spec <file>`.

It allows copying the spec file in `ckb init` from `<file>`. Specially, when `<file>` is "-", read the spec file from stdin encoded in base64.


doitian:
bors r=quake,u2

u2:
bors r-

conflict

doitian:
bors r=quake,u2

PR #1683, opened 2019-10-08
fix: Remove descendants of committed txs from pending pool

keroro520:
* fix(tx-pool): Remove descendants of committed txs from pending-pool
* test(tx_pool): Add case of handling propose


test/src/specs/tx_pool/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+        self.get(3)
+    }
+
+    #[allow(dead_code)]
+    pub fn e(&self) -> &TransactionView {
+        self.get(4)
+    }
+}
+
+pub fn prepare_txfamily(node: &Node) -> TxFamily {
```
u2: prepare_tx_family

test/src/specs/tx_pool/descendant.rs
```diff
@@ -0,0 +1,261 @@
...
+                block
+                    .union_proposal_ids()
+                    .contains(&family.b().proposal_short_id()),
+                "Miner should propose tx_family.b since it has never been proposed, actual: {:?}",
+                block.union_proposal_ids(),
+            );
+
+            node.submit_block(&blank(node)); // continuously submit blank blocks.
+        });
+        (0..10).for_each(|_| {
```
u2: What's the `10` represents for?

test/src/specs/tx_pool/descendant.rs
```diff
@@ -0,0 +1,261 @@
...
+            let block = node.new_block(None, None, None);
+            assert_eq!(
+                vec![
+                    family.a().proposal_short_id(),
```
u2: Why does the new block propose `a` again after it is proposed before?

test/src/specs/tx_pool/descendant.rs
```diff
@@ -0,0 +1,261 @@
...
+    // Case: This case intends to test that submit a transaction which its parent is in gap-pool
+    fn run(&self, net: &mut Net) {
+        let node = &net.nodes[0];
+        let window = node.consensus().tx_proposal_window();
+        node.generate_blocks((window.farthest() + 2) as usize);
+
+        // 1. Propose `tx_family.a` into gap-pool.
+        let family = prepare_txfamily(node);
+        node.submit_transaction(family.a());
+        node.submit_block(&propose(node, &[family.a()]));
```
u2: I think after a new block, the `a` will be in the gap-pool.

keroro520: Yes. But here explicitly proposing is better.

test/src/specs/tx_pool/descendant.rs
keroro520: It is a meaningless number.
I have changed it. Please have a look.

test/src/specs/tx_pool/descendant.rs
u2: I meant that one more block after proposing.

u2:
bors r=quake,u2

PR #1684, opened 2019-10-09
fix(test): First 11 cellbases have not outputs

keroro520:
https://github.com/nervosnetwork/ckb/pull/1605 did not correctly resolve conflict before merging. Hence it breaks CI for develop-branch. Here comes this PR to fix it.

PR #1685, opened 2019-10-09
fix: fix construct_uncle in integration test

u2:
Fix panic of ForksContainSameUncle. The reason is that the uncle block has the same hash with the main chain block.

### DIFF
--- a/test/src/node.rs
+++ b/test/src/node.rs
@@ -274,6 +274,22 @@ impl Node {
         self.submit_block(&self.new_block(None, None, None).data())
     }
 
+    // Convenient way to construct an uncle block
+    pub fn construct_uncle(&self) -> BlockView {
+        let block = self.new_block(None, None, None);
+        // Make sure the uncle block timestamp is different from
+        // the next block timestamp in main fork.
+        // Firstly construct uncle block which timestamp
+        // is less than the current time, and then generate
+        // the new block in main fork which timestamp is greater than
+        // or equal to the current time.
+        let timestamp = block.timestamp() - 1;
+        block
+            .as_advanced_builder()
+            .timestamp(timestamp.pack())
+            .build()
+    }
+
     // generate a transaction which spend tip block's cellbase and send it to pool through rpc.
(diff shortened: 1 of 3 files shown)


u2:
> I don't get the reason why the generated uncle block has the same hash with the main chain block?
> As I understand, if two blocks have the same hash value, then their heights and timestamp should be the same too; but the problem is when the uncle and main-block have the same height, [`UncleVerifier` should report `UnclesError::InvalidNumber`](https://github.com/nervosnetwork/ckb/blob/9c60a0f00d8a7ba82e87cc2f0553e5df467619aa/verification/src/uncles_verifier.rs#L92) but not `UnclesError::DoubleInclusion`.
> 
> Could you please give an example?

```
    //                     /-> A(U)
    // genesis -> 1 -> 2 ->
    //             \       \-> B(U)
    //              \-> U
```
The problem is that uncle `U` has the same timestamep with block `2`. If you print out the timestamps of uncle block and the  node_a's next block, you will find that the timestamp is very near. 

In my local computer

```
[2019-10-09T01:35:35Z INFO  ckb_test::specs::sync::chain_forks] timestamp 1570584935849
[2019-10-09T01:35:35Z INFO  ckb_test::node] bv.timestamp 1570584935843
```
They are easily collisional.

If you modify this part of code 
https://github.com/nervosnetwork/ckb/blob/develop/test/src/specs/sync/chain_forks.rs#L586-L589

```
        info!("(1) Construct an uncle before fork point");
        let uncle = construct_uncle(node_a);
        node_a.generate_block();
        node_b.generate_block();
```
and make the `uncle` and `node_a.generate_block()` have the same timestamps, the `UnclesError::DoubleInclusion` error will always occur.

doitian:
It is better to add an ID in the cellbase witness instead of relying on different timestamps.

u2:
bors r-

will conflict with https://github.com/nervosnetwork/ckb/pull/1684

u2:
bors r=quake,keroro520

PR #1696, opened 2019-10-10
fix: set `next_epoch_diff` to one instead of panic when it is zero

doitian:


spec/src/consensus.rs
```diff
@@ -622,35 +622,34 @@ impl Consensus {
...
             }
         } else {
-            let denominator = (orphan_rate_target + U256::one()) * next_epoch_length_u256;
-            (diff_numerator / denominator).into_u256()
+            (orphan_rate_target + U256::one()) * next_epoch_length_u256
         };
 
-        debug_assert!(
-            next_epoch_diff > U256::zero(),
-            "next_epoch_diff should greater than one"
```
u2: I don't understand the before code. why it's `should greater than one` while the code is `> U256::zero()`? I think it should be `greater than zero` or `greater than or equal to one`.

u2:
bors r=zhangsoledad,u2

spec/src/consensus.rs
doitian: The comment is wrong, diff cannot be zero.

PR #1700, opened 2019-10-10
fix: fix panic when submit tx with low fee rate

jjyr:


zhangsoledad:
bors r=xxuejie,u2,doitian

PR #1702, opened 2019-10-10
test: fix compact_block_missing_fresh_txs integration test

u2:
There are some relayer-notify jobs in sync, see https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/mod.rs#L44-L46. This would dislocate the message order.

### DIFF
--- a/test/src/net.rs
+++ b/test/src/net.rs
@@ -217,6 +217,23 @@ impl Net {
     pub fn receive_timeout(&self, timeout: Duration) -> Result<NetMessage, RecvTimeoutError> {
         self.controller.as_ref().unwrap().1.recv_timeout(timeout)
     }
+
+    pub fn should_receive<F>(&self, f: F, message: &str) -> PeerIndex
+    where
+        F: Fn(&Bytes) -> bool,
+    {
+        let mut peer_id: PeerIndex = Default::default();
+        let received = wait_until(30, || {
+            let (receive_peer_id, _, data) = self
+                .receive_timeout(Duration::from_secs(30))
+                .expect("receive msg");
+            peer_id = receive_peer_id;
+            f(&data)
+        });
+
+        assert!(received, message.to_string());
+        peer_id
+    }
 }
 
 pub struct DummyProtocolHandler {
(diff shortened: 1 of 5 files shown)


test/src/specs/relay/compact_block.rs
```diff
@@ -176,12 +177,28 @@ impl Spec for CompactBlockMissingFreshTxs {
...
-        let message = RelayMessage::from_slice(&data).unwrap();
-        assert_eq!(
-            message.to_enum().item_name(),
-            packed::GetBlockTransactions::NAME,
-            "Node0 should send GetBlockTransactions message for missing transactions",
+        let mut received = false;
+
+        // There is some relayer notify job in sync
+        // see: https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/mod.rs#L44-L46
+        for _ in 0..=3 {
```
keroro520: It is better to use `wait_until`.

```rust
let received_get_block_transactions = wait_until(10s, || {
  let (_, _, data) = net.receive_timeout(10s);
  let message = RelayMessage::from_slice(&data).unwrap();
  message.to_enum().item_name() ==  packed::GetBlockTransactions::NAME
});
```

u2: updated.

test/src/net.rs
```diff
@@ -217,6 +217,23 @@ impl Net {
...
     }
+
+    pub fn should_receive<F>(&self, f: F, message: String) -> PeerIndex
+    where
+        F: Fn(&Bytes) -> bool,
+    {
+        let mut peer_id: PeerIndex = Default::default();
+        let received = wait_until(10, || {
+            let (receive_peer_id, _, data) = self
+                .receive_timeout(Duration::from_secs(3))
```
keroro520: 3s is too short to run on azure machines... maybe 30s is safer

test/src/net.rs
```diff
@@ -217,6 +217,23 @@ impl Net {
     pub fn receive_timeout(&self, timeout: Duration) -> Result<NetMessage, RecvTimeoutError> {
         self.controller.as_ref().unwrap().1.recv_timeout(timeout)
     }
+
+    pub fn should_receive<F>(&self, f: F, message: String) -> PeerIndex
```
keroro520: Suggest change the message type to ‘ToString’ or other convenient types.

u2:
bors r=quake,driftluo,keroro520

PR #1705, opened 2019-10-11
feat: verify genesis block specific rules on start

jjyr:
* Add more genesis block specific verify cases
* Add a test to verify embed chain specs


verification/src/genesis_verifier.rs
```diff
@@ -0,0 +1,213 @@
...
+        UnclesVerifier::new().verify(consensus.genesis_block())?;
+        DAOVerifier::new(consensus).verify(consensus.genesis_block())?;
+        BlockVerifier::new(consensus).verify(consensus.genesis_block())
+    }
+}
+
+#[derive(Clone)]
+pub struct NumberVerifier {}
+
+impl Default for NumberVerifier {
```
quake: Please remove all `Default` trait, you should only add it when needed

jjyr: This is only for eliminate the warning from clippy.

quake: `NumberVerifier` and other verifiers are empty struct, no need to impl `new` fn and `Default` trait, you can use it directly:
```rust
pub struct NumberVerifier;

impl NumberVerifier {
    pub fn verify(&self, block: &BlockView) -> Result<(), Error> {
        ...
    }
}

NumberVerifier.verify(consensus.genesis_block())?;
```

btw, I prefer static method to instance method, since these verifiers are non-contextual and empty structs.
```rust
pub struct NumberVerifier;

impl NumberVerifier {
    pub fn verify(block: &BlockView) -> Result<(), Error> {
        ...
    }
}

NumberVerifier::verify(consensus.genesis_block())?;
```

jjyr: You are right, updated.

jjyr:
@quake @keroro520 rebased, Please re-approve

jjyr:
bors r=keroro520, quake

jjyr:
bors r=keroro520, quake

PR #1707, opened 2019-10-11
fix: resolve uncles hash calculation issue

quake:
Uncles hash is the blake2b on concatenated uncle block hashes.

### DIFF
--- a/util/types/src/extension/calc_hash.rs
+++ b/util/types/src/extension/calc_hash.rs
@@ -124,7 +124,13 @@ impl<'r> packed::UncleBlockVecReader<'r> {
         if self.is_empty() {
             packed::Byte32::zero()
         } else {
-            blake2b_256(self.as_slice()).pack()
+            let mut ret = [0u8; 32];
+            let mut blake2b = new_blake2b();
+            for uncle in self.iter() {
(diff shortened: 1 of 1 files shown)


util/types/src/extension/calc_hash.rs
```diff
@@ -214,6 +220,78 @@ mod tests {
...
+            .raw(uncle2_raw_header)
+            .nonce(0x2f39_2d41_cd70_024fu128.pack())
+            .build();
+        let uncle2 = packed::UncleBlock::new_builder()
+            .header(uncle2_header)
+            .build();
+
+        let uncles = vec![uncle1, uncle2].pack();
+        let expect = h256!("0x0135d01f169a870bd9c92b2b37aecfa0fbfb7c1862cc176e03bb525fab0649d9");
+        assert_eq!(uncles.calc_uncles_hash(), expect.pack());
```
u2: I don't like this test case, there are too many hard codes, it's not maintainable.

quake: It's just a test vector for reference, which can be used to verify the correctness of different language implementations.

u2: ok

PR #1708, opened 2019-10-12
chore: use primitive u128 for genesis nonce

quake:


spec/src/lib.rs
```diff
@@ -99,7 +99,7 @@ pub struct Genesis {
     pub compact_target: u32,
     pub uncles_hash: H256,
     pub hash: Option<H256>,
-    pub nonce: U128,
+    pub nonce: u128,
```
keroro520: https://github.com/nervosnetwork/ckb/blob/develop/resource/specs/dev.toml#L9

Should `"0x0"` -> `0` ? And integration.toml as well.

quake: I didn't notice that. now fixed, thanks.

driftluo: `toml-rs` is not supported parse u128 now

quake: WTH...

PR #1710, opened 2019-10-12
refactor: Enforce txpool consistency

keroro520:
This refactoring PR intends to reduce the probability of inconsistency between `SortedTxMap::entries` and `SortedTxMap::links`.

1. Merge entries and links into one hashmap: The `SortedTxMap.entries` and `SortedTxMap.links` must keep consistent. The previous code enforces the consistency carefully and correctly, but I think that merging them into one slot, so that it reduces the probability of inconsistency, is a better management way.
2. Create a trait `RelativeTraversal`, a helper to traverse all the in-pool ancestors and descendants of the target proposal-id.
3. Try to wrap some update operations into functions to reduce the error probability.

### DIFF
--- a/tx-pool/src/component/container.rs
+++ b/tx-pool/src/component/container.rs
@@ -4,7 +4,7 @@
 use crate::component::entry::TxEntry;
 use ckb_types::{core::Capacity, packed::ProposalShortId};
 use std::cmp::Ordering;
-use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 /// A struct to use as a sorted key
 #[derive(Eq, PartialEq, Clone, Debug)]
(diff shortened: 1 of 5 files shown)


tx-pool/src/component/container.rs
```diff
@@ -161,118 +134,90 @@ impl SortedTxMap {
...
-                for c_id in link.children {
-                    self.links
-                        .get_mut(&c_id)
-                        .map(|link| link.parents.remove(&id));
-                }
-            }
-            entry
-        })
-    }
+        if !self.contains_key(id) {
```
jjyr:
```suggestion
        let result = remove(id);
        if result.is_none() {
            return None;
        }
        let (entry, link) = result;
```

tx-pool/src/component/container.rs
```diff
@@ -161,118 +134,90 @@ impl SortedTxMap {
...
-        removed
+        let descendants = self.get_descendants(id);
+        removal.extend(descendants.iter().map(|desc_id| {
+            self.get(desc_id)
+                .expect("found inconsistency when remove_entry_and_descendants")
+                .to_owned()
+        }));
+
+        removal.iter().for_each(|e| {
+            self.remove_entry(&e.transaction.proposal_short_id());
```
jjyr: Obviously, this implementation has `log(N * N)` complexity compare with the old `log(N)`, please don't do this.

tx-pool/src/component/container.rs
```diff
@@ -161,118 +134,90 @@ impl SortedTxMap {
...
+                    .expect("found inconsistency when remove_entry");
+                let old_sorted_key = desc_entry.as_sorted_key();
+
+                desc_entry.sub_entry_weight(&entry);
+                let new_sorted_key = desc_entry.as_sorted_key();
+                (old_sorted_key, new_sorted_key)
+            };
+
+            self.sorted_index.insert(new_sorted_key);
+            self.sorted_index.remove(&old_sorted_key);
```
jjyr: Must remove first, **logically** the `new_sorted_key` may equal to `old_sorted_key`.

tx-pool/src/component/relative.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::collections::HashSet;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) enum Relation {
+    Ancestor,
+    Descendant,
+}
+
+/// This crate is used for recursively searching in-pool ancestors or descendants
+pub(crate) trait RelativeTraversal {
```
jjyr: Can we just implement these on `SortedTxMap`?

jjyr: This file should be named `relation.rs` or `tx_relation.rs`

tx-pool/src/component/relative.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fn get_descendants(&self, id: &ProposalShortId) -> HashSet<ProposalShortId> {
+        self.get_relative_ids(id, Relation::Descendant)
+    }
+}
+
+impl RelativeTraversal for SortedTxMap {
+    fn get_parents(&self, id: &ProposalShortId) -> &HashSet<ProposalShortId> {
+        &self
```
jjyr: return Option instead of panic. then use `result.map(ToOwned::to_owned).unwrap_or_default` outside.

keroro520: If an entry is in the tx_pool, then `get_parents` will be ok;
If an entry is not in the pool, then `get_parents` get panicked.

It is the duty for the outside caller to make sure that 
The outside caller should make ensure that the given entry is already in tx_pool before calling `RelativeTraversal's methods. I think the assert style is helpful for enforcing the consistency.

jjyr: Because `RelativeTraversal` is a public trait, and `get_parents ` is a public function. panic from a public function is not usually the rust idiomatic way, as you see in the std the more common style is `map.get(i).unwrap()`, the invoker control when to panic or not. This function looks don't like it'll panic, this is a pitfall for the invoker.

PR #1712, opened 2019-10-14
fix: fix tx_pool_info

u2:


u2:
bors r=quake,keroro520

u2:
bors r=quake,keroro520

PR #1718, opened 2019-10-15
[HOLD] test: Add cases of tx_hash collision

keroro520:
Add some test-cases about tx_hash collision.

There are some bugs in txpool, to make the added cases fail to run. But I did not fix that, just comment on the cases.

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -297,6 +297,9 @@ fn all_specs() -> SpecMap {
         Box::new(ProposeTransactionButParentNot),
         Box::new(ProposalExpireRuleForCommittingAndExpiredAtOneTime),
         Box::new(ReorgHandleProposals),
+        // Box::new(TransactionHashCollisionDifferentWitnessHashes1),
+        // Box::new(TransactionHashCollisionDifferentWitnessHashes2),
+        // Box::new(TransactionHashCollisionDifferentWitnessHashes3),
     ];
     specs.into_iter().map(|spec| (spec.name(), spec)).collect()
 }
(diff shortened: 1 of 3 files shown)


u2:
The reason is that these two transactions have the same transaction hash and short_id, but the tx_pool use short_id as hash_map/hash_set key.

https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L145
The two transactions have different `sorted_keys` because the `vbytes` is different.

If one transaction is removed at https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L164 or other places.

https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L146
When we fetch the other transaction, https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L146 it will panic.

I think we should return an error if the later transaction's short_id has a collision with the ones in the transaction pool.

keroro520:
> The reason is that these two transactions have the same transaction hash and short_id, but the tx_pool use short_id as hash_map/hash_set key.
> 
> https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L145
> The two transactions have different `sorted_keys` because the `vbytes` is different.
> 
> If one transaction is removed at https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L164 or other places.
> 
> https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L146
> When we fetch the other transaction, https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process/chain_reorg.rs#L146 it will panic.
> 
> I think we should return an error if the later transaction's short_id has a collision with the ones in the transaction pool.

Out-of-scope of this PR.

u2:
it's the reason for node panic.

doitian:
Hold until we have a fix

keroro520:
Already cherry-picked by #1736

PR #1719, opened 2019-10-15
test: Add tests for pool handling conflict txs

keroro520:


test/src/specs/tx_pool/conflict.rs
```diff
@@ -0,0 +1,95 @@
...
+        node.generate_blocks(window.farthest() as usize + 2);
+
+        let (txa, txb) = conflict_transactions(node);
+        node.submit_transaction(&txa);
+
+        node.submit_block(&propose(node, &[&txa]));
+        (0..window.closest()).for_each(|_| {
+            node.submit_block(&blank(node));
+        });
+        node.submit_transaction(&txb);
```
u2: I think the submit tx will return an error result here.

keroro520: Updated. PTAL.

keroro520:
> Why not any assertion after `generate_blocks`?

Assert committed transactions? Or what do you suggest to assert?
IMO, `generate_blocks` already includes that miner pick transactions into new blocks and node verify whether the new blocks are valid, which means if `generate_blocks` do not panic, then the blockchain runs ok.

jjyr:
bors r=quake, jjy

jjyr:
cherry pick to https://github.com/nervosnetwork/ckb/pull/1736

PR #1720, opened 2019-10-16
feat: Add load transaction syscall

xxuejie:


script/src/syscalls/mod.rs
```diff
@@ -730,22 +732,29 @@ mod tests {
         machine.set_register(A2, 0); // offset
         machine.set_register(A7, LOAD_TX_HASH_SYSCALL_NUMBER); // syscall number
 
-        let hash = blake2b_256(&data);
+        let transaction_view = TransactionBuilder::default()
+            .output_data(data.pack())
```
jjyr: This is weird, smile like a bug.

jjyr: Oh, I see.

xxuejie: It's in test, just a way to generate different transactions in proptest

PR #1728, opened 2019-10-17
fix: node should return duplicated when tx_hash collision

u2:
https://github.com/nervosnetwork/ckb/pull/1718#issuecomment-542601807

### DIFF
--- a/tx-pool/src/process/submit_txs.rs
+++ b/tx-pool/src/process/submit_txs.rs
@@ -53,6 +53,8 @@ impl Future for PreResolveTxsProcess {
                 let snapshot = tx_pool.cloned_snapshot();
                 let tip_hash = snapshot.tip_hash();
 
+                check_transaction_hash_collision(&tx_pool, &txs)?;
+
                 let mut txs_provider = TransactionsProvider::default();
                 let resolved = txs
                     .iter()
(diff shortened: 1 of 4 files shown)


tx-pool/src/component/container.rs
```diff
@@ -202,6 +202,13 @@ impl SortedTxMap {
         self.entries.get(id)
     }
 
+    pub fn get_sorted_key(&self, id: &ProposalShortId) -> Option<AncestorsScoreSortKey> {
```
jjyr: Please delete this function.

tx-pool/src/process/submit_txs.rs
```diff
@@ -242,6 +242,11 @@ fn resolve_tx<'a>(
     }
 
     let short_id = tx.proposal_short_id();
+
+    if tx_pool.contains_proposal_id(&short_id) {
```
jjyr: Can you move this condition to `process_txs`.

tx-pool/src/component/container.rs
u2: done.

tx-pool/src/process/submit_txs.rs
u2: I move it in PreResolveTxsProcess

jjyr:
cherry pick to https://github.com/nervosnetwork/ckb/pull/1736

PR #1732, opened 2019-10-17
refactor: support run integration tests concurrently

jjyr:
Use `make integration CKB_TEST_ARGS="-c 8"` to run integration tests in 8 concurrent workers.

The default concurrent number is 4, which can reduce 3~4x times of integration tests on my machine.
For CI the concurrent number is set to 1 (our CI sucks), we can increase this value when we enhance the performance of CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ MOLC    := moleculec
 MOLC_VERSION := 0.3.1
 VERBOSE := $(if ${CI},--verbose,)
 CLIPPY_OPTS := -D warnings -D clippy::clone_on_ref_ptr -D clippy::enum_glob_use -D clippy::fallible_impl_from
+CKB_TEST_ARGS := -c 4
+INTEGRATION_RUST_LOG := ckb-network=error
 
 ##@ Testing
 .PHONY: test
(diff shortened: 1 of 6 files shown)


jjyr:
bors try

jjyr:
bors try

jjyr:
bors try

Makefile
```diff
@@ -33,7 +34,7 @@ submodule-init:
 .PHONY: integration
 integration: submodule-init setup-ckb-test ## Run integration tests in "test" dir.
 	cargo build
-	cd test && RUST_BACKTRACE=1 cargo run -- --bin ../target/debug/ckb ${CKB_TEST_ARGS}
+	cd test && RUST_BACKTRACE=1 RUST_LOG=ckb-network=error cargo run -- --bin ../target/debug/ckb ${CKB_TEST_ARGS}
```
yangby-cryptape: Please remove this `RUST_LOG`, so, we can set it via `export` environment variable.

jjyr:
bors try

jjyr:
bors try

test/src/main.rs
```diff
@@ -45,73 +54,101 @@ fn main() {
...
+        vendor.clone(),
+    );
+    workers.start();
+
+    let mut rerun_specs = Vec::new();
+    let mut worker_running = worker_count;
+    let mut done_specs = 0;
+    while worker_running > 0 {
+        if max_time > 0 && start_time.elapsed().as_secs() > max_time {
+            // shutdown, specs running to long
```
u2: too long

test/src/worker.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.is_shutdown = true;
+    }
+
+    /// wait all workers to shutdown
+    pub fn join_all(&mut self) {
+        if self.join_handles.is_none() {
+            return;
+        }
+        // make sure shutdown all workers
+        self.shutdown();
```
u2: why call `shutdown` before `join`?

jjyr: https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#method.join

Need to use join to free threads resource.

PR #1736, opened 2019-10-18
Fix tx pool inconsistent when receive duplicated hash txs.

jjyr:


tx-pool/src/component/container.rs
```diff
@@ -154,6 +154,12 @@ impl SortedTxMap {
     pub fn add_entry(&mut self, mut entry: TxEntry) -> Option<TxEntry> {
         let short_id = entry.transaction.proposal_short_id();
 
+        let removed_entry = if self.contains_key(&short_id) {
```
u2: I think if there is tx_hash collision, it should return None.

jjyr: Solve tx_hash collision is not a responsibility for this struct, we need to handle collision in the pool logic, here the struct only guarantees internal consistency.

The behavior is just like https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert

u2: ok

quake:
bors r=quake,u2,zhangsoledad

PR #1740, opened 2019-10-20
feat: Add multisig system script cell

jjyr:
https://github.com/nervosnetwork/ckb-system-scripts/pull/60

* Add multisig to genesis


script/src/syscalls/mod.rs
```diff
@@ -31,9 +31,9 @@ pub const ITEM_MISSING: u8 = 2;
 pub const SLICE_OUT_OF_BOUND: u8 = 3;
 
 pub const LOAD_TRANSACTION_SYSCALL_NUMBER: u64 = 2051;
+pub const LOAD_SCRIPT_SYSCALL_NUMBER: u64 = 2052;
```
doitian: Why change the syscall number? cc @xxuejie

doitian:
Please rebase onto the latest change in https://github.com/nervosnetwork/ckb-system-scripts/pull/60

script/src/syscalls/mod.rs
jjyr: To keep syscall number consistent, the number `205x` is used to describe load the whole struct, and `206x` is used to describe load the hash of the struct.

PR #1745, opened 2019-10-22
fix: block 11 should issue block reward for genesis

doitian:


chain/src/tests/reward.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) fn create_cellbase(
         .input(CellInput::new_cellbase_input(number))
         .witness(miner_lock.into_witness());
 
-    if (parent.number() + 1) <= consensus.finalization_delay_length() {
+    if (parent.number() + 1) < consensus.finalization_delay_length() {
```
u2:
```
#[allow(clippy::int_plus_one)]
```
This can be removed.

doitian:
Creating cell locked by all zeros has the same effect to burn it.

PR #1756, opened 2019-10-24
fix: fix benchmark overflow

u2:
The reason is that `U` in genesis DAO is too small. When generating transactions that occupy many capacities than `U`, the DAO calculation overflow.


yangby-cryptape:
bors r=keroro520,quake

PR #1757, opened 2019-10-24
feat: shutdown when protocol handle panic

driftluo:
Shutdown when protocol handle panic


yangby-cryptape:
bors r=doitian,TheWaWaR

driftluo:
bors retry

PR #1760, opened 2019-10-24
refactor: replace non-maintained jsonrpc client

zhangsoledad:
* replace https://crates.io/crates/jsonrpc-client-core
This https://crates.io/crates/jsonrpc-client-core is unruly and still uses tokio-core which is deprecated, 
the maintainers are not around to work on this project. 
https://dev.azure.com/nervosnetwork/ckb/_build/results?buildId=7366
we catch some  `Canceled` error is not being handled elegantly, lead to lock been poisoned.

* singleton http-clinet 
We don't need to instantiate http-clinet for every test-case, every simulate node.

### DIFF
--- a/test/Cargo.toml
+++ b/test/Cargo.toml
@@ -25,14 +25,15 @@ ckb-dao-utils = { path = "../util/dao/utils" }
 ckb-test-chain-utils = { path = "../util/test-chain-utils" }
 ckb-resource = { path = "../resource" }
 tempfile = "3.0"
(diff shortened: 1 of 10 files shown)


yangby-cryptape:
bors r=quake,yangby-cryptape

zhangsoledad:
bors r=quake,yangby-cryptape

zhangsoledad:
bors r=quake,yangby-cryptape

PR #1763, opened 2019-10-26
fix: fix cli output and ban reason

### LINKED ISSUE #1746: ckb cli secp256k1-lock should return a single bytes as args
**Environment**

- **CKB version**: v0.23.0
- **Chain**: testnet
- **Operating system**: macOS
- **Arch**: x64

**Describe the bug**

```
$ target/release/ckb cli secp256k1-lock 049b34df88f4e1364d32ab7f5fe5b25d60c56c686500925e3b87e2f5d989a8b0f9d8b8376d352ee5a1c70f061b59c4b6799
68fe84b3f09ab2ee03f760e21150ef9
# args = [ "ckb cli blake160 <compressed-pubkey>" ]                                                                                                                
args = [ "0xde9ed127a494d92342d68b0424986dd69530be30" ]
```

driftluo:
1. fix #1746
2. add reason on `ban_peer` in cx trait

### DIFF
--- a/sync/src/net_time_checker.rs
+++ b/sync/src/net_time_checker.rs
@@ -141,7 +141,11 @@ impl CKBProtocolHandler for NetTimeProtocol {
             Some(timestamp) => timestamp,
             None => {
                 info!("Peer {} sends us malformed message", peer_index);
-                nc.ban_peer(peer_index, BAD_MESSAGE_BAN_TIME);
+                nc.ban_peer(
+                    peer_index,
+                    BAD_MESSAGE_BAN_TIME,
+                    String::from("send us a malformed message"),
+                );
                 return;
             }
         };
--- a/sync/src/synchronizer/get_headers_process.rs
+++ b/sync/src/synchronizer/get_headers_process.rs
@@ -89,7 +89,11 @@ impl<'a> GetHeadersProcess<'a> {
                 warn!("unknown block headers from peer {} {}", self.peer, hash);
             }
             // Got 'headers' message without known blocks
-            self.nc.ban_peer(self.peer, SYNC_USELESS_BAN_TIME);
+            self.nc.ban_peer(
+                self.peer,
+                SYNC_USELESS_BAN_TIME,
+                String::from("send us headers with unknown-block"),
+            );
         }
         Ok(())
     }
--- a/sync/src/synchronizer/mod.rs
+++ b/sync/src/synchronizer/mod.rs
@@ -94,7 +94,11 @@ impl Synchronizer {
     ) {
         if let Err(err) = self.try_process(nc, peer, message) {
             debug!("try_process error: {}", err);
-            nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+            nc.ban_peer(
+                peer,
+                BAD_MESSAGE_BAN_TIME,
+                format!("sync message process error: {}", err),
+            );
         }
     }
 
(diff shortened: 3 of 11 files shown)


sync/src/synchronizer/mod.rs
```diff
@@ -94,7 +94,11 @@ impl Synchronizer {
     ) {
         if let Err(err) = self.try_process(nc, peer, message) {
             debug!("try_process error: {}", err);
-            nc.ban_peer(peer, BAD_MESSAGE_BAN_TIME);
+            nc.ban_peer(
+                peer,
+                BAD_MESSAGE_BAN_TIME,
+                format!("syn message process error: {}", err),
```
yangby-cryptape:
```suggestion
                format!("sync message process error: {}", err),
```

sync/src/synchronizer/get_headers_process.rs
```diff
@@ -89,7 +89,11 @@ impl<'a> GetHeadersProcess<'a> {
                 warn!("unknown block headers from peer {} {}", self.peer, hash);
             }
             // Got 'headers' message without known blocks
-            self.nc.ban_peer(self.peer, SYNC_USELESS_BAN_TIME);
+            self.nc.ban_peer(
+                self.peer,
+                SYNC_USELESS_BAN_TIME,
+                String::from("send use headers with unknown block"),
```
yangby-cryptape:
```suggestion
                String::from("send us headers with unknown block"),
```

util/network-alert/src/alert_relayer.rs
```diff
@@ -138,7 +146,11 @@ impl CKBProtocolHandler for AlertRelayer {
                 "Peer {} sends us a alert with invalid signatures, error {:?}",
                 peer_index, err
             );
-            nc.ban_peer(peer_index, BAD_MESSAGE_BAN_TIME);
+            nc.ban_peer(
+                peer_index,
+                BAD_MESSAGE_BAN_TIME,
+                String::from("send us a alert with invalid signatures"),
```
yangby-cryptape:
```suggestion
                String::from("send us an alert with invalid signatures"),
```

sync/src/net_time_checker.rs
```diff
@@ -141,7 +141,11 @@ impl CKBProtocolHandler for NetTimeProtocol {
             Some(timestamp) => timestamp,
             None => {
                 info!("Peer {} sends us malformed message", peer_index);
-                nc.ban_peer(peer_index, BAD_MESSAGE_BAN_TIME);
+                nc.ban_peer(
+                    peer_index,
+                    BAD_MESSAGE_BAN_TIME,
+                    String::from("send us malformed message"),
```
yangby-cryptape: Some are `"send us a malformed message"`, some are `"send us malformed message"` (`send us invalid block`), I think we could use error messages in a same format.

driftluo: Most of them are deserialization failures, some are able to locate that block is illegal.

Ban does not need to locate why it fails, just need to know,  invalid data, or can locate specific data is invalid. Error message doesn't make sense

yangby-cryptape: You seems not to have caught what I meant.
Forgive my unclear description that leading your misunderstanding.

1. I mean, you use both `"send us malformed message"` and `"send us a malformed message"` (the difference is the `a`), it's a bit weird.
2. In `"send us malformed message"`, IMHO,  a definite article is required. (also in `"send us invalid block"`, and other messages)

PR #1770, opened 2019-10-28
fix: skip cellbase short-id collision validation

quake:


sync/src/relayer/compact_block_verifier.rs
```diff
@@ -75,9 +75,11 @@ impl ShortIdsVerifier {
             return Err(Error::Misbehavior(Misbehavior::DuplicatedShortIds));
         }
 
-        // Check intersection of prefilled transactions and short ids
+        // Check intersection of prefilled transactions and short ids.
+        // Cellbase is skipped since it's always prefilled and has the chances of collision with other txs
         let is_intersect = prefilled_transactions
             .into_iter()
+            .skip(1)
```
u2: The `cellbase` is always in the first position? I am not sure about that.

quake: Yes
https://github.com/nervosnetwork/ckb/blob/a875ce11246df4acfdd047b5bc114eede2ee64ed/sync/src/relayer/compact_block_verifier.rs#L26-L33

u2: ok

quake:
bors retry

PR #1772, opened 2019-10-29
fix: Limit p2p protocol message size

TheWaWaR:


driftluo:
`send_bulk_of_tx_hashes` must be pruned when too large

sync/src/relayer/get_transactions_process.rs
```diff
@@ -65,6 +67,15 @@ impl<'a> GetTransactionsProcess<'a> {
                         .transaction(tx.data())
                         .build()
                 })
+                .take_while(|relay_tx| {
+                    // If the size of RelayTransactions response is too large we
+                    // just skip it, since this case is rare and we can get
+                    // missing transactions by BlockTransactions message later.
+                    total_size += relay_tx.as_slice().len();
```
doitian: may overflow

quake:
I am working on it, will create a new pr later

sync/src/relayer/get_transactions_process.rs
TheWaWaR: Actually it's not possible, total_size must not greater than `MAX_RELAY_TXS_SIZE`

`total_size <= MAX_RELAY_TXS_SIZE`

doitian:
bors r=quake,doitian,driftluo,u2

PR #1773, opened 2019-10-29
chore: code cleanup and add limit checking to `GetBlocks` message

quake:
cleanup code for unused `PER_FETCH_BLOCK_LIMIT` const, it should be replaced by `MAX_BLOCKS_IN_TRANSIT_PER_PEER `, and add limit checking to `GetBlocks` message.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -1,12 +1,11 @@
 use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
 use crate::types::{HeaderView, SyncSnapshot};
-use crate::{MAX_BLOCKS_IN_TRANSIT_PER_PEER, PER_FETCH_BLOCK_LIMIT};
+use crate::MAX_BLOCKS_IN_TRANSIT_PER_PEER;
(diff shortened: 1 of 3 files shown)


doitian:
@zhangsoledad please take a look at this PR

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -113,17 +112,12 @@ impl BlockFetcher {
         debug_assert!(best_known_header.number() > fixed_last_common_header.number());
 
         let mut index_height = fixed_last_common_header.number();
-        // Read up to 128, get_ancestor may be as expensive
-        // as iterating over ~100 entries anyway.
-        let mut fetch = Vec::with_capacity(PER_FETCH_BLOCK_LIMIT);
+        let mut fetch = Vec::with_capacity(MAX_BLOCKS_IN_TRANSIT_PER_PEER);
```
doitian: Is `PER_FETCH_BLOCK_LIMIT` no long used? It's better to remove the const as well.

quake: yes, removed already

doitian:
bors r=doitian,u2,keroro520

PR #1784, opened 2019-10-31
feat: support limit max_tx_verify_cycles

jjyr:
The purpose is to limit max verify cycles on single tx, to reduce DDOS vulnerability.

* Add a new tx pool config option `max_tx_verify_cycles`, default is 70_000_000 (20 times 2-in-2-out tx verify cycles), (take about ~160ms to run on my machine).
* txs that submit via RPC will be fail if consume more than `max_tx_verify_cycles` cycles
* txs that received in relay protocol will be fail if consume more than `max_tx_verify_cycles` cycles
* block propogation should not be affected


sync/src/relayer/transactions_process.rs
```diff
@@ -77,16 +77,23 @@ impl<'a> TransactionsProcess<'a> {
         }
 
         let mut notify_txs = Vec::with_capacity(txs.len());
-
+        let max_tx_verify_cycles = self.relayer.max_tx_verify_cycles;
         let relay_cycles_vec: Vec<_> = txs
             .into_iter()
-            .map(|(tx, relay_cycles)| {
+            .filter_map(|(tx, relay_cycles)| {
+                // skip txs which consume too much cycles
```
xxuejie: Can we log them at least instead of simply ignoring them?

jjyr:
bors r= xxuejie driftluo

PR #1788, opened 2019-11-01
feat: limit tx max ancestors count

jjyr:
Txs with long ancestors chain affect tx pool performance. we limit max ancestors count of a single tx to resolve this issue, tx pool will reject txs which ancestors count large than the limit.

The default `max_ancestors_count` is 25.


tx-pool/src/component/proposed.rs
```diff
@@ -300,22 +306,24 @@ mod tests {
...
 
         pool.add_entry(TxEntry::new(
             tx1.clone(),
             MOCK_CYCLES,
             MOCK_FEE,
             MOCK_SIZE,
             get_related_dep_out_points(&tx1, |_| None).unwrap(),
-        ));
+        ))
+        .unwrap();
```
xxuejie: There's a lot of unwrap added here, any chance we can properly deal with the errors? Or are we sure they won't happen?

jjyr: This is test code.

tx-pool/src/component/container.rs
```diff
@@ -179,6 +183,18 @@ impl SortedTxMap {
...
+        if entry.ancestors_count > self.max_ancestors_count {
+            return Err(SubmitTxError::ExceededMaximumAncestorsCount);
+        }
+
+        // check duplicate tx
+        let removed_entry = if self.contains_key(&short_id) {
+            self.remove_entry(&short_id)
+        } else {
+            None
+        };
```
u2: Why do you move this part of the code?

tx-pool/src/process/plug.rs
```diff
@@ -40,12 +41,16 @@ impl Future for PlugEntryProcess {
                 match self.target {
                     PlugTarget::Pending => {
                         for entry in entries {
-                            tx_pool.add_pending(entry);
+                            if let Err(err) = tx_pool.add_pending(entry) {
+                                error!("plug entry error {}", err);
```
u2: plug?

tx-pool/src/component/container.rs
jjyr: If ancestors reach the limit, we should just return an error and not remove the conflict tx. Other wise we should go to the old logic.

tx-pool/src/process/plug.rs
jjyr: Yes, see the file name.

PR #1792, opened 2019-11-05
fix: Add back to celloutput lost type script

driftluo:
Add back to celloutput lost type script


driftluo:
bors r=doitian,u2,keroro520

PR #1797, opened 2019-11-07
feat: allow specify single consensus param in spec

zhangsoledad:


zhangsoledad:
bors r=driftluo,xxuejie

PR #1799, opened 2019-11-07
feat: disable jemalloc on MacOS

jjyr:
jemalloc is crashed on MacOS catalina.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,5 +66,5 @@ members = [
 debug = true
 overflow-checks = true
 
-[target.'cfg(all(unix, not(target_os="macos")))'.dependencies]
+[target.'cfg(all(not(target_env = "msvc"), not(target_os="macos")))'.dependencies]
 jemallocator = { version = "0.3.0" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -66,5 +66,5 @@ members = [
 debug = true
 overflow-checks = true
 
-[target.'cfg(unix)'.dependencies]
+[target.'cfg(all(unix, not(target_os="macos")))'.dependencies]
```
doitian:
```suggestion
[target.'cfg(and(not(target_env = "msvc"), not(target_os="macos")))'.dependencies]
```

src/main.rs
```diff
@@ -1,7 +1,7 @@
 use ckb_bin::run_app;
 use ckb_build_info::Version;
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "macos")))]
```
doitian:
```suggestion
#[cfg(all(not(target_env = "msvc"), not(target_os = "macos")))]
```

As suggested in https://github.com/gnzlbg/jemallocator

PR #1804, opened 2019-11-08
fix: retrieve few missing ckb

yangby-cryptape:


zhangsoledad:
bors r=zhangsoledad,doitian

PR #1805, opened 2019-11-08
fix: proposal table bound

zhangsoledad:


doitian:
Confirmed locally that the transaction proposed in block 1 will not be committed in block 2 template.

PR #1806, opened 2019-11-09
chore: remove code name and rename testnet genesis message

doitian:
- Remove code name.
- Rename testnet genesis message.


doitian:

PR #1807, opened 2019-11-12
feat: Include raw header bytes in get_block_template

crackcomm:
As requested and discussed by Greg from icemining.ca.

I'm open to suggestions eq. as to how to remove additional `molecule` dependency.

### DIFF
--- a/tx-pool/src/process/block_template.rs
+++ b/tx-pool/src/process/block_template.rs
@@ -7,7 +7,7 @@ use crate::config::BlockAssemblerConfig;
 use crate::error::BlockAssemblerError;
 use crate::pool::TxPool;
 use ckb_dao::DaoCalculator;
-use ckb_jsonrpc_types::BlockTemplate;
+use ckb_jsonrpc_types::{Uint32, JsonBytes, BlockTemplate, EpochNumberWithFraction};
 use ckb_logger::info;
 use ckb_snapshot::Snapshot;
 use ckb_store::ChainStore;
@@ -16,7 +16,7 @@ use ckb_types::{
         cell::{resolve_transaction, OverlayCellProvider, TransactionsProvider},
         Cycle, EpochExt, ScriptHashType, TransactionView, UncleBlockView, Version,
     },
-    packed::{CellbaseWitness, ProposalShortId, Script},
+    packed::{RawHeader, CellbaseWitness, ProposalShortId, Script},
     prelude::*,
 };
 use failure::Error as FailureError;
@@ -29,6 +29,7 @@ use std::sync::Arc;
 use std::{cmp, iter};
 use tokio::prelude::{Async, Poll};
 use tokio::sync::lock::Lock;
+use molecule::prelude::Entity;
 
 type Args = (u64, u64, Version);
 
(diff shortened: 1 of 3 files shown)


yangby-cryptape:
Thanks, but we cannot accept it.

### Reasons

- Redundant

  Why we have two format of same data in one API?
  For example: version, compact_target, number and epoch.

- Human-readable

  The data used in JSON RPC should be human readable as much as possible.
  (Only the epoch and since is not human readable, in current version)
  Can u read the raw bytes without any program?

- We don't want to implement all functions (in ckb repo).

  We only want to do the core part, we only offer the necessary functions (keep the core part be simple).
  We only provide opportunities to implement various functions (depend on the technological ability of developers).

  Keep API (JSON RPC) be easy to use is not our goal (in ckb repo). 
  We will offer these combined functions in our various SDKs (java, javascript, ruby, and so on).

  If you think it's too difficult to solve this problem with our current API, please be patience, our SDK teams are working hard on our various SDKs, you can use these SDKs in the near future.

crackcomm:
Is there any existing function that would serialize header in javascript SDK?

yangby-cryptape:
> Is there any existing function that would serialize header in javascript SDK?

I had asked one of our Javascript team member for that, I have been told that:
- We don't have that function, now.
- And, we don't have plan for that in a short term.
- **As an open-source project, we welcome contributions.**
  If anyone submit commits about these, we will accept that after review and audit.

---

I feel really sorry for that.
But we really don't have enough manpower and resources to cover all languages in a limited time. So, we don't focus on implement miners or pools since the project was started. We believe the community.
There are lots of intelligent developers who can do that better than us. In our testnet, there already are a lot of pools.
(Few months later, maybe we will have resources to do it, correctness and security are the most important thing in recent weeks.)

---

As a matter of expediency, in your scene, if u feel it's too difficult to serialize  and deserialize headers, I strongly recommend that your can use FFI to do it.

I have a very simple [example] to show how to call it in C, u can try it.
This [example] covers enough functions for a miner or a pool (I provide it as a fallback solution).

I really wish I could help you more.

[example]: https://github.com/yangby-cryptape/ckb-ffi

tx-pool/src/process/block_template.rs
```diff
@@ -273,19 +274,35 @@ impl Future for BlockTemplateBuilder {
...
         // Should recalculate current time after create cellbase (create cellbase may spend a lot of time)
         let current_time = cmp::max(unix_time_as_millis(), tip_header.timestamp() + 1);
+
+        let epoch: EpochNumberWithFraction = self
+            .current_epoch
+            .number_with_fraction(candidate_number)
+            .into();
+
+
+        let raw = RawHeader::new_builder()
```
doitian: This step misses the cellbase transaction, and the raw header is wrong. There's a easy way to get raw header from block template.

```
let block_template = BlockTemplate {
  // omit
};

let block: packed::Block = block_template.into();
```

crackcomm: Thank you, that simplifies it a lot.

crackcomm:
I appreciate your comments and understand the reasons with which I agree.

I think for node.js we will use neon bindings and it will simplify it a lot, thanks.

PR #1813, opened 2019-11-13
fix: Fix get locator performance bug

TheWaWaR:
When get header from main chain we can get it from snapshot

### DIFF
--- a/store/src/store.rs
+++ b/store/src/store.rs
@@ -188,7 +188,7 @@ pub trait ChainStore<'a>: Send + Sync {
     }
 
     fn is_main_chain(&'a self, hash: &packed::Byte32) -> bool {
-        self.get_block_number(&hash).is_some()
+        self.get(COLUMN_INDEX, hash.as_slice()).is_some()
     }
 
     fn get_tip_header(&'a self) -> Option<HeaderView> {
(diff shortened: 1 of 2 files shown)


store/src/store.rs
```diff
@@ -188,7 +188,7 @@ pub trait ChainStore<'a>: Send + Sync {
     }
 
     fn is_main_chain(&'a self, hash: &packed::Byte32) -> bool {
-        self.get_block_number(&hash).is_some()
+        self.get(COLUMN_INDEX, hash.as_slice()).is_some()
```
keroro520: What is the difference between `self.get_block_number(&hash).is_some()` and `self.get(COLUMN_INDEX, hash.as_slice()).is_some()`?

zhangsoledad: `get_block_number` will do deserialize which is redundant for this function

zhangsoledad:
bors r=zhangsoledad,keroro520

zhangsoledad:
bors r=quake,keroro520,zhangsoledad

doitian:
bors r=quake,keroro520,zhangsoledad

doitian:
bors retry

PR #1820, opened 2019-11-14
test(ci): share travis cache between jobs

doitian:


zhangsoledad:
bors r=quake,zhangsoledad

PR #1821, opened 2019-11-15
fix: common header should be updated when block received

zhangsoledad:


sync/src/synchronizer/block_process.rs
```diff
@@ -37,31 +37,30 @@ impl<'a> BlockProcess<'a> {
...
+                self.nc.ban_peer(
+                    self.peer,
+                    BAD_MESSAGE_BAN_TIME,
+                    String::from("send us an invalid block"),
+                );
+            }
+        } else if snapshot.contains_block_status(&block.hash(), BlockStatus::BLOCK_STORED) {
+            state
+                .peers()
+                .set_last_common_header(self.peer, block.header().clone());
```
keroro520: Is this likely to happen that `get_last_common_header(self.peer).number() > block.header.number()`, which means we replace with an old header.

zhangsoledad: This peer may be reorganized, receive a smaller number header is fine.
We don't need do number comparison here.
We will fix it before we use it.
https://github.com/nervosnetwork/ckb/pull/1821/files#diff-86d69964b227ca1248fadb84da967a96R54-R55

keroro520: That's cool!

zhangsoledad:
bors retry

PR #1830, opened 2019-11-17
chore: ensure commit abbrev in version consistent

doitian:
The default abbrev length is different in different git versions.


doitian:
bors retry

PR #1833, opened 2019-11-19
test: Add test vector for extract/pack dao field

keroro520:


jjyr:
Can we prepare some more DAO bytes(5 ~ 10) in the test?

keroro520:
> Can we prepare some more DAO bytes(5 ~ 10) in the test?

Yeah, of course. I think 3 examples are enough(ckb-mainnet block[0], block[1], block[5892]).

PR #1837, opened 2019-11-20
chore: omit calculate_dao_maximum_withdraw log

keroro520:
Neuron will request `calculate_dao_maximum_withdraw` to calculate NervosDAO interest. So when a fresh node does not synchronize completely, a `calculate_dao_maximum_withdraw` request may trigger an error.
It is unnecessary to write this error into logs.


doitian:
bors r=quake,doitian

doitian:
bors r-

cargo fmt failed

TheWaWaR:
bors r=quake,doitian,zhangsoledad

PR #1838, opened 2019-11-21
feat: add `output_data_len` and `cellbase` to get_cells_by_lock_hash rpc

quake:


rpc/README.md
```diff
@@ -383,13 +383,22 @@ http://localhost:8114
...
 ### `get_cells_by_lock_hash`
 
-Returns the information about cells collection by the hash of lock script.
+Returns the information about live cells collection by the hash of lock script.
 
 #### Parameters
 
     lock_hash - Cell lock script hash
     from - Start block number
     to - End block number
```
TheWaWaR: New line

yangby-cryptape:
I disapprove of adding `cellbase_maturity`.

IMHO, we should only offer the necessary and general functions (keep the core part be simple).
`cellbase_maturity` is too special, we can't get any other information from it.
If users need some data next time, we have to add another field.
If things go on like this, the rpc interfaces will be huge.

IMHO, we could return `epoch` (and `block_number`?).
It's more general, and they can calculate `cellbase_maturity` by themself.
And it provide more information in only one field.

quake:
epoch is not enough, user can't find out this cell is created by cellbase or not.

yangby-cryptape:
> epoch is not enough, user can't find out this cell is created by cellbase or not.

So, I'm afraid that users may ask for a field to distinguish whether it's a cellbase or not in some day.

rpc/README.md
```diff
@@ -383,13 +383,22 @@ http://localhost:8114
...
+#### Returns
+
+    block_hash - Refer to block
+    capacity - Cell capacity
+    cellbase_maturity - Cellbase maturity, optional boolean, null means non-cellbse
```
doitian: RPC should return some determined result, instead of a value which varies overtime.

rpc/README.md
```diff
@@ -417,6 +426,7 @@ http://localhost:8114
         {
             "block_hash": "0xf293d02ce5e101b160912aaf15b1b87517b7a6d572c13af9ae4101c1143b22ad",
             "capacity": "0x2ca86f2642",
+            "cellbase_maturity": false,
```
doitian: It's better to add a new RPC such as `get_cells_by_lock_hash_v2` instead of breaking the existing one.

rpc/README.md
quake: how about change `cellbase_maturity` to enum: none / yes / no ?

rpc/README.md
quake: Creating a new RPC name for compatibility upgrade is not common practice, I would like to follow other blockchains (btc/eth)

zhangsoledad: I incline to add new RPC, move the old one to Deprecated module

quake:
@yangby-cryptape @doitian I changed field `cellbase_maturity` to `cellbase` according to your review comment. Users may use other RPC to calculate maturity. Please review again, thanks.

rpc/README.md
doitian: I prefer returning the epoch with fraction where the cell is created.

PR #1840, opened 2019-11-25
refactor: db migration

quake:
Database migration may involve multiple iterations and different db (indexer/chain store), this PR added a `Migration` trait and improve the API.


db/src/migration.rs
```diff
@@ -0,0 +1,75 @@
...
+            .map_err(|err| {
+                internal_error(format!("failed to get the version of database: {}", err))
+            })?
+            .map(|version_bytes| unsafe { String::from_utf8_unchecked(version_bytes.to_vec()) });
+
+        match db_version {
+            Some(v) => self
+                .migrations
+                .iter()
+                .filter(|m| m.version() > v.as_str())
```
doitian: Comparing string? `"0.0.15" < "0.0.2"`

db/src/migration.rs
```diff
@@ -0,0 +1,75 @@
...
+                .try_for_each(|m| {
+                    info!("Migrating database version to {}", m.version());
+                    m.migrate(db)
```
doitian: The migration script may forget to save the version. It is safer to set the version here. Also it is more consistent, because the `None` branch will write the version.

db/src/migration.rs
```diff
@@ -0,0 +1,75 @@
...
+                    })
+                } else {
+                    Ok(())
+                }
+            }
+        }
+    }
+}
+
+pub trait Migration {
```
doitian: There are versions we cannot do migrations, such as from "0.1.0" to "1.0.0". Another issue is that this PR accepts an old database such as "0.2000.0" as a compatible version and happily upgrades to "0.2500.1".

I thinks the trait should have a method `upgradable(&self, current_version: Option<&str>) -> bool`  to check whether it is possible to do a migration. The default implementation can use semver to decide whether it is possible to migrate the database.

db/src/migration.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub fn migrate(&self, db: &OptimisticTransactionDB) -> Result<()> {
+        let db_version = db
+            .get(VERSION_KEY)
+            .map_err(|err| {
+                internal_error(format!("failed to get the version of database: {}", err))
+            })?
+            .map(|version_bytes| unsafe { String::from_utf8_unchecked(version_bytes.to_vec()) });
+
+        match db_version {
+            Some(v) => self
```
doitian: We also need to check that if there's no applicable migrations, v must equal to the version of the last migration.

doitian:
I think we don't need to use semver as the database version. A simple sequence number is OK. The migration script itself is responsible to check whether it is possible to do the automatic migration.

quake:
> I think we don't need to use semver as the database version. A simple sequence number is OK. The migration script itself is responsible to check whether it is possible to do the automatic migration.

Changed to using timestamp as db migration version, please review again, thanks.

db/src/db.rs
```diff
@@ -316,34 +268,91 @@ mod tests {
...
+            );
+        }
+    }
+
+    struct CustomizedMigration;
+
+    impl Migration for CustomizedMigration {
+        fn migrate(&self, db: &RocksDB) -> Result<()> {
+            let txn = db.transaction();
+            // append 1u8 to each value of column `0`
```
yangby-cryptape: `0` to be a `const`?

quake:
> I have an **important** question: how to apply this change to existed database?
> Can we treat `0.2500.0` as a special version, and test the migration for it?

it will be migrated to version `20191127135521` :

https://github.com/nervosnetwork/ckb/pull/1840/files#diff-4047562fc22a6fbdfa0888718605fbeaR248

I have tested it on v0.25 release

db/src/db.rs
quake: Updated, please review again, thanks.

yangby-cryptape:
> > I have an **important** question: how to apply this change to existed database?
> > Can we treat `0.2500.0` as a special version, and test the migration for it?
> 
> it will be migrated to version `20191127135521` :
> 
> https://github.com/nervosnetwork/ckb/pull/1840/files#diff-4047562fc22a6fbdfa0888718605fbeaR248
> 
> I have tested it on v0.25 release

> > I have an **important** question: how to apply this change to existed database?
> > Can we treat `0.2500.0` as a special version, and test the migration for it?
> 
> it will be migrated to version `20191127135521` :
> 
> https://github.com/nervosnetwork/ckb/pull/1840/files#diff-4047562fc22a6fbdfa0888718605fbeaR248
> 
> I have tested it on v0.25 release

It's controlled by `m.version() > v.as_str()`, I think the function of this PR is correct, but there are two hidden danger:
1. In old code, we only set version at step of database initialization.

   If a database is already existed, and data is corrupt, no version number is existed, old code will throw an exception, but new code will treat it as a normal database.

2. Also, when the data is corrupt,  `m.version() > v.as_str()` is not enough to prove the database can be migrated, I think we have to check the previous version number.

   For example:

   `Init -> v1 -> v2 -> v3`

   - When we set first version number `v1`, we have to check the database is a new database.
   - When we migrate it to `v2`, the current database should be `v1`,  an initialized database is not allowed.
   - When we migrate it `v3`, the current database should be `v2`, `v2` is not allowed.
   - In future, we can add some methods to allow database do migrate cross several serial version as a shortcut: `v1 -> v2 -> v3 -> v4 -> v5` -> `v1 -> shortcut -> v5`.

quake:
> In short, my opinion is: **the current version of database should always be one of our known version.**

Current db version is always known unless user modifies db directly, I don't think that we should handle this special case. Database corruption checking is not migration's responsibility, and migration is incremental, no need to handle the batch processing.

db/src/migration.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    pub fn migrate(&self, db: &RocksDB) -> Result<()> {
+        let db_version = db
+            .inner
+            .get(VERSION_KEY)
+            .map_err(|err| {
+                internal_error(format!("failed to get the version of database: {}", err))
+            })?
+            .map(|version_bytes| unsafe { String::from_utf8_unchecked(version_bytes.to_vec()) });
```
jjyr: Why unsafe?
`String::from_utf8(..).expect("version bytes to utf8")` also works

PR #1846, opened 2019-11-27
doc: add RPC notices

doitian:


doitian:
bors r=quake,zhangsoledad

PR #1847, opened 2019-11-27
fix(ci): Fix security audit job in CI

doitian:


quake:
fixed in https://github.com/nervosnetwork/ckb/pull/1838 already

PR #1848, opened 2019-11-27
feat: add a new json rpc method `get_block_economic_state`

yangby-cryptape:
### Purpose

Replace the JSON-RPC method [`get_cellbase_output_capacity_details`].

##### Secondary Issuance

According to [CKB RFC-0015](https://github.com/nervosnetwork/rfcs/blob/dda4478/rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md#collecting-state-rent-with-secondary-issuance-and-the-nervosdao):
> Then 60% of the secondary issuance goes to the miners, 35% of the issuance goes to the NervosDAO to be distributed to the locked tokens proportionally.

Only 60% of the secondary issuance goes to the miners.
But we can't see it from the response of [`get_cellbase_output_capacity_details`].

##### Transaction Fees

According to [CKB v0.25.2 Code](https://github.com/nervosnetwork/ckb/blob/dda4ed9/util/reward-calculator/src/lib.rs#L87-L215):

```rust
    /// Miner get (tx_fee - 40% of tx fee) for tx commitment.
    /// Be careful of the rounding, tx_fee - 40% of tx fee is different from 60% of tx fee.
    pub fn txs_fees(&self, target: &HeaderView) -> Result<Capacity, Error> {
        /* ... omitted ... */
    }

    /// Earliest proposer get 40% of tx fee as reward when tx committed
    ///  ... omitted ...
    pub fn proposal_reward(&self, parent: &HeaderView, target: &HeaderView) -> Result<Capacity, Error> {
        /* ... omitted ... */
    }
```

The transaction fees which paid to miner is not same as the total transaction fee of all transactions in the mined block.
The `tx_fee` in [`get_cellbase_output_capacity_details`] is only about 60% of total transaction fee in the mined block.
And transaction fees in miner reward have two part: committed fee/reward and proposal fee/reward.

#### Help for Doing Statistics

- Currently, there is no interface can query total issuance of a block.
- Currently, there is no interface can query total transaction fee of a block.

#### Simplify the Usage

Currently, if we want to know which miner mined block-N and how many rewards he received for block-N, **we have to use a magic number 11 (it's a bit obscure, and reading the source code is the only way to get it)** and call JSON-RPC 4 times:
```rust
/* pseudo-code */
let miner =  rpc
    .get_block_hash(N)
    .map(|hash| rpc.get_block(hash).cellbase().witness().parse_miner())
    .unwrap();
let reward = rpc
    .get_block_hash(N+11)
    .map(|hash| rpc.get_cellbase_output_capacity_details(hash))
    .unwrap();
```

The new JSON-RPC method will hide the magic number and return the block hash which finalized the miner reward:
```rust
/* pseudo-code */
let hash =  rpc.get_block_hash(N).unwrap();
let miner = rpc.get_block(hash).cellbase().witness().parse_miner()).unwrap();
let reward = rpc.get_block_economic_state(hash).unwrap();
```

### Further

Currently, the `Transaction Fee` on the ckb explorer is the sum of committed reward and proposal reward in miner reward of a block, not the total transaction fee of all transactions in that block.
**It's really hard to understand the `Transaction Fee` for newbies.**

For example:
- [block-14 on mainnet](https://explorer.nervos.org/block/0x5b78979ee71a7a32868e6256874571b2b59a9e40e38e6923e653db4ae5fd3d1d) has no proposal transaction, and total transaction fee of it is `66.00000000 CKB`, but the `Transaction Fee` on the ckb explorer is `39.60000000 CKB`.
- [block-334 on mainnet](https://explorer.nervos.org/block/0xbaf6c624282a7a9c5bc673ea79d6b612de822f68cb8e3f8004bf887dd11583de) has no transaction expect cellbase, but the `Transaction Fee` on the ckb explorer is `0.40522815 CKB`.

**I think we should unify the names.**
And, I didn't find any description about how to partition the total transaction fee of all transactions in a block into committed fee/reward and proposal fee/reward in [CKB RFCs](https://github.com/nervosnetwork/rfcs/tree/dda4478/rfcs).
Did I miss it?

[`get_cellbase_output_capacity_details`]: https://github.com/nervosnetwork/ckb/tree/dda4ed9/rpc#get_cellbase_output_capacity_details

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -658,6 +658,16 @@ Returns increased issuance, miner reward and total transaction fee of a block.
 #### Parameters
 
     hash - Block hash
+#### Returns
+
+    finalized_at - The hash of the block which finalized
+    issuance::primary - Primary issuance in this block
+    issuance::secondary - Secondary issuance in this block
+    miner_reward::committed - Committed fee in miner reward
+    miner_reward::proposal - Proposal fee in miner reward
+    miner_reward::primary - Primary issuance in miner reward
+    miner_reward::secondary - Secondary issuance in miner reward
+    txs_fee - The total transaction fee of all transactions in the this block
 
 #### Examples
 
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -16,6 +16,7 @@ NOTE: This file is auto-generated. Please don't update this file directly; inste
     *   [`get_live_cell`](#get_live_cell)
     *   [`get_transaction`](#get_transaction)
     *   [`get_cellbase_output_capacity_details`](#get_cellbase_output_capacity_details)
+    *   [`get_block_economic_state`](#get_block_economic_state)
     *   [`get_block_by_number`](#get_block_by_number)
 *   [`Experiment`](#experiment)
     *   [`dry_run_transaction`](#dry_run_transaction)
@@ -640,6 +641,51 @@ http://localhost:8114
         "secondary": "0x17b93605",
         "total": "0x18e64b61cf",
         "tx_fee": "0x0"
+    }
+}
+```
+
+### `get_block_economic_state`
+
+Returns increased issuance, miner reward and total transaction fee of a block.
+
+#### Parameters
+
+    hash - Block hash
+
+#### Examples
+
+```bash
+echo '{
+    "id": 2,
+    "jsonrpc": "2.0",
+    "method": "get_block_economic_state",
+    "params": [
+        "0x02530b25ad0ff677acc365cb73de3e8cc09c7ddd58272e879252e199d08df83b"
+    ]
+}' \
+| tr -d '\n' \
+| curl -H 'content-type: application/json' -d @- \
+http://localhost:8114
+```
+
+```json
+{
+    "id": 2,
+    "jsonrpc": "2.0",
+    "result": {
+        "finalized_at": "0xa5f5c85987a15de25661e5a214f2c1449cd803f071acc7999820f25246471f40",
+        "issuance": {
+            "primary": "0x18ce922bca",
+            "secondary": "0x7f02ec655"
+        },
+        "miner_reward": {
+            "committed": "0x0",
+            "primary": "0x18ce922bca",
+            "proposal": "0x0",
+            "secondary": "0x17b93605"
+        },
+        "txs_fee": "0x0"
     }
 }
 ```
--- a/rpc/json/rpc.json
+++ b/rpc/json/rpc.json
@@ -752,6 +752,33 @@
         ]
     },
     {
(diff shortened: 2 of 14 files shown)


rpc/json/rpc.json
```diff
@@ -751,6 +751,33 @@
...
+                "proposal": "0x0",
+                "secondary": "0x17b93605"
+            },
+            "txs_fee": "0x0"
+        },
+        "types": [
+            {
+                "hash": "Block hash"
+            }
+        ]
```
quake: could you add a `returns` section to descript the result, reference:
https://github.com/nervosnetwork/ckb/blob/develop/rpc/json/rpc.json#L242

yangby-cryptape: Yes! I will do it later.
I'm not anxious for merging this PR.
**I'm waiting for more opinions.**

rpc/README.md
```diff
@@ -650,6 +651,51 @@ http://localhost:8114
 }
 ```
 
+### `get_block_economic_state`
+
+Returns increased issuance, miner reward and total transaction fee of a block.
```
doitian: What's the behavior when the proposal rewards are not finalized yet?

rpc/json/rpc.json
```diff
@@ -751,6 +751,33 @@
...
+                "primary": "0x18ce922bca",
+                "secondary": "0x7f02ec655"
+            },
+            "miner_reward": {
+                "committed": "0x0",
+                "primary": "0x18ce922bca",
+                "proposal": "0x0",
+                "secondary": "0x17b93605"
+            },
+            "txs_fee": "0x0"
```
doitian: There's another field we may want to include in the response, total withdrawn DAO rewards of all the txs.

rpc/README.md
yangby-cryptape: [Return `None` if not finalized](https://github.com/nervosnetwork/ckb/pull/1848/files#diff-c80a342630d6634092de83806d174b3eR306-R309), same as [`get_cellbase_output_capacity_details`](https://github.com/nervosnetwork/ckb/tree/dda4ed9/rpc#get_cellbase_output_capacity_details).

rpc/json/rpc.json
yangby-cryptape: [DONE!](https://github.com/nervosnetwork/ckb/pull/1848/commits/deb5d3f4ac3d37edc9a615723efe6b35dc7baa5c)

doitian:
bors r=quake,doitian

doitian:
bors ping

doitian:
bors try

doitian:
bors r=quake,doitian

PR #1851, opened 2019-11-29
chore: upgrade p2p to 0.2.7

driftluo:
I hard code the default maximum number of connections to 1024, because I haven't figured out whether it needs to be handled as a configuration item by the user

see https://github.com/nervosnetwork/p2p/blob/master/CHANGELOG.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -657,7 +657,7 @@ dependencies = [
 name = "ckb-network"
 version = "0.25.3-pre"
 dependencies = [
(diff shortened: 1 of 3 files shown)


doitian:
Is it enough for the boot nodes?

cc @TheWaWaR

TheWaWaR:
> Is it enough for the boot nodes?
> 
> cc @TheWaWaR

The number is OK for me.

doitian:

PR #1852, opened 2019-11-29
docs: add RPC warning in config file

doitian:


doitian:
bors r=quake,yangby-cryptape

PR #1853, opened 2019-11-30
feat: Update error message and prompt of ckb init subcommand

ashchan:


doitian:
bors r=doitian,keroro520,yangby-cryptape

PR #1856, opened 2019-12-03
feat: Define StatusCode to indicate the result of sync operation

keroro520:
Learned from HTTP Response, use `StatusCode` to indicate the result of sync-operation, try to replace original `Result<T, future::Error>`.
More detail: https://github.com/keroro520/ckb/blob/a66ab9a7d5eec69211a9a5073ca32691113a0c2e/sync/src/status.rs#L31-L53


sync/src/status.rs
```diff
@@ -0,0 +1,187 @@
...
+///   - x06xx: BlockTransactions
+///   - x07xx: GetBlockProposal
+///   - x08xx: BlockProposal
+///   - x09xx: GetHeaders
+///   - x10xx: SendHeaders
+///   - x11xx: GetBlocks
+///   - x12xx: SendBlock
+///   - x13xx: InIBD
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum StatusCode {
```
zhangsoledad: Under the default representation, the specified discriminant is interpreted as an isize value although the compiler is allowed to use a smaller type in the actual memory layout. The size and thus acceptable values can be changed by using a primitive representation.
`#[repr(u16)]`, suggestion

sync/src/status.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn ignored() -> Self {
+        Self::new::<&str>(StatusCode::Ignored, None)
+    }
+
+    pub fn is_ok(&self) -> bool {
+        self.code == StatusCode::OK
+    }
+
+    pub fn should_ban(&self) -> Option<Duration> {
+        if (self.code as u16) < 40000 || 50000 <= (self.code as u16) {
```
zhangsoledad: `!(40000..50000).contains(&(self.code as u16))`, suggestion

sync/src/relayer/block_transactions_process.rs
```diff
@@ -111,42 +94,32 @@ impl<'a> BlockTransactionsProcess<'a> {
...
+                        .into_iter()
+                        .flat_map(|u| u.data().proposals().into_iter())
+                        .collect();
+                    self.relayer.request_proposal_txs(
+                        self.nc.as_ref(),
+                        self.peer,
+                        block_hash.clone(),
+                        proposals,
                     );
-                };
```
zhangsoledad: Silently ignore`request_proposal_txs` error, maybe we should improve debug log rather than delete it.

sync/src/relayer/block_transactions_verifier.rs
```diff
@@ -21,24 +21,23 @@ impl BlockTransactionsVerifier {
...
-            return Err(Error::Misbehavior(
-                Misbehavior::InvalidBlockTransactionsLength {
-                    expected: missing_short_ids.len(),
-                    actual: transactions.len(),
-                },
+            return StatusCode::UnmatchedBlockTransactionsLength.with_context(format!(
+                "Expected({}) != actual({})",
+                missing_short_ids.len(),
+                transactions.len(),
             ));
```
zhangsoledad: same as above

sync/src/relayer/compact_block_process.rs
```diff
@@ -178,30 +154,25 @@ impl<'a> CompactBlockProcess<'a> {
...
-                self.nc.as_ref(),
-                self.peer,
-                block_hash.clone(),
-                proposals,
-            ) {
-                debug_target!(
-                    crate::LOG_TARGET_RELAY,
-                    "[CompactBlockProcess] request_proposal_txs: {}",
-                    err
+            {
```
zhangsoledad: same as above

zhangsoledad:
LGTM

sync/src/relayer/block_transactions_process.rs
keroro520: I had change the function from `request_proposal_txs(...) -> Result<(), FailureError>` to `request_proposal_txs (...);`, and print out the returned error inside. [diff code](https://github.com/nervosnetwork/ckb/pull/1856/files#diff-fd90e5b183cb29b0901377bac66658e0L156-L159)

So the error will not be ignored after changing.

sync/src/status.rs
```diff
@@ -0,0 +1,187 @@
...
+    ($code:expr) => {{
+        let ret = $code;
+        if !ret.is_ok() {
+            return ret;
+        }
+        ret
+    }};
+}
+
+/// StatusCodes indicate whether a specific operation has been successfully completed.
```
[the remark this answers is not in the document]
keroro520: Thanks for your suggestions.
The last 2 commits make updates based on the above suggestions:

  * Category status into 3 classes, informational(1xx), malformed error(4xx), warning(5xx)
  * Rename the code with verbose style. I also agree that a verbose name brings more readability. Please help review it again.

doitian:
bors r=doitian,zhangsoledad

PR #1865, opened 2019-12-07
chore: send_transaction rpc doc tweak

quake:
resolve doc copy-paste error and add tx struct link to rfc


doitian:
bors r=doitian,keroro520

PR #1873, opened 2019-12-11
fix: ban peer that are not on the same network

driftluo:
If it cannot be parser, ban it, only two possibilities:
1. message format error（molecule）
2. not on same net

### DIFF
--- a/network/src/protocols/identify.rs
+++ b/network/src/protocols/identify.rs
@@ -10,9 +10,10 @@ use p2p::{
     utils::{is_reachable, multiaddr_to_socketaddr},
 };
 use p2p_identify::{Callback, MisbehaveResult, Misbehavior};
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 const MAX_RETURN_LISTEN_ADDRS: usize = 10;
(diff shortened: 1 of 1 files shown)


network/src/protocols/identify.rs
```diff
@@ -56,7 +56,15 @@ impl Callback for IdentifyCallback {
         identify: &[u8],
     ) -> MisbehaveResult {
         match self.identify.verify(identify) {
-            None => MisbehaveResult::Disconnect,
+            None => {
+                self.network_state.ban_session(
+                    &context.control(),
+                    context.session.id,
+                    Duration::from_secs(3600 * 1024),
```
jjyr: * Use a constant
* 1024 hours is too long, few hours should be enough.

network/src/protocols/identify.rs
```diff
@@ -56,7 +57,15 @@ impl Callback for IdentifyCallback {
         identify: &[u8],
     ) -> MisbehaveResult {
         match self.identify.verify(identify) {
-            None => MisbehaveResult::Disconnect,
+            None => {
+                self.network_state.ban_session(
+                    &context.control(),
```
yangby-cryptape: `pub fn control(&self) -> &ServiceControl`?

PR #1875, opened 2019-12-11
feat: move feeler behind identify

driftluo:
after this pr, all protocol will open after `identify` open, avoid feeler interacting with different networks and compatible with older versions

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -459,7 +459,7 @@ impl NetworkState {
             p2p_control,
             peer_id,
             addr,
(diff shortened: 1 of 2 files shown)


jjyr:
Need an integration test to ensure the feeler protocol wound not connect to a node which from other network_id(or genesis hash).

driftluo:
`test_identify_behavior` has tested identify will not be connected by different networks
`test_feeler_behavior` has test after identify open，open feeler will close the session

PR #1876, opened 2019-12-13
chore(deps): remove dependency `serde_derive`

yangby-cryptape:
- If a crate use `serde_derive`, then any other crates which are dependent on that crate have to add `serde_derive` as a dependency too.
- [Use `serde_derive`directly is not recommended since rustc v1.31](https://github.com/serde-rs/serde/tree/d540e72fc784e459c75d167973a9ac54ef1bd731/serde_derive#serde-in-action).


yangby-cryptape:
bors r=doitian,quake

yangby-cryptape:
bors retry

Issue #1867, opened 2019-12-07
JSON-RPC subscriptions


quake:
Add JSON-RPC subscriptions, this allows clients to wait for events instead of polling for them.

### `subscribe`

Subscribe to a topic, if successful it returns the subscription id.

#### Examples

```bash
echo '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "subscribe",
    "params": ["new_tip_header"]
}' \
| tr -d '\n' \
| curl -H 'content-type: application/json' -d @- \
http://localhost:8114
```

```json
{
    "id": 2,
    "jsonrpc": "2.0",
    "result": "0x5a2e9c09c436eb1b70050a4591ce2b35"
}
```

### `unsubscribe`

Unsubscribe from a topic.

#### Examples

```bash
echo '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "unsubscribe",
    "params": ["0x5a2e9c09c436eb1b70050a4591ce2b35"]
}' \
| tr -d '\n' \
| curl -H 'content-type: application/json' -d @- \
http://localhost:8114
```

```json
{
    "id": 2,
    "jsonrpc": "2.0",
    "result": true
}
```

### new_tip_header

Send a notification with header struct each time a new tip is appended to the chain, including chain reorganizations.

### new_tip_block

Send a notification with block struct each time a new tip is appended to the chain, including chain reorganizations. 

## Possible future topics

1. new_transaction

PR #1882, opened 2019-12-17
feat: add tcp and websocket to rpc service

quake:
This PR adds tcp and websocket support for rpc, this is required for #1867 

test tcp:
```shell
telnet localhost 18114
> {"id":2,"jsonrpc":"2.0","method":"get_tip_block_number","params":[]}
< {"jsonrpc":"2.0","result":"0x337e","id":2}
```

test websocket:
```javascript
let socket = new WebSocket("ws://localhost:28114");
socket.onmessage = function(event) {
  console.log(`Data received from server: ${event.data}`);
};
socket.send('{"id":2,"jsonrpc":"2.0","method":"get_tip_block_number","params":[]}');
```

### DIFF
--- a/rpc/src/test.rs
+++ b/rpc/src/test.rs
@@ -291,7 +291,7 @@ fn setup_node(height: u64) -> (Shared, ChainController, RpcServer) {
         }
         .to_delegate(),
     );
-    let server = ServerBuilder::new(io)
+    let http = ServerBuilder::new(io)
         .cors(DomainsValidation::AllowOnly(vec![
             AccessControlAllowOrigin::Null,
             AccessControlAllowOrigin::Any,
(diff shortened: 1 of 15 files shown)


rpc/src/test.rs
```diff
@@ -544,6 +548,4 @@ fn test_rpc() {
             pretty_assert_eq!(actual, expected, "Assert results of jsonrpc",);
         }
     }
-
-    server.close();
```
doitian: If we don't close the server explicitly, the test may panic randomly when dropping objects.

quake: I think this issue has been fixed by https://github.com/paritytech/jsonrpc/pull/495

quake:
bors try

quake:
bors try

doitian:
bors r=doitian,keroro520

PR #1889, opened 2019-12-23
fix: get_cell_meta should return None if output index is not exists

jjyr:


doitian:
bors r=quake,doitian,xxuejie

Issue #1860, opened 2019-12-05
Is there any setting on configuration file just like  "blocknotify=myurl" of bitcoind.conf


loloxiao:
Is there any setting on configuration file just like  "blocknotify=myurl" of bitcoind.conf  ,  which can post new block info to my myurl  automatically?

doitian:
Not yet but in our list.

PR #1891, opened 2019-12-24
feat: notify service

quake:
This PR resolve https://github.com/nervosnetwork/ckb/issues/1860 and refactor network alert script notification by adding a notify service, and it's required to implement https://github.com/nervosnetwork/ckb/issues/1867

**configuration  file breaking change**

```diff
-# [alert_notifier]
-# # Script will be notified when node received an alert, first arg is alert message string.
-# notify_script = "echo"
+# [notifier]
+# # Execute command when the new tip block changes, first arg is block struct in json format string.
+# new_block_notify_script = "your_new_block_notify_script.sh"
+# # Execute command when node received an network alert, first arg is alert message string.
+# network_alert_notify_script = "your_network_alert_notify_script.sh"
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -385,6 +385,10 @@ impl ChainService {
                     error!("notify new_uncle error {}", e);
                 }
             }
+            let block_ref: &BlockView = &block;
+            self.shared
+                .notify_controller()
+                .notify_new_block(block_ref.clone());
             if log_enabled!(ckb_logger::Level::Debug) {
                 self.print_chain(10);
             }
--- a/notify/src/lib.rs
+++ b/notify/src/lib.rs
@@ -6,7 +6,6 @@ use ckb_types::{
 };
 use crossbeam_channel::{bounded, select, Receiver, RecvError, Sender};
 use serde::{Deserialize, Serialize};
-use serde_json::to_string;
 use std::collections::HashMap;
 use std::process::Command;
 use std::thread;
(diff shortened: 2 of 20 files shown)


notify/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+                // notify script
+                if let Some(script) = self.config.new_block_notify_script.as_ref() {
+                    let block_json: ckb_jsonrpc_types::BlockView = block.into();
+                    let args =
+                        [serde_json::to_string(&block_json).expect("serializing should be ok")];
+                    if let Err(err) = Command::new(script).args(&args).status() {
+                        error!(
+                            "failed to run new_block_notify_script script, error: {}",
+                            err
+                        );
```
keroro520: It's better to log the `command` and `args` also.

chain/src/chain.rs
```diff
@@ -385,6 +385,10 @@ impl ChainService {
                     error!("notify new_uncle error {}", e);
                 }
             }
+            let block_ref: &BlockView = &block;
+            self.shared
+                .notify_controller()
+                .notify_new_block(block_ref.clone());
```
jjyr:
```suggestion
                .notify_new_block(block);
```

notify/src/lib.rs
quake: Added `command` to the error log, `args` is too large, was skipped. Please review again, thanks.

notify/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+            Ok(block) => {
+                trace!("event new block {:?}", block);
+                // notify all subscribers
+                for subscriber in self.new_block_subscribers.values() {
+                    let _ = subscriber.send(block.clone());
+                }
+                // notify script
+                if let Some(script) = self.config.new_block_notify_script.as_ref() {
+                    let block_json: ckb_jsonrpc_types::BlockView = block.into();
+                    let args = [to_string(&block_json).expect("serializing should be ok")];
```
jjyr: A block may be very large like a few hundred KB, it may exceed the limits of Linux. we should pass the block hash instead.

quake: Updated, please review again, thanks

chain/src/chain.rs
quake: I prefer breaking change in the configuration file than maintaining multiple conf version compatibility.

PR #1896, opened 2019-12-26
chore: move protocols to ckb

driftluo:
Since the compatibility of the protocol is closely related to the behavior of the protocol itself, the version and behavior must be bound to each other for compatibility upgrades, so all external protocols are moved into the ckb project


network/Cargo.toml
```diff
@@ -17,12 +17,11 @@ tokio-threadpool = "0.1"
 futures = "0.1"
 crossbeam-channel = "0.3"
 p2p = { version="0.2.7", package="tentacle", features = ["molc"] }
-p2p-ping = { version="0.3.9", package="tentacle-ping", features = ["molc"] }
-p2p-discovery = { version="0.2.9", package="tentacle-discovery", features = ["molc"] }
-p2p-identify = { version="0.2.10", package="tentacle-identify", features = ["molc"] }
+p2p-ping = { version="0.26.2-pre", package="ckb-ping", path = "../protocols/ping" }
```
quake: I don't think we need 3 separate packages, can we move these files to `network/src/protocols`?

driftluo: I think the first step is to move into ckb without any major changes, and then merge into the network. The biggest workload here is the integration of discovery.

quake: Get, please rebase and change the version to 0.27

PR #1900, opened 2019-12-30
feat: add rpc subscription

### LINKED ISSUE #1867: JSON-RPC subscriptions

quake:
resolve #1867 

tcp rpc subscription example:
```
telnet localhost 18114
> {"id": 2, "jsonrpc": "2.0", "method": "subscribe", "params": ["new_tip_header"]}
< {"jsonrpc":"2.0","result":0,"id":2}
< {"jsonrpc":"2.0","method":"subscribe","params":{"result":"...block header json...","subscription":0}}
< {"jsonrpc":"2.0","method":"subscribe","params":{"result":"...block header json...","subscription":0}}
< ...
> {"id": 2, "jsonrpc": "2.0", "method": "unsubscribe", "params": [0]}
< {"jsonrpc":"2.0","result":true,"id":2}
```

websocket rpc subscription example:
```
let socket = new WebSocket("ws://localhost:28114")

socket.onmessage = function(event) {
  console.log(`Data received from server: ${event.data}`);
}

socket.send(`{"id": 2, "jsonrpc": "2.0", "method": "subscribe", "params": ["new_tip_header"]}`)

socket.send(`{"id": 2, "jsonrpc": "2.0", "method": "unsubscribe", "params": [0]}`)
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -2132,7 +2132,7 @@ http://localhost:8114
 
 ### `subscribe`
 
-Subscribe to a topic, if successful it returns the subscription id. For each event that matches the subscription a notification with relevant data (JSON-formatted string) is send together with the subscription id. Example: {"jsonrpc":"2.0","method":"subscribe","params":{"result":"...block header JSON-formatted string...","subscription":42}}
+Subscribe to a topic, if successful it returns the subscription id. For each event that matches the subscription a notification with relevant data (JSON-formatted string) is send together with the subscription id. Example: {"jsonrpc":"2.0","method":"subscribe","params":{"result":"...block header JSON-formatted string...","subscription":"0x2a"}}
 
 #### Parameters
 
@@ -2161,7 +2161,7 @@ http://localhost:8114
 {
     "id": 2,
     "jsonrpc": "2.0",
-    "result": 42
+    "result": "0x2a"
 }
 ```
 
@@ -2184,7 +2184,7 @@ echo '{
     "jsonrpc": "2.0",
     "method": "unsubscribe",
     "params": [
-        42
+        "0x2a"
     ]
 }' \
 | tr -d '\n' \
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -8,6 +8,8 @@ Allowing arbitrary machines to access the JSON-RPC port (using the `rpc.listen_a
 
 CKB JSON-RPC only supports HTTP now. If you need SSL, please setup a proxy via Nginx or other HTTP servers.
 
+Subscriptions require a full duplex connection. CKB offers such connections in the form of tcp (enable with `rpc.tcp_listen_address` configuration option) and websockets (enable with `rpc.ws_listen_address`).
+
 
 *   [`Chain`](#chain)
     *   [`get_tip_block_number`](#get_tip_block_number)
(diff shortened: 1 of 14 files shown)


rpc/README.md
```diff
@@ -2122,3 +2127,75 @@ http://localhost:8114
...
+
+## Subscription
+
+### `subscribe`
+
+Subscribe to a topic, if successful it returns the subscription id.
+
+#### Parameters
+
+    topic - Subscription topic (enum: new_tip_header | new_tip_block)
```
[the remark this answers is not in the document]
quake: Updated, please review again, thanks.
https://github.com/nervosnetwork/ckb/pull/1900/files#diff-622e6d119ac5d43f7eb41cb596159f9fR11

rpc/src/module/subscription.rs
```diff
@@ -0,0 +1,170 @@
...
+#[derive(Default)]
+pub struct SubscriptionRpcImpl {
+    pub(crate) id_generator: AtomicUsize,
+    pub(crate) subscribers: Arc<RwLock<HashMap<Topic, Subscribers>>>,
+}
+
+impl SubscriptionRpc for SubscriptionRpcImpl {
+    type Metadata = Option<SubscriptionSession>;
+
+    fn subscribe(&self, meta: Self::Metadata, subscriber: Subscriber<String>, topic: Topic) {
```
keroro520: What if the subscriber makes a duplicated subscription to the same topic? It seems that it will receive the same notify-messages multiple times. So my question is, should we check the duplicated subscription?

rpc/src/module/subscription.rs
```diff
@@ -0,0 +1,170 @@
...
+                session
+                    .subscription_ids
+                    .write()
+                    .expect("acquiring subscription_ids write lock")
+                    .insert(id);
+            }
+        }
+    }
+
+    fn unsubscribe(&self, meta: Option<Self::Metadata>, id: SubscriptionId) -> Result<bool> {
```
keroro520: What about using `id: Uint64`?

rpc/src/module/subscription.rs
quake: I don't have any preference about duplicate checking, however, it's not easy to implement the checking, because of the active subscriptions of session are private: https://github.com/paritytech/jsonrpc/blob/master/pubsub/src/subscription.rs#L18 , and we need to create a PR for jsonrpc crate first, then add some extra logic code.

rpc/src/module/subscription.rs
quake: Updated subscription id to hex string, please review again, thanks.

keroro520:
@quake Fix CI, `make fmt` failed.

doitian:
bors r=doitian,keroro520

PR #1902, opened 2020-01-02
chore: Rename withdraw_header to withdrawing_header in DaoCalculator

TheWaWaR:
`withdrawing_header` is the header of block where dao prepare transaction in.


doitian:
bors r=doitian,driftluo

PR #1904, opened 2020-01-02
chore: print disconnect reason as info level log

doitian:
The disconnect reason from peer is a useful information to check out network
issues.


doitian:
bors retry

PR #1908, opened 2020-01-12
feat: periodically disconnect peers which open invalid sub-protocols

jjyr:
Add a new background network service, ProtocolTypeChecker:

CKB evicts inactive peers in `sync` protocol; but due to P2P connection design, a malicious peer may choose not to open `sync` protocol, to sneak from the eviction mechanism; this service periodically check peers opened sub-protocols, to make sure no malicious connection.

Currently, 2 sub-protocols types are valid:

1. fully-opened: all sub-protocols(except feeler) are opened.
2. feeler: only open feeler protocol is open.

Other protocols will be closed after a timeout.

### DIFF
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -109,9 +109,13 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
             Arc::clone(&network_state),
         ),
     ];
+
+    let required_protocol_ids = vec![NetworkProtocol::SYNC.into()];
+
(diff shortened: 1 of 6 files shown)


network/Cargo.toml
```diff
@@ -36,6 +36,7 @@ ckb-types = { path = "../util/types" }
 ipnetwork = "0.14"
 serde_json = "1.0"
 tempfile = "3.0.7"
+failure = "0.1"
```
driftluo: In terms of code usage, failure is an unnecessary dependency

PR #1915, opened 2020-01-17
feat: Reject new scripts with known bugs

xxuejie:
For compatibility reasons, there're certain bugs that we have to leave
to the next hardfork to fix. However those bugs, especially VM bugs
might lead to surprising unexpected behaviors. This change adds a new
checker that checks against newly created cells for scripts with bugs,
and reject those transaction when we can. This way we can alert users
about the bugs as early as we can.


script/src/known_bugs_checker.rs
```diff
@@ -0,0 +1,137 @@
...
+use ckb_vm::{
+    instructions::{extract_opcode, i, m, rvc, Instruction, Itype, Stype},
+    registers::RA,
+};
+use ckb_vm_definitions::instructions as insts;
+use goblin::elf::{section_header::SHF_EXECINSTR, Elf};
+
+const CKB_VM_ISSUE_92: &str = "https://github.com/nervosnetwork/ckb-vm/issues/92";
+
+pub struct KnownBugsChecker<'a> {
```
janx: `IllTransactionChecker` might be a better name.

xxuejie: What does `Ill` mean?

janx: Means some transaction are considered "unhealthy".

xxuejie: Oops, silly me😂😂😂

jjyr: IMO IllInstructionChecker is more clear.

jjyr:
I still think we should not check the relayed txs. Here are the reasons:

1. The intention of this mechanism is to hint the user to avoid potential VM bugs. If a contract author sends tx via a recent client he already gets the error message hint. Instead, if the user sends tx via an old client there maybe two results: 1 he observes that the tx gets confirmed without an error message, the potential bugs will still exists in his script. 2 he observes that the tx never gets confirmed, and fall into self-doubt. both situations do not help.
2. We scan and decode every instruction to check the potential error, its OK in RPC since every user sends tx from the user's own node, it naturally decentralized; but for relayed txs, I doubt there may be a potential weak point to DDOS, and may drawback the relay performance.

xxuejie:
> I still think we should not check the relayed txs. Here are the reasons:
> 
> 1. The intention of this mechanism is to hint the user to avoid potential VM bugs. If a contract author sends tx via a recent client he already gets the error message hint since we check instructions in RPC. Instead, if the user sends tx via an old client there maybe two results: 1 he observes that the tx gets confirmed without an error message, the potential bugs will still exist in his script. 2 he observes that the tx never gets confirmed since other nodes use a new version client, and fall into self-doubt. both situations do not help the author.
> 2. We scan and decode every instruction to check the potential error, its OK in RPC since every user sends tx from the user's own node, it naturally decentralized; but for relayed txs, I doubt there may be a potential weak point to DDOS, and may drawback the relay performance.

Personally, I'm good either direction. Thoughts? @doitian @janx

tx-pool/src/config.rs
```diff
@@ -27,6 +27,8 @@ pub struct TxPoolConfig {
     pub max_committed_txs_hash_cache_size: usize,
     // txs with lower fee rate than this will not be relayed or be mined
     pub min_fee_rate: FeeRate,
+    // Rejects txs with scripts that might trigger known bugs
+    pub reject_known_bugs: bool,
```
doitian: It will break existing config file because by default serde will require the field `reject_known_bugs`.  It is recommended to add `#[serde(default)]`. See https://serde.rs/field-attrs.html for details.

script/src/known_bugs_checker.rs
```diff
@@ -0,0 +1,137 @@
...
+    tx: &'a TransactionView,
+}
+
+impl<'a> KnownBugsChecker<'a> {
+    pub fn new(tx: &'a TransactionView) -> Self {
+        KnownBugsChecker { tx }
+    }
+
+    pub fn check(&self) -> Result<(), Error> {
```
doitian: It can be more helpful if the error can tell the output index.

doitian:
Maybe we need separate switchers for RPC and Relay

sync/src/relayer/transactions_process.rs
```diff
@@ -92,6 +94,18 @@ impl<'a> TransactionsProcess<'a> {
                     );
                     return None;
                 }
+                // skip txs with known bugs
+                if reject_known_bugs {
+                    if let Err(e) = IllTransactionChecker::new(&tx).check() {
```
doitian: Does it be renamed back to the original name?

xxuejie: Could you elaborate here? I'm not sure I get what this means

doitian: I thought the name was KnownBugsChecker or alike when I reviewed last time

xxuejie: Yes there's a suggestion to rename `KnownBugsChecker` to `IllTransactionChecker`. Do you mean to rename `reject_known_bugs` as well?

doitian: My fault, I thought the comment 
 https://github.com/nervosnetwork/ckb/pull/1915#pullrequestreview-344408399 was asking to rename the class to KnownBugsChecker, instead it is the reverse.

doitian:
bors r=doitian,zhangsoledad

PR #1916, opened 2020-01-21
fix: transaction should be relayed when node connects peers

quake:


doitian:
bors r=doitian,zhangsoledad,driftluo

PR #1922, opened 2020-01-29
fix: CKBProtocolContext#connected_peers should filter peers by protoc…

quake:
…ol id


doitian:
bors r=doitian,TheWaWaR

PR #1926, opened 2020-02-07
chore(deps): upgrade molecule, ckb-vm and bytes

yangby-cryptape:
### Features

- bump `molecule` from 0.4.2 to 0.5.0
- bump `ckb-vm` from 1.18.2 to 1.19.0
- bump `bytes` from 0.4.x to 0.5.x

### Next

Since `bytes` in `p2p` is not upgraded in this PR, I add few code to convert `bytes 0.4` and `bytes 0.5`, we have to remove them when we upgrade all `bytes` to 0.5.

- `TODO update-after-upgrade-p2p` (in another PR)


yangby-cryptape:
bors r=doitian,driftluo

PR #1929, opened 2020-02-10
chore: ignore spin warning

driftluo:
`spin` is a stable synchronization primitive implementation, not updating for 5 months is not a security issue, just ignore this warning

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,8 @@ clippy: setup-ckb-test ## Run linter to examine Rust source codes.
 
 .PHONY: security-audit
 security-audit: ## Use cargo-audit to audit Cargo.lock for crates with security vulnerabilities.
-	cargo audit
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -117,7 +117,8 @@ clippy: setup-ckb-test ## Run linter to examine Rust source codes.
 
 .PHONY: security-audit
 security-audit: ## Use cargo-audit to audit Cargo.lock for crates with security vulnerabilities.
-	cargo audit
+	# https://rustsec.org/advisories/RUSTSEC-2019-0031: spin is no longer actively maintained, it's not a problem
+	cargo audit --ignore RUSTSEC-2019-0031
```
doitian: Could you test whether we can pass `--deny-warnings` now? It's better to fail the CI on any warning as well.

driftluo: Now if there is a warning message, ci cannot pass

PR #1931, opened 2020-02-11
feat: collect metrics by logger

keroro520:


doitian:
I prefer `ckb-logger`. If we really want to save the metrics into another file, we can add dispatch feature in logger later. The log may support key value in the [future](https://docs.rs/log/0.4.10/log/kv/index.html), which can simplify logging metrics via logger.

util/logger/src/lib.rs
```diff
@@ -106,6 +106,17 @@ macro_rules! log_enabled_target {
     };
 }
 
+#[macro_export(local_inner_macros)]
+macro_rules! metric {
+    ($class:expr, $context:expr) => {
```
doitian: Better avoid the name `class`, which have too many meanings. 

For the metrics, I recommend that it should contains 3 kinds of data:

* label: which type of metric
* tags: used to filter metrics, for example, tx hash, peer id are good candidate of tags
° values: metric values, such as served requests, validated transaction counts

FYI:

- https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/
- https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md

A sample JSON may look like:

```json
{"metric":"p2p_requests", "tags":{"peer_id": "0"}, "values": {"requests": 1}, "timestamp": "..."}
```

util/logger/src/lib.rs
```diff
@@ -144,33 +155,67 @@ impl Logger {
...
+                                }),
+                        )
+                    } else {
+                        None
+                    }
+                });
 
                 while let Ok(Message::Record(record)) = receiver.recv() {
+                    // log to metrics-file
+                    if record.starts_with("ckb-metrics ") {
```
doitian: It's better to setup different message enum variant like `Message::Metric`.

keroro520:
> I prefer `ckb-logger`. If we really want to save the metrics into another file, we can add dispatch feature in logger later. The log may support key value in the [future](https://docs.rs/log/0.4.10/log/kv/index.html), which can simplify logging metrics via logger.

Yet there's no important reason to save into another file. Thanks!

sync/src/relayer/mod.rs
```diff
@@ -142,7 +142,25 @@ impl Relayer {
                 ban_time,
                 status
             );
+            metric!({
+                "measurement": "error",
+                "tags": {"source": item_name, "status": format!("{:?}", status.code()) },
+                "fields": {"count": 1},
```
jjyr: I noticed sometimes the fields are `count: 1` other time the fields just be empty, what's the difference?

util/logger/src/lib.rs
```diff
@@ -57,6 +58,15 @@ macro_rules! error {
     }
 }
 
+/// Enable metrics collection feature by setting `ckb-metrics=trace` in logger filter.
+#[macro_export(local_inner_macros)]
+macro_rules! metric {
+    ($( $args:tt )*) => {
+        let output = $crate::json!($( $args )*);
+        $crate::internal::trace!(target: "ckb-metrics", "{}", output);
```
jjyr: This PR use "measurement": "error" everywhere, if the output does not include `CARGO_PKG_NAME`, it will be hard to distinguish these metrics.

I suggest attach `"package": "CARGO_PKG_NAME"` to the output json. Or add a prefix `${CARGO_PKG_NAME}_error` to "measurement" fields.

jjyr: Also, IMO the "measurement" is not a good name, we can just use "name" or "topic" as the key.

sync/src/relayer/mod.rs
keroro520: It's my fault that uses different style at different places.
I will add `{"count": 1}` to these empty fields. Thanks.

util/logger/src/lib.rs
keroro520: > This PR use "measurement": "error" everywhere, if the output does not include `CARGO_PKG_NAME`, it will be hard to distinguish these metrics.
> 
> I suggest attach `"package": "CARGO_PKG_NAME"` to the output json. Or add a prefix `${CARGO_PKG_NAME}_error` to "measurement" fields.

Adding `"package": "CARGS_PKG_NAME"` is a good filter to identify the source of metrics!  We can reuse "logger target" as the package names.

keroro520: > Also, IMO the "measurement" is not a good name, we can just use "name" or "topic" as the key.

Here use the keywords as same as https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/
Could you please explain why "name"/"topic" is better than "measurement"?

sync/src/relayer/mod.rs
jjyr: Why not just empty the fields since the key pair have no actually meaning?

util/logger/src/lib.rs
doitian: Reusing the log target as measure name is a good idea. We can use `measure.{measure_name}` as the logger target. The log body only need to include measure value and tags.

doitian: This way we can enable/disable each measure via logger filter

jjyr: "name"/"topic" is shorter, easier to type, no chance typo.

Do we have any reason to use the same key to a specific backend DB?

jjyr: > Reusing the log target as measure name is a good idea. We can use `metric.{metric_name}` as the logger target. The log body only need to include measure value and tags.

But how do we enable all metrics? Does RUST_LOG support this `key.*=trace`?

sync/src/relayer/mod.rs
keroro520: Keep consistency. So that when integrates these metrics, we can sum the total events by simply `sum += metric["field"]["count"]`.

The field "count" will be used to sum the number of happened events. For example, `{ "topic": "fresh_transaction", "field": { "count" : 2 } }` means that we require `2` fresh transactions.

util/logger/src/lib.rs
keroro520: I don't think we have the requirements to enable/disable specific measures.
All or never.
I think that adding `tags: { "package": LOGGER_TARGET }` is just for distinguishing and filtering.

sync/src/relayer/mod.rs
jjyr: IMO we should only keep the consistency under the same "topic". It's unavoidable to extend specific fields, for example:

```
                metric!({
                     "measurement": "reorg",
                     "tags": { "source": "chain", },
                     "fields": { "attached": fork.attached_blocks.len(), "detached": fork.detached_blocks.len(), },
                 });
```

The "reorg" also has no `count: 1` field.

For every "fresh_transaction" it makes sense to keep the key "count", but for "error" or "warn", the field is meaningless, just junk outputs to increase the log size.

keroro520: Got it!

util/logger/src/lib.rs
keroro520: Update in last commit 630ad552e10867a7275c131f730cb4eaa37c4022
    * Remove useless field `{"count": 1}`
    * Add new tag `{"target": crate::env!("CARGO_PKG_NAME")}` by default
    * Rename "measurement" to "topic"
    * Rename "source" to "input"
    * Rename "warn" to "warning"

Please take a look. Thanks.

sync/src/relayer/mod.rs
```diff
@@ -142,7 +142,25 @@ impl Relayer {
                 ban_time,
                 status
             );
+            metric!({
+                "topic": "error",
+                "tags": { "target": crate::LOG_TARGET_RELAY, "input": item_name, "status": format!("{:?}", status.code()) },
+                "fields": {},
```
jjyr: We can remove the entire `fields` key pair

keroro520: I think that keeping a consistent format is good for processing.
`"count"` can be omitted since different kinds of metrics could have their own fields;
But `"fields"` as a part of the metric format is better to show explicitly.

As for the log size, metrics logs are few. I don't think it brings the benefit that omits `{"fields": {}}`

jjyr: I guess we must have a program to convert the output and feed it to statistic DB? so the `fields: {}` actually not help for processing.

IMO it is not about the "consistent", normally we ignore the "target" key in logging macro, it is more likely for convenience and I believe nobody thinks its broken the consistency of the code. It is the same thing for the `metrics`, we just empty the field to a default value, we do not either break the `json` format or the scheme, just empty it. The `"fields": {}` in the source code and output log can be very confusing, most people will think some bug causes the output `{}`.

keroro520: `fields` is a part of the format.
Just like `cell_deps` is a part of a transaction format, we will keep `cell_deps: []` in the responded transaction in JSON even if it is empty.

keroro520: Update in d14635b4058b63b06a7d3dcaf2ad5488d33f8431:

* Fill `fields: {}` for metrics by default.

keroro520:
bors r= doitian, jjyr

PR #1941, opened 2020-02-13
feat: uses feature flags to enable deadlock detection

zhangsoledad:
we should disable deadlock detection by default.  
use the `deadlock_detection` feature flag enable  deadlock detection.


zhangsoledad:
bors r= doitian, quake , driftluo

PR #1944, opened 2020-02-18
test: add a unit test about getting uncle block body

keroro520:
This PR just provides a test case to produce the problem: Getting the body of an uncle block, which has been inserted at once, will return empty.

A possible reason is that we don't restore the new snapshot after the uncle blocks inserting [code](https://github.com/nervosnetwork/ckb/blob/1447929d4516e972e6f6bb7c16f31bbd33ccc2c8/chain/src/chain.rs#L369).

@zhangsoledad Please help review&debug it.

### DIFF
--- a/chain/src/tests/mod.rs
+++ b/chain/src/tests/mod.rs
@@ -3,4 +3,5 @@ mod block_assembler;
 mod delay_verify;
 mod find_fork;
 mod reward;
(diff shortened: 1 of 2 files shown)


keroro520:
https://github.com/nervosnetwork/ckb/pull/1947

PR #1947, opened 2020-02-19
refactory: repair using of snapshot

zhangsoledad:
## Fixes

  - snapshot not refresh after insert block
  - snapshot `get_pinned` isolation

## Refactory
Reimplement `sync-snapshot`, rename to `active-chain`,  prevent misuse of snapshot.
chain data are immutable, we don't need snapshot freeze it.
also, access chain data from snapshot in network-handler may inconsistent with memory state, cause network process are asynchronize, it hard to keep snapshot sync with memory state.

## Bench
 https://github.com/nervosnetwork/ckb-bench/issues/7

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -659,6 +659,86 @@ impl SyncSharedState {
     pub fn consensus(&self) -> &Consensus {
         self.shared.consensus()
     }
+
+    pub fn insert_new_block(
+        &self,
+        chain: &ChainController,
+        pi: PeerIndex,
+        block: Arc<core::BlockView>,
+    ) -> Result<bool, FailureError> {
+        // Insert the given block into orphan_block_pool if its parent is not found
+        if !self.snapshot().known_parent(&block) {
+            debug!(
+                "insert new orphan block {} {}",
+                block.header().number(),
+                block.header().hash()
+            );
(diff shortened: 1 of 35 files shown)


util/snapshot/src/lib.rs
```diff
@@ -73,6 +73,18 @@ impl Snapshot {
         }
     }
 
+    pub fn refresh(&self, store: StoreSnapshot) -> Snapshot {
```
doitian: Why we need `refresh`? When should we `refresh`? Please add some comment here to explain it.

sync/src/types.rs
```diff
@@ -838,41 +988,45 @@ impl SyncState {
...
+
+impl ActiveChain {
+    pub fn store(&self) -> &ChainDB {
+        self.shared.store()
     }
 
-    pub fn get_block(&self, hash: &packed::Byte32) -> Option<core::BlockView> {
-        self.store.get_block(hash)
+    pub fn snapshot(&self) -> &Snapshot {
+        &self.snapshot
```
keroro520: It brings a potential risk of misusing the exposed `snapshot` and `store`.

What about keeping `snapshot` and `store` private and wrapping `pub fn get_block_hash`/`pub fn get_block_header`/`pub fn get_block` .

PR #1948, opened 2020-02-20
feat: use module disable error instead of method not found

### LINKED ISSUE #1901: Return RPC disabled instead of not found when the module is not enabled

driftluo:
use module disable error instead of method not found

close #1901

### DIFF
--- a/rpc/src/server.rs
+++ b/rpc/src/server.rs
@@ -8,7 +8,7 @@ use jsonrpc_server_utils::cors::AccessControlAllowOrigin;
 use jsonrpc_server_utils::hosts::DomainsValidation;
 use jsonrpc_tcp_server;
 use jsonrpc_ws_server;
-use std::net::ToSocketAddrs;
+use std::{collections::HashSet, net::ToSocketAddrs};
 
 pub struct RpcServer {
     pub(crate) http: jsonrpc_http_server::Server,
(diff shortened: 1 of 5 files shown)


rpc/src/server.rs
```diff
@@ -104,3 +104,126 @@ impl RpcServer {
...
     }
 }
+
+use crate::error::RPCError;
+use futures::{future::ok, future::Either, Async, Future, Poll};
+use jsonrpc_core::{middleware::Middleware, Call, Id, Metadata, Output, Request, Response};
+use std::sync::Arc;
+
+#[derive(Clone, Default)]
+pub struct ModuleEnableCheck(Arc<HashSet<String>>);
```
quake: suggest changing `HashSet<String>` to `HashMap<String, String>`  and change related module code to 
```
if self.config.chain_enable() {
  ...
} else {
  rpc_method.into_iter().each(|(method, _)| disabled_module_methods.insert("Indexer", method)) 
}
```

and we can output the detail disabled module name in the response message:

```
you need to enable `xxx` module to invoke `xxx` rpc, please modify `rpc.modules` of configuration file ckb.toml and restart the ckb node
```

driftluo:
```
rpc_method.into_iter().each(|(method, _)| disabled_module_methods.insert("Indexer", method))
```
This can only record one method, maybe it should be `hashmap<method, Arc<module>>`, but this will increase the complexity of the code

quake: I modified comment code please check again

driftluo: Done it, please review again

quake:
LGTM, one suggestion to using a static hashmap:
```
lazy_static!{
    static ref DISABLED_METHODS_MODULE: RwLock<HashMap<String>> = RwLock::new(HashMap::new());
}
```

open for other suggestions

driftluo:
I don't think it is necessary to use rwlock.

The current approach is globally unique and read-only. After use rwlock, you need to do a read lock operation before reading. I don't think it makes any sense.

rpc/src/server.rs
```diff
@@ -104,3 +104,128 @@ impl RpcServer {
...
+                        "You need to enable `{module}` module to invoke `{method}` rpc, \
+                        please modify `rpc.modules` {miner_info} of configuration file ckb.toml and restart the ckb node",
+                        method = method.method, module = module, miner_info = if module == "Miner" {"and `block_assembler`"} else {""}
+                    ),
+                )),
+                method.id.clone(),
+                method.jsonrpc,
+            )
+        };
+        match request {
```
doitian: Instead of adding a pre-middleware, it's better to implement it as a post-middleware. The middleware only checks whether the module is enabled when the handler returns `Error::method_not_found`.

driftluo: post-middleware is not a better solution, at least not here.

1. There is no request information in the response, which means you have to save(clone） a piece of request information for possible error return
2. Especially for batch requests, the complexity is doubled

quake:
> I don't think it is necessary to use rwlock.
> 
> The current approach is globally unique and read-only. After use rwlock, you need to acquire a read lock operation before reading. I don't think it makes any sense.

sorry for the late reply, `Arc<HashMap>` looks good to me after a more detail review. Please check the review suggestions below:

1.  rename `disable_method` to `disabled_methods` and change its type to `HashMap<String, String>`

2. remove inner `Arc` of  HashMap value since the map itself is `Arc` reference already:
```rust
pub struct ModuleEnableCheck(Arc<HashMap<String, String>>)
```

3. add a helper method to `ServiceBuilder`

```rust
    fn update_disabled_methods<I, M>(&mut self, module: &str, io_delegate: I)
    where
        I: IntoIterator<Item = (String, M)>,
    {
        io_delegate
            .into_iter()
            .map(|(method, _)| method)
            .for_each(|method| {
                self.disabled_methods.insert(method, module.to_string());
            });
    }
```

it will help to DRY

```rust
        if self.config.chain_enable() {
            self.io_handler.extend_with(rpc_method);
        } else {
            self.update_disabled_methods("Chain", rpc_method);
        }
```

driftluo:
I think any other comments are okay, only the second one is worth considering:
1. The module name and method name have a one-to-many relationship, and there is no need to store multiple strings.
2. Middleware may be used on multiple threads, there is no problem using inner `Arc` here

rpc/src/service_builder.rs
```diff
@@ -43,16 +52,17 @@ impl<'a> ServiceBuilder<'a> {
...
-                    shared,
-                    sync_shared_state,
-                    min_fee_rate,
-                    reject_ill_transactions,
-                )
-                .to_delegate(),
-            );
+            self.io_handler.extend_with(rpc_method);
+        } else {
+            self.update_disabled_methods("Pool", rpc_method);
```
doitian: Why not just fill the disabled methods with a handler which just returns the error message?

driftluo: I don't know what you mean

doitian:
```
for each method in rpc_method:
  hander.extend(method, |...| { return "Disabled" })
```

driftluo: This is also a solution, I can't decide which is better, but I chose middleware interception at the time

doitian: I vote against middleware since it will have an overhead.

driftluo: ok, I'll change to add disable method

PR #1957, opened 2020-03-06
feat: concurrent download blocks on ibd

driftluo:
As the chain height rises gradually, the synchronization duration increases linearly. This PR is the beginning of improving the synchronization time.

ckb synchronization is roughly divided into two blocks:
- ibd(init block download), when tip block timestamp < now - 24h
- Normal synchronization, when tip block timestamp > now - 24h

In the ibd process, currently only request a header map from a random node, and then initiate a download block request to the node based on the node's best-known header and the last common block calculated by the current tip.

In principle, block download does not limit the number of nodes requested, but because the ibd process only interacts with a node's header map during the ibd process, currently ibd can only download blocks from one node.

This pr uses the `headers_locator_hash_list` in the `getheader` request from the other node to try to restore the best-known header of the node through the locally synchronized header map and then expands the number of downloadable block nodes from one to the current outbound number

I used a Hong Kong cloud host for testing, its configuration is:
- 4c 8g 100G hard drive 10M bandwidth
- 4 Intel (R) Xeon (R) Platinum 8163 CPU @ 2.50GH

before：

Sync from 0 to 1013645, took 184 minutes, average speed 5509/min
Sync from 0 to 1017469，took 148 minutes，average speed 6874/min

after:

8 outbound peer, sync from 0 to 1015441，took 72 minutes，average speed 14103/min
8 outbound peer, sync from 0 to 1017339，took 102 minutes，average speed 9974/min
8 outbound peer, sync from 0 to 1017339，took 100 minutes，average speed 10173/min
16 outbound peer, sync from 0 to 1023994，took 93 minutes，average speed 11010/min
32 outbound peer, sync from 0 to 494467，took 110 minutes，average speed 4495/min

When the number of sync nodes reaches 32, it seems that the download logic itself is a bit problem and further observation is needed, but this does not affect the modification of pr itself.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -1,6 +1,6 @@
 use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
-use crate::types::{ActiveChain, HeaderView};
+use crate::types::{ActiveChain, HeaderView, IBDState};
 use crate::MAX_BLOCKS_IN_TRANSIT_PER_PEER;
 use ckb_logger::{debug, trace};
 use ckb_network::PeerIndex;
@@ -10,17 +10,20 @@ pub struct BlockFetcher {
     synchronizer: Synchronizer,
     peer: PeerIndex,
     active_chain: ActiveChain,
+    ibd: IBDState,
 }
 
 impl BlockFetcher {
-    pub fn new(synchronizer: Synchronizer, peer: PeerIndex) -> Self {
+    pub fn new(synchronizer: Synchronizer, peer: PeerIndex, ibd: IBDState) -> Self {
         let active_chain = synchronizer.shared.active_chain();
         BlockFetcher {
             peer,
             synchronizer,
             active_chain,
+            ibd,
         }
     }
+
     pub fn reached_inflight_limit(&self) -> bool {
         let inflight = self.synchronizer.shared().state().read_inflight_blocks();
 
@@ -68,6 +71,13 @@ impl BlockFetcher {
                 self.peer
             );
             return None;
+        }
+
+        if let IBDState::In = self.ibd {
+            self.synchronizer
+                .shared
+                .state()
+                .try_update_best_known_with_unknown_header_list(self.peer)
         }
 
         let best_known_header = match self.peer_best_known_header() {
--- a/sync/src/synchronizer/mod.rs
+++ b/sync/src/synchronizer/mod.rs
@@ -12,7 +12,7 @@ use self::get_headers_process::GetHeadersProcess;
 use self::headers_process::HeadersProcess;
 use self::in_ibd_process::InIBDProcess;
 use crate::block_status::BlockStatus;
-use crate::types::{HeaderView, PeerFlags, Peers, SyncShared};
+use crate::types::{HeaderView, IBDState, PeerFlags, Peers, SyncShared};
 use crate::{
     Status, StatusCode, BAD_MESSAGE_BAN_TIME, CHAIN_SYNC_TIMEOUT, EVICTION_HEADERS_RESPONSE_TIME,
     HEADERS_DOWNLOAD_TIMEOUT_BASE, HEADERS_DOWNLOAD_TIMEOUT_PER_HEADER, MAX_HEADERS_LEN,
@@ -150,8 +150,12 @@ impl Synchronizer {
         }
     }
 
-    pub fn get_blocks_to_fetch(&self, peer: PeerIndex) -> Option<Vec<packed::Byte32>> {
-        BlockFetcher::new(self.clone(), peer).fetch()
+    pub fn get_blocks_to_fetch(
+        &self,
+        peer: PeerIndex,
+        ibd: IBDState,
+    ) -> Option<Vec<packed::Byte32>> {
+        BlockFetcher::new(self.clone(), peer, ibd).fetch()
     }
 
     fn on_connected(&self, nc: &dyn CKBProtocolContext, peer: PeerIndex) {
@@ -344,13 +348,20 @@ impl Synchronizer {
         }
     }
 
-    fn find_blocks_to_fetch(&self, nc: &dyn CKBProtocolContext) {
+    fn find_blocks_to_fetch(&self, nc: &dyn CKBProtocolContext, ibd: IBDState) {
         let peers: Vec<PeerIndex> = {
             self.peers()
                 .state
                 .read()
                 .iter()
-                .filter(|(_, state)| state.sync_started)
+                .filter(|(_, state)| match ibd {
+                    IBDState::In => {
+                        state.peer_flags.is_outbound
+                            || state.peer_flags.is_whitelist
+                            || state.peer_flags.is_protect
+                    }
+                    IBDState::Out => state.sync_started,
+                })
                 .map(|(peer_id, _)| peer_id)
                 .cloned()
                 .collect()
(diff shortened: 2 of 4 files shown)


sync/src/types.rs
```diff
@@ -836,6 +868,16 @@ impl SyncState {
     pub fn get_orphan_block(&self, block_hash: &Byte32) -> Option<core::BlockView> {
         self.orphan_block_pool.get_block(block_hash)
     }
+
+    pub fn insert_peer_unknown_header_list(&self, pi: PeerIndex, header_list: Vec<Byte32>) {
```
quake: It's more common to pass slice as arg rather than vector when you just want to provide read-only access

```suggestion
    pub fn insert_peer_unknown_header_list(&self, pi: PeerIndex, header_list: &[Byte32]) {
```
then we can remove the mut modifier `block_locator_hashes` from `get_headers_process.rs`

```
insert_peer_unknown_header_list(self.peer, &block_locator_hashes[1..])
```

sync/src/types.rs
```diff
@@ -432,6 +437,33 @@ impl Peers {
...
+
+    pub fn clear_unknown_list(&self) {
+        self.state.write().values_mut().for_each(|state| {
+            if !state.unknown_header_list.is_empty() {
+                state.unknown_header_list.clear()
+            }
+        })
+    }
+
+    pub fn get_unknown_list(&self, peer: PeerIndex) -> Option<Vec<Byte32>> {
```
quake:
```suggestion
    pub fn take_unknown_list(&self, peer: PeerIndex) -> Option<Vec<Byte32>> {
        self.state.write().get_mut(&peer).map(|state| ::std::mem::take(&mut state.unknown_header_list))
    }
```

sync/src/synchronizer/mod.rs
```diff
@@ -345,12 +345,21 @@ impl Synchronizer {
...
     fn find_blocks_to_fetch(&self, nc: &dyn CKBProtocolContext) {
+        let is_ibd = { self.shared.snapshot().is_initial_block_download() };
         let peers: Vec<PeerIndex> = {
             self.peers()
                 .state
                 .read()
                 .iter()
-                .filter(|(_, state)| state.sync_started)
+                .filter(|(_, state)| {
+                    if is_ibd {
```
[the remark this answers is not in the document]
driftluo: Can't use this can sync function. In ibd, this function will always return false except the header sync node.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -34,6 +34,16 @@ impl BlockFetcher {
     }
 
     pub fn peer_best_known_header(&self) -> Option<HeaderView> {
+        // update best known header and clean up unknown list
```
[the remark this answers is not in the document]
doitian: I agree it. We also can skip checking the unknown list when the node is not in IBD.

sync/src/types.rs
```diff
@@ -432,6 +437,33 @@ impl Peers {
     pub fn disconnected(&self, peer: PeerIndex) -> Option<PeerState> {
         self.state.write().remove(&peer)
     }
+
+    pub fn insert_unknown_list(&self, peer: PeerIndex, hash: Byte32) {
```
doitian: The name is confusing, I suggest changing it to `insert_into_unknown_list` or `insert_unknown_block_hash`.

sync/src/synchronizer/block_fetcher.rs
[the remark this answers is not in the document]
doitian: It seems inefficient that we take the whole list out, and then insert back one by one. We can peek the unknown hash backward. If there's a hash found in the local index, we pop it and add it as a new received block. Otherwise the loop exists early.

```
while (!peer.unknown_list.empty()) {
  if index.exists(peer.unknown_list[-1]) {
    receive_new_block(peer.unknown_list.pop());
  } else {
    return;
  }
}
```

sync/src/types.rs
driftluo: types.rs#877 it need ownership

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -34,6 +34,16 @@ impl BlockFetcher {
...
+        let list = self
+            .synchronizer
+            .peers()
+            .get_unknown_list(self.peer)
+            .unwrap();
```
TheWaWaR: Use `.expect()` instead of `.unwrap()`, when panic we can see a specific message.

sync/src/synchronizer/mod.rs
```diff
@@ -344,13 +344,21 @@ impl Synchronizer {
...
+    fn find_blocks_to_fetch(&self, nc: &dyn CKBProtocolContext, is_ibd: bool) {
         let peers: Vec<PeerIndex> = {
             self.peers()
                 .state
                 .read()
                 .iter()
-                .filter(|(_, state)| state.sync_started)
+                .filter(|(_, state)| {
+                    if is_ibd {
+                        state.peer_flags.is_outbound
```
TheWaWaR: Why can not fetch blocks from inbound peers?

driftluo: > Why can not fetch blocks from inbound peers?

`32 outbound peer, sync from 0 to 494467，took 110 minutes，average speed 4495/min` must fix this first

doitian: I think the reason is that outbounds are chosen randomly, which are consider safer than inbounds.

sync/src/synchronizer/mod.rs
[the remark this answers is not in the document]
doitian: I think we can use the IBD condition for non-IBD state as well. We should download blocks from peer whoever has better chain work than us.

sync/src/synchronizer/mod.rs
```diff
@@ -150,8 +150,8 @@ impl Synchronizer {
         }
     }
 
-    pub fn get_blocks_to_fetch(&self, peer: PeerIndex) -> Option<Vec<packed::Byte32>> {
-        BlockFetcher::new(self.clone(), peer).fetch()
+    pub fn get_blocks_to_fetch(&self, peer: PeerIndex, ibd: bool) -> Option<Vec<packed::Byte32>> {
```
doitian: I really hate boolean parameter in function, it makes the place invoking the function hard to read. An enum or a new type is much better than a boolean.

Ex.

```
enum IBDState {
  IBD,
  NotIBD,
}

struct IBDState(bool);
impl IBDState {
  bool is_ibd(&self) -> { self.0 }
}
```

sync/src/synchronizer/mod.rs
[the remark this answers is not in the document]
doitian: If `is_outbound || is_whitelist || is_protect` does not cover `sync_started`, we can use `sync_started || is_outbound || is_whitelist || is_protect`

sync/src/types.rs
```diff
@@ -836,6 +867,32 @@ impl SyncState {
...
+            } else {
+                self.peers.insert_unknown_header_hash(pi, hash)
+            }
+        }
+    }
+
+    pub fn try_update_best_known_with_unknown_header_list(&self, pi: PeerIndex) {
+        // header list Is an ordered list, sorted from highest to lowest,
+        // when header hash unknown, break loop is ok
+        while let Some(hash) = self.peers().take_unknown_last(pi) {
```
doitian: :+1:

doitian:
Hey folks, please review @quake  @keroro520 @zhangsoledad

doitian:
Please rebase

PR #1959, opened 2020-03-09
perf: improve get_ancestor efficiency

keroro520:
This PR introduces 2 optimizations related to `get_ancestor`

#### 1. `BlockFetcher::fetch` save times of involving `get_ancestor` with the help of tracing parents

  https://github.com/nervosnetwork/ckb/pull/1959/files#diff-86d69964b227ca1248fadb84da967a96

* [Original]  `BlockFetcher::fetch` would involve `get_ancestor` multiple times for continuous headers.
For example, assume `block-chain-tip = 100, header-chain-tip=999`, `get_ancestor` would be involved at least 16(`MAX_BLOCKS_IN_TRANSIT_PER_PEER = 16`) times:
    - `get_ancestor(101)`
    - `get_ancestor(102)`
    - ...
    - `get_ancestor(115)`

* [Now] While we can save 15 times of involving `get_ancestor` with the help of tracing parents:

    - `get_ancestor(115)`
    - get `headers[100...114]` by tracing parents

#### 2. `get_ancestor` shortcuts scanning for main-chain headers.

  https://github.com/nervosnetwork/ckb/pull/1959/files#diff-9912cd1cc8b513bd52d8bfe461cadfd0

* [Original] `get_ancestor` locate the header of the target number via skip-scanning. While the scan may get block-headers from storage frequently, which may make `get_ancestor` become a heavy operation.

* [Now] This PR introduces a shortcut: If we already skip into the main-chain scope, then we can use `store.get_header_by_number` to locate the target header directly.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -2,7 +2,7 @@ use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
 use crate::types::{ActiveChain, HeaderView, IBDState};
 use crate::MAX_BLOCKS_IN_TRANSIT_PER_PEER;
-use ckb_logger::{debug, trace, metric};
+use ckb_logger::{debug, metric, trace};
 use ckb_network::PeerIndex;
 use ckb_types::{core, packed};
 use faketime::unix_time_as_millis;
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -2,10 +2,10 @@ use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
 use crate::types::{ActiveChain, HeaderView, IBDState};
 use crate::MAX_BLOCKS_IN_TRANSIT_PER_PEER;
-use ckb_logger::{debug, metric, trace};
+use ckb_logger::{debug, trace};
 use ckb_network::PeerIndex;
 use ckb_types::{core, packed};
-use faketime::unix_time_as_millis;
+use std::cmp::min;
 
 pub struct BlockFetcher {
     synchronizer: Synchronizer,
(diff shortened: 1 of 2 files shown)


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +123,42 @@ impl BlockFetcher {
...
-                    );
-                    fetch.push(to_fetch.hash());
+                // Try to get [from, to) headers, return the corresponding headers in reversed order.
+                let from = index_height;
+                let to = from + (count - fetch.len()) as u64;
+                let candidates = {
+                    let mut candidates = Vec::with_capacity((to - from) as usize);
+                    let mut parent_hash = self
+                        .snapshot
+                        .get_ancestor(&best_known_header.hash(), to)?
```
TheWaWaR: The block number `to` may overflow.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +123,42 @@ impl BlockFetcher {
...
+                    }
+
+                    if inflight.insert(self.peer, to_fetch.hash()) {
+                        trace!(
+                            "[Synchronizer] inflight insert {:?}------------{}",
+                            to_fetch.number(),
+                            to_fetch.hash(),
+                        );
+                        fetch.push(to_fetch.hash());
+                    }
```
TheWaWaR: Why two level of loop?

keroro520: For example, when we search block[100..116], the original code will involve `16` times of `get_ancestor`, `get_ancestor(100), get_ancestor(101), ... get_ancestor(105)`.

And now changed to: 
(1) `get_ancestor(16)`
(2) get header[100..115] via parent tracing.

That will save 15 times of `get_ancestor` involving.

sync/src/synchronizer/block_fetcher.rs
keroro520: why `to` may overflow?

`to = from + (count - fetch.len())`, `from <= tip_number`, `count - fetch.len() < 16`, so I think `to` is safe.

TheWaWaR: Here `to = from + 16`, assume `from = 0, fetch.len() = 0, to = 16`,  in Rust `0..16` means [0, 16), there are 16 items, if 16 included,  there are 17 items. There is possible `to > best_known_header.number()`.

sync/src/synchronizer/block_fetcher.rs
TheWaWaR: I think `get_ancestor` should be a cheap action, when it's not a hot code, better with simple code. I worry about the two level loop introduce new bug.

sync/src/synchronizer/block_fetcher.rs
keroro520: Thanks. Already fixed now.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +124,43 @@ impl BlockFetcher {
...
-                    .get_ancestor(&best_known_header.hash(), index_height)?;
-
-                // NOTE: Filtering `BLOCK_STORED` but not `BLOCK_RECEIVED`, is for avoiding
-                // stopping synchronization even when orphan_pool maintains dirty items by bugs.
-                if self
-                    .snapshot
-                    .contains_block_status(&to_fetch.hash(), BlockStatus::BLOCK_STORED)
-                {
-                    continue;
+                // Try to get [from, to] headers, return the corresponding headers in reversed order.
```
doitian: We could increase the stripe of each fetch as well, like bitcoin, which uses 128. The method `get_ancestor` may be as heavy as traversing 100 blocks, we should take the advantage of each `get_ancestor`.

However, the change requires more benchmarks, we could introduce it in a new PR.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +124,43 @@ impl BlockFetcher {
...
                 }
 
-                if inflight.insert(self.peer, to_fetch.hash()) {
-                    trace!(
-                        "[Synchronizer] inflight insert {:?}------------{}",
-                        to_fetch.number(),
-                        to_fetch.hash(),
-                    );
-                    fetch.push(to_fetch.hash());
+                index_height = to;
```
doitian: Please simplify the logic to update `index_height`.

1. On init, set it to `fixed_last_common_header.number()`.
2. Update it in the top of the loop:

    ```
    if index_height > best_known_header.number() {
      break;
    }

    let from = index_height;
    let to = ...;
    index_height = to + 1;
    ```

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +124,43 @@ impl BlockFetcher {
...
-                    .contains_block_status(&to_fetch.hash(), BlockStatus::BLOCK_STORED)
-                {
-                    continue;
+                // Try to get [from, to] headers, return the corresponding headers in reversed order.
+                let from = index_height;
+                let to = min(
+                    from + (count - fetch.len()) as u64 - 1,
+                    best_known_header.number(),
+                );
+                let mut candidates = Vec::with_capacity((to - from + 1) as usize);
```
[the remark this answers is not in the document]
doitian: We can introduce this in anther PR, after sufficient benchmark.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -123,26 +124,43 @@ impl BlockFetcher {
...
-
-                // NOTE: Filtering `BLOCK_STORED` but not `BLOCK_RECEIVED`, is for avoiding
-                // stopping synchronization even when orphan_pool maintains dirty items by bugs.
-                if self
-                    .snapshot
-                    .contains_block_status(&to_fetch.hash(), BlockStatus::BLOCK_STORED)
-                {
-                    continue;
+                // Try to get [from, to] headers, return the corresponding headers in reversed order.
+                let from = index_height;
```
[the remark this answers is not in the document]
keroro520: Yes, PR description updated.

sync/src/synchronizer/block_fetcher.rs
keroro520: I have rewrite the code. PLTA again.

doitian:
@keroro520 please resolve conflicts

sync/src/types.rs
```diff
@@ -945,15 +953,21 @@ impl SyncSnapshot {
...
-        }
-        self.get_header_view(base)?
-            .get_ancestor(number, |hash| self.get_header_view(hash))
+        let tip_number = self.tip_number();
+        self.get_header_view(base)?.get_ancestor(
+            number,
+            |hash| self.get_header_view(hash),
+            |current| {
+                // shortcut to return a ancestor block
+                if current.number() <= tip_number && self.store().is_main_chain(&current.hash()) {
```
doitian: This may have a problem. If the chain forks between `if` and `get_block_hash`, this will return wrong result. We should use snapshot here.

keroro520: Done

sync/src/types.rs
```diff
@@ -1112,17 +1120,21 @@ impl ActiveChain {
...
-        self.shared
-            .get_header_view(base)?
-            .get_ancestor(number, |hash| self.shared.get_header_view(hash))
+        let tip_number = self.tip_number();
+        self.shared.get_header_view(base)?.get_ancestor(
+            number,
+            |hash| self.shared.get_header_view(hash),
+            |current| {
+                // shortcut to return a ancestor block
+                if current.number() <= tip_number && self.snapshot().is_main_chain(&current.hash())
```
doitian: @zhangsoledad Should we hold the snapshot instance locally to make sure `get_block_hash` will use the same snapshot?

zhangsoledad: rebase will resolve this.

doitian: Checked the code. The Arc is cloned when creating the `active_chain`, and will not change in the `active_chain` instance.

PR #1960, opened 2020-03-10
test: add options for integration tests and fix a bug

yangby-cryptape:
1. Add an option to run all tests regardless of failure.
2. Fix a bug: when the nodes are panicked, the exit code is 0.

   For example: add `if block.number() > 0 { panic!(); }` to function `insert_block(..)` in `ckb-store`, all integration tests will be passed (`exit code == 0`).

   Since the follow two pieces of code are not match (`panicked_error` is skipped when exit):

   https://github.com/nervosnetwork/ckb/blob/e76cbe076586d362ce7f676635caf1e84f1cc9d4/test/src/main.rs#L150-L152

   https://github.com/nervosnetwork/ckb/blob/e76cbe076586d362ce7f676635caf1e84f1cc9d4/test/src/main.rs#L165-L167

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -16,7 +16,6 @@ use std::collections::HashMap;
 use std::env;
 use std::fs::{read_to_string, File};
 use std::io::{BufRead, BufReader};
-use std::panic;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


keroro520:
bors r= quake , driftluo , keroro520

PR #1961, opened 2020-03-10
fix: fix bug on last common marked

driftluo:
Fix the stuttering problem of multi-node download on ibd concurrent download and no-ibd concurrent download.

If the `last common` tag is incorrect, it will cause the `index_height` and `tip` to be too far apart, which will cause `get_ancestor` and `contains_block_status` to be penetrated to RocksDB. When the gap exceeds 500k, it means that each time you look for the `can fetch` block, it will take up to Millions of cache breakdowns, this problem will increase rapidly with the number of synchronizable nodes and chain height, which will lead to pauses.

https://github.com/nervosnetwork/ckb/blob/e76cbe076586d362ce7f676635caf1e84f1cc9d4/sync/src/synchronizer/block_fetcher.rs#L40-L52

https://github.com/nervosnetwork/ckb/blob/e76cbe076586d362ce7f676635caf1e84f1cc9d4/sync/src/synchronizer/block_fetcher.rs#L121-L137

Currently, our strategy for downloading blocks is to request a block from two nodes at the same time. However, we only recorded the last common update information of one node. This is incorrect. At the same time, the same problem occurs in the compact block

The problem here will cause the local sync protocol to fall into the pseudo-death stage, unable to send or receive any messages

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -86,6 +86,19 @@ impl<'a> CompactBlockProcess<'a> {
 
         let status = active_chain.get_block_status(&block_hash);
         if status.contains(BlockStatus::BLOCK_STORED) {
+            // update last common header and best known
+            let parent = shared
+                .get_header_view(&header.data().raw().parent_hash())
+                .expect("parent block must exist");
+            let header_view = {
+                let total_difficulty = parent.total_difficulty() + header.difficulty();
+                crate::types::HeaderView::new(header.clone(), total_difficulty)
+            };
+
+            let state = shared.state().peers();
+            state.new_header_received(self.peer, &header_view);
+            state.set_last_common_header(self.peer, header);
+
             return StatusCode::CompactBlockAlreadyStored.with_context(block_hash);
         } else if status.contains(BlockStatus::BLOCK_INVALID) {
             return StatusCode::BlockIsInvalid.with_context(block_hash);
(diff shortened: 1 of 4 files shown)


sync/src/relayer/compact_block_process.rs
```diff
@@ -85,6 +85,19 @@ impl<'a> CompactBlockProcess<'a> {
 
         let status = snapshot.get_block_status(&block_hash);
         if status.contains(BlockStatus::BLOCK_STORED) {
+            // update last common header and best known
+            let parent = snapshot
+                .get_header_view(&header.data().raw().parent_hash())
+                .unwrap();
```
keroro520: Better use `expect`. I am afraid that [snapshot problem](https://github.com/nervosnetwork/ckb/pull/1947) may cause here panic.

driftluo: No, when the block exists locally, its parent block must exist. If it does not exist, it is a very serious bug.

doitian: Anyway, `expect` is better than `unwrap`.

doitian: Does `BLOCK_STORED` guarantee that all the ancestors are also stored?

driftluo: > Does `BLOCK_STORED` guarantee that all the ancestors are also stored?

yes

driftluo:
> Should check whether the new header is higher than last common header in `fn set_last_common_header`?

How to check it?
the last common does not accurately describe the status of the other peer, such as fork, orphan block

Our chain may short-fork. At this time, common is not strictly incremented, similar to the difficulty.

At the same time, the update of the compact block only receives the header of the current epoch.

I think that if we can receive the uncle block of the same epoch, we can tolerate the last common marking mistake in the same epoch

doitian:
@TheWaWaR please review again

doitian:
@driftluo please resolve conflicts

quake:
bors r=quake,doitian

PR #1965, opened 2020-03-12
chore: send failed integration test to sentry

doitian:
Collecting the failed specs so we can known which specs fail most and should be disabled until fixed.

#2bm04k[closed]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,8 @@ tags
 
 # tmp files
 *.pyc
+/test/integration.log
+/test/integration.failure
 
 # generated files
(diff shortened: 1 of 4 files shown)


keroro520:
Oh, what about log `GITHUB_RUN_ID` as well? It can help us to locate the job logs. @doitian

doitian:
> Oh, what about log `GITHUB_RUN_ID` as well? It can help us to locate the job logs. @doitian

It will capture all environment variables

doitian:
bors r=quake,keroro520,yangby-cryptape

PR #1966, opened 2020-03-12
feat: enhanced locator

driftluo:
After allowing ibd multi-node download, the premise of enabling multi-node download is to update the best-known header of the corresponding node as soon as possible.

However, as the chain height continues to increase, the exponentially descending abstraction header of the locator will cause the minimum height header in the locator to become higher and higher, that is, it will take longer and longer to enable ibd multi-node download.

So, slightly improve the step insertion method of the locator so that the minimum height within it is as low as possible


yangby-cryptape:
bors r=quake,yangby-cryptape,keroro520

PR #1968, opened 2020-03-17
refactor: Simplify network protocols

TheWaWaR:
- Move p2p protocols to network module
- Change some discovery log message's log level from debug to trace
- Remove DiscoveryService struct


yangby-cryptape:
Reuse a big crate is not easy.
If anyone want to use these protocols, they have to import the whole `network` crate (why they have to use `sentry`) or write other implements?

If we move all protocols away from `network`, people can use there protocols in their own way.
We communicate through protocols, not implements.

Since I have already write another implement of these protocols.
I don't oppose this change, just express my opinion.

driftluo:
please rebase @TheWaWaR

network/src/protocols/discovery/addr.rs
```diff
@@ -33,23 +29,24 @@ pub enum Misbehavior {
 
 /// Misbehavior report result
 pub enum MisbehaveResult {
-    /// Continue to run
-    Continue,
+    // /// Continue to run
+    // Continue,
```
driftluo, line 33: delete it if it doesn't use

network/src/protocols/ping.rs
```diff
@@ -1,16 +1,322 @@
...
+            .nth2(nonce_le[2].into())
+            .nth3(nonce_le[3].into())
+            .build();
+        let pong = packed::Pong::new_builder().nonce(nonce).build();
+        let payload = packed::PingPayload::new_builder().set(pong).build();
+        // TODO update-after-upgrade-p2p
+        packed::PingMessage::new_builder()
+            .payload(payload)
+            .build()
+            .as_bytes()
```
driftluo, line 298: Wrong here, already upgraded, no conversion required.

TheWaWaR: I’ll just redo those changed.

TheWaWaR: I’ll just redo those changed.

PR #1975, opened 2020-03-23
chore: reset testnet aggron genesis

doitian:


doitian:
Integration ran on staging2

PR #1977, opened 2020-03-24
fix: Fix false positive in IllTransactionChecker

xxuejie:


doitian:
bors r=doitian,TheWaWaR

doitian:
bors r=doitian,TheWaWaR

PR #1979, opened 2020-03-24
perf: Fix build_skip performance bug

TheWaWaR:
When build skip list want a header not exists in header map, we should get it from main chain directly, otherwise it will search it by parent hash linearly.

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -512,13 +512,18 @@ impl HeaderView {
         self.inner
     }
 
-    pub fn build_skip<F>(&mut self, mut get_header_view: F)
+    pub fn build_skip<F, G>(&mut self, mut get_header_view: F, fast_scanner: G)
     where
         F: FnMut(&Byte32) -> Option<HeaderView>,
+        G: Fn(BlockNumber, &HeaderView) -> Option<HeaderView>,
     {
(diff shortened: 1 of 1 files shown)


sync/src/types.rs
```diff
@@ -762,7 +767,22 @@ impl SyncShared {
             HeaderView::new(header.clone(), total_difficulty)
         };
 
-        header_view.build_skip(|hash| self.get_header_view(hash));
+        let snapshot = Arc::clone(&self.shared.snapshot());
+        header_view.build_skip(
+            |hash| self.get_header_view(hash),
+            |number, current| {
+                // shortcut to return an ancestor block
+                if current.number() <= snapshot.tip_number()
```
driftluo: I think this condition is wrong, it will cause the skip table construction to almost read RocksDB.

I think the limit that must be added is `(current.number () >>= 1)> number`

sync/src/types.rs
```diff
@@ -762,7 +767,22 @@ impl SyncShared {
...
+                    && snapshot.is_main_chain(&current.hash())
+                {
+                    snapshot
+                        .get_block_hash(number)
```
driftluo: Here should be `get_block_hash(number-1)`, the condition for building a skip list is `while current.number ()> number`

doitian:

PR #1985, opened 2020-03-25
perf: Speed up fetch collect

driftluo:
Speed up fetch collect

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -131,16 +131,6 @@ impl BlockFetcher {
         let mut fetch = Vec::with_capacity(count);
         let mut start = fixed_last_common_header.number() + 1;
 
-        // Judge whether we should fetch the target block, neither stored nor in-flighted
-        let mut should_fetch = |block_hash| {
-            // NOTE: Filtering `BLOCK_STORED` but not `BLOCK_RECEIVED`, is for avoiding
-            // stopping synchronization even when orphan_pool maintains dirty items by bugs.
-            let stored = self
-                .active_chain
-                .contains_block_status(&block_hash, BlockStatus::BLOCK_STORED);
-            !stored && inflight.insert(self.peer, block_hash)
-        };
-
         while fetch.len() < count && start <= best_known_header.number() {
             let span = min(
                 best_known_header.number() - start + 1,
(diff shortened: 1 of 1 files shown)


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -131,14 +131,26 @@ impl BlockFetcher {
...
                 .active_chain
                 .contains_block_status(&block_hash, BlockStatus::BLOCK_STORED);
-            !stored && inflight.insert(self.peer, block_hash)
+            if stored {
+                ShouldFetch::Stored
+            } else if !inflight.insert(self.peer, block_hash) {
+                ShouldFetch::Inflightlimit
+            } else {
+                ShouldFetch::Fetch
+            }
```
keroro520: It can be simplified to

```
should_fetch -> (bool, bool) {
  return (stored, inflight)
}

match should_fetch() {
  (false, true) =>  // to fetch
  (false, false) => continue
  (true, _) => break;
}
```

driftluo: Too much bool causes difficulty in understanding, suggesting naming

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -151,9 +141,22 @@ impl BlockFetcher {
...
+            // Judge whether we should fetch the target block, neither stored nor in-flighted
             for _ in 0..span {
                 let parent_hash = header.parent_hash();
-                if should_fetch(header.hash()) {
+                // NOTE: Filtering `BLOCK_STORED` but not `BLOCK_RECEIVED`, is for avoiding
+                // stopping synchronization even when orphan_pool maintains dirty items by bugs.
+                let stored = self
+                    .active_chain
+                    .contains_block_status(&parent_hash, BlockStatus::BLOCK_STORED);
+                if stored {
```
doitian:
```
if stored {
  break;
} else if inflight.insert(...) {
  fetch.push(header);  
}
```

or 

```
if stored {
  break;
}
if inflight.insert(...) {
  fetch.push(header);  
}
```

doitian: I personally prefer giving the `insert` result a variable name. I never know whether the `true` returns from `insert` means the key already exists or the key has successfully inserted.  

🤣

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -151,9 +141,20 @@ impl BlockFetcher {
...
+                    // If the block is stored, its ancestor must on store
+                    // So we can skip the search of this space directly
+                    break;
```
doitian: Should we break outer while as well?

sync/src/synchronizer/block_fetcher.rs
driftluo: > I personally prefer giving the `insert` result a variable name. I never know whether the `true` returns from `insert` means the key already exists or the key has successfully inserted.
> 

may refactoring on download scheduler

sync/src/synchronizer/block_fetcher.rs
driftluo: > Should we break outer while as well?

No, this logic is segmented, the parent here is just the current segment can be ignored. 

something like:

```
|  first | second | ....
|  <--   |   <--  |....
```

doitian:
bors r=doitian,keroro520

PR #1988, opened 2020-03-26
refactor: use a new method to detect headers sync timeout

yangby-cryptape:
### Purpose

To avoid possible performance issues on headers synchronization.

---

#### The Timeout is Ambiguous

- [The first commit introduced the `headers_sync_timeout` treat `headers_sync_timeout` as a **total** timeout for all headers.](https://github.com/nervosnetwork/ckb/commit/b7db483cb519c37d15dfbd868d4621091c61d6a2#diff-af95c26bdbe8ee3013958651bd249995R143-R148)

- [But, the commit add a upper bound to `estimated_headers_count`, try to treat the timeout as a **partial** timeout, but didn't finish it.](https://github.com/nervosnetwork/ckb/commit/f14c8dc0fb5bd77e5935e818d2340f38c3802ab6#diff-67a34c19141a7c62f189385b4c2bacf7R181-R185)

#### The Timeout Could be Lacked

The timeout has upper bound, but the headers need to sync doesn't have a upper bound, it could be a fatal error in someday.

- Current timeout is `base_timeout + estimated_headers_count * timeout_per_header`: https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/lib.rs#L50-L53 https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/mod.rs#L128

- And, the `estimated_headers_count` has a upper bound: https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/mod.rs#L124-L127 https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/lib.rs#L24

So, if the timeout wasn't update after set, **the max timeout is 362 seconds for syncing all headers**.

  - There are only 3 conditions that will update the `headers_sync_timeout` in whole project:

    - [Receive an empty SendHeaders](https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/headers_process.rs#L147-L158
)
    - [The difference of current time and tip header timestamp smaller than](https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/mod.rs#L217-L225) [`MAX_TIP_AGE`](https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/types.rs#L966-L968) 

    - [Suspend syncing](https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/types.rs#L130-L135)

#### Timeout could be Disable by an Empty `SendHeaders`

https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/headers_process.rs#L147-L158

#### The Best Known Tip Header Wouldn't Update Again in IBD

At first time when do headers sync, we will know the current best header, but it will not update in IBD again. (I did a test for this: print the best known tip header, it was a same value during the IBD.)

So if IBD is too slow, the best known tip header will be drop behind real tip header too much.

---

### My Solution

- Not use `base_timeout`.
- Check both of **the global average speed** and **the instantaneous speed** for headers synchronization.
- Only check speeds after a fixed interval, to avoid a too small time range for checking speeds.
- Not reset headers sync timeout during IBD.
- Send `GetHeaders` if the best known tip header was drop behind the estimated real tip header too much.

---

#### Why use Both Global Average Speed and Instantaneous Speed

- A fixed timeout isn't enough for all conditions.
  - If `estimated_headers_count` is too small , we shouldn't care about the timeout / speed;
  - If `estimated_headers_count` is big enough, we should check the timeout / speed again;
  - If only check the timeout for a `SendHeaders` message, not check how many headers in a `SendHeaders` message, the sync node can send only 1 header in each SendHeaders, then need `timeout * tip-number` milliseconds to sync to the tip.
- If we only check the global average speed, if the headers sync is fast at beginning, but it became slow or it was stopped after enough time, the global average speed will still keep elevated in a long period.
- If we only check the instantaneous speed, the effect of a unstable network will be increased.
  If a node send headers very fast in a long period, we should tolerate more bias for the instantaneous speed in a short period. 

#### What does `expected_headers_per_sec * inspect_window * POW_INTERVAL / 1000` Mean

The old code estimates a height number for now: https://github.com/nervosnetwork/ckb/blob/26e2599b90063fdde209039b2c6e8c69b9b96bb5/sync/src/synchronizer/mod.rs#L126

The current code use `time-in-blockchain` instead of it.

So, `let y = expected_headers_per_sec * x * POW_INTERVAL / 1000` means:
after `x` milliseconds in reality, the timestamp in tip header should be increase `y` milliseconds.

#### Preset Parameters

- Lowest Global Average Speed: `300 KiB/second`
  Sync 1 year headers in blockchain will be in 32.85 minutes in reality
- Lowest Instantaneous Speed: `75.0 KiB/second`

### DIFF
--- a/sync/src/lib.rs
+++ b/sync/src/lib.rs
@@ -46,10 +46,18 @@ pub(crate) const LOW_INDEX: usize = TIME_TRACE_SIZE * 9 / 10;
 
 pub(crate) const LOG_TARGET_RELAY: &str = "ckb-relay";
 
-//  Timeout = base + per_header * (expected number of headers)
-pub const HEADERS_DOWNLOAD_TIMEOUT_BASE: u64 = 6 * 60 * 1000; // 6 minutes
-pub const HEADERS_DOWNLOAD_TIMEOUT_PER_HEADER: u64 = 1; // 1ms/header
-pub const POW_SPACE: u64 = 10_000; // 10s
+// Inspect the headers downloading every 2 minutes
+pub const HEADERS_DOWNLOAD_INSPECT_WINDOW: u64 = 2 * 60 * 1000;
+// Global Average Speed
+//      Expect 300 KiB/second
+//          = 1600 headers/second (300*1024/192)
+//          = 96000 headers/minute (1600*60)
+//          = 11.11 days-in-blockchain/minute-in-reality (96000*10/60/60/24)
+//      => Sync 1 year headers in blockchain will be in 32.85 minutes (365/11.11) in reality
+pub const HEADERS_DOWNLOAD_HEADERS_PER_SECOND: u64 = 1600;
+// Acceptable Lowest Instantaneous Speed: 75.0 KiB/second (300/4)
+pub const HEADERS_DOWNLOAD_TOLERABLE_BIAS_FOR_SINGLE_SAMPLE: u64 = 4;
+pub const POW_INTERVAL: u64 = 10;
 
 // Protect at least this many outbound peers from disconnection due to slow
 // behind headers chain.
--- a/sync/src/synchronizer/headers_process.rs
+++ b/sync/src/synchronizer/headers_process.rs
@@ -127,6 +127,37 @@ impl<'a> HeadersProcess<'a> {
         acceptor.accept()
     }
 
+    fn debug(&self) {
+        if log_enabled!(Level::Debug) {
+            // Regain the updated best known
+            let shared_best_known = self.synchronizer.shared.state().shared_best_header();
+            let peer_best_known = self.synchronizer.peers().get_best_known_header(self.peer);
+            debug!(
+                "chain: num={}, diff={:#x};",
+                self.active_chain.tip_number(),
+                self.active_chain.total_difficulty()
+            );
+            debug!(
+                "shared best_known_header: num={}, diff={:#x}, hash={};",
+                shared_best_known.number(),
+                shared_best_known.total_difficulty(),
+                shared_best_known.hash(),
+            );
+            if let Some(header) = peer_best_known {
+                debug!(
+                    "peer's best_known_header: peer: {}, num={}; diff={:#x}, hash={};",
+                    self.peer,
+                    header.number(),
+                    header.total_difficulty(),
+                    header.hash()
+                );
+            } else {
+                debug!("state: null;");
+            }
+            debug!("peer: {}", self.peer);
+        }
+    }
+
     pub fn execute(self) -> Status {
         debug!("HeadersProcess begin");
         let shared = self.synchronizer.shared();
@@ -198,34 +229,7 @@ impl<'a> HeadersProcess<'a> {
             }
         }
 
-        if log_enabled!(Level::Debug) {
-            // Regain the updated best known
-            let shared_best_known = self.synchronizer.shared.state().shared_best_header();
-            let peer_best_known = self.synchronizer.peers().get_best_known_header(self.peer);
-            debug!(
-                "chain: num={}, diff={:#x};",
-                self.active_chain.tip_number(),
-                self.active_chain.total_difficulty()
-            );
-            debug!(
-                "shared best_known_header: num={}, diff={:#x}, hash={};",
-                shared_best_known.number(),
-                shared_best_known.total_difficulty(),
-                shared_best_known.hash(),
-            );
-            if let Some(header) = peer_best_known {
-                debug!(
-                    "peer's best_known_header: peer: {}, num={}; diff={:#x}, hash={};",
-                    self.peer,
-                    header.number(),
-                    header.total_difficulty(),
-                    header.hash()
-                );
-            } else {
-                debug!("state: null;");
-            }
-            debug!("peer: {}", self.peer);
-        }
+        self.debug();
 
         if headers.len() == MAX_HEADERS_LEN {
             let start = headers.last().expect("empty checked");
--- a/sync/src/synchronizer/headers_process.rs
+++ b/sync/src/synchronizer/headers_process.rs
@@ -145,15 +145,15 @@ impl<'a> HeadersProcess<'a> {
         }
 
         if headers.is_empty() {
-            // Reset headers sync timeout
-            self.synchronizer
-                .peers()
-                .state
-                .write()
-                .get_mut(&self.peer)
-                .expect("Peer must exists")
-                .headers_sync_timeout = None;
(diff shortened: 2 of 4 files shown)


doitian:
Sync timeout is for the whole chain, using max headers in a single SendHeaders request as upper bound is definitely a bug.

On the other side, it means the sync issue is not caused by the too large header sync timeout, because it is at most 6'2''.

sync/src/lib.rs
```diff
@@ -47,9 +47,11 @@ impl Into<ProtocolId> for NetworkProtocol {
     }
 }
 
-//  Timeout = base + per_header * (expected number of headers)
-pub const HEADERS_DOWNLOAD_TIMEOUT_BASE: u64 = 6 * 60 * 1000; // 6 minutes
-pub const HEADERS_DOWNLOAD_TIMEOUT_PER_HEADER: u64 = 1; // 1ms/header
+pub const HEADERS_DOWNLOAD_TIMEOUT_BASE: u64 = 2 * 60 * 1000; // 2 minutes
```
doitian, line 50: The name is borrowed from Bitcoin, however, I think it is misleading. It is not really a timeout to download headers, it is a timeout to sync the tip.

sync/src/synchronizer/mod.rs
```diff
@@ -207,21 +213,20 @@ impl Synchronizer {
...
+            if let Some(ref mut process) = state.headers_sync_process {
+                if let Some(is_timeout) = process.is_timeout(&better_tip, now) {
+                    if is_timeout && !state.disconnect {
                         eviction.push(*peer);
                         state.disconnect = true;
                         continue;
                     }
                 } else {
-                    state.headers_sync_timeout = None
+                    active_chain.send_getheaders_to_peer(nc, *peer, &better_tip);
```
doitian, line 229: This is duplicated with the feature `CHAIN_SYNC_TIMEOUT`. Maybe we can merge these two features into one. Or reimplemented these two features, one focuses on the headers sync and another handles the active chain tip.

sync/src/synchronizer/mod.rs
```diff
@@ -207,21 +213,20 @@ impl Synchronizer {
...
+        let better_tip = self.better_tip();
         for (peer, state) in peer_states.iter_mut() {
             let now = unix_time_as_millis();
 
-            // headers_sync_timeout
-            if let Some(timeout) = state.headers_sync_timeout {
-                if is_initial_header_sync {
-                    if now > timeout && !state.disconnect {
+            if let Some(ref mut process) = state.headers_sync_process {
+                if let Some(is_timeout) = process.is_timeout(&better_tip, now) {
```
doitian, line 222: As I mentioned in the comment on `HEADERS_DOWNLOAD_TIMEOUT_BASE`, this is not a timeout on downloaded headers, but indeed the timeout on the tip. So the timeout should also use active chain tip to check.

yangby-cryptape:
I will update this PR after few minor refactoring.

sync/src/lib.rs
yangby-cryptape: I will rename this const variable.

sync/src/synchronizer/headers_process.rs
```diff
@@ -145,15 +145,18 @@ impl<'a> HeadersProcess<'a> {
...
-                .peers()
-                .state
-                .write()
-                .get_mut(&self.peer)
-                .expect("Peer must exists")
-                .headers_sync_timeout = None;
-            debug!("HeadersProcess is_empty (synchronized)");
+            debug!("HeadersProcess is_empty");
+            let ibd = self.active_chain.is_initial_block_download();
+            if !ibd {
```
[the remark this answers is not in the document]
yangby-cryptape: Since I had implement a mechanism base on the sync speed, I thought we can allow keep the connection here.
We don't reject any nodes for one single fault, we measure their credit base on a period of time.

I had confirmed this part with @TheWaWaR last week. (Maybe I made a mistake on understanding his explanation,) I thought may be this message just means that I had reach the tip.

sync/src/synchronizer/mod.rs
yangby-cryptape: I should remove the timeout of header sync when IBD was done. Then both features are independent.

I will update this.

sync/src/synchronizer/mod.rs
yangby-cryptape: I think there still need two timeouts or two mechanisms:
1. In IBD, I known some blocks were already existed, so I hoped I could download headers of them more quickly（far greater than `1 minute-in-blockchain/minute-in-reality`).
  So, I can reach the tip in a reasonable time.
2. After IBD, I was on the tip, so I only expected I could download headers not too slow. So slightly slower than `1 minute-in-blockchain/minute-in-reality` is acceptable.

sync/src/synchronizer/mod.rs
```diff
@@ -342,8 +333,7 @@ impl Synchronizer {
                 let mut state = self.peers().state.write();
                 if let Some(peer_state) = state.get_mut(&peer) {
                     if !peer_state.sync_started {
-                        let headers_sync_timeout = self.predict_headers_sync_time(&tip);
-                        peer_state.start_sync(headers_sync_timeout);
+                        peer_state.start_sync(IBDHeadersSyncProcess::from_header(&tip));
```
driftluo, line 336: It may not be IBD here, but  IBD's judgment has been added

sync/src/synchronizer/headers_process.rs
```diff
@@ -127,6 +127,7 @@ impl<'a> HeadersProcess<'a> {
         acceptor.accept()
     }
 
+    #[allow(clippy::cognitive_complexity)]
```
doitian, line 130: Refactor the method?

driftluo:
@doitian

driftluo:
bors r=quake,driftluo

PR #1989, opened 2020-03-29
[ᚬrc/v0.30] perf: Fix build_skip performance bug

TheWaWaR:
Fix https://github.com/nervosnetwork/ckb/issues/1986


quake:
bors r=quake,doitian

PR #1990, opened 2020-03-30
chore: export OutputsValidator

yangby-cryptape:
### The Issue
- `OutputsValidator` is in `ckb_rpc::module::pool`.
- `OutputsValidator` is `public` but `ckb_rpc::module::pool` is private.
- `OutputsValidator` is not re-exported.
- So, no one can use `OutputsValidator`.

### Changes
- Move `OutputsValidator` to `ckb-jsonrpc-types`
- Export `OutputsValidator`

### DIFF
--- a/rpc/src/module/pool.rs
+++ b/rpc/src/module/pool.rs
@@ -1,6 +1,6 @@
 use crate::error::RPCError;
 use ckb_chain_spec::consensus::Consensus;
-use ckb_jsonrpc_types::{Transaction, TxPoolInfo};
+use ckb_jsonrpc_types::{OutputsValidator, Transaction, TxPoolInfo};
 use ckb_logger::error;
 use ckb_network::PeerIndex;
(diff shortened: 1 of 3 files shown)


util/jsonrpc-types/src/blockchain.rs
```diff
@@ -391,6 +391,13 @@ impl TxStatus {
     }
 }
 
+#[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum OutputsValidator {
```
quake, line 396: Please move this enum to `util/jsonrpc-types/src/pool.rs`

yangby-cryptape: Oh, I'm sorry! I forgot this. I will done it in 10 minutes!

PR #1992, opened 2020-03-30
feat: add an option to control how many blocks the miner has to mine

yangby-cryptape:
### Usage

```bash
ckb miner -C . --limit 10 # Exit after 10 nonces found
ckb miner -C . -l 5       # Exit after 5 nonces found
ckb miner -C .            # Run forever
ckb miner -C . --limit 0  # Run forever, too
```

Ref:
- [CKB-Cli Issue-256: Mining with CKB-Cli (for dev chain)](https://github.com/nervosnetwork/ckb-cli/issues/259)

### DIFF
--- a/ckb-bin/src/lib.rs
+++ b/ckb-bin/src/lib.rs
@@ -28,7 +28,7 @@ pub fn run_app(version: Version) -> Result<(), ExitCode> {
 
     match app_matches.subcommand() {
         (cli::CMD_RUN, Some(matches)) => subcommand::run(setup.run(&matches)?, version),
-        (cli::CMD_MINER, _) => subcommand::miner(setup.miner()?),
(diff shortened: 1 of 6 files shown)


doitian:
Although it does not matter in this PR, I just want to mention that Option is more explicit than limit 0

yangby-cryptape:
> Although it does not matter in this PR, I just want to mention that Option is more explicit than limit 0

Yes, I used `Option<u128>` at the begining, too.
Since the direct input which meant forever / maximum from users is `0`, I changed it to keep it same in whole program just before I pushed request.

yangby-cryptape:
bors r=quake,doitian

yangby-cryptape:
bors retry

PR #1996, opened 2020-04-02
fix: wait for RPC server to cleanup on shutdown

zhangsoledad:
RPC servers currently don't wait for threads to cleanup, leading rpc-test to segfaults.
Should be reverted once upstream fix is released.


doitian:

PR #2015, opened 2020-04-15
chore: use self-hosted sentry for testnet

doitian:


doitian:
bors r=quake,zhangsoledad

PR #2018, opened 2020-04-16
feat: prof command support specify execution path

zhangsoledad:
## Proposed Changes
Previously, the prof command can not configure the execution path. it hard coding in system temporary folder. 

But the system temporary folder usually has some limit, such as Linux putting a limit on `/tmp` size if it is mounted as `tmpfs` is to prevent a DOS attack. although the settings can be changed, I'm afraid it's not worth to do so. 

So, this PR proposal appends a new arg to prof command to support specify execution path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,6 +378,7 @@ dependencies = [
  "ctrlc",
  "serde",
  "serde_plain",
+ "tempfile",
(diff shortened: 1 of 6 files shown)


doitian:
bors r=quake,doitian

PR #2028, opened 2020-04-21
feat: make minor change to pow algorithm for testnet

yangby-cryptape:
[CKB Testnet Aggron PoW Variant Proposal](https://talk.nervos.org/t/4551):

> PoW testnets are prone to the denial-of-service attack. Although there’s an proposal to use PoA 3 in Aggron, it requires significant changes to the codebase and has been hold.


quake:
`PowEngine` is designed to describe the algorithm instead of describing where it is used, I suggest to add a new enum type:
```
pub enum Pow {
    Dummy,
    Eaglesong,
    EaglesongBlake2b
}
```

reusing `EaglesongSimpleConfig` looks OK to me since nobody will run this built-in miner for mainnet, however, I suggest to change the config:

```
pub struct EaglesongSimpleConfig {
    pub threads: usize,
    pub extra_hash_function: Option<ExtraHashFunction>,
}

pub enum ExtraHashFunction {
    Blake2b,
}
```

resource/ckb-miner.toml
```diff
@@ -58,3 +58,6 @@ worker_type = "EaglesongSimple" # {{
 threads     = 1 # {{
 # dev => delay_type = "Constant"\nvalue = 5000
 # }}
+# # {{
+# extra_hash_function => "Blake2b"
```
quake, line 62:
```suggestion
# testnet => extra_hash_function = "Blake2b"
```

doitian:
bors r=quake,doitian

PR #2042, opened 2020-04-24
feat: return filename of jemalloc_profiling_dump

keroro520:
RPC `jemalloc_profiling_dump` will generate the target heap file with a timestamp to the working directory. Returning the target filename will be convenient for the command-sender to know which heap file is the current one.

Examples of the proposed responses of RPC `jemalloc_profiling_dump`:

* Success : `{"jsonrpc":"2.0","result":"ckb-jeprof.1587964630.heap","id":2}`
* ~~Fail: `{"jsonrpc":"2.0","error":{"code":-32603,"message":"jemalloc profiling dump: unsupported"},"id":2}`~~ // It never return errors at the current implementation.

### DIFF
--- a/rpc/src/module/debug.rs
+++ b/rpc/src/module/debug.rs
@@ -1,5 +1,5 @@
 use ckb_logger::configure_logger_filter;
-use jsonrpc_core::{Error, ErrorCode::InternalError, Result};
+use jsonrpc_core::Result;
 use jsonrpc_derive::rpc;
 use std::time;
 
@@ -20,14 +20,8 @@ impl DebugRpc for DebugRpcImpl {
             .unwrap()
             .as_secs();
         let filename = format!("ckb-jeprof.{}.heap", timestamp);
-        match ckb_memory_tracker::jemalloc_profiling_dump(&filename) {
-            Ok(()) => Ok(filename),
-            Err(err) => Err(Error {
-                code: InternalError,
-                message: err,
-                data: None,
-            }),
-        }
+        ckb_memory_tracker::jemalloc_profiling_dump(filename.clone());
+        Ok(filename)
     }
 
     fn set_logger_filter(&self, filter: String) -> Result<()> {
--- a/rpc/src/module/debug.rs
+++ b/rpc/src/module/debug.rs
@@ -1,12 +1,12 @@
 use ckb_logger::configure_logger_filter;
-use jsonrpc_core::Result;
+use jsonrpc_core::{Error, ErrorCode::InternalError, Result};
 use jsonrpc_derive::rpc;
 use std::time;
 
 #[rpc(server)]
 pub trait DebugRpc {
     #[rpc(name = "jemalloc_profiling_dump")]
-    fn jemalloc_profiling_dump(&self) -> Result<()>;
+    fn jemalloc_profiling_dump(&self) -> Result<String>;
     #[rpc(name = "set_logger_filter")]
     fn set_logger_filter(&self, filter: String) -> Result<()>;
 }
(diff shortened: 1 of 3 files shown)


rpc/src/module/debug.rs
```diff
@@ -5,19 +5,19 @@ use std::time;
...
-    fn jemalloc_profiling_dump(&self) -> Result<()> {
+    fn jemalloc_profiling_dump(&self) -> Result<String> {
         let timestamp = time::SystemTime::now()
             .duration_since(time::SystemTime::UNIX_EPOCH)
             .unwrap()
             .as_secs();
         let filename = format!("ckb-jeprof.{}.heap\0", timestamp);
-        ckb_memory_tracker::jemalloc_profiling_dump(filename);
-        Ok(())
+        ckb_memory_tracker::jemalloc_profiling_dump(filename.clone());
```
yangby-cryptape, line 20: **(I haven't tested.)** I think this `clone()` is unnecessary, pass a `&mut str` should be enough.

https://github.com/nervosnetwork/ckb/blob/3f27eae7998ee24b282da54eb8d9cbe5b5e2d4cc/util/memory-tracker/src/jemalloc.rs#L7-L13

https://github.com/gnzlbg/jemallocator/blob/c27a859e98e3cb790dc269773d9da71a1e918458/jemalloc-sys/src/lib.rs#L470-L476
(I don't know why this link can't preview, because it has been transferred?)

rpc/src/module/debug.rs
```diff
@@ -5,19 +5,19 @@ use std::time;
...
 pub(crate) struct DebugRpcImpl {}
 
 impl DebugRpc for DebugRpcImpl {
-    fn jemalloc_profiling_dump(&self) -> Result<()> {
+    fn jemalloc_profiling_dump(&self) -> Result<String> {
         let timestamp = time::SystemTime::now()
             .duration_since(time::SystemTime::UNIX_EPOCH)
             .unwrap()
             .as_secs();
         let filename = format!("ckb-jeprof.{}.heap\0", timestamp);
```
yangby-cryptape, line 19: **(I haven't tested.)** Maybe we have to remove this `\0` before return?

rpc/src/module/debug.rs
```diff
@@ -5,19 +5,19 @@ use std::time;
 #[rpc(server)]
 pub trait DebugRpc {
     #[rpc(name = "jemalloc_profiling_dump")]
-    fn jemalloc_profiling_dump(&self) -> Result<()>;
+    fn jemalloc_profiling_dump(&self) -> Result<String>;
```
yangby-cryptape, line 8: When `target_env = "msvc"`, `target_os="macos"` or feature `profiling` wasn't enable, it doesn't make sense to return a `String`.
https://github.com/nervosnetwork/ckb/blob/3f27eae7998ee24b282da54eb8d9cbe5b5e2d4cc/util/memory-tracker/src/lib.rs#L6-L10
https://github.com/nervosnetwork/ckb/blob/3f27eae7998ee24b282da54eb8d9cbe5b5e2d4cc/util/memory-tracker/src/jemalloc-mock.rs#L3-L5

keroro520: You are right, it is useless for these cases. I think it will be better to return an error message when the feature `profiling` is disabled. And it already updated, please take a look. Thanks.

Examples of the proposed responses of RPC jemalloc_profiling_dump:

  * Success : `{"jsonrpc":"2.0","result":"ckb-jeprof.1587964630.heap","id":2}`
  * Fail: `{"jsonrpc":"2.0","error":{"code":-32603,"message":"jemalloc profiling dump: unsupported"},"id":2}`

doitian:
Please fix the conflicts

rpc/src/module/debug.rs
yangby-cryptape: Since u have to fix the lint errors, remove this `clone()`?

rpc/src/module/debug.rs
yangby-cryptape: I agree with u.
Please change the mock version of `jemalloc_profiling_dump`, let it return an error:
https://github.com/nervosnetwork/ckb/blob/e79f7ca87fbfbdd67e121c740dd7eafa9cc12f0a/util/memory-tracker/src/lib.rs#L15-L17

doitian:
bors r=quake,doitian

doitian:
bors try

quake:
bors r=quake,doitian

rpc/src/module/debug.rs
```diff
@@ -1,27 +1,33 @@
...
+    fn jemalloc_profiling_dump(&self) -> Result<String> {
         let timestamp = time::SystemTime::now()
             .duration_since(time::SystemTime::UNIX_EPOCH)
             .unwrap()
             .as_secs();
-        let filename = format!("ckb-jeprof.{}.heap\0", timestamp);
-        ckb_memory_tracker::jemalloc_profiling_dump(filename);
-        Ok(())
+        let filename = format!("ckb-jeprof.{}.heap", timestamp);
+        match ckb_memory_tracker::jemalloc_profiling_dump(filename.clone()) {
```
yangby-cryptape, line 23:
```suggestion
        match ckb_memory_tracker::jemalloc_profiling_dump(&filename) {
```

doitian:
Linter fails, please fix.

keroro520:
> Please fix linter

Updated. Something strange, `make clippy` works well on my computer but fails on CI. Wait for CI run on the latest update.

doitian:
bors r=quake,doitian

PR #2052, opened 2020-04-30
fix: return connected address in RPC `get_peers`

keroro520:
The RPC `get_peers` miss the peer connected address. Hence it may be empty addresses returned for inbound peers. e.g.

```json
      {
         "addresses" : [],
         "version" : "0.29.0 (a6733e6 2020-02-26)",
         "node_id" : "QmPoPDhGPbNfAuDM1kmQFyY8cocC2SXipi6gYV6BmAj2w7",
         "is_outbound" : false
      }
```

This PR:

* Includes connected address in the returning addresses
* Uses the score recorded in `peer_store` as the returning peer score

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -1,7 +1,7 @@
 use crate::errors::Error;
 use crate::peer_registry::{ConnectionStatus, PeerRegistry};
 use crate::peer_store::{
-    types::{BannedAddr, MultiaddrExt},
+    types::{AddrInfo, BannedAddr, IpPort, MultiaddrExt},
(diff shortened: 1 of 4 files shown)


doitian:
Is it possible to add test cases to cover more branches instead of just relying on the RPC documentation test?

keroro520:
> Is it possible to add test cases to cover more branches instead of just relying on the RPC documentation test?

Adding integration tests can cover more detail. It seems hard to do it in unit tests.

quake:
bors r=quake,doitian

PR #2053, opened 2020-04-30
test: save integration test logs in a specified directory

yangby-cryptape:
Currently, it's too hard to distinguish between integration test logs and other temporary logs, especially when make integration test to be automated.


quake:
bors r=keroro520,doitian

PR #2061, opened 2020-05-06
test: use unique ports for each node in integration tests

yangby-cryptape:
- Fix `AddrInUse: Only one usage of each socket address (protocol/network address/port) is normally permitted.` when running integration tests in windows.

  CI Result Preview: [CKB GitHub Action](https://github.com/nervosnetwork/ckb/runs/650909401)

- In fact, `scoop install msys2` doesn't install `msys2`, it just installs a `msys2-installer`, so we can just remove this task from Azure Windows CI.

  CI Result Preview: [CKB GitHub Action](https://github.com/nervosnetwork/ckb/runs/651041405)

### DIFF
--- a/devtools/azure/windows-dependencies.yml
+++ b/devtools/azure/windows-dependencies.yml
@@ -10,8 +10,6 @@ steps:
   displayName: Add scoop to path
 - script: scoop install llvm
   displayName: Install LLVM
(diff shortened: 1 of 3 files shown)


doitian:
bors try

quake:
bors r=quake,doitian

PR #2062, opened 2020-05-07
chore: Only use gow instead of msys2 in Windows CI

xxuejie:
We don't need the full msys2 to build, gow can help us save some
dependency installation time


yangby-cryptape:
In fact, `scoop install msys2` doesn't install `msys2`, it just installs a `msys2-installer`, so we can just remove this task.
I had test it yesterday.

CI Result Preview (tested with #2061 in Azure CI): [CKB GitHub Action](https://github.com/nervosnetwork/ckb/runs/651041405)

PR #2063, opened 2020-05-10
fix: should use an empty peer store when failed to load data from file

quake:
`PeerStore#load_from_dir` should return an empty peer store instead of panic when failed to load data from file system.

### DIFF
--- a/network/src/peer_store/peer_store_db.rs
+++ b/network/src/peer_store/peer_store_db.rs
@@ -7,7 +7,7 @@ use crate::{
         PeerStore,
     },
 };
-use ckb_logger::debug;
+use ckb_logger::{debug, error};
 use std::fs::{copy, create_dir_all, remove_file, rename, File, OpenOptions};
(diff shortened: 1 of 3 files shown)


network/src/peer_store/peer_store_db.rs
```diff
@@ -49,25 +49,19 @@ impl BanList {
 }
 
 impl PeerStore {
-    pub fn load_from_dir<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
+    pub fn load_from_dir<P: AsRef<Path>>(path: P) -> Self {
```
doitian, line 52: I prefer a new method `load_from_dir_or_default`. It just not feels right to silently discard the error.

quake: renamed method accordingly and added the error log when failed to load data from file (empty or broken), test case is added also, please review again, thanks

doitian:

PR #2064, opened 2020-05-11
feat: add RPC truncate

keroro520:
For convenient to reproduce a specified environment when test, this PR adds RPC `truncate(target_tip_hash)` to roll-back the blockchain downto the target block. It trigger the below behaviours:

  * Roll-back the blockchain status, like cellset and transactions, downto the height corresponding to `target_tip_hash`. In short, reset the tip to the target block.
  * Delete the truncated blocks from database.
  * Clear the tx-pool, which may hold dirty transactions/proposals.

### DIFF
--- a/tx-pool/src/process.rs
+++ b/tx-pool/src/process.rs
@@ -29,7 +29,6 @@ use failure::Error as FailureError;
 use faketime::unix_time_as_millis;
 use std::collections::HashSet;
 use std::collections::{HashMap, VecDeque};
-use std::ops::DerefMut;
 use std::sync::atomic::Ordering;
 use std::sync::{atomic::AtomicU64, Arc};
 use std::{cmp, iter};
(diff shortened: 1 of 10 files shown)


zhangsoledad:
https://github.com/nervosnetwork/ckb/pull/2120

keroro520:
> * The code is a bit too complex. The chain service should provide a interface to allow setting any stored block as tip.

I think the [`ChainController::truncate`](https://github.com/nervosnetwork/ckb/pull/2064/files#diff-dc8555c8638c6e8bf7723ae776ce254bR61) is that interface.

Thanks for @zhangsoledad doing some refactoring, it may be more simple. PTAL.

> * Also it is better to leave the stored blocks untouched. Once a node receives a stored block, and it has better cumulative difficulty, the chain service should set tip to this new block instead.

RPC "truncate" is to reproduce a specified environment when test. The "environment" includes tx-pool, chain, and database. So I think removing the truncated blocks from database is better.

As for "sync" a stored block and the stored block has a better cumulative difficulty, It is out of this PR.

tx-pool/src/process.rs
```diff
@@ -501,6 +501,14 @@ impl TxPoolService {
             }
         });
     }
+
+    pub(crate) async fn clear_pool(&self) {
```
doitian, line 505: This method also deserves an RPC method.

doitian:
bors r=quake,doitian

PR #2067, opened 2020-05-12
Optimize scheduler

driftluo:
Problems with the current scheduler:
1. The calculation is too frequent
2. Inability to adapt to complex network environments, e.g., increased network latency as the number of nodes increases, network latency exceeding expectations, etc.

This PR implements an adaptive scheduler based on past data, removing most of the redundant calculations. Translating the overall data into 4 intervals:

| fast | normal | penalty | double penalty |

The dividing line is the 1/3 position, 4/5 position, 1/10 position.

There is a 14/30 normal area, 1/10 penalty area, 1/10 double penalty area, 1/3 accelerated reward area.

Replace the 30-second clear-out strategy with a doubling penalty.

10%-20% increase in sync performance from testing, from 220-240 block/s to 250-270 block/s on a dual-core machine

the simulation test: https://github.com/nervosnetwork/sync-scheduler-simulation-test

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -3,8 +3,8 @@ use crate::orphan_block_pool::OrphanBlockPool;
 use crate::BLOCK_DOWNLOAD_TIMEOUT;
 use crate::{NetworkProtocol, SUSPEND_SYNC_TIME};
 use crate::{
-    FIRST_LEVEL_MAX, INIT_BLOCKS_IN_TRANSIT_PER_PEER, MAX_BLOCKS_IN_TRANSIT_PER_PEER,
-    MAX_HEADERS_LEN, MAX_TIP_AGE, RETRY_ASK_TX_TIMEOUT_INCREASE,
+    INIT_BLOCKS_IN_TRANSIT_PER_PEER, LOW_INDEX, MAX_BLOCKS_IN_TRANSIT_PER_PEER, MAX_HEADERS_LEN,
+    MAX_TIP_AGE, MEDIAN_INDEX, NORMAL_INDEX, RETRY_ASK_TX_TIMEOUT_INCREASE, TIME_TRACE_SIZE,
 };
 use ckb_chain::chain::ChainController;
 use ckb_chain_spec::consensus::Consensus;
@@ -280,11 +280,90 @@ impl From<(BlockNumber, Byte32)> for BlockNumberAndHash {
     }
 }
 
+enum SchedulerOutput {
+    DoubleInCrease,
+    InCrease,
+    Decrease,
+    DoubleDeCrease,
+}
+
+/// Using 512 blocks as a period, dynamically adjust the scheduler's time standard
+/// Divided into three time periods, including:
+///
+/// | fast | normal | penalty | double penalty |
+///
+/// The dividing line is, median position, 4/5 position, 1/10 position.
+///
+/// There is 3/10 normal area, 1/10 penalty area, 1/10 double penalty area, 1/2 accelerated reward area.
+///
+/// Most of the nodes that fall in the normal and accelerated reward area will be retained,
+/// while most of the nodes that fall in the normal and penalty zones will be slowly eliminated
+///
+/// The purpose of dynamic tuning is to reduce the consumption problem of sync networks
+/// by retaining the vast majority of nodes with stable communications and
+/// cleaning up nodes with significantly lower response times than a certain level
+#[derive(Clone)]
+struct DynamicTimeLimit {
+    trace: [u64; TIME_TRACE_SIZE],
+    index: usize,
+    median_time: u64,
+    normal_time: u64,
+    low_time: u64,
+}
+
+impl fmt::Debug for DynamicTimeLimit {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("DynamicTimeLimit")
+            .field("median_time", &self.median_time)
+            .field("normal_time", &self.normal_time)
+            .field("low_time", &self.low_time)
+            .finish()
+    }
+}
+
+impl Default for DynamicTimeLimit {
+    fn default() -> Self {
+        // Block max size about 700k, Under 10m/s bandwidth it may cost 1s to response
+        Self {
+            trace: [0; TIME_TRACE_SIZE],
+            index: 0,
+            median_time: 1000,
+            normal_time: 1250,
+            low_time: 1500,
+        }
+    }
+}
+
+impl DynamicTimeLimit {
+    fn push_time(&mut self, time: u64) -> SchedulerOutput {
+        if self.index < TIME_TRACE_SIZE {
+            self.trace[self.index] = time;
+            self.index += 1;
+        } else {
+            self.trace.sort_unstable();
+            self.median_time = (self.median_time + self.trace[MEDIAN_INDEX]) >> 1;
+            self.normal_time = (self.normal_time + self.trace[NORMAL_INDEX]) >> 1;
+            self.low_time = (self.low_time + self.trace[LOW_INDEX]) >> 1;
+            self.trace[0] = time;
+            self.index = 1;
+        }
+
+        if time <= self.median_time {
+            SchedulerOutput::DoubleInCrease
+        } else if time <= self.normal_time {
+            SchedulerOutput::InCrease
+        } else if time > self.low_time {
+            SchedulerOutput::DoubleDeCrease
+        } else {
+            SchedulerOutput::Decrease
+        }
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct DownloadScheduler {
     task_count: usize,
     timeout_count: usize,
-    breakthroughs_count: usize,
     hashes: HashSet<BlockNumberAndHash>,
 }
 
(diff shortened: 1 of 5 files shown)


driftluo:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 359.98
  - Samples Count: 51
  - CKB Version: 70b6265ae1d0a969e0559859ef35f8d0b04ac304
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

driftluo:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 294.03
  - Samples Count: 51
  - CKB Version: 25e539c75ceab25802cfbbd3bbdb70276100a5b8
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

driftluo:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 354.71
  - Samples Count: 50
  - CKB Version: 70b6265ae1d0a969e0559859ef35f8d0b04ac304
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

sync/src/types.rs
```diff
@@ -280,11 +280,90 @@ impl From<(BlockNumber, Byte32)> for BlockNumberAndHash {
...
 }
 
+enum SchedulerOutput {
+    DoubleInCrease,
+    InCrease,
+    Decrease,
+    DoubleDeCrease,
+}
+
+/// Using 512 blocks as a period, dynamically adjust the scheduler's time standard
```
quake, line 290: This structure looks like a typical quantiles query algorithm on streaming data,  I'm curious about why [0, 0.5, 0.8, 0.9] quantiles are using here, is it better than [0, 0.25, 0.5, 0.75] ?

sync/src/types.rs
```diff
@@ -308,65 +386,30 @@ impl DownloadScheduler {
...
+    fn increase(&mut self, num: usize) {
+        if self.task_count < MAX_BLOCKS_IN_TRANSIT_PER_PEER {
+            self.task_count += num
+        }
+    }
+
+    fn decrease(&mut self, num: usize) {
         // But the mark is on the same time, so multiple tasks may be affected by one task,
         // causing all responses to be in the range of reduced tasks.
         // So the adjustment to reduce the number of tasks will be calculated by modulo 3 with the actual number of triggers.
```
quake, line 402: outdated comment?

sync/src/types.rs
```diff
@@ -280,11 +280,90 @@ impl From<(BlockNumber, Byte32)> for BlockNumberAndHash {
     }
 }
 
+enum SchedulerOutput {
+    DoubleInCrease,
```
quake, line 284: inconsistent naming style, `InCrease` vs `Decrease` vs `DoubleDeCrease`
```suggestion
    DoubleIncrease,
    Increase,
    Decrease,
    DoubleDecrease
```

and I don't like the enum name here, it's a quantile, not related to scheduler, how about renaming the `DynamicTimeLimit` to `TimeAnalyzer`, and return a `TimeQuantile` enum?
```
MinToLower
LowerToMedian
MedianToUpper
UpperToMax
```

sync/src/types.rs
driftluo: Yes, the original plan was actually using 0.5 as the boundary, but the results were very poor and would result in a much slower speed. There are several reasons for this.

The purpose of the scheduler is to filter out the relatively good nodes and maintain communication with them. 

If 0.5 is used as a boundary, it may not be a big problem at first, but at a certain point, there will be self-competition, resulting in the number of tasks per node not reaching the desired maximum, which affects the overall efficiency.

Similarly, sampling at 512 intervals is also the result of comparative experiments. If it's too small, it's calculated too often. If it is too large, the purpose of the screening cannot be achieved in the default configuration. Guaranteed number of tasks for nearly 4 nodes with maximum task*4

doitian:
@quake plans to give a better distribution models in simulation test to find better parameters after @keroro520 has collected new metrics. Before that we can merge this PR when there's no problems on the algorithm itself.

sync/src/types.rs
```diff
@@ -308,65 +386,26 @@ impl DownloadScheduler {
...
-            // Above 1500ms, relatively poor network, reduced
-            _ => {
-                self.timeout_count += 1;
-                if self.timeout_count > 3 {
-                    self.task_count = self.task_count.saturating_sub(1);
-                    self.timeout_count = 0;
-                }
-            }
+    fn increase(&mut self, num: usize) {
+        if self.task_count < MAX_BLOCKS_IN_TRANSIT_PER_PEER {
```
doitian, line 394: It's wrong when `num > 1`.

sync/src/types.rs
```diff
@@ -308,65 +386,26 @@ impl DownloadScheduler {
...
+            self.task_count += num
         }
```
doitian, line 396:
```suggestion
        self.task_count = self.task_count.saturating_add(num).min(MAX_BLOCKS_IN_TRANSIT_PER_PEER)
```

sync/src/types.rs
```diff
@@ -308,65 +386,26 @@ impl DownloadScheduler {
...
+        self.timeout_count += num;
+        if self.timeout_count > 2 {
+            self.task_count = self.task_count.saturating_sub(1);
+            self.timeout_count = 0;
         }
     }
 
-    fn punish(&mut self) {
-        self.task_count >>= 1
+    fn punish(&mut self, num: usize) {
```
doitian, line 407: Name `exp` is better than `num`.

sync/src/types.rs
```diff
@@ -308,65 +386,26 @@ impl DownloadScheduler {
...
-        if self.breakthroughs_count > 4 {
-            if self.task_count < MAX_BLOCKS_IN_TRANSIT_PER_PEER {
-                self.task_count += 1;
-            }
-            self.breakthroughs_count = 0;
+    }
+
+    fn decrease(&mut self, num: usize) {
+        self.timeout_count += num;
+        if self.timeout_count > 2 {
```
doitian, line 401: Why changed from 3 to 2? Why not decreasing `task_count` directly?

doitian: Since it does not decrease `task_count` like what `increase` does, a name like `on_timeout` is better.

sync/src/types.rs
```diff
@@ -280,11 +280,90 @@ impl From<(BlockNumber, Byte32)> for BlockNumberAndHash {
...
+    fn push_time(&mut self, time: u64) -> TimeQuantile {
+        if self.index < TIME_TRACE_SIZE {
+            self.trace[self.index] = time;
+            self.index += 1;
+        } else {
+            self.trace.sort_unstable();
+            self.fast_time = (self.fast_time + self.trace[FAST_INDEX]) >> 1;
+            self.normal_time = (self.normal_time + self.trace[NORMAL_INDEX]) >> 1;
+            self.low_time = (self.low_time + self.trace[LOW_INDEX]) >> 1;
+            self.trace[0] = time;
```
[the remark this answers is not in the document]
driftluo: Only records are not sorted in each window, and sorting at the moment of switching windows, after the value is taken, all data is useless

sync/src/types.rs
driftluo: Nothing wrong, it is allowed to be greater than `MAX_BLOCKS_IN_TRANSIT_PER_PEER`, only need to control not to rise indefinitely

sync/src/types.rs
driftluo: > Why changed from 3 to 2? Why not decreasing `task_count` directly?

Experience value, it will affect efficiency after changing to direct operation

driftluo: > Since it does not decrease `task_count` like what `increase` does, a name like `on_timeout` is better.

I think the behavior is consistent

driftluo: By the way, the penalty and bonus values are matched when the penalty range hits about 1.5 times the bonus range's the `task_count` must be trending downward until it is finally expelled.

sync/src/types.rs
doitian: This issue is that the upper is checked before adding `num` to it. So if a node may have a higher task count when adding 1 then adding 2. It's not issue since we only use 1 and 2 now. But later if we decide to use a higher `num`, it can cause strange result.

driftluo: I don't think it will be a problem, but I'll change it if you insist.

doitian:
Sorry that I misunderstood https://github.com/nervosnetwork/ckb/pull/2067#discussion_r442596402 , the original code looks good.

doitian:
bors r=quake,doitian

doitian:
The latest test result. The unit is block counts per second.

  | avg | median | min | max | count
-- | -- | -- | -- | -- | --
eu_33_80_90 | 696 | 704 | 454 | 798 | 200
eu_50_75_90 | 674 | 683 | 483 | 786 | 193
eu_33_75_90 | 670 | 683 | 464 | 793 | 192
ap_33_80_90 | 662 | 666 | 475 | 759 | 190
ap_50_75_90 | 648 | 648 | 475 | 764 | 186
ap_33_75_90 | 636 | 650 | 474 | 720 | 182

PR #2070, opened 2020-05-12
chore: remove line with only # in generated ckb-miner.toml

doitian:


resource/ckb-miner.toml
```diff
@@ -57,7 +57,5 @@ worker_type = "EaglesongSimple" # {{
 # }}
 threads     = 1 # {{
 # dev => delay_type = "Constant"\nvalue = 5000
-# }}
-# # {{
-# testnet => extra_hash_function = "Blake2b"
+# testnet => threads     = 1\nextra_hash_function = "Blake2b"
```
quake, line 60: duplicate line: `thread = 1`  ?

doitian: No, the content before `# {{` will be replaced.

yangby-cryptape:
bors r=quake,driftluo,yangby-cryptape

PR #2074, opened 2020-05-13
fix: orphan block pool deadlock

quake:
resolved potential deadlock in `OrphanBlockPool`  and a unit test which simulates deadlocking was added also.

### DIFF
--- a/sync/src/orphan_block_pool.rs
+++ b/sync/src/orphan_block_pool.rs
@@ -58,9 +58,10 @@ impl OrphanBlockPool {
     }
 
     pub fn get_block(&self, hash: &packed::Byte32) -> Option<core::BlockView> {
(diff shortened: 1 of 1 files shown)


doitian:
> I'm OK with the fix. Just raise a question: how to avoid the similar bugs in the future?

Since every function requires the lock of `blocks` first, maybe an internal structure as below will work.

```
internal: RwLock<(BlocksMap, RwLock<ParentsMap>)>,
```

quake:
> I'm OK with the fix. Just raise a question: how to avoid the similar bugs in the future?

I think it can be detected by feature flag `deadlock_detection`, however it's not enabled in release build

doitian:
https://github.com/tokio-rs/loom may help here to reproduce the bug deterministically.

doitian:
bors r=doitian,zhangsoledad

PR #2075, opened 2020-05-14
fix: Fix collaboration issues between two protocol

driftluo:
relayer and sync have some known problems in the critical zone of switching, which are inherent to the request/response architecture.

Previous attempts to fix the problem by allowing duplicate downloads, but this solution results in wasted network resources and slowed sync speed.

This typical problem can be solved by the node itself, and the current PR gives a temporary solution to try to check the orphan pool with a very low-frequency timer.

### DIFF
--- a/sync/src/relayer/mod.rs
+++ b/sync/src/relayer/mod.rs
@@ -43,6 +43,7 @@ use std::time::{Duration, Instant};
 pub const TX_PROPOSAL_TOKEN: u64 = 0;
 pub const ASK_FOR_TXS_TOKEN: u64 = 1;
 pub const TX_HASHES_TOKEN: u64 = 2;
+pub const SEARCH_ORPHAN_POOL_TOKEN: u64 = 3;
 
 pub const MAX_RELAY_PEERS: usize = 128;
 pub const MAX_RELAY_TXS_NUM_PER_BATCH: usize = 32767;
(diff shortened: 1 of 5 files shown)


sync/src/relayer/mod.rs
```diff
@@ -711,6 +715,12 @@ impl CKBProtocolHandler for Relayer {
             }
             ASK_FOR_TXS_TOKEN => self.ask_for_txs(nc.as_ref()),
             TX_HASHES_TOKEN => self.send_bulk_of_tx_hashes(nc.as_ref()),
+            SEARCH_ORPHAN_POOL_TOKEN => tokio::task::block_in_place(|| {
+                self.shared.try_search_orphan_pool(
+                    &self.chain,
+                    &self.shared.active_chain().tip_header().hash(),
+                )
+            }),
```
zhangsoledad, line 723: https://github.com/nervosnetwork/p2p/blob/master/src/protocol_handle_stream.rs#L197
p2p already issuing `notify` handle as a blocking call, so I think  `tokio::task::block_in_place(` is unnecessary, also the `TX_PROPOSAL_TOKEN` handle.

driftluo: https://github.com/nervosnetwork/ckb/blob/develop/ckb-bin/src/subcommand/run.rs#L91
For performance and control, p2p blocking is disabled at the ckb level and it is controlled by itself

doitian:
@zhangsoledad Do you still have any concern about this PR?

doitian:
bors r=quake,doitian

PR #2076, opened 2020-05-16
feat: Expose fields from setup struct

xxuejie:
Previously, the fields in Setup in app-config package remains private,
meaning the only way we can initialize a Setup struct, is via the
exposed from_matches method. But since from_matches only accepts
matches initialized from clap, this means the only way we can
initialize CKB, would be from a binary via argv. This commit changes
the code so fields from Setup struct are exposed, which can enable us
to use much of CKB code as a library, not necessarily a binary.


xxuejie:
bors r=quake,yangby-cryptape,keroro520

PR #2080, opened 2020-05-18
chore: skip audit for a deprecated crate: net2

yangby-cryptape:
The `net2` crate has been deprecated and users are encouraged to considered `socket2` instead.
But there are but a lot of required crates are dependent on it.


doitian:
bors r=doitian,driftluo

PR #2081, opened 2020-05-18
feat: update last_common_header only in find_blocks_to_fetch

keroro520:
* `peer.best_known_block` refers to the best-known block we know this peer has announced; `peer.last_common_header` refers to the last block we both stored between local and peer. This PR proposes the logic to update the two fields:

  - When we receive `CompactBlock` or `SentHeader` from `peer`, we update the `peer.best_known_block` if the arrival one has greater `total_difficulty`.
  - When we insert a block which is from `peer`, we don't update the corresponding `last_common_header`(remove lines [CompactBlock](https://github.com/nervosnetwork/ckb/pull/2081/files#diff-98eccb89a5611099d6adca819c9c8b08L100), [SendBlock](https://github.com/nervosnetwork/ckb/pull/2081/files#diff-402ee94ba649e0f53da25c4da84ec58dL50-L56)).
  - We update the `last_common_header` only in `find_blocks_to_fetch`. There are 2 places, 
    1. [`fn update_last_common_header` fixes `last_common_header` when the peer has reorganized](https://github.com/nervosnetwork/ckb/pull/2081/files#diff-86d69964b227ca1248fadb84da967a96R62-R64);
    2. [Updates `last_common_header` when we know a better block has already stored](https://github.com/nervosnetwork/ckb/pull/2081/files#diff-86d69964b227ca1248fadb84da967a96R120-R124).

* Fix bug: [`status == BLOCK_STORED` => `status.contains(BLOCK_STORED)`](https://github.com/nervosnetwork/ckb/pull/2081/files#diff-86d69964b227ca1248fadb84da967a96L151)

* Enhence `last_common_ancestor`.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -65,7 +65,9 @@ impl<'a> BlockFetcher<'a> {
         let fixed_last_common_header = self
             .active_chain
             .last_common_ancestor(&last_common_header, &best.inner())?;
-        self.synchronizer.peers().set_last_common_header(self.peer, fixed_last_common_header.clone());
+        self.synchronizer
+            .peers()
+            .set_last_common_header(self.peer, fixed_last_common_header.clone());
 
         Some(fixed_last_common_header)
     }
@@ -152,7 +154,9 @@ impl<'a> BlockFetcher<'a> {
                 if status.contains(BlockStatus::BLOCK_STORED) {
                     // If the block is stored, its ancestor must on store
                     // So we can skip the search of this space directly
-                    self.synchronizer.peers().set_last_common_header(self.peer, header.clone());
+                    self.synchronizer
+                        .peers()
+                        .set_last_common_header(self.peer, header.clone());
                     break;
                 } else if status.contains(BlockStatus::BLOCK_RECEIVED) {
                     // Do not download repeatedly
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -2,7 +2,7 @@ use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
 use crate::types::{ActiveChain, HeaderView, IBDState};
 use crate::BLOCK_DOWNLOAD_WINDOW;
-use ckb_logger::{debug, trace};
+use ckb_logger::trace;
 use ckb_network::PeerIndex;
 use ckb_types::{core, packed};
 use std::cmp::min;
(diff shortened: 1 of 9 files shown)


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -40,41 +40,38 @@ impl<'a> BlockFetcher<'a> {
...
+        self.synchronizer
+            .peers()
+            .set_last_common_header(self.peer, last_common.clone());
+
+        Some(last_common)
     }
 
     pub fn fetch(self) -> Option<Vec<Vec<packed::Byte32>>> {
-        trace!("[block downloader] BlockFetcher process");
+        let best_known = self.peer_best_known_header()?;
```
driftluo, line 70: The wrong call location of this function will cause the ibd multi-node download to fail, must after `try_update_best_known_with_unknown_header_list`

keroro520: Fixed: b32b91de1503264ad6fa5ca2cd82b82263c8d4a3

doitian:
Is there any benchmark to show the different among the three different strategies?

1. Update last common immediately for all peers whenever the node receives a new block.
2. Lazy update only in fetch
3. A hybrid solution. Do some quick updates in 1 and use 2 as a fallback.

quake:
@keroro520 CI failed, may need to rebase, please check the log

keroro520:
> Is there any benchmark to show the differences among the three different strategies?
> 
> 1. Update last common immediately for all peers whenever the node receives a new block.
> 2. Lazy update only in fetch
> 3. A hybrid solution. Do some quick updates in 1 and use 2 as a fallback.

No benchmark for this change, as I don't think it is necessary to bench it.

1. In my opinion, the strategy to update `last_common_header` cause a very small effect on the sync-performance. It's hard to make this kind of benchmark.

2. "Update last common immediately" is prone to error. For example, when [`chain.process_block` returns `Ok(false)` because the given block is not a main-chain block, but the original code will still update the `last_common_header`](https://github.com/nervosnetwork/ckb/blob/868312d73dc275ccd1d03610bde4bf31ba79f21a/sync/src/types.rs#L1070).

3. "Lazy update only in fetch" is more simple and easy to ensure the correctness.

doitian:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 345.21
  - Samples Count: 50
  - CKB Version: e99b150daee3b942b6d36753a8a59fb2a7620cae
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

sync/src/synchronizer/block_process.rs
```diff
@@ -9,7 +9,7 @@ use ckb_types::{packed, prelude::*};
 pub struct BlockProcess<'a> {
     message: packed::SendBlockReader<'a>,
     synchronizer: &'a Synchronizer,
-    peer: PeerIndex,
+    _peer: PeerIndex,
```
doitian, line 12: Why?

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -151,6 +121,9 @@ impl<'a> BlockFetcher<'a> {
                 if status.contains(BlockStatus::BLOCK_STORED) {
                     // If the block is stored, its ancestor must on store
                     // So we can skip the search of this space directly
+                    self.synchronizer
+                        .peers()
+                        .set_last_common_header(self.peer, header.clone());
```
doitian, line 126: It's better to also slide the window as well.


```suggestion
                    end = min(best_known.number(), header.number() + 1 + BLOCK_DOWNLOAD_WINDOW);
                    self.synchronizer
                        .peers()
                        .set_last_common_header(self.peer, header.clone());
```

This also requires change end to `let mut end`.

sync/src/synchronizer/block_process.rs
keroro520: - https://github.com/nervosnetwork/ckb/pull/2081/files/e99b150daee3b942b6d36753a8a59fb2a7620cae#diff-402ee94ba649e0f53da25c4da84ec58dL41

- https://github.com/nervosnetwork/ckb/pull/2081/files/e99b150daee3b942b6d36753a8a59fb2a7620cae#diff-402ee94ba649e0f53da25c4da84ec58dL56

The above two code has changed/removed, `peer` is unnecessary for `BlockProcess`.
However, removing `peer` from `BlockProcess` would affect many test codes. I just change to the field name to `_peer` to disable the compiler noise.

sync/src/synchronizer/block_fetcher.rs
keroro520: 1493a7773969bcf5f565ff8bf3c2c7bd6eaa085d merges the suggested change manually.

keroro520:
bors r=quake,doitian

PR #2082, opened 2020-05-18
chore: remove duplicate code of peer protocols insert/remove

quake:
Peer protocols are updated via: https://github.com/nervosnetwork/ckb/blob/b8a4743c4ae9368f5d56880c74d7edb7580d0d45/network/src/network.rs#L765 and https://github.com/nervosnetwork/ckb/blob/b8a4743c4ae9368f5d56880c74d7edb7580d0d45/network/src/network.rs#L792

it's unnecessary to update it again in each protocol handler

=========update=========
Changed the sync/relay protocol hander mode to `ProtocolHandle::Callback` according to review comment

### DIFF
--- a/network/src/protocols/mod.rs
+++ b/network/src/protocols/mod.rs
@@ -20,7 +20,7 @@ use std::{
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
-    time::Duration,
+    time::{Duration, Instant},
 };
 use tokio_util::codec::length_delimited;
 
@@ -167,7 +167,7 @@ impl CKBProtocol {
             })
             .support_versions(supported_versions)
             .service_handle(move || {
-                ProtocolHandle::Both(Box::new(CKBHandler {
+                ProtocolHandle::Callback(Box::new(CKBHandler {
                     proto_id: self.id,
                     network_state: Arc::clone(&self.network_state),
                     handler: self.handler,
@@ -201,6 +201,12 @@ impl ServiceProtocol for CKBHandler {
     }
 
     fn connected(&mut self, context: ProtocolContextMutRef, version: &str) {
+        self.network_state.with_peer_registry_mut(|reg| {
+            if let Some(peer) = reg.get_peer_mut(context.session.id) {
+                peer.protocols.insert(self.proto_id, version.to_owned());
+            }
+        });
+
         let pending_data_size = context.session.pending_data_size();
         let send_paused = pending_data_size >= self.network_state.config.max_send_buffer();
         let nc = DefaultCKBProtocolContext {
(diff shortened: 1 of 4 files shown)


driftluo:
https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/mod.rs#L170

I think the comment is correct, it is wrong here, it should not be both

TheWaWaR:
> https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/mod.rs#L170
> 
> I think the comment is correct, it is wrong here, it should not be `Both`, it is useless in sync and relay protocols, but lower performance

`Both` is required, because [`EventHandler`](https://github.com/nervosnetwork/ckb/blob/develop/network/src/network.rs#L551) depend on it.

driftluo:
> > https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/mod.rs#L170
> > I think the comment is correct, it is wrong here, it should not be `Both`, it is useless in sync and relay protocols, but lower performance
> 
> `Both` is required, because [`EventHandler`](https://github.com/nervosnetwork/ckb/blob/develop/network/src/network.rs#L551) depend on it.

Not needed for sync relay

network/src/protocols/mod.rs
```diff
@@ -201,6 +201,12 @@ impl ServiceProtocol for CKBHandler {
     }
 
     fn connected(&mut self, context: ProtocolContextMutRef, version: &str) {
+        self.network_state.with_peer_registry_mut(|reg| {
+            if let Some(peer) = reg.get_peer_mut(context.session.id) {
+                peer.protocols.insert(self.proto_id, version.to_owned());
```
driftluo, line 206: version use an enum may be better `impl From<'a str> for Version`

network/src/protocols/mod.rs
```diff
@@ -167,7 +167,7 @@ impl CKBProtocol {
             })
             .support_versions(supported_versions)
             .service_handle(move || {
-                ProtocolHandle::Both(Box::new(CKBHandler {
+                ProtocolHandle::Callback(Box::new(CKBHandler {
```
keroro520, line 170: I wonder which behaviour changed after modifying `ProtocolHandle::Both` to `ProtocolHandle::Callback`?

quake: Yes, please check @driftluo 's review comment, it should use `Callback` mode here.

doitian: We need some documents about Both and Callback

doitian:

quake:

quake:

PR #2090, opened 2020-05-25
Add fail-points to test robustness

keroro520:
- Specify the actions of fail-points via the configuration `fail-points` in `ckb.toml`.
  - Add fail-points at the front of sending/receiving network messages.
  - Create `sync/src/utils.rs` to wrap most of the operations of sending messages.
  - Refactor some code.
  - Run 2 nodes from the fresh state, the one configured with `0.1%` probability of network messages lost, another is normal. The former takes `170` minutes and the latter takes `120` minutes to synchronize to the TIP.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -379,6 +379,7 @@ dependencies = [
  "clap",
  "crossbeam-channel",
  "ctrlc",
+ "fail",
  "serde",
  "serde_plain",
  "tempfile",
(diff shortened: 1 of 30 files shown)


doitian:
> Please add some doc in docs/ckb-core-dev.md about how to enable this feature.

@xxuejie should we start to add docs to new docs site instead?

PR #2091, opened 2020-05-25
feat: logger filter parse crate name leniently

yangby-cryptape:


util/logger/src/lib.rs
```diff
@@ -157,14 +157,46 @@ fn enable_ansi_support() {
...
+            }
+        }
+        filter_new.push(part.to_owned());
+    }
+    filter_new.join(",")
+}
+
+#[test]
+fn test_convert_compatible_crate_name() {
+    let filter = "info,a-b=trace,c-d_e-f=warn/*[0-9],g-h-i=debug/[^-]*,jkl=trace";
```
quake, line 186: There is only one regex filter for all modules, the correct format should be `info,a-b=trace,c-d=warn,g-h-i=debug/one-single-regex-for-all`, that means you may simplified the `convert_compatible_crate_name` fn to `filter.split('/').next().replace('-', '_') + next()`

util/logger/src/lib.rs
```diff
@@ -157,14 +157,37 @@ fn enable_ansi_support() {
 #[cfg(not(target_os = "windows"))]
 fn enable_ansi_support() {}
 
+// Parse crate name leniently in logger filter: convert "-" to "_".
+fn convert_compatible_crate_name(spec: &str) -> String {
+    let mut parts = spec.split('/');
```
quake, line 162:
```suggestion
    let mut parts = spec.splitn(2, '/');
    match (parts.next(), parts.next()) {
        (Some(mods), Some(regex)) => [&mods.replace("-", "_"), regex].join("/"),
        _ => spec.replace("-", "_") 
    }
```

quake:
bors r=quake,keroro520

PR #2095, opened 2020-05-29
test: Need to fully confirm the getblocks request to continue the test

driftluo:
Need to fully confirm the `getblocks` request to continue the test

### DIFF
--- a/test/src/specs/sync/block_sync.rs
+++ b/test/src/specs/sync/block_sync.rs
@@ -8,7 +8,7 @@ use ckb_network::{bytes::Bytes, PeerIndex};
 use ckb_sync::NetworkProtocol;
 use ckb_types::{
     core::BlockView,
-    packed::{self, Byte32, SyncMessage},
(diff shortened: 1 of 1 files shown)


test/src/specs/sync/block_sync.rs
```diff
@@ -278,9 +278,20 @@ impl Spec for BlockSyncOrphanBlocks {
         });
 
         // Wait for block fetch timer
-        let (_, _, _) = net
-            .receive_timeout(Duration::new(10, 0))
-            .expect("net receive timeout");
+        let time = Instant::now();
```
quake, line 281: Please use `should_receive` helper:
```
        net.should_receive(
            |data: &Bytes| {
                SyncMessage::from_slice(&data)
                    .map(|message| message.to_enum().item_name() == GetBlocks::NAME)
                    .unwrap_or(false)
            },
            "Test node should receive GetBlocks message from node0",
        );
```

quake:
bors r=quake,keroro520

PR #2096, opened 2020-06-02
refactor: move network protocol related variables to SupportProtocols

quake:
To eliminate dependence of `ckb-sync` crate,  this PR refactored network protocol related variables and move them to a new enum: `SupportProtocols`

### DIFF
--- a/network/src/protocols/support_protocols.rs
+++ b/network/src/protocols/support_protocols.rs
@@ -1,0 +1,134 @@
+use crate::ProtocolId;
+use p2p::{
+    builder::MetaBuilder,
+    service::{BlockingFlag, ProtocolHandle, ProtocolMeta},
+    traits::ServiceProtocol,
+};
+use tokio_util::codec::length_delimited;
+
+pub enum SupportProtocols {
+    Ping,
+    Discovery,
+    Identify,
+    Feeler,
+    DisconnectMessage,
+    Sync,
+    Relay,
+    Time,
+    Alert,
+}
+
+impl SupportProtocols {
+    pub fn protocol_id(&self) -> ProtocolId {
+        match self {
+            SupportProtocols::Ping => 0,
+            SupportProtocols::Discovery => 1,
+            SupportProtocols::Identify => 2,
+            SupportProtocols::Feeler => 3,
+            SupportProtocols::DisconnectMessage => 4,
+            SupportProtocols::Sync => 100,
+            SupportProtocols::Relay => 101,
+            SupportProtocols::Time => 102,
+            SupportProtocols::Alert => 110,
+        }
+        .into()
+    }
+
+    pub fn name(&self) -> String {
+        match self {
+            SupportProtocols::Ping => "/ckb/ping",
+            SupportProtocols::Discovery => "/ckb/discovery",
+            SupportProtocols::Identify => "/ckb/identify",
+            SupportProtocols::Feeler => "/ckb/flr",
+            SupportProtocols::DisconnectMessage => "/ckb/disconnectmsg",
+            SupportProtocols::Sync => "/ckb/syn",
+            SupportProtocols::Relay => "/ckb/rel",
+            SupportProtocols::Time => "/ckb/tim",
+            SupportProtocols::Alert => "/ckb/alt",
+        }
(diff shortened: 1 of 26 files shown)


network/src/protocols/support_protocols.rs
```diff
@@ -0,0 +1,131 @@
...
+            SupportProtocols::DisconnectMessage => "/ckb/disconnectmsg",
+            SupportProtocols::Sync => "/ckb/syn",
+            SupportProtocols::Relay => "/ckb/rel",
+            SupportProtocols::Time => "/ckb/tim",
+            SupportProtocols::Alert => "/ckb/alt",
+        }
+        .to_owned()
+    }
+
+    pub fn support_versions(&self) -> Vec<String> {
```
keroro520, line 52: Why there are different versions, `"0.0.1"` and `"1"`?

quake: we didn't invoke `MetaBuilder#support_versions` fn for these protocols (ping/discovery/etc) in previous code, so the default `0.0.1` value is used:
https://github.com/nervosnetwork/tentacle/blob/master/src/builder.rs#L312

have to keep `0.0.1` for compatibility...

keroro520: Please add some comments to describe this history.

quake: copied my comment to code, please review again, thanks.

test/src/specs/sync/invalid_locator_size.rs
```diff
@@ -11,6 +11,9 @@ use log::info;
 
 pub struct InvalidLocatorSize;
 
+// test same value as ckb_sync::MAX_LOCATOR_SIZE;
+const MAX_LOCATOR_SIZE: usize = 101;
+
 impl Spec for InvalidLocatorSize {
```
zhangsoledad, line 17: https://github.com/nervosnetwork/ckb/blob/63a699b65df1e542107956835c64eadc0d460a68/sync/src/lib.rs#L53

zhangsoledad: This refractory does not affect  `MAX_LOCATOR_SIZE`, should remain unchanged

quake: This PR want to remove the dependence of `ckb-sync` crate in the integration test, 3 constants are copied to test code: RETRY_ASK_TX_TIMEOUT_INCREASE / BLOCK_DOWNLOAD_TIMEOUT / MAX_LOCATOR_SIZE, we may move them and more shared constants to a new crate (ckb-constants), and change `ckb-sync` and integration test to use this new crate in the future PR. 

I prefer to re-define it in test code regarding this PR, let's keep changes to a minimum.

zhangsoledad: I insist on this should remain unchanged, since this PR does not move it.
I think this is the so-called minimum change.

quake:
> I insist on this should remain unchanged, since this PR does not move it.
> I think this is the so-called minimum change.

OK, I have reverted related change and will create a new PR for integration-test change, please review again, thanks.

network/src/protocols/support_protocols.rs
```diff
@@ -0,0 +1,134 @@
+use crate::ProtocolId;
+use p2p::{
+    builder::MetaBuilder,
+    service::{BlockingFlag, ProtocolHandle, ProtocolMeta},
+    traits::ServiceProtocol,
+};
+use tokio_util::codec::length_delimited;
+
+pub enum SupportProtocols {
```
keroro520, line 9: FYI. `SupportProtocols` is just like a "constants wrapper", which wraps the protocols and their contributions. It seems that the protocols and contributions are pre-defined. I think the plain style would be more intuitive(define as constants):

  ```
  pub const PING_PROTOCOL_NAME = "ping";
  pub const PING_SUPPORTED_VERSIONS = vec!["1"];
  ```

doitian:
ping @zhangsoledad

quake:
@doitian @zhangsoledad rebased this PR, please review again

doitian:
bors r=doitian,zhangsoledad

PR #1973, opened 2020-03-20
chore: Cleanup Environments

doitian:
* Make the bootnode location names consistent in the config file
* Remove obsoleted staging environment. The bootnode listed as staging has joined the testnet Aggron.

PR #2097, opened 2020-06-02
chore: bring staging environment back

yangby-cryptape:
Revert #1973 but has few changes:
- No bootnodes is set, keep this chain be private.
- ~~Use `EaglesongBlake2b` as staging pow algorithm to avoid denial-of-service attacks.~~


resource/specs/staging.toml
```diff
@@ -0,0 +1,97 @@
+name = "ckb_staging"
+
+[genesis]
+version = 0
+parent_hash = "0x0000000000000000000000000000000000000000000000000000000000000000"
+timestamp = 1564207200792
+compact_target = 0x1e015555
```
doitian, line 7: It may be too large for staging. Let's make it 1000 times easier.

resource/ckb-miner.toml
```diff
@@ -58,4 +61,5 @@ worker_type = "EaglesongSimple" # {{
 threads     = 1 # {{
 # dev => delay_type = "Constant"\nvalue = 5000
 # testnet => threads     = 1\nextra_hash_function = "Blake2b"
+# staging => threads     = 1\nextra_hash_function = "Blake2b"
```
quake, line 64: I think we should use same pow engine on staging as mainnet, it won't be DoSed by asic miner since it's a private env.

resource/specs/staging.toml
```diff
@@ -0,0 +1,97 @@
+name = "ckb_staging"
+
+[genesis]
+version = 0
+parent_hash = "0x0000000000000000000000000000000000000000000000000000000000000000"
+timestamp = 1564207200792
+compact_target = 0x1f010000
```
yangby-cryptape, line 7: Difficulty:
- mainnet: 8318741404228533
- testnet: 12582960
- **staging: 65536**
- dev: 256

Calculate difficulty via `ckb-cli` use next function:

```bash
function compact_to_difficulty () { python -c "print($(ckb-cli util compact-to-difficulty --compact-target ${1-0} | awk '{ print $2 }'))"; }
```

doitian:
bors r=quake,doitian

PR #2101, opened 2020-06-03
fix: resolve ProtocolHandleError scope issue

quake:
this PR resolved an unexpected shutdown issue when we got a `ProtoHandleBlock` error in p2p layer


quake:
> Add a test case? How easy it is to initialize a `EventHandler` for test?

it's not easy to write a test case for this bug ...

quake:
bors r=doitian,driftluo

PR #2103, opened 2020-06-04
refactor: use generic type in NetworkService 

quake:
This PR changed `NetworkService`'s exit_condvar to generic type and removed node_version from `start` fn, make it easier to use `ckb-network` crate as a lib

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -21,7 +21,6 @@ use crate::{
     MAX_FRAME_LENGTH_IDENTIFY, MAX_FRAME_LENGTH_PING,
 };
 use ckb_app_config::NetworkConfig;
-use ckb_build_info::Version;
 use ckb_logger::{debug, error, info, trace, warn};
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_util::{Condvar, Mutex, RwLock};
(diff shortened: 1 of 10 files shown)


network/src/network.rs
```diff
@@ -1026,10 +1047,10 @@ impl NetworkService {
         }
     }
 
-    pub fn start<S: ToString>(
+    pub fn start<NV: ToString, TN: ToString>(
```
doitian, line 1050: It's not necessary to use generic here. It's simpler to just ask for callee to pass in String.

quake: I found this parameter is unnecessary, removed it in new commit, please review again, thanks

network/src/network.rs
```diff
@@ -549,7 +570,7 @@ impl EventHandler {
     }
 }
 
-impl ServiceHandle for EventHandler {
+impl<T: ExitHandler> ServiceHandle for EventHandler<T> {
```
doitian, line 573: Since `ExitHandler` is only called once, there's no performance issue. I think we can use trait box instead of generic to make the code a bit simpler.

keroro520: > Since `ExitHandler` is only called once, there's no performance issue, we can use trait box instead of generic to make the code a bit simpler.

I agree with ian.

keroro520: And I think it is better to define `ExitHandler` and `DefaultExitHandler` in `ckb-util` (`ckb-util` is small enough).

quake:
bors r=doitian,keroro520

doitian:
bors r=doitian,keroro520

PR #2106, opened 2020-06-05
chore: restrict secret file permissions

driftluo:
Restrict the permissions of secret key files

### DIFF
--- a/util/app-config/src/configs/network.rs
+++ b/util/app-config/src/configs/network.rs
@@ -56,14 +56,49 @@ pub(crate) fn write_secret_to_file(secret: &[u8], path: PathBuf) -> Result<(), E
         .create(true)
         .write(true)
         .open(path)
-        .and_then(|mut file| file.write_all(&secret))
+        .and_then(|mut file| {
+            file.write_all(&secret)?;
+            #[cfg(unix)]
+            {
+                use std::os::unix::fs::PermissionsExt;
+                file.set_permissions(fs::Permissions::from_mode(0o400))
+            }
+            #[cfg(not(unix))]
+            {
+                let mut permissions = file.metadata()?.permissions();
+                permissions.set_readonly(true);
+                file.set_permissions(permissions)
+            }
+        })
 }
 
 pub(crate) fn read_secret_key(path: PathBuf) -> Result<Option<secio::SecioKeyPair>, Error> {
-    let mut file = match fs::File::open(path) {
+    let mut file = match fs::File::open(path.clone()) {
         Ok(file) => file,
         Err(_) => return Ok(None),
     };
+    let warn = |m: bool, d: &str| {
+        if m {
+            ckb_logger::warn!(
+                "Your network secret file's permission is not {}, path: {:?}, \
+                please fix it as soon as possible",
+                d,
+                path
+            )
+        }
+    };
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        warn(
+            file.metadata()?.permissions().mode() & 0o177 != 0,
+            "less than 0o600",
+        );
+    }
+    #[cfg(not(unix))]
+    {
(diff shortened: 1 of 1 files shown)


util/app-config/src/configs/network.rs
```diff
@@ -105,7 +116,14 @@ impl Config {
...
             .open(path)
-            .and_then(|mut file| file.write_all(&random_key_pair))
+            .and_then(|mut file| {
+                #[cfg(unix)]
+                {
+                    use std::os::unix::fs::PermissionsExt;
+                    file.set_permissions(fs::Permissions::from_mode(0o600))?;
+                }
+                file.write_all(&random_key_pair)
+            })
```
zhangsoledad, line 126: Maybe we can use `std::fs::set_permissions` here, this function works on Windows. rust said platform-specific behavior may change in the future, but that's future.

util/app-config/src/configs/network.rs
```diff
@@ -105,7 +116,14 @@ impl Config {
             .create(true)
             .write(true)
             .open(path)
-            .and_then(|mut file| file.write_all(&random_key_pair))
+            .and_then(|mut file| {
+                #[cfg(unix)]
```
doitian, line 120: Does it also cover mac?

driftluo: `#[cfg(unix)]` contains mac

`cfg(target_os = "linux")` does not contain mac

util/app-config/src/configs/network.rs
driftluo: Do we need to make separate settings for non-UNIX systems? I think the `read_only` restriction is not appropriate. What I want here is that only owner can access

util/app-config/src/configs/network.rs
```diff
@@ -90,6 +90,17 @@ impl Config {
             Ok(file) => file,
             Err(_) => return Ok(None),
         };
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            if file.metadata()?.permissions().mode() != 0o600 {
+                ckb_logger::warn!(
```
quake, line 97: should we set the permission to 0o600 directly without warning log?

util/app-config/src/configs/network.rs
```diff
@@ -90,6 +90,17 @@ impl Config {
             Ok(file) => file,
             Err(_) => return Ok(None),
         };
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            if file.metadata()?.permissions().mode() != 0o600 {
```
doitian, line 96:
```suggestion
            if file.metadata()?.permissions().mode() & 0o177 != 0 {
```

0400 should also be accepted.

util/app-config/src/configs/network.rs
```diff
@@ -105,7 +116,14 @@ impl Config {
             .create(true)
             .write(true)
             .open(path)
-            .and_then(|mut file| file.write_all(&random_key_pair))
+            .and_then(|mut file| {
+                #[cfg(unix)]
+                {
+                    use std::os::unix::fs::PermissionsExt;
+                    file.set_permissions(fs::Permissions::from_mode(0o600))?;
```
doitian, line 123:
```suggestion
                    file.set_permissions(fs::Permissions::from_mode(0o400))?;
```

util/app-config/src/configs/network.rs
```diff
@@ -90,6 +90,26 @@ impl Config {
             Ok(file) => file,
             Err(_) => return Ok(None),
         };
+
+        let warn = |m: bool| {
+            if m {
+                ckb_logger::warn!(
+                    "Your network secret file's permission is not readonly, path: {:?}, \
```
yangby-cryptape, line 97: If a `secret_key` is `0444`, it's read only but it will cause a warning.

doitian:

PR #2107, opened 2020-06-05
refactor: use generic key / value in template context

quake:
This PR changed `TemplateContext` key/value from fixed field to hashmap, it made the `ckb-resource` crate easier to use in 3rd party applications


doitian:

PR #2111, opened 2020-06-08
perf: spawn fee_estimate

zhangsoledad:
`fee_estimator` acquire tx-pool lock previously, in some cases, may cause performance issues.
move out `fee_estimator` from tx-pool, spawn fee_estimator process for now.


quake:
close via https://github.com/nervosnetwork/ckb/pull/2126

PR #2112, opened 2020-06-08
chore: remove useless `Arc`

yangby-cryptape:
`Shared::store` doesn't need `Arc`, since `ChainDB` only contains `Arc`.
https://github.com/nervosnetwork/ckb/blob/12d8812c80cf6712dcdbae138c85b114788dded2/shared/src/shared.rs#L23-L24
https://github.com/nervosnetwork/ckb/blob/12d8812c80cf6712dcdbae138c85b114788dded2/store/src/db.rs#L18-L21
https://github.com/nervosnetwork/ckb/blob/12d8812c80cf6712dcdbae138c85b114788dded2/db/src/db.rs#L16-L18

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -13,6 +13,7 @@ use std::sync::Arc;
 
 pub const VERSION_KEY: &str = "db-version";
 
+#[derive(Clone)]
 pub struct RocksDB {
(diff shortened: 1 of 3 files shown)


quake:
bors r=quake,keroro520

PR #2114, opened 2020-06-09
feat: add generate peer id command

driftluo:
add generate peer id command

```
$ ./target/debug/ckb peer-id gen --secret_path ./a.txt
$ ./target/debug/ckb peer-id from-secret --secret-path ./a.txt
peer_id: QmYKM9CJhSsn8kBoAmUGdKiS9kbnJA2f9TXohgNBPKxogB
```


util/app-config/src/cli.rs
```diff
@@ -350,6 +355,33 @@ fn init() -> App<'static, 'static> {
         )
 }
 
+fn peer_id() -> App<'static, 'static> {
+    SubCommand::with_name(CMD_PEERID)
+        .about("about peer id, base on Secp256k1")
```
yangby-cryptape, line 360: When I try `ckb --help`, I found that all other about messages were capitalized.

util/app-config/src/cli.rs
```diff
@@ -11,6 +11,9 @@ pub const CMD_PROF: &str = "prof";
 pub const CMD_STATS: &str = "stats";
 pub const CMD_LIST_HASHES: &str = "list-hashes";
 pub const CMD_RESET_DATA: &str = "reset-data";
+pub const CMD_PEERID: &str = "peer-id";
+pub const CMD_GEN_SECRET: &str = "gen";
+pub const CMD_FROM_SECRET: &str = "from_secret";
```
yangby-cryptape, line 16: All other arguments use `-` to split words, except this.

util/app-config/src/cli.rs
```diff
@@ -350,6 +355,33 @@ fn init() -> App<'static, 'static> {
...
+        .about("About peer id, base on Secp256k1")
+        .subcommand(
+            SubCommand::with_name(CMD_FROM_SECRET)
+                .about("Generate peer id from secret file")
+                .arg(
+                    Arg::with_name(ARG_SECRET_PATH)
+                        .takes_value(true)
+                        .long(ARG_SECRET_PATH)
+                        .required(true)
+                        .help("Generate peer id from privkey file path"),
```
quake, line 369: inconsistent use of `secret file` in about and `privkey file` in help

quake:
I prefer to use stdin/stdout for these helper functions:
```
./ckb peer-id gen > key
cat key | ./ckb peer-id show
```

yangby-cryptape:
> I prefer to use stdin/stdout for these helper functions:
> 
> ```
> ./ckb peer-id gen > key
> cat key | ./ckb peer-id show
> ```

When the output is binary,  usually the default method is output to a file, for example, gzip, bzip2, ssh-key and so on. It's a protection behavior.
Otherwise, after people try `./ckb peer-id gen` before they read the manual, they have to reset their terminal. For some terminal tools, they have to restart their terminal tools (both tmux and screen has bug when stdout is binary data). They may lost some data which on terminal before run `./ckb peer-id gen`.

driftluo:
bors r=quake,yangby-cryptape

PR #2123, opened 2020-06-16
chore: split json rpc type `Node` to `LocalNode` and `RemoteNode`

quake:
We are planing to add more fields to network related rpc `local_node_info` and `get_peers`, for example, add uptime to `local_node_info`, add best_known_header to `get_peers`, this PR split json rpc type `Node` to `LocalNode` and `RemoteNode`, make it easier to add different fields to these rpc in the future.

This PR is compatible with previous version, only one field `is_outbound` is removed from `local_node_info`, which is always null in old rpc.

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -1866,7 +1866,6 @@ http://localhost:8114
                 "score": "0x1"
             }
         ],
-        "is_outbound": null,
         "node_id": "QmTRHCdrRtgUzYLNCin69zEvPvLYdxUZLLfLYyHVY3DZAS",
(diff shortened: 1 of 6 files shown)


quake:
> * [ ]  Please add BREAKING CHANGE in commit message and PR description
> * [ ]  Please explain how RPC will change in both commit message and PR description

I think this PR is compatible with previous version, only one field `is_outbound` is removed from `local_node_info`, which is always null in old rpc

keroro520:
bors r=doitian,keroro520

PR #2126, opened 2020-06-16
perf: remove fee estimator

zhangsoledad:
This `estimate_fee_rate` RPC is experimental, due to availability and performance issues, we decide to remove it.


doitian:
bors r=quake,driftluo

PR #2131, opened 2020-06-18
chore: upgrade tentacle/tokio

driftluo:
tentacle:
- Performance optimization
- Split error type

tokio：
- Fix task budgeting with `block_in_place`
- Fix notify when resetting a Delay to a time in the past

### DIFF
--- a/network/src/errors.rs
+++ b/network/src/errors.rs
@@ -1,5 +1,12 @@
 use crate::ProtocolId;
-use p2p::{error::Error as P2PError, secio::PeerId, SessionId};
+use p2p::{
+    error::{
+        DialerErrorKind, ListenErrorKind, ProtocolHandleErrorKind, SendErrorKind,
+        TransportErrorKind,
+    },
+    secio::PeerId,
+    SessionId,
+};
 use std::fmt;
 use std::fmt::Display;
 use std::io::Error as IoError;
@@ -16,6 +23,15 @@ pub enum Error {
     Dial(String),
     PeerStore(PeerStoreError),
     Shutdown,
+}
+
+#[derive(Debug)]
+pub enum P2PError {
+    Transport(TransportErrorKind),
+    Protocol(ProtocolHandleErrorKind),
+    Dail(DialerErrorKind),
+    Listen(ListenErrorKind),
+    Send(SendErrorKind),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


network/src/errors.rs
```diff
@@ -18,6 +25,15 @@ pub enum Error {
     Shutdown,
 }
 
+#[derive(Debug)]
+pub enum P2PError {
```
doitian, line 29: Can we just use the `P2PError` in `tentacle`?

network/src/network.rs
```diff
@@ -579,18 +582,12 @@ impl ServiceHandle for EventHandler {
...
-                        id,
-                        Duration::from_secs(300),
-                        message,
-                    );
-                } else if let Err(err) =
-                    disconnect_with_message(context.control(), id, message.as_str())
-                {
-                    debug!("Disconnect failed {:?}, error {:?}", id, err);
-                }
+                self.network_state.ban_session(
```
doitian, line 585: It's hard to tell why is `if else` is removed in this PR. Could you please give more context on this?

network/src/network.rs
```diff
@@ -1114,7 +1111,9 @@ impl NetworkService {
                                     addr.clone(),
                                     err
                                 );
-                                start_sender.send(Err(Error::Io(err))).unwrap();
+                                start_sender
+                                    .send(Err(Error::P2P(P2PError::Transport(err))))
+                                    .unwrap();
```
doitian, line 1116: Make it better when touching the code. Please use `expect` instead of `unwrap` here.

network/src/network.rs
driftluo: tentacle splits the error, the output is more detailed, only one error will be output here, which is previous `if let` error.

This is the reason why a big Error output is not used

network/src/errors.rs
driftluo: see https://github.com/nervosnetwork/ckb/pull/2131#discussion_r442120466

doitian: Got it. I though there was an error enum like this in tentacle but I'm wrong.

@quake What's your opinion? Should P2P define this error enum even it self does not use it in any function?

quake: I prefer to let the lib user define the error combinations

doitian:
bors r=quake,doitian

PR #2134, opened 2020-06-22
fix: update proposal table after chain reorg

zhangsoledad:
Previously, proposal-table update not considered in chain rollback, it's almost impossible to happen in hashrate-based chain. But can be triggered by `truncate` RPC.

Proposal fix: 
* check whether need reload proposal_table from block when the fork occurs.
* make sure fork blocks is sorted

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -465,6 +465,42 @@ impl ChainService {
             self.proposal_table
                 .insert(blk.header().number(), blk.union_proposal_ids());
         }
+
+        self.reload_proposal_table(&fork);
+    }
+
+    // if rollback happen, go back check whether need reload proposal_table from block
+    pub(crate) fn reload_proposal_table(&mut self, fork: &ForkChanges) {
+        if fork.has_detached() {
+            let proposal_window = self.shared.consensus().tx_proposal_window();
+            let detached_front = fork
+                .detached_blocks()
+                .front()
+                .map(|blk| blk.header().number())
+                .expect("detached_blocks is not empty");
+            if detached_front < 2 {
+                return;
+            }
+            let common = detached_front - 1;
+            let new_tip = fork
+                .attached_blocks()
+                .back()
+                .map(|blk| blk.header().number())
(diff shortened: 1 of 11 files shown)


keroro520:
image

The modified version above will fail.

chain/src/chain.rs
```diff
@@ -465,6 +465,43 @@ impl ChainService {
...
                 .insert(blk.header().number(), blk.union_proposal_ids());
         }
+
+        self.reload_proposal_table(&fork);
+    }
+
+    // if rollback happen, go back check whether need reload proposal_table from block
+    pub(crate) fn reload_proposal_table(&mut self, fork: &ForkChanges) {
+        if fork.has_detached() {
+            println!("reload_proposal_table");
```
quake, line 475: remove println

quake, line 475: please remove println

doitian:
bors r=quake,keroro520

chain/src/chain.rs
```diff
@@ -99,6 +99,15 @@ impl ForkChanges {
     pub fn verified_len(&self) -> usize {
         self.attached_blocks.len() - self.dirty_exts.len()
     }
+
+    pub fn is_sorted(&self) -> bool {
```
doitian, line 103: This is inefficient. If it's only used in `debug_assert`, please guard it using `cfg!(debug_assertions)`.

doitian: Alternative, we can just copy the logic from nightly rust

https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#3205

zhangsoledad: https://github.com/nervosnetwork/ckb/pull/2134/commits/aa8b759209340376efd5f791be43ea7c7888fabb
I bring `rust unstable port` back, introduce `https://github.com/gnzlbg/is_sorted`.

doitian:
bors r=quake,doitian

doitian:
bors retry

quake:
bors r=quake,keroro520

PR #2135, opened 2020-06-23
test: construct valid uncle blocks no matter how well the host is

yangby-cryptape:
### Issue

Integration tests are randomly failed at spec [pack_uncles_into_epoch_starting](https://github.com/nervosnetwork/ckb/blob/199c643b16f78f0915e8a0ca51f553a4e0401d1b/test/src/specs/mining/uncle.rs#L133-L169):

> ERROR ckb_rpc::module::miner  [] submit_block error: Error { kind: HeaderError { kind: BlockTimeTooOld { min: N, actual: N }

### How to Reproduce

- You need a **high-performance** machine.

- Build ckb  in release mode.
  ```bash
  cargo build --release --features deadlock_detection
  ```
- Build ckb-test in release mode.
  ```bash
  cd test; cargo build --release
  ```
- Run tests:
  ```bash
  while true; do sleep 1; target/release/ckb-test --bin ../target/release/ckb --port 9000 pack_uncles_into_epoch_starting && rm -rf /tmp/ckb-it-* || break; done
  ```
- Leave your machine alone and go to drink a cup of tea or coffee.

### Reason

- When your machine is excellent, `block.template.timestamp = tip.timestamp() + 1` since:
  https://github.com/nervosnetwork/ckb/blob/199c643b16f78f0915e8a0ca51f553a4e0401d1b/tx-pool/src/process.rs#L213

- When construct uncles, the `uncle.timestamp = tip.timestamp() - 1` since 
 https://github.com/nervosnetwork/ckb/blob/199c643b16f78f0915e8a0ca51f553a4e0401d1b/test/src/node.rs#L290-L296

- So `block_median_time` will return the same timestamp as the uncle (in this spec, there always be two timestamps, and the first is `0`, because it's the genesis block):
https://github.com/nervosnetwork/ckb/blob/199c643b16f78f0915e8a0ca51f553a4e0401d1b/traits/src/block_median_time_context.rs#L35

- Then this check will return an error:
  https://github.com/nervosnetwork/ckb/blob/199c643b16f78f0915e8a0ca51f553a4e0401d1b/verification/src/header_verifier.rs#L94-L103

### DIFF
--- a/test/src/node.rs
+++ b/test/src/node.rs
@@ -293,7 +293,17 @@ impl Node {
         // is less than the current time, and then generate
         // the new block in main fork which timestamp is greater than
         // or equal to the current time.
-        let timestamp = block.timestamp() - 1;
+        let timestamp = block.timestamp();
+        loop {
+            let timestamp_next: u64 = self
+                .rpc_client()
+                .get_block_template(None, None, None)
+                .current_time
+                .into();
+            if timestamp_next > timestamp {
+                break;
+            }
+        }
         block
             .as_advanced_builder()
             .timestamp(timestamp.pack())


yangby-cryptape:
bors r=keroro520,quake

yangby-cryptape:
bors r=keroro520,quake

PR #2136, opened 2020-06-23
feat: add RPC `clear_tx_pool`

keroro520:
The adding RPC `cleat_tx_pool` removes all the memoried transactions from tx-pool.


doitian:
bors r=quake,doitian

PR #2137, opened 2020-06-23
chore: add memory leak fixing in CHANGELOG

doitian:


yangby-cryptape:
bors r=quake,yangby-cryptape

PR #2139, opened 2020-06-24
chore: add make script for Windows PowerShell

doitian:
The script uses PowerShell and can run tasks similar to the targets defined in Makefile, thus we don't need to install make and other bash command line utilities.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,13 +37,6 @@ integration: submodule-init setup-ckb-test ## Run integration tests in "test" di
 	cargo build --features deadlock_detection
 	RUST_BACKTRACE=1 RUST_LOG=${INTEGRATION_RUST_LOG} test/run.sh -- --bin ../target/debug/ckb ${CKB_TEST_ARGS}
 
-.PHONY: integration-windows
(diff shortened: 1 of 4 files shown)


doitian:
> Replace the follow line to check if this script works well in CI:
> https://github.com/nervosnetwork/ckb/blob/3a29d26b94e49d018b3ebb63dbf0a06d246ad768/azure-pipelines.yml#L67

Let me try whether the default bash script can call ps1 script

doitian:
> > Replace the follow line to check if this script works well in CI:
> > https://github.com/nervosnetwork/ckb/blob/3a29d26b94e49d018b3ebb63dbf0a06d246ad768/azure-pipelines.yml#L67
> 
> Let me try whether the default bash script can call ps1 script

The default script uses `cmd`. I have changed to `powershell`.

Now azure pipelines only use `make.ps1`  to run tasks.

doitian:
I'll recreate the PR to bypass the CI

doitian:
https://github.com/nervosnetwork/ckb/pull/2142

PR #2140, opened 2020-06-26
fix: shrink state map

zhangsoledad:
Cause rust hash table capacity does not shrink automatically, we need explicit call `shrink` for predictable limit memory usage.

### DIFF
--- a/sync/src/orphan_block_pool.rs
+++ b/sync/src/orphan_block_pool.rs
@@ -1,8 +1,10 @@
 use ckb_types::{core, packed};
+use ckb_util::shrink_to_fit;
 use ckb_util::RwLock;
 use std::collections::{HashMap, VecDeque};
 
 pub type ParentHash = packed::Byte32;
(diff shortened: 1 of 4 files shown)


keroro520:
Why not periodically shrink based [notify mechanism]( https://github.com/nervosnetwork/ckb/blob/56709f7c364c5571cf17a7f5e372a954a34a1f57/sync/src/relayer/mod.rs#L730)?

zhangsoledad:
> Why not periodically shrink based [notify mechanism](https://github.com/nervosnetwork/ckb/blob/56709f7c364c5571cf17a7f5e372a954a34a1f57/sync/src/relayer/mod.rs#L730)?

Because it's bad design, different map should have different shrink rates according to occupied memory, shrink rate should be configurable, this also attempts on smooth out potential issues incurred by simultaneously accessing multiple locks.

quake:
bors r=quake,keroro520,driftluo,yangby-cryptape

PR #2142, opened 2020-06-29
chore: add make script for Windows PowerShell

doitian:
Add a PowerShell script for Windows so developers can perform the tasks without installing too many dependencies.


doitian:
bors r=quake,yangby-cryptape

PR #2145, opened 2020-07-01
improvement: don't cache all block status in memory

yangby-cryptape:
When a node is start from number 0, it will sync all headers at first, then the `block_status_map` will be full quickly.
Then, along with the block sync, all data in `block_status_map` will be removed.
When the IBD is done, there will be nothing left in `block_status_map`.

But when another new-started node try to sync data from this node, this node will fetch all block statuses from database and insert them into `block_status_map` without deletions. And full block status will store in memory until the node is shutdown.
https://github.com/nervosnetwork/ckb/blob/b2b2276478619b1ecd6d9eeefb3186013c430b9a/sync/src/synchronizer/get_blocks_process.rs#L46
https://github.com/nervosnetwork/ckb/blob/b2b2276478619b1ecd6d9eeefb3186013c430b9a/sync/src/types.rs#L1752-L1755

Up to now, there all 2,222,026 blocks in mainnet.
Each key in `block_status_map` has [4 pointer](https://docs.rs/bytes/0.5.5/bytes/struct.Bytes.html#sharing) and 32 bytes, and each value is `u32`, so it costs about 150 MiB at least (7 months data).

It's not a problem in the near term. Just save more memory.

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -1737,8 +1737,7 @@ impl ActiveChain {
     }
 
     pub fn get_block_status(&self, block_hash: &Byte32) -> BlockStatus {
-        let mut locked = self.state.block_status_map.lock();
-        match locked.get(block_hash).cloned() {
+        match self.state.block_status_map.lock().get(block_hash).cloned() {
             Some(status) => status,
             None => {
                 let verified = self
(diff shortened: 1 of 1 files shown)


yangby-cryptape:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 342.26
  - Samples Count: 50
  - CKB Version: b75dbb97f5f28cee83be6c11c7229022e4b7771c
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

quake:
bors r=quake,keroro520,zhangsoledad

PR #2147, opened 2020-07-01
improvement: don't cache all data of header map in memory during IBD

yangby-cryptape:
- Open a temporary rocksdb (default is `${ckb_data_dir}/tmp/ckb-tmp-*`) to store a part data of header map which exceeded the configured quota.

- Why not use the same rocksdb as block data?

  - It is just temporary data, all data was destroyed when ckb shutdown, so the format of data can be changed in any commits, we don't have to do migrations on it.

  - I don't want to create columns, it requires migrations. Also too many columns will affect performance. 

  - If I use current existed columns, then traverse all data will be difficult.

    At present, no data in chain database have key-prefix.
    For example, if we use key `header-${header_hash}` to save headers, key `uncle-${uncle_hash}` to save uncles,  key `block-${block_hash}` to save blocks, then we can save them in a same column and we can traverse each of them use the key prefix (`header-*`, `uncle-*`, `block-*`) easily.
    But we save almost all data just via `${hash}`, we can't know the type of data if we put them in a same column.

    So I can't use current existed columns without breaking the data traversing.

  - After IBD, there is about 450 MiB for current mainnet which height is about 2_200_000 with default rocksdb configuration (we can optimize the configuration later), but the size of header map is very small (in most of the time).
    So we can just delete the entire rocksdb, we don't care about how to compress data and free space for rocksdb.

- I use two thresholds to control the balance of memory and disk:
  - The count of headers in memory is less than or equal to `primary_limit`.
  - If the rocksdb will be closed when the count header of headers in memory is less than `primary_limit - backend_close_threshold` and there is no headers in rocksdb.

- I use a lot of `panic`, same as what we did in `ckb-store`. I trust the rocksdb and assume it always works well.

- I wrote two simple serialization and deserialization functions, because I think it is a private data structure, there is no need to make it too public (move to `ckb-types`) or bring heavy serialization and deserialization dependencies into this crate.

- Enable the `stats` feature to collect the statistics, the result of syncing mainnet from `0` to `2247420` is :
  ```
  2020-07-06 15:53:32.875 +08:00 NetworkRuntime TRACE ckb_sync::types::header_map::kernel_lru  Header Map Statistics
  >       | storage | length  |  limit  | contain |   select   | insert  | delete  |
  >       |---------+---------+---------+---------+------------+---------+---------|
  >       | primary |     4974|   300000|  2252392|   196946199|  4243176|  4238382|
  >       | backend |        0|    false|        0|           0|  1990779|  2457795|
  ```

### DIFF
--- a/sync/src/types/header_map/kernel_lru.rs
+++ b/sync/src/types/header_map/kernel_lru.rs
@@ -1,0 +1,275 @@
+use std::path;
+
+#[cfg(feature = "stats")]
+use ckb_logger::trace;
+use ckb_types::packed::Byte32;
+
+use super::{KeyValueBackend, KeyValueMemory};
+use crate::types::HeaderView;
+
+pub(crate) struct HeaderMapLruKernel<Backend>
+where
+    Backend: KeyValueBackend,
+{
+    primary: KeyValueMemory<Byte32, HeaderView>,
+    backend: Backend,
+    // Configuration
+    primary_limit: usize,
+    backend_close_threshold: usize,
+    // Statistics
+    #[cfg(feature = "stats")]
+    stats: HeaderMapLruKernelStats,
+}
+
+#[cfg(feature = "stats")]
+#[derive(Default)]
+struct HeaderMapLruKernelStats {
+    frequency: usize,
+
+    trace_progress: usize,
+
+    primary_contain: usize,
+    primary_select: usize,
+    primary_insert: usize,
+    primary_delete: usize,
+
+    backend_contain: usize,
+    backend_insert: usize,
+    backend_delete: usize,
+}
+
+impl<Backend> HeaderMapLruKernel<Backend>
+where
+    Backend: KeyValueBackend,
+{
+    pub(crate) fn new<P>(
+        tmpdir: Option<P>,
+        primary_limit: usize,
+        backend_close_threshold: usize,
+    ) -> Self
+    where
+        P: AsRef<path::Path>,
+    {
+        let primary = Default::default();
+        let backend = Backend::new(tmpdir);
+
+        #[cfg(not(feature = "stats"))]
+        {
+            Self {
+                primary,
+                backend,
+                primary_limit,
+                backend_close_threshold,
+            }
+        }
+
+        #[cfg(feature = "stats")]
+        {
+            Self {
+                primary,
+                backend,
+                primary_limit,
+                backend_close_threshold,
+                stats: HeaderMapLruKernelStats::new(50_000),
+            }
+        }
+    }
+
+    pub(crate) fn contains_key(&mut self, hash: &Byte32) -> bool {
+        #[cfg(feature = "stats")]
+        {
+            self.mut_stats().tick_primary_contain();
+        }
+        if self.primary.contains_key(hash) {
+            return true;
+        }
+        if !self.backend.is_opened() {
+            return false;
+        }
+        #[cfg(feature = "stats")]
+        {
+            self.mut_stats().tick_backend_contain();
+        }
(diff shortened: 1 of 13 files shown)


yangby-cryptape:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 350.48
  - Samples Count: 51
  - CKB Version: 63e1cc88c32fa3989e1f9d7ad526f713956afa31
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

doitian:
It's better to default to the CKB data directory. Also expose the default in the generated config file.

yangby-cryptape:
benchmark

yangby-cryptape:
@quake I re-push this PR and update the description, I try several strategies and compare the statistics, this is the easiest strategy, my original strategy is too complex but only a little benefit.

nervos-bot-user:
#### Benchmark Result
  - TPS: 347.23
  - Samples Count: 51
  - CKB Version: d4c0ba58cb991fc842b4cde66af7d397fa15b767
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

sync/src/types/header_map/kernel_lru.rs
```diff
@@ -0,0 +1,275 @@
...
+                primary,
+                backend,
+                primary_limit,
+                backend_close_threshold,
+                stats: HeaderMapLruKernelStats::new(50_000),
+            }
+        }
+    }
+
+    pub(crate) fn has(&mut self, hash: &Byte32) -> bool {
```
quake, line 78: suggest to follow naming convention
```
has => contains_key / add => insert / rm => remove
```

yangby-cryptape: > suggest to follow naming convention
> 
> ```
> has => contains_key / add => insert / rm => remove
> ```

To avoid typo: use `map.inner.insert` instead of `map.insert`.
If they have different names, if we use the wrong method, it's a compile error.

```rust
fn insert(&mut self, value: HeaderView) -> Option<HeaderView>;
pub(crate) fn add(&mut self, view: HeaderView) -> Option<HeaderView>;
```

I could accept your suggestion only if these structs in different crates and we don't have chance to make mistakes.

yangby-cryptape: DONE and CI passed!

sync/src/types/mod.rs
```diff
@@ -947,6 +952,73 @@ impl HeaderView {
...
+    fn to_vec(&self) -> Vec<u8> {
+        let mut v = Vec::new();
+        let inner: packed::HeaderView = self.inner.pack();
+        let total_difficulty: packed::Uint256 = self.total_difficulty.pack();
+        let skip_hash: packed::Byte32Opt = Pack::pack(&self.skip_hash);
+        let inner_len: packed::Uint32 = (inner.as_slice().len() as u32).pack();
+        let total_difficulty_len: packed::Uint32 =
+            (total_difficulty.as_slice().len() as u32).pack();
+        let skip_hash_len: packed::Uint32 = (skip_hash.as_slice().len() as u32).pack();
+        v.extend_from_slice(inner_len.as_slice());
```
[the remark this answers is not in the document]
yangby-cryptape: DONE and CI passed!

doitian:
@yangby-cryptape please resolve the conflicts

yangby-cryptape:
@quake @driftluo @doitian Rebased develop onto this PR. No more updates.

doitian:
bors r=quake,driftluo

PR #1927, opened 2020-02-07
feat: trace memory for some data structures

yangby-cryptape:
### Features

- There are many data structures (mostly `HashMap`) to store states in memory. Some of them have no size limit.
  This PR add a special log target to trace the size of these data structures, so that we can analyze them by running performance test.
- This PR is base on #1940. The first three commits are same as the commits in #1904.
  [**Only the last commit is a new commit.**](https://github.com/nervosnetwork/ckb/pull/1927/commits/63f72c6fd96fe5add39763b24de1c7394edf253f)
  So, please review #1904 at first, then review the last commit in this PR.

### Usage

This function will be enabled when compile the executable with feature `measure-collections`.

PR #2148, opened 2020-07-02
chore: trace memory for some data structures

yangby-cryptape:
### Brief

Rebase from the closed #1927, but remove the feature `measure-collections`.
So, **the operations of tracing those data structures will always exist**.

But if you set `[memory_tracker] interval = 0`, the statistics will be ignore as soon as possible.

```rust
pub fn measure(tag: Arc<String>, record: MeasureRecord) {
    if crate::interval() > 0 {
        // ... send the record to the collector thread ....
    }
}
```

### Features

- There are many data structures (mostly `HashMap`) to store states in memory. Some of them have no size limit.
  This PR add a logs to trace the size of these data structures, so that we can analyze them.
  I find #2145 with this function.

- **[NEW!]** Before this PR, the whole memory tracker function was disabled in Windows and MacOS.
  In this PR, I split those statistics into four parts: process (1), jemalloc (2), rocksdb (3), collections (4).
  In Linux , all of them are enabled, otherwise only 1, 3 and 4 enabled.

- Can trace `HeaderMap` in #2147 via the follow struct:
  ```rust
  MeasureRecord::SpooledHashMapWithRocksDB {
      mem_len: usize,
      mem_cap: usize,
      disk_cnt: usize,
      db_stats: RocksDBMemoryStatistics,
  },
  ```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -617,10 +617,13 @@ version = "0.34.0-pre"
 dependencies = [
  "ckb-db",
  "ckb-logger",
+ "ckb-util",
+ "crossbeam-channel",
  "futures 0.3.4",
  "heim",
  "jemalloc-ctl",
  "jemalloc-sys",
+ "lazy_static",
 ]
 
 [[package]]
@@ -963,6 +966,7 @@ dependencies = [
  "ckb-error",
  "ckb-fee-estimator",
  "ckb-logger",
+ "ckb-memory-tracker",
  "ckb-network",
  "ckb-shared",
  "ckb-store",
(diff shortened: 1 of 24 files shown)


yangby-cryptape:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 340.47
  - Samples Count: 51
  - CKB Version: 6348fa4d0709390a47aecd6dca5a0c604b35f7b0
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

PR #2150, opened 2020-07-03
improvement: don't query store twice in method chaining

yangby-cryptape:
- `fn_get_pending_header` always try call `get_header_view`:
https://github.com/nervosnetwork/ckb/blob/6678f812fd3f3263010298388f73cafd2e92590a/sync/src/relayer/compact_block_process.rs#L153-L164
- `get_header_view` always try call `get_block_header`
  https://github.com/nervosnetwork/ckb/blob/6678f812fd3f3263010298388f73cafd2e92590a/sync/src/types.rs#L1170-L1179
- Always try call `get_block_header` after `fn_get_pending_header`: 
 https://github.com/nervosnetwork/ckb/blob/6678f812fd3f3263010298388f73cafd2e92590a/sync/src/relayer/compact_block_process.rs#L305-L307


quake:
bors r=quake,TheWaWaR

PR #2151, opened 2020-07-03
improvement: reduce times of querying header map

yangby-cryptape:
I added some log outputs when do tests, I got the follow results:

- Sync chain from height `0` to `tip`:
  - If always query the header map first.
    - Query the header map first but not found: about `tip/2 ~ tip` times (mark as `x`)
    - If not found in the header map, it must be in the store.
  - If query header from store first when the block in that height was already inserted into the chain database.
    - Query the header map first but not found: about `tip/4 ~ tip/6` times (mark as `y`)
    - Query the store first but not found: `0`
- The result of my last test: `tip = 2_230_953; x = 1_179_787; y = 552_325`

### DIFF
--- a/sync/src/types.rs
+++ b/sync/src/types.rs
@@ -870,14 +870,20 @@ impl HeaderView {
         self.inner
     }
 
-    pub fn build_skip<F, G>(&mut self, mut get_header_view: F, fast_scanner: G)
-    where
-        F: FnMut(&Byte32) -> Option<HeaderView>,
+    pub fn build_skip<F, G>(
+        &mut self,
+        tip_number: BlockNumber,
+        mut get_header_view: F,
+        fast_scanner: G,
+    ) where
+        F: FnMut(&Byte32, Option<bool>) -> Option<HeaderView>,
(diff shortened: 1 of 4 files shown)


sync/src/types.rs
```diff
@@ -1167,15 +1178,30 @@ impl SyncShared {
         self.state.may_set_shared_best_header(header_view);
     }
 
-    pub fn get_header_view(&self, hash: &Byte32) -> Option<HeaderView> {
+    pub fn get_header_view(
+        &self,
+        hash: &Byte32,
+        store_first_opt: Option<bool>,
```
quake, line 1184: is it possible to use `BlockStatus` here?

```
BlockStatus::UNKNOWN => None,
BlockStatus::STORED => Some(true),
others => Some(false),
```

yangby-cryptape: This assume that `tip.number() >= header.number()` => `BlockStatus::STORED`, I'm not sure about that.
I just search store first when `tip.number() >= header.number()`, but not mean `BlockStatus::STORED`.

quake:
bors r=quake,TheWaWaR,driftluo

PR #2152, opened 2020-07-05
pref: change storage molecule table to struct

quake:
storage structs `HeaderView`, `EpochExt` and `TransactionInfo` are fixed size, we should use molecule `struct` instead of `table`, it reduces storage size and improves the performance a little bit.

run a simple test with `ckb replay --profile 1 20000 --tmp-target /tmp` on my local pc:
```
before: avg 37.2 sec
after: avg 36.1 sec
```

f.y.i, this migration took 15 ~ 20 sec to finish on my local pc with current tip number, I think it's acceptable.

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -1,3 +1,4 @@
+mod migrations;
 pub mod shared;
 
 pub use ckb_snapshot::{Snapshot, SnapshotMgr};
--- a/shared/src/migrations.rs
+++ b/shared/src/migrations.rs
@@ -1,0 +1,54 @@
+use ckb_db::{Migration, Result, RocksDB};
+use ckb_store::{
+    COLUMN_BLOCK_HEADER, COLUMN_EPOCH, COLUMN_META, COLUMN_TRANSACTION_INFO, COLUMN_UNCLES,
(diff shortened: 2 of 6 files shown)


quake:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 337.60
  - Samples Count: 51
  - CKB Version: 8cdb577124380bb7d9e3f4aae5923c47ac474bf5
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

quake:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 346.85
  - Samples Count: 51
  - CKB Version: 54f25843514502c61daa25daed97f59358ffcec1
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

doitian:
It will break tools that directly depend on database schema. Cc @xxuejie

xxuejie:
For my case it will be okay, since I'm not using the schema directly. I'm only using `ckb-storage` package, which should be good after this PR.

PR #2153, opened 2020-07-06
refactor: decoupling migration from db

zhangsoledad:
Previously, migration coupling with DB, this sacrifice flexibility. In a case like this, open a read-only DB will be trouble.
this PR proposal split migration.


db/src/db.rs
```diff
@@ -97,15 +88,26 @@ impl RocksDB {
...
-            .unwrap_or_else(|err| panic!("{}", err))
+        Self::open_with_check(&config, columns).unwrap_or_else(|err| panic!("{}", err))
     }
 
     pub fn get_pinned(&self, col: Col, key: &[u8]) -> Result<Option<DBPinnableSlice>> {
         let cf = cf_handle(&self.inner, col)?;
         self.inner.get_pinned_cf(cf, &key).map_err(internal_error)
     }
 
+    pub fn get_pinned_defalt(&self, key: &[u8]) -> Result<Option<DBPinnableSlice>> {
```
quake, line 99: typo?
```suggestion
    pub fn get_pinned_default(&self, key: &[u8]) -> Result<Option<DBPinnableSlice>> {
```

zhangsoledad: @quake  nice catch!

zhangsoledad:
bors r=quake,driftluo

PR #2154, opened 2020-07-06
refactor: split chain iter

zhangsoledad:


yangby-cryptape:
bors r=quake,yangby-cryptape

PR #2158, opened 2020-07-10
fix: panic if db options is empty

yangby-cryptape:
Append `[db.options]` to `ckb.toml` and don't add any options, then `ckb run -C .` will panic since:
- In `rust-rocksdb`:
https://github.com/nervosnetwork/ckb/blob/32042c81f1ec4fdaf7358b63dc2417d9363d9944/db/src/db.rs#L72-L74
- In [`rocksdb::db/c.cc::rocksdb_set_options(..)`](https://github.com/facebook/rocksdb/blob/a9a973869afd987a6b728b833b2896dbb38bc02f/db/c.cc#L2216-L2217)
  ```c++
          SaveError(errptr,
              db->rep->SetOptions(options_map));
  ```
- In [`rocksdb::db/db_impl/db_impl.cc::SetOptions(..)`](
https://github.com/facebook/rocksdb/blob/a9a973869afd987a6b728b833b2896dbb38bc02f/db/db_impl/db_impl.cc#L951-L956)
  ```c++
    if (options_map.empty()) {
      ROCKS_LOG_WARN(immutable_db_options_.info_log,
                     "SetOptions() on column family [%s], empty input",
                     cfd->GetName().c_str());
      return Status::InvalidArgument("empty input");
    }
  ```

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -64,8 +64,9 @@ impl RocksDB {
                 }
             })?;
 
-        if let Some(db_opt) = config.options.as_ref() {
-            let rocksdb_options: Vec<(&str, &str)> = db_opt
+        if !config.options.is_empty() {
+            let rocksdb_options: Vec<(&str, &str)> = config
+                .options
                 .iter()
                 .map(|(k, v)| (k.as_str(), v.as_str()))
                 .collect();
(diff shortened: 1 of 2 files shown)


quake:
bors r=quake,zhangsoledad

PR #2161, opened 2020-07-12
chore: remove unused fn from `DataLoader` trait

quake:
`get_block_ext` fn is unused and epoch related information should be loaded via header's epoch field, cleanup code and tweak unit test.

### DIFF
--- a/script/src/syscalls/load_header.rs
+++ b/script/src/syscalls/load_header.rs
@@ -114,20 +114,18 @@ impl<'a, DL: DataLoader + 'a> LoadHeader<'a, DL> {
         header: &HeaderView,
     ) -> Result<(u8, u64), VMError> {
         let field = HeaderField::parse_from_u64(machine.registers()[A5].to_u64())?;
-        let epoch = match self.data_loader.get_block_epoch(&header.hash()) {
-            Some(epoch) => epoch,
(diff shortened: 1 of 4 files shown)


script/src/syscalls/load_header.rs
```diff
@@ -114,16 +114,14 @@ impl<'a, DL: DataLoader + 'a> LoadHeader<'a, DL> {
...
+        let epoch = header.epoch();
 
         let result = match field {
             HeaderField::EpochNumber => (SUCCESS, store_u64(machine, epoch.number())?),
-            HeaderField::EpochStartBlockNumber => {
-                (SUCCESS, store_u64(machine, epoch.start_number())?)
-            }
+            HeaderField::EpochStartBlockNumber => (
+                SUCCESS,
+                store_u64(machine, header.number() - epoch.index())?,
```
zhangsoledad, line 123: I suggest using `overflowing_sub` here, even those data are verified in `ckb`, but syscalls may be used out of ckb and also more friendly for mock return an error rather than panic.

quake: Resolved, returns a `VMError::Unexpected` here like other syscall, please review again, thanks.

quake:

PR #2176, opened 2020-07-22
chore: unify logger targets

yangby-cryptape:
- Since `tx-pool` is split from `ckb-shared` after #1567, we can remove all user-specified targets for `tx-pool`.
  https://github.com/nervosnetwork/ckb/blob/dfc944a1ce7be98285cd34a45772e671f1434a6c/shared/src/lib.rs#L20
- After [CKB PR 2078: chore: use fully qualified module name as the default log target](https://github.com/nervosnetwork/ckb/pull/2078), almost all logs use `$crate::env!("CARGO_PKG_NAME").replace("-", "_")` as first part of log targets.


quake:
bors r=quake,driftluo

PR #2177, opened 2020-07-22
fix: fix the lenient logger filter parser

yangby-cryptape:
If using log targets with `-`, we **shouldn't** just replace `-` by `_`, we should keep the original value, too.

For example:
If the **real** log target is `ckb-relay`, log filter is `info,ckb-relay=trace`, the lenient log filter should be `info,ckb-relay=trace,ckb_relay=trace`, not `info,ckb_relay=trace`.


yangby-cryptape:
bors r=quake,TheWaWaR

PR #2178, opened 2020-07-23
feat: add `generate_block` rpc to IntegrationTest module

quake:
allows user to generate block through rpc, it's a convenient feature for dApp integration test (like `truncate` rpc)

```shell
# generate a block with default block assembler config in ckb.toml
curl -d '{"id": 1, "jsonrpc": "2.0", "method":"generate_block","params": []}' \
  -H 'content-type:application/json' 'http://localhost:8114'

# generate a block with different block assember config
curl -d '{"id": 1, "jsonrpc": "2.0", "method":"generate_block","params": [{"code_hash": "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8", "args": "0xc8328aabcd9b9e8e64fbc566c4385c3bdeb219d7", "hash_type": "type"}]}' \
  -H 'content-type:application/json' 'http://localhost:8114'
```

### DIFF
--- a/rpc/src/module/test.rs
+++ b/rpc/src/module/test.rs
@@ -1,6 +1,7 @@
 use crate::error::RPCError;
+use ckb_app_config::BlockAssemblerConfig;
 use ckb_chain::{chain::ChainController, switch::Switch};
-use ckb_jsonrpc_types::{Block, BlockView, Cycle, Transaction};
+use ckb_jsonrpc_types::{Block, BlockView, Cycle, JsonBytes, Script, Transaction};
 use ckb_logger::error;
 use ckb_network::{NetworkController, SupportProtocols};
 use ckb_shared::shared::Shared;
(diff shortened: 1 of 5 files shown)


chuijiaolianying:
Can this RPC divide into `generate_block` and `insert_and_announce_block`? I'm not sure if dAPP test has this condition, in CKB integration test, there often only create a new block but will later or never submit it.

quake:
> Can this RPC divide into `generate_block` and `insert_and_announce_block`? I'm not sure if dAPP test has this condition, in CKB integration test, there often only create a new block but will later or never submit it.

dApp may use existing `get_block_template` and `submit_block` rpc and I think the requirement of get a block template and modify it is very rare.

chuijiaolianying:
> dApp may use existing `get_block_template` and `submit_block` rpc and I think the requirement of get a block template and modify it in dApp integration test is very rare.

I see, looks good to me

rpc/src/module/test.rs
```diff
@@ -111,6 +119,49 @@ impl IntegrationTestRpc for IntegrationTestRpcImpl {
...
+        let block_view = Arc::new(block.into_view());
+        let content = packed::CompactBlock::build_from_block(&block_view, &HashSet::new());
+        let message = packed::RelayMessage::new_builder().set(content).build();
+
+        // insert block to chain
+        self.chain
+            .process_block(Arc::clone(&block_view))
+            .map_err(|err| RPCError::custom(RPCError::CKBInternalError, err.to_string()))?;
+
+        // announce new block
```
doitian, line 154: I think for integration test, it's better to add a option to control whether to broadcast the block, and add a new RPC to just broadcast the current tip. However, it is out of the scope of the current PR.

doitian:
bors r=doitian,zhangsoledad

PR #2179, opened 2020-07-23
refactor: integration service - mining part

chuijiaolianying:
Refactor about integration service mining relate cases:
1. Add comment for explain what does the case do;
2. Update some assertions and add relate info if assert failed;

Besides there also add a new folder `test/src/assertion` which contains some assert functions, a new folder `test/src/util` which contains some functions used in cases setup.


chuijiaolianying:
bors r=quake,keroro520

PR #2180, opened 2020-07-23
refactor: integration test - alert

chuijiaolianying:
Add case description and some assertion for alert_propagation


keroro520:
bors r=quake,keroro520

PR #2175, opened 2020-07-22
feat: support multiple file loggers

yangby-cryptape:
### Features

- Support multiple file loggers in `ckb.toml`.
  For example, the follow configuration will write all `info` logs into `run.log` and `stdout`, all `trace` logs into `trace.log` and `ckb_chain=info` logs into `chain_info.log`:
  ```toml
  [logger]
  filter = "info"
  color = true
  log_to_file = true
  log_to_stdout = true
  [logger.extra.trace]
  filter = "trace"
  [logger.extra.chain_info]
  filter = "off,ckb_chain=info"
  ```
- Support configuring extra file loggers via RPC (in module `Debug`).
  ```shell
  # Add or update a file logger to write `ckb_chain=info` logs into `chain_info.log`.
  curl -H 'content-type: application/json' -d \
      '{ "id": 2, "jsonrpc": "2.0", "method": "set_extra_logger", "params": ["chain_info", {"filter": "off,ckb_chain=info"}] }' \
      http://localhost:8114
  # Remove the file logger `ckb_chain`
  curl -H 'content-type: application/json' -d \
      '{ "id": 2, "jsonrpc": "2.0", "method": "set_extra_logger", "params": ["chain_info", null] }' \
      http://localhost:8114
  ```

PR #2182, opened 2020-07-24
feat: take full control of main logger filter via RPC

yangby-cryptape:
### Issues:
- After #2175 , the name of rpc `set_logger_filter` is not proper (`the main logger` or `a extra logger`), change to `update_main_logger`  (in module `Debug`).
  And let `update_main_logger` take full control of main logger filter, not only filter.
- `panic` when failed to open a log file doesn't output any error messages.
  Use `eprintln!(..)` and `process::exit(..)` instead of `panic`.
- `panic` in logger service thread doesn't stop ckb process, but only stop logger service.
  Just ignore the error logger, keep the logger service running.

### Usages:
```shell
# Disable output logs to file.
curl -H 'content-type: application/json' -d \
  '{ "id": 2, "jsonrpc": "2.0", "method": "update_main_logger", "params": [{"to_file": false}] }' \
  http://localhost:8114 
# Enable output logs to stdout with color.
curl -H 'content-type: application/json' -d \
  '{ "id": 2, "jsonrpc": "2.0", "method": "update_main_logger", "params": [{"to_stdout": false, "color": true}] }' \
  http://localhost:8114 
# Update the filter of main logger.
curl -H 'content-type: application/json' -d \
  '{ "id": 2, "jsonrpc": "2.0", "method": "update_main_logger", "params": [{"filter": "error,ckb-chain=info"}] }' \
  http://localhost:8114 
```

### DIFF
--- a/rpc/src/module/debug.rs
+++ b/rpc/src/module/debug.rs
@@ -1,5 +1,5 @@
-use ckb_jsonrpc_types::ExtraLoggerConfig;
-use ckb_logger::{configure_logger_filter, Logger};
+use ckb_jsonrpc_types::{ExtraLoggerConfig, MainLoggerConfig};
+use ckb_logger::Logger;
 use jsonrpc_core::{Error, ErrorCode::InternalError, Result};
 use jsonrpc_derive::rpc;
 use std::time;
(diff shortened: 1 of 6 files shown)


yangby-cryptape:
bors r=quake,TheWaWaR

PR #2188, opened 2020-07-28
feat: add sync state rpc

driftluo:


rpc/json/rpc.json
```diff
@@ -413,6 +413,22 @@
...
+            "best_known_block_number": "0x248623",
+            "fast_time": "0x3e8",
+            "ibd": false,
+            "inflight_blocks_size": "0x0",
+            "low_time": "0x5dc",
+            "normal_time": "0x4e2",
+            "orphan_block_pool_size": "0x0"
+        },
+        "skip": true
+    },
```
quake, line 431: Please add a `returns` section to describe the returning data fields:
https://github.com/nervosnetwork/ckb/blob/develop/rpc/README.md#returns-1

rpc/json/rpc.json
```diff
@@ -413,6 +413,22 @@
...
+        "description": "Returns sync state of this client",
+        "method": "sync_state",
+        "module": "net",
+        "params": [],
+        "result": {
+            "best_known_block_number": "0x248623",
+            "fast_time": "0x3e8",
+            "ibd": false,
+            "inflight_blocks_size": "0x0",
+            "low_time": "0x5dc",
```
quake, line 426: suggest to add `pending_get_headers_size` field

driftluo: `pending_get_headers` is meaningless data. It’s lru. It will only be cleaned up when the upper limit is reached, and the upper limit will always be maintained.

doitian:
@CipherWang can you assign an engineer in explorer team to review the RPC interface?

rpc/README.md
```diff
@@ -2001,6 +2002,41 @@ http://localhost:8114
...
+| curl -H 'content-type: application/json' -d @- \
+http://localhost:8114
+```
+
+```json
+{
+    "id": 2,
+    "jsonrpc": "2.0",
+    "result": {
+        "best_known_block_number": "0x248623",
```
doitian, line 2029: I think the best known block timestamp is also useful. The client can use the timestamp to check whether the node has synchronized to the recent tip header.

doitian: Another field suggestion: inflight_peers_count

rpc/README.md
```diff
@@ -2001,6 +2002,41 @@ http://localhost:8114
...
+        "fast_time": "0x3e8",
+        "ibd": false,
+        "inflight_blocks_size": "0x0",
```
doitian, line 2032: size of count?

rpc/README.md
driftluo: > Another field suggestion: inflight_peers_count

I don’t think it’s useful, unless you can display all the synced peer information

doitian:
bors r=doitian,quake

quake:
bors r=quake,doitian

doitian:
bors r=quake,doitian

PR #2189, opened 2020-07-28
refactor: integration test - consensus

chuijiaolianying:
Add case description and update case assertions


doitian:
bors r=quake,doitian

PR #2197, opened 2020-07-30
fix: should exit with error code when setup failed

yangby-cryptape:
Issue: if the config was malformed and an error was thrown in `setup_app`, the process wouldn't exit.


doitian:
bors r=quake,doitian

PR #2198, opened 2020-07-30
chore: several minor changes about thread and runtime

yangby-cryptape:
- Simplify dependencies of `ckb-async-runtime`.
  - Use `::std::sync::Barrier` instead of `crossbeam-channel`.
- Let method `ckb-async-runtime::new_runtime` more general.
  - Accept name prefix for threads.
  - Accept `tokio::runtime::Builder` as an optional parameter, so we can set options for the runtime.
- After we upgrade `tokio` from 0.1 to 0.2, the thread name of `tokio` runtime threads has a redundant suffix "-".
  - [`tokio-0.1.x::runtime::Builder::name_prefix(...)`](https://docs.rs/tokio/0.1.22/tokio/runtime/struct.Builder.html#method.name_prefix)
  - [`tokio-0.2.x::runtime::Builder::thread_name(...)`](https://docs.rs/tokio/0.2.22/tokio/runtime/struct.Builder.html#method.thread_name)
- Use a default thread name in logs instead of an empty string.
  Without color, a log record with empty thread name is a bit hard to parse.
- Give name to rayon threads to improve the debuggability slightly.

### DIFF
--- a/util/runtime/src/lib.rs
+++ b/util/runtime/src/lib.rs
@@ -1,29 +1,44 @@
-use std::future::Future;
-use std::thread;
-use tokio::runtime;
-pub use tokio::runtime::Handle;
+use std::{future::Future, sync, thread};
 
-pub fn new_runtime<F, R>(block: F) -> (Handle, thread::JoinHandle<()>)
+pub use tokio::runtime::{Builder, Handle};
+
+pub fn new_runtime<F, R>(
+    name_prefix: &str,
+    runtime_builder_opt: Option<Builder>,
+    block: F,
+) -> (Handle, thread::JoinHandle<()>)
 where
     F: FnOnce(Handle) -> R + Send + 'static,
     R: Future,
 {
-    let (tx, rx) = crossbeam_channel::bounded(1);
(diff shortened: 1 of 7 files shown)


util/runtime/src/lib.rs
```diff
@@ -1,29 +1,47 @@
...
-                .thread_name("GlobalRuntime-")
-                .build()
-                .expect("Global tokio runtime init");
+    let barrier = sync::Arc::new(sync::Barrier::new(2));
+    let barrier_clone = sync::Arc::clone(&barrier);
+
+    let service_name = format!("{}Service", name_prefix);
+    let runtime_name = format!("{}Runtime", name_prefix);
+
+    let mut builder = if let Some(runtime_builder) = runtime_builder_opt {
```
quake, line 20:
```suggestion
    let mut runtime = runtime_builder_opt
        .unwrap_or_else(|| Builder::new().threaded_scheduler())
        .thread_name(&runtime_name)
        .build()
```

yangby-cryptape: DONE!

doitian:
bors r=quake,driftluo

PR #2199, opened 2020-07-30
feat: add metrics service

yangby-cryptape:
- Features:
  - 4 kinds of recorders: counter, guage, value and timing.
  - 2 kinds of exporters: log and http.
  - 3 kinds of outputs format: json, yaml and prometheus.
- No more threads if no exporters.
  https://github.com/nervosnetwork/ckb/blob/0db57dafaad73efbfcf5330ec289efba94fd6975/util/metrics-service/src/lib.rs#L35-L37
- Usages to enable the metrics service:
  https://github.com/nervosnetwork/ckb/blob/0db57dafaad73efbfcf5330ec289efba94fd6975/util/metrics-config/src/lib.rs#L5-L22


util/metrics-service/src/config.rs
```diff
@@ -0,0 +1,191 @@
...
+            Err(err) => {
+                return Err(format!("failed to check the name of exporter: {}", err));
+            }
+        }
+        Ok(())
+    }
+}
+
+impl Exporter {
+    pub(crate) fn run<C>(self, handle: &Handle, c: C) -> Result<(), String>
```
keroro520, line 115: Just a question, why not put the function `run` into "service.rs".

util/metrics/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+        Self(Instant::now())
+    }
+
+    pub fn stop(self) -> Duration {
+        Instant::now() - self.0
+    }
+}
+
+#[macro_export(local_inner_macros)]
+macro_rules! metrics {
```
keroro520, line 18: Suggest commenting an example to use `metrics!`

util/metrics-service/src/config.rs
```diff
@@ -0,0 +1,191 @@
...
+        match Regex::new(r"^[0-9a-zA-Z_-]+$") {
+            Ok(re) => {
+                if !re.is_match(&name) {
+                    return Err(format!(
+                        "invaild exporter name \"{}\", only \"0-9a-zA-Z_-\" are allowed",
+                        name
+                    ));
+                }
+            }
+            Err(err) => {
```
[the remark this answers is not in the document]
yangby-cryptape: Example:
```toml
[metrics.exporter."I+can~use*any#characters^between@double%quotes"] 
```

util/metrics-service/src/config.rs
yangby-cryptape: - **`Exporter` is defined in this file.** And 90% code in `run` is to parse the configuration (just as the file name `config.rs`).
- At first, I want to return a generic `MetricsExporter` from the `ConfigExporter`, then I can `async_run()` the generic `MetricsExporter` in "service.rs".
  **But**, I failed.
  Ideally, the code should be:
  ```rust
  impl ConfigExporter {
      pub(crate) fn into_metrics_exporter(self, observe) -> MetricsExporter {
          let Self { target, format } = self;
          let builder = Builder::from(format);
          let exporter = MetricsExporter::new(observe, builder, target);
          return exporter;
      }
  }

  pub fn init(config) {
      /* ... omit ... */
      for (name, exporter) in config.exporter {
          Config::check_exporter_name(&name)?;
          let metrics_exporter = exporter.into_metrics_exporter();
          handle.spawn(async {
              tokio::spawn(metrics_exporter.async_run(controller.clone()));
          });
      }
      /* ... omit ... */
  }
  ```

util/metrics-service/src/config.rs
```diff
@@ -0,0 +1,191 @@
...
+pub enum LogLevel {
+    Error,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+}
+
+impl Config {
+    // .map_err(|err| format!("metrics exporter {}: {}", name, err))?;
```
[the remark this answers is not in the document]
yangby-cryptape: This line should be removed.
It was added automatically in one `git rebase` operation during my development.
I will remove it.

yangby-cryptape:
The last "force-pushed" just removed a useless comment.

util/metrics-service/src/config.rs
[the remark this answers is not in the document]
keroro520: What you talk about is `re.is_match`. I mean `Re::new()` will not return errors in our case.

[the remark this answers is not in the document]
yangby-cryptape: Since `Regex::new()` return `Result` and we can handle it in this function, I think `match Error` is OK.
I think `panic` and `expect` should be only used in the context  which can't do error handling.

util/app-config/src/app_config.rs
```diff
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 
 use ckb_chain_spec::ChainSpec;
 use ckb_logger_service::Config as LogConfig;
+use ckb_metrics_service::Config as MetricsConfig;
```
doitian, line 15: It's better to create a config crate so that app config does not  depend on services. The same is the logger config.

util/metrics-service/src/config.rs
```diff
@@ -0,0 +1,189 @@
...
+            Ok(re) => {
+                if !re.is_match(&name) {
+                    return Err(format!(
+                        "invaild exporter name \"{}\", only \"0-9a-zA-Z_-\" are allowed",
+                        name
+                    ));
+                }
+            }
+            Err(err) => {
+                return Err(format!("failed to check the name of exporter: {}", err));
```
doitian, line 105: The message `"invalid regular expression {}: {}", re, err` is more accurate.

util/metrics-service/src/config.rs
doitian: As I suggest that config should be its own crate, the `run` method should be in the service crate instead in the config crate.

util/metrics-service/src/service.rs
```diff
@@ -0,0 +1,88 @@
...
+use metrics_runtime::Receiver;
+use tokio::sync::oneshot;
+
+use ckb_async_runtime::{new_runtime, Builder, Handle};
+use ckb_stop_handler::{SignalSender, StopHandler};
+
+use crate::Config;
+
+#[must_use]
+pub enum Guard {
```
doitian, line 12: Nice pattern

doitian:
Since it heavily depends on a 3rd-party crate, please and link and some introductions in the PR description.

util/metrics/src/lib.rs
doitian, line 18: Please add some documents for the metric, at least list the available metrics types.

yangby-cryptape:
Hold for updating, according to review  suggestions.

doitian:
@yangby-cryptape Ready to unhold?

yangby-cryptape:
> @yangby-cryptape Ready to unhold?

I will rebase and force push this PR after #2220 merged.

util/app-config/src/app_config.rs
yangby-cryptape: DONE!

util/metrics-service/src/config.rs
yangby-cryptape: DONE!

util/metrics/src/lib.rs
yangby-cryptape: DONE!

yangby-cryptape:
Updated according to review suggestions.

The most major changes is: create a config crate so that app config does not depend on services.

yangby-cryptape:
@quake I updated the format of dependencies' versions in `Cargo.toml` in the last `force-push`.

yangby-cryptape:
bors r=quake,keroro520

PR #2208, opened 2020-08-05
fix: rpc get_peers deadlock caused by conflicting lock order

quake:
This PR fixes a deadlock bug caused by conflicting lock order in `block_fetcher.rs` and `net.rs`

`block_fetcher.rs`
SyncState#inflight_blocks write lock acquisition
https://github.com/nervosnetwork/ckb/blob/37dac610489e2a15636cb6d192f831b9af2fc8f5/sync/src/synchronizer/block_fetcher.rs#L98

SyncState#peers#state write lock acquisition
https://github.com/nervosnetwork/ckb/blob/37dac610489e2a15636cb6d192f831b9af2fc8f5/sync/src/synchronizer/block_fetcher.rs#L124

v.s

`net.rs`
SyncState#peers#state read lock acquisition
https://github.com/nervosnetwork/ckb/blob/719d937fbdddaeca8d9b70c96344b548e889eb7b/rpc/src/module/net.rs#L124

SyncState#inflight_blocks read lock acquisition
https://github.com/nervosnetwork/ckb/blob/719d937fbdddaeca8d9b70c96344b548e889eb7b/rpc/src/module/net.rs#L150

This deadlock has a chance of being triggered by calling `get_peers` rpc at the same time as IBD, it's found on integration test log: https://gist.github.com/quake/78f9db87ebd0a8e776adcb24b881869a

### DIFF
--- a/rpc/src/module/net.rs
+++ b/rpc/src/module/net.rs
@@ -108,6 +108,7 @@ impl NetworkRpc for NetworkRpcImpl {
                     }
                 }
 
+                let inflight_blocks = self.sync_shared.state().read_inflight_blocks();
                 RemoteNode {
                     is_outbound: peer.is_outbound(),
                     version: peer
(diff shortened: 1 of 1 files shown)


doitian:
bors r=doitian,driftluo,yangby-cryptape

PR #2209, opened 2020-08-05
chore: add log to `try_search_orphan_pool`

quake:
Try to resolve integration test `block_sync_relayer_collaboration` that will occasionally go wrong, add a line of log where I suspect there may be a problem.


zhangsoledad:

PR #2196, opened 2020-07-30
feat: add more fields to rpc `get_peers`

quake:
added `connected_duration`, `last_ping_duration`, `protocols` and `sync_state` to `get_peers` rpc

PR #2212, opened 2020-08-06
feat: advance last_common_header even the peer is worse than us

keroro520:
`last_common_header` is like a cursor of block-sync mechanism between the local node and remote peers. The local node only fetches blocks from peers who have a better chain. It means it is not necessary to timely advance the `last_common_header` of peers whose chain is worse than the local. 

However, https://github.com/nervosnetwork/ckb/pull/2196 adds `last_common_header` to RPC `get_peers`. It expects that the returned `last_common_header` to provide a more realistic picture. Hence I create this PR, timely advance the `last_common_header` of peers even their chains are worse than the local.

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -87,6 +87,23 @@ impl<'a> BlockFetcher<'a> {
         // This peer has nothing interesting.
         let best_known = self.peer_best_known_header()?;
         if !self.is_better_chain(&best_known) {
+            // Advancing this peer's last_common_header is unnecessary for block-sync mechanism.
+            // However, RPC `get_peers`, returns peers information which includes
+            // last_common_header, is expected to provide a more realistic picture. Hence here we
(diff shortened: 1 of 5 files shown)


sync/src/synchronizer/block_fetcher.rs
```diff
@@ -87,6 +87,23 @@ impl<'a> BlockFetcher<'a> {
...
+            // Advancing this peer's last_common_header is unnecessary for block-sync mechanism.
+            // However, RPC `get_peers`, returns peers information which includes
+            // last_common_header, is expected to provide a more realistic picture. Hence here we
+            // specially advance this peer's last_common_header at the case of both us on the same
+            // active chain.
+            if self
+                .active_chain
+                .get_block_hash(best_known.number())
+                .map(|block_hash| block_hash == best_known.hash())
+                .unwrap_or(true)
```
quake, line 99:
```suggestion
            if self
                .active_chain
                .snapshot()
                .is_main_chain(best_known.hash())
```

doitian:
bors r=quake,doitian

PR #2213, opened 2020-08-07
perf: reduce useless field and reduce get time call

driftluo:


network/src/protocols/identify/mod.rs
```diff
@@ -151,33 +149,26 @@ impl<T: Callback> IdentifyProtocol<T> {
         observed: Multiaddr,
     ) -> MisbehaveResult {
         let session = context.session;
-        let mut info = self
+        let info = self
             .remote_infos
             .get_mut(&session.id)
             .expect("RemoteInfo must exists");
 
-        if info.observed_addr.is_some() {
```
doitian, line 159: The code has bad smell. Function `process_observed` does not check duplicated message because it is ways called after `process_listens`. I recommend adding a function `process_message` to check the duplication in it first and then call the two sub-routines.

driftluo:
bors r=doitian,TheWaWaR

PR #2214, opened 2020-08-07
chore: Apply changes to help locating Windows integration test failure reason

doitian:


devtools/windows/make.ps1
```diff
@@ -76,7 +76,7 @@ function run-integration {
   pushd test
   Set-Env RUST_BACKTRACE 1
   Set-Env RUST_LOG $env:INTEGRATION_RUST_LOG
-  iex "cargo run -- --bin target/debug/ckb $($env:CKB_TEST_ARGS)"
+  iex "& cargo run -- --bin target/debug/ckb $($env:CKB_TEST_ARGS)"
```
yangby-cryptape, line 79: Ref: [`Invoke-Expression` (whose built-in alias is `iex`) vs `&` (the call operator)](https://stackoverflow.com/a/50023419)

_Maybe I'm not the only one who don't know the different between `iex` and `&`._

PR #2215, opened 2020-08-10
chore: prefix integration test log with case name and node index

chuijiaolianying:
Update integration test log from `ckb-it-xxx` to `ckb-it-case_name-node_index-xxx`.

Because of node_id initialize after the node start, and node start need a working_dir which contains the node log, so I use `node_index` in the test case instead of `node_id`.


chuijiaolianying:
bors r=keroro520,doitian

doitian:
bors r=doitian,keroro520

PR #2216, opened 2020-08-11
fix: integration test random failures

quake:
While waiting for the `get_blocks` message in the integration test, we should determine whether the last block hash is equal or not, otherwise, in the case of sending `get_blocks` in batches (which is triggered by timer), subsequent messages will be processed when the first batch is received, which may cause the random integration test failure.


yangby-cryptape:
bors r=driftluo,keroro520,yangby-cryptape

yangby-cryptape:
`bors` failed because no code owner approvals, even though this PR was in the `Approved` state (and the merge button is green) :confused:.

yangby-cryptape:
bors r=driftluo,keroro520,yangby-cryptape,zhangsoledad

PR #2217, opened 2020-08-11
chore: document DEFAULT_BYTES_PER_CYCLES and increase precision

doitian:
CKB block has two dimensional limits on transactions, the total cycles and total bytes. Pool uses vbytes by uniforming these two dimension:

* Convert cycles and bytes to percentage of the block respectively.
* Pick the larger one
* Multiplying the larger percent with the block bytes limit gives the vbytes.

Thus the ratio between bytes and cycles is block cycles limit dividing the bytes limit.

I increase the precision to 7 digits so that

```
abs(MAX_BLOCK_BYTES - MAX_BLOCK_CYCLES * DEFAULT_BYTES_PER_CYCLES) < 1
```

### DIFF
--- a/tx-pool/src/component/mod.rs
+++ b/tx-pool/src/component/mod.rs
@@ -8,7 +8,10 @@ pub(crate) mod proposed;
 
 pub use self::entry::{DefectEntry, TxEntry};
 
-const DEFAULT_BYTES_PER_CYCLES: f64 = 0.000_17f64;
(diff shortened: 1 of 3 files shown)


doitian:
bors try

doitian:
Fixed clippy and unit tests

zhangsoledad (approved):

zhangsoledad:
bors r=quake,zhangsoledad

PR #2226, opened 2020-08-12
chore: upgrade tentacle

driftluo:
Features

- Refactor secio to reduce channel overhead
- Split listener from service struct
- Implement and replace priority channel
- Removal of side-effects of set delay tasks, resulting in significant performance improvements
- Rewrite stream poll
- Rewrite buffer cache
- Remove protocol on service
- Add more test

Bug Fix

- Fix yamux leak mem
- Fix yamux send go away

### DIFF
--- a/network/src/protocols/mod.rs
+++ b/network/src/protocols/mod.rs
@@ -95,8 +95,10 @@ pub trait CKBProtocolHandler: Sync + Send {
     }
     /// Called when the Service receives the notify task
     fn notify(&mut self, _nc: Arc<dyn CKBProtocolContext + Sync>, _token: u64) {}
-    /// Behave like `Stream::poll`, but nothing output
-    fn poll(&mut self, _nc: Arc<dyn CKBProtocolContext + Sync>) {}
+    /// Behave like `Stream::poll`
+    fn poll(&mut self, _nc: Arc<dyn CKBProtocolContext + Sync>) -> Poll<Option<()>> {
+        Poll::Ready(None)
+    }
 }
 
 pub struct CKBProtocol {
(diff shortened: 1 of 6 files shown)


driftluo:
benchmark

network/src/protocols/mod.rs
```diff
@@ -96,7 +96,9 @@ pub trait CKBProtocolHandler: Sync + Send {
     /// Called when the Service receives the notify task
     fn notify(&mut self, _nc: Arc<dyn CKBProtocolContext + Sync>, _token: u64) {}
     /// Behave like `Stream::poll`, but nothing output
```
yangby-cryptape, line 98: Is this comment deprecated? Not nothing output, now.

util/app-config/Cargo.toml
```diff
@@ -21,7 +21,7 @@ ckb-resource = { path = "../../resource"}
 ckb-build-info = { path = "../build-info" }
 ckb-types = { path = "../types" }
 ckb-fee-estimator = { path = "../fee-estimator" }
-p2p = { version="0.3.0-alpha.4", package="tentacle", features = ["molc"] }
+p2p = { version="0.3.0", package="tentacle", features = ["molc"] }
```
yangby-cryptape, line 24: In current commit, this version is not as same as `ckb-network`,  and only `p2p::secio::PeerId` is used in `ckb-app-config`.
So I think the dependencies in `ckb-network` and `ckb-app-config` could be different.

_As @doitian said in https://github.com/nervosnetwork/ckb/pull/2199#discussion_r465498134_:
> It's better to create a config crate so that app config does not depend on services.

Thus, could you change this to a lightweight crate?

nervos-bot-user:
#### Benchmark Result
  - TPS: 330.62
  - Samples Count: 50
  - CKB Version: ba813ac27d4506c1c6aad24c41537685509c9519
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

quake:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 346.42
  - Samples Count: 51
  - CKB Version: b466a2843b492d7f7bfade801f46712773c6501a
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: 2in2out

driftluo:
bors r=quake,yangby-cryptape

PR #2227, opened 2020-08-12
Add test cases for security GHSA-q73f-w3h7-7wcc

zhangsoledad:
Add test cases:
* tx-pool reject tx which calls syscall load_cell_data_hash from input
* missing output_data tx verify
* update block cells correctly which contains tx-chain


zhangsoledad:

PR #2236, opened 2020-08-21
refactor: rewrite discovery

driftluo:
1. Discovery remove `Both` registration
2. Rewrite discovery protocol, the original implementation is too complicated and difficult to read, and there is a lot of unnecessary consumption
3. Use bloom filter to filter known address

Later, I will write these implementations back into Tentacle to avoid these problems with reference implementations


network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,409 @@
...
+                                    self.addr_mgr.add_new_addr(session.id, addr.clone());
+                                }
+                            }
+
+                            while items.len() > 1000 {
+                                if let Some(last_item) = items.pop() {
+                                    let idx = rand::random::<usize>() % 1000;
+                                    items[idx] = last_item;
+                                }
+                            }
```
quake, line 149:
```suggestion
let items = items.choose_multiple(&mut rand::thread_rng(), 1000);
```

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,409 @@
...
+                                    .collect::<Vec<_>>();
+
+                                let items_clone = addrs
+                                    .iter()
+                                    .filter_map(|addr| {
+                                        if let Some(addr) = multiaddr_to_socketaddr(addr) {
+                                            Some(RawAddr::from(addr))
+                                        } else {
+                                            None
+                                        }
```
quake, line 232:
```suggestion
multiaddr_to_socketaddr(addr).map(Into::into)
```

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,409 @@
...
+                                .collect::<Vec<_>>();
+
+                            let items_clone = addrs
+                                .iter()
+                                .filter_map(|addr| {
+                                    if let Some(addr) = multiaddr_to_socketaddr(addr) {
+                                        Some(RawAddr::from(addr))
+                                    } else {
+                                        None
+                                    }
```
quake, line 290:
```suggestion
multiaddr_to_socketaddr(addr).map(Into::into)
```

quake (dismissed):

network/src/protocols/discovery/addr.rs
```diff
@@ -44,63 +45,87 @@ impl MisbehaveResult {
...
 
 // bitcoin: bloom.h, bloom.cpp => CRollingBloomFilter
-pub struct AddrKnown {
+pub struct AddrKnown<T> {
     max_known: usize,
-    addrs: HashSet<RawAddr>,
-    addr_times: HashMap<RawAddr, Instant>,
-    time_addrs: BTreeMap<Instant, RawAddr>,
+    addrs: HashSet<T>,
+    order_addrs: Vec<T>,
```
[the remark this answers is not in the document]
driftluo: StableBloomFilter has no interface for batch insertion, but, batch insert is used a lot here

doitian:
The bloom filter change deserves its own PR

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
+
+        let dynamic_query_cycle = self.dynamic_query_cycle.unwrap_or(ANNOUNCE_INTERVAL);
+        let addr_mgr = &self.addr_mgr;
+
+        // get announce list
+        let announce_list: Vec<_> = self
+            .sessions
+            .iter_mut()
+            .filter_map(|(id, state)| {
+                // send all announce addr to remote
```
doitian, line 285: The function names are confusing here.

The method `send_messages` send the Nodes message as announce. And then we check the timer to update `state.announce`. It seems there are two different concepts of `announce`s. We can rename the latter `announce` to `state.to_be_anounced` to make it clear.

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
+                state.send_messages(context, *id);
+                // check timer
+                state.check_timer(now, dynamic_query_cycle);
```
doitian, line 288: The pattern can be added as just one public interface in state, instead exposing these fields and methods separately.

```
state.should_be_announced_as().filter(|addr| addr_mgr.is_valid_addr(addr))

impl State {
    pub fn should_be_announced_as(&mut self) -> Option<Addr>;
}
```

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
+                for i in 0..2 {
+                    if let Some(key) = remain_keys.get(i) {
+                        if let Some(value) = self.sessions.get_mut(key) {
+                            trace!(
+                                ">> send {} to: {:?}, contains: {}",
+                                announce_multiaddr,
+                                value.remote_addr,
+                                value.addr_known.contains(&announce_multiaddr)
+                            );
+                            if value.announce_multiaddrs.len() < 10
```
doitian, line 322: Hardcoded, use `ANNOUNCE_THRESHOLD` instead

```suggestion
                            if value.announce_multiaddrs.len() < ANNOUNCE_THRESHOLD
```

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
                 } else {
-                    warn!("other channel error: {:?}", err);
+                    None
+                }
+            })
+            .collect();
+
+        if !announce_list.is_empty() {
+            let mut rng = rand::thread_rng();
+            let mut remain_keys = self.sessions.keys().cloned().collect::<Vec<_>>();
```
doitian, line 310: The name suggests we are going to remove something from the vec, but we don’t.

network/src/protocols/discovery/addr.rs
```diff
@@ -52,50 +46,36 @@ pub trait AddressManager {
...
 }
 
 impl AddrKnown {
     pub(crate) fn new(max_known: usize) -> AddrKnown {
         AddrKnown {
-            max_known,
-            addrs: HashSet::default(),
-            addr_times: HashMap::default(),
-            time_addrs: BTreeMap::default(),
+            filters: StableBloomFilter::new(
```
doitian, line 55: Please explain why choosing these parameters

network/src/protocols/discovery/addr.rs
```diff
@@ -52,50 +46,36 @@ pub trait AddressManager {
...
+                max_known,
```
doitian, line 56: The name `max_known` no longer makes sense, it is the number of buckets in the filter.

network/src/protocols/discovery/addr.rs
doitian: Please refer to http://webdocs.cs.ualberta.ca/~drafiei/papers/DupDet06Sigmod.pdf chapter 4.1 and explain how the parameters are set.

This is why I suggest to create a new PR for bloom filter.

doitian (changes requested):

doitian:
Some review questions may need help from @TheWaWaR

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
+            for announce_multiaddr in announce_list {
+                remain_keys.shuffle(&mut rng);
+                for i in 0..2 {
```
doitian, line 313: Easier way:

`for key in remain_keys.iter().take(3)`

doitian: Why 3?

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,364 @@
...
-            .expect("Discovery init only once");
-        if let Err(err) = context.future_task(discovery_task) {
-            error!("Start discovery_task failed: {:?}", err);
+        if context
+            .set_service_notify(
+                context.proto_id,
+                self.check_interval.unwrap_or(CHECK_INTERVAL),
+                0,
+            )
+            .is_err()
```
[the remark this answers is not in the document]
driftluo: Probably impossible, because the init call is at startup

network/src/protocols/discovery/addr.rs
```diff
@@ -52,50 +46,36 @@ pub trait AddressManager {
...
     }
 
-    pub(crate) fn contains(&self, addr: &RawAddr) -> bool {
-        self.addrs.contains(addr)
+    pub(crate) fn extend<'a, T: 'a + ::std::hash::Hash>(
+        &mut self,
+        keys: impl Iterator<Item = &'a T>,
+    ) {
+        for key in keys {
+            self.filters.insert(key)
```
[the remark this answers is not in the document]
driftluo: This is unrealizable. Bloomfilter's extend needs to traverse the hash once. Even if it provides a method called extend, it also traverses and inserts. 

And, as a table filter, when inserting, it will randomly delete an occupation, which means that it can ensure the existence of recent inserts as much as possible, and randomly delete more distant inserts

network/src/protocols/discovery/mod.rs
driftluo: maybe this is a secrete，just keep random

network/src/protocols/discovery/addr.rs
driftluo: Memory usage = m * d = 5000 * 3 bytes
Verifiable data volume = m / d = 5000 / 3 =  1600+

doitian:
🎉 I'll review again tomorrow (Sep 3)

doitian (approved):

TheWaWaR:
Will review this PR today.

network/src/protocols/discovery/mod.rs
```diff
@@ -1,426 +1,312 @@
...
+                    } => {
+                        if let Some(state) = self.sessions.get_mut(&session.id) {
+                            if state.received_get_nodes && check(Misbehavior::DuplicateGetNodes) {
+                                return;
+                            }
+
+                            state.received_get_nodes = true;
+                            // must get the item first, otherwise it is possible to load
+                            // the address of peer listen.
+                            let mut items = self.addr_mgr.get_random(2500);
```
TheWaWaR, line 124: The magic number `2500` better with a `const` value.

TheWaWaR (approved):

driftluo:
bors r=TheWaWaR,doitian

PR #2239, opened 2020-08-24
feat: support to control memory usage for header map

yangby-cryptape:


ckb-bin/src/subcommand/run.rs
```diff
@@ -30,6 +30,14 @@ pub fn run(args: RunArgs, version: Version) -> Result<(), ExitCode> {
         .consensus(args.consensus)
         .tx_pool_config(args.config.tx_pool)
         .notify_config(args.config.notify)
+        .header_map_config(
+            args.config
+                .network
+                .header_map
+                .as_ref()
+                .cloned()
+                .unwrap_or_else(Default::default),
```
quake, line 39:
```suggestion
                .unwrap_or_default,
```

shared/src/shared.rs
```diff
@@ -200,6 +205,10 @@ impl Shared {
     pub fn store(&self) -> &ChainDB {
         &self.store
     }
+
+    pub fn header_map_config(&self) -> &HeaderMapConfig {
```
quake, line 209: I don't think we need a fn to get `header_map_config` here, the config should be passed to `Shared::init` directly like other config objects

util/app-config/src/configs/network.rs
```diff
@@ -37,6 +37,24 @@ pub struct Config {
     pub bootnode_mode: bool,
     // Max send buffer size
     pub max_send_buffer: Option<usize>,
+    pub header_map: Option<HeaderMapConfig>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HeaderMapConfig {
```
quake, line 44: suggest to rename it SyncConfig, rename the fields and add prefix:
```
pub struct SyncConfig {
  header_map_primary_limit
  header_map_backend_close_threshold
}
```

This is useful for config file compatibility when adding a block map in the future.

quake (changes requested):

util/app-config/src/configs/network.rs
yangby-cryptape: DONE!

In future, we can add `BlockStatusMapConfig` as follow:

```rust
#[derive(Clone, Debug, Serialize, Deserialize, Default)]
pub struct SyncConfig {
    #[serde(default)]
    pub header_map: HeaderMapConfig,
    #[serde(default)]
    pub block_status_map: BlockStatusMapConfig,
}

#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct MapConfig {
    // The maximum size of data in memory
    pub primary_limit: usize,
    // Disable cache if the size of data in memory less than this threshold
    pub backend_close_threshold: usize,
}

#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct HeaderMapConfig {
    #[serde(flatten)]
    pub map: MapConfig,
}

#[derive(Clone, Debug, Serialize, Deserialize)]
pub struct BlockStatusMapConfig {
    #[serde(flatten)]
    pub map: MapConfig,
}

impl Default for HeaderMapConfig {
    fn default() -> Self {
        let map = MapConfig {
            primary_limit: 300_000,
            backend_close_threshold: 20_000,
        };
        Self { map }
    }
}

impl Default for BlockStatusMapConfig {
    fn default() -> Self {
        // we can use different default values
        let map = MapConfig {
            primary_limit: 600_000,
            backend_close_threshold: 40_000,
        };
        Self { map }
    }
}
```

quake (approved):

keroro520 (approved):

quake:
bors r=quake,keroro520

quake:
bors r=quake,keroro520

PR #2240, opened 2020-08-24
feat: create log monitor for integration test check status between nodes

chuijiaolianying:
1. Reading node's log directly is an easier way to implement log monitor than using `Stdio::piped`. 
2. Because we could get result by checking if there exists a specify keyword in log at the most, so I use `str::contains()` instead of regex.
3. I updated the cases in `block_relay.rs` as an example, and I think all `wait_util()` can be updated with using log monitor, and some RPC call also can be replaced.

### DIFF
--- a/test/src/node.rs
+++ b/test/src/node.rs
@@ -1,5 +1,5 @@
 use crate::rpc::RpcClient;
-use crate::utils::{temp_path, wait_until};
+use crate::utils::{node_log, sleep, temp_path, tweaked_duration, wait_until};
 use crate::{DEFAULT_TX_PROPOSAL_WINDOW, SYSTEM_CELL_ALWAYS_SUCCESS_INDEX};
 use ckb_app_config::{BlockAssemblerConfig, CKBAppConfig};
 use ckb_chain_spec::consensus::Consensus;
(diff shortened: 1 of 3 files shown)


keroro520 (changes requested):
I think the current implementation is more like "log grep" but not "log monitor". For example, when a node receives the same `CompactBlock` twice, the "log monitor" will notify twice **at once**.

test/src/node.rs
```diff
@@ -94,6 +96,43 @@ impl Node {
         &self.working_dir
     }
 
+    pub fn log_file(&self) -> String {
+        node_log(&self.working_dir()).display().to_string()
+    }
+
+    pub fn log_monitor(&self, expected: &str, timeout: u64) -> Receiver<String> {
```
doitian, line 103: We can remember the file size when the monitor is created. The monitor only needs to parse the new appended text content (content after the remembered file size). It can save time when the log file is large and we know exactly the event happens after the monitor is created. It also helps when we want to wait on several events in sequence.

test/src/node.rs
```diff
@@ -94,6 +96,43 @@ impl Node {
...
+        let (log_sender, log_notifier) = bounded(1);
+        let timeout = tweaked_duration(timeout);
+        let start = Instant::now();
+        let filename = &self.log_file();
+        let mut skip_lines = 0;
+        let mut modified = SystemTime::now();
+        loop {
+            let filename = fs::File::open(&filename).unwrap();
```
doitian, line 111: It's faster to keep the file handle, and poll the new content.

I found several crates that you can refer to:

* https://crates.io/crates/chase async file following
* https://crates.io/crates/logwatcher tail -F clone
* https://crates.io/crates/logwatch another tail -F

Issue #2247, opened 2020-08-26
CKB RPC Method Deprecation Process


doitian:
The project is under active development. Some RPC methods no longer make sense in the new version. We need a way to deprecate the RPC methods and finally remove them.

## CKB RPC Method Deprecation Process

A CKB RPC method is deprecated in three steps.

First the method is marked as deprecated in the CKB release notes and RPC document. However, the RPC method is still available. The RPC document will have the suggestion of the alternative solutions.

The CKB dev team will disable any deprecated RPC methods starting from the next minor version release. Users can enable the methods via the command line argument `--enable-deprecated-rpc` or the config file option `enable_deprecated_rpc`.

Once a deprecated method is disabled, the CKB dev team will remove it in a future minor version release.

For example, a method is marked as deprecated in 0.35.0, it can be disabled in 0.36.0 and removed in 0.37.0. The minor versions are released monthly, so there's at least two month buffer for a deprecated RPC method.

PR #2251, opened 2020-08-27
feat: deprecated rpc method by adding `deprecated.` prefix to the rpc name

quake:
resolve https://github.com/nervosnetwork/ckb/issues/2247, rpc method can be deprecated by adding `deprecated.` prefix to the rpc name:
```diff
+    #[rpc(name = "deprecated.get_cells_by_lock_hash")]
-    #[rpc(name = "get_cells_by_lock_hash")]
    fn get_cells_by_lock_hash
```

### DIFF
--- a/devtools/doc/rpc.py
+++ b/devtools/doc/rpc.py
@@ -14,7 +14,11 @@ def sort_cases_by_module(cases):
 
 
 def print_title(case):
-    print("### `{}`".format(case["method"]))
+    if case.get("deprecated") is None:
+        print("### `{}`".format(case["method"]))
+    else:
+        print("### ~~`{}`~~".format(case["method"]))
+        print("**DEPRECATED** {}".format(case["deprecated"]))
     newline(1)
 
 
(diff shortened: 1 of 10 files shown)


doitian:
According to #2247, we need a window to notify that an RPC has been deprecated. During the window, users must be able to continue using the methods without any changes.

And I think renaming the RPC method name is not a friend deprecation mechanism, it requires updating client code. I prefer the solution used by bitcoin, in which user must listed the enabled deprecated methods in the command line arguments.

doitian (changes requested):
We need a two stage deprecation mechanism, see https://github.com/nervosnetwork/ckb/issues/2247

quake:
> According to #2247, we need a window to notify that an RPC has been deprecated. During the window, users must be able to continue using the methods without any changes.
> 
> And I think renaming the RPC method name is not a friend deprecation mechanism, it requires updating client code. I prefer the solution used by bitcoin, in which user must listed the enabled deprecated methods in the command line arguments.

this PR doesn't rename rpc method, please review again, thanks.

doitian (dismissed):

yangby-cryptape (dismissed):

quake:

doitian (approved):

zhangsoledad (approved):

zhangsoledad:
bors r=doitian,zhangsoledad,yangby-cryptape

PR #2257, opened 2020-09-02
feat: add ping_peers rpc

quake:
This PR added a `ping_peers` rpc, which requests that a ping be sent to all connected peers, to measure ping time. 

for reviewer's reference: the code change of `network/src/network.rs` is using struct deconstruction and scoped clone to replace the original multiple rename clone.

before:
```
let foo = Arc::new(...);
let foo_clone = foo.clone();
thread::spawn(move || {
    let foo_clone_1 = foo_clone.clone();
    thread::spawn(move || {
        foo_clone_1.bar();
    });
    foo_clone.bar();
});
foo.bar();
```

now:
```
let foo = Arc::new(...);
{
    let foo = foo.clone();
    thread::spawn(move || {
        {
            let foo = foo.clone();
            thread::spawn(move || {
                foo.bar();
            });
        }
        foo.bar();
    });
}
foo.bar();
```

makes it easier to read and add new variables.

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -23,7 +23,10 @@ use ckb_app_config::NetworkConfig;
 use ckb_logger::{debug, error, info, trace, warn};
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_util::{Condvar, Mutex, RwLock};
-use futures::{channel::oneshot, Future, StreamExt};
+use futures::{
+    channel::{mpsc::Sender, oneshot},
+    Future, StreamExt,
+};
 use ipnetwork::IpNetwork;
 use p2p::{
     builder::ServiceBuilder,
@@ -854,6 +857,7 @@ impl<T: ExitHandler> ServiceHandle for EventHandler<T> {
 pub struct NetworkService<T> {
     p2p_service: Service<EventHandler<T>>,
     network_state: Arc<NetworkState>,
+    ping_controller: Sender<()>,
     // Background services
     bg_services: Vec<Pin<Box<dyn Future<Output = ()> + 'static + Send>>>,
     version: String,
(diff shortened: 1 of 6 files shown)


network/src/protocols/ping.rs
```diff
@@ -244,6 +258,40 @@ impl ServiceProtocol for PingHandler {
...
+        context: &mut ProtocolContext,
+    ) -> Poll<Option<()>> {
+        self.control_receiver
+            .poll_next_unpin(cx)
+            .map(|control_message| {
+                control_message.map(|_| {
+                    let now = SystemTime::now();
+                    let peers: Vec<SessionId> = self
+                        .connected_session_ids
+                        .iter_mut()
```
driftluo, line 274: https://github.com/nervosnetwork/ckb/pull/2257/files#diff-d189679dd379fe6aac5b2b82d73c9e22L181

must filter processing peer，other wise it will be disconnected

quake: 😬  resolved, please review again, thanks

network/src/network.rs
```diff
@@ -1275,6 +1288,11 @@ impl NetworkController {
     pub fn protocols(&self) -> Vec<(ProtocolId, String, Vec<String>)> {
         self.network_state.protocols.read().clone()
     }
+
+    pub fn ping_peers(&self) {
```
driftluo, line 1292: why not mut here？

quake: need to change the filed of `NetworkController` to `Arc` if we using mut here:
pub struct NetworkController {
    ping_controller: Arc<Sender<()>>,
}

quake: it's limited by rpc macro:
https://github.com/nervosnetwork/ckb/blob/670bdddbf343b639a22fe2c06a545fd45b522c6b/rpc/src/module/net.rs#L58

driftluo: I probably know the reason its trait requires send + sync

driftluo (dismissed):

quake:
@driftluo rebased with latest develop branch, please review again, thanks

driftluo (approved):

doitian (approved):

quake:
bors r=doitian,driftluo

PR #2260, opened 2020-09-04
feat: add get_transaction_proof / verify_transaction_proof rpc

quake:


util/types/Cargo.toml
```diff
@@ -10,7 +10,7 @@ molecule = "=0.6.0"
 ckb-fixed-hash = { path = "../fixed-hash" }
 numext-fixed-uint = { version = "0.1", features = ["support_rand", "support_heapsize", "support_serde"] }
 bytes = { version="0.5.4", features = ["serde"] }
-merkle-cbt = "0.2"
+merkle-cbt = { git = "https://github.com/nervosnetwork/merkle-tree", branch = "quake/tree-verify-proof" }
```
quake, line 13: waiting for https://github.com/nervosnetwork/merkle-tree/pull/20, will use a published crate version later

rpc/src/module/chain.rs
```diff
@@ -364,4 +377,121 @@ impl ChainRpc for ChainRpcImpl {
...
+        if let Some(specified_block_hash) = block_hash {
+            if !retrieved_block_hash.eq(&specified_block_hash.pack()) {
+                return Err(RPCError::invalid_params(
+                    "Not all transactions found in specified block",
+                ));
+            }
+        }
+
+        snapshot
+            .get_block(&retrieved_block_hash)
```
doitian, line 430: I'm not sure whether it is a performance concern to read the whole block to only verify several leaves. As I mentioned in https://github.com/nervosnetwork/merkle-tree/pull/20/files#r484305194 , its possible to only read the transactions in the proof.

rpc/src/module/chain.rs
```diff
@@ -364,4 +377,121 @@ impl ChainRpc for ChainRpcImpl {
...
+                    }
+                }
+            })
+    }
+
+    fn verify_transaction_proof(&self, tx_proof: TransactionProof) -> Result<Vec<H256>> {
+        let snapshot = self.shared.snapshot();
+
+        snapshot
+            .get_block(&tx_proof.block_hash.pack())
```
doitian, line 459: I'm not sure whether it is a performance concern to read the whole block to only verify several leaves. As I mentioned in https://github.com/nervosnetwork/merkle-tree/pull/20/files#r484305194 , its possible to only read the transactions in the proof.

quake: It may be a bottleneck when this RPC is called frequently, and the optimized code is able to directly read the tx hash stored inside `COLUMN_BLOCK_BODY`, but I don't see the need here to optimize early, it makes the code more complex.

doitian (dismissed):

doitian (approved):

TheWaWaR (approved):

quake:
bors r=doitian,TheWaWaR

PR #2262, opened 2020-09-07
[HOLD] feat: add some util functions

keroro520:
Wrap some common behaviors, make tests more readable and programable.


doitian:
Blocked by #2264

test/src/utils/check.rs
```diff
@@ -0,0 +1,30 @@
...
+        .get_transaction(transaction.hash())
+        .map(|txstatus| txstatus.tx_status.status == Status::Committed)
+        .unwrap_or(false)
+}
+
+pub fn is_transaction_unknown(node: &Node, transaction: &TransactionView) -> bool {
+    node.rpc_client()
+        .get_transaction(transaction.hash())
+        .is_none()
+}
```
chuijiaolianying, line 30: how about move this file into `test/src/assertion/` dir

test/src/utils/mine.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/// The `[1, PROPOSAL_WINDOW.farthest()]` of chain is called as bootstrap period. Cellbases w
+/// this period are zero capacity.
+///
+/// This function will generate blank blocks until node.tip_block_number > PROPOSAL_WINDOW.fa
+///
+/// Typically involve this function at the beginning of test.
+pub fn mine_until_out_bootstrap_period(node: &Node) {
+    let farthest = node.consensus().tx_proposal_window().farthest();
```
chuijiaolianying, line 16: there have a `DEFAULT_TX_PROPOSAL_WINDOW` in `test/src/lib.rs`, maybe we can init this in the beginning of case execution, or just directly use the default value of this variable.

test/src/utils/sugar.rs
```diff
@@ -0,0 +1,17 @@
+use crate::utils::mine::mine_until_bool;
+use crate::Node;
+
+pub fn out_ibd_mode(nodes: &[Node]) {
+    for node in nodes.iter() {
+        mine_until_bool(node, || node.get_tip_block_number() > 0)
+    }
+}
```
chuijiaolianying, line 8: There already have `exit_ibd_mode()` in `test/src/net.rs`, so why need another `out_ibd_mode()` function here?

chuijiaolianying (commented):
Those utils functions look good, but is there any plan how to use it?

Directly use those functions in case like
```
let node = net.node(0);
mine(node, 1);
```

Or put it in high layer like in `src/node.rs`
```
pub fn generate_block(&self, count) {
    min(&self, count)
}
```
and call it in case file
```
let node = net.node(0);
node.generate_block(1);
```

I wish there have a unified way to write cases.

test/src/utils/check.rs
keroro520: We do agree that `test/src/assertion` and `check.rs` have many duplicate codes.
I would like to remove the utils in `test/src/assertion` or move into `test/src/utils/assertion.rs`. But to avoid more changes in this PR, I didn't do that.

// Maybe utils in `test/src/assertion` are good for code reuse, but less readability.

test/src/utils/mine.rs
keroro520: Oh, I think the`DEFAULT_TX_PROPOSAL_WINDOW` in integration-test is a legacy and redundant constant, I just intend to remove it. `node.consensus().tx_proposal_window()` returns the most realistic value.

keroro520:
> Those utils functions look good, but is there any plan how to use it?
> 
> Directly use those functions in case like
> 
> ```
> let node = net.node(0);
> mine(node, 1);
> ```
> 
> Or put it in high layer like in `src/node.rs`
> 
> ```
> pub fn generate_block(&self, count) {
>     min(&self, count)
> }
> ```
> 
> and call it in case file
> 
> ```
> let node = net.node(0);
> node.generate_block(1);
> ```
> 
> I wish there have a unified way to write cases.

Thanks for your feedback. I think I should split this PR into several small PRs and adjust the existing test cases. I will do it today.

keroro520:
https://github.com/nervosnetwork/ckb/pull/2268
https://github.com/nervosnetwork/ckb/pull/2271

PR #2255, opened 2020-09-01
feat: chain spec check

zhangsoledad:

PR #2265, opened 2020-09-08
feat: add nMinimumChainWork config

driftluo:
The mainnet has been online for a long time, and it is time to add a minimum workload proof to prevent possible node attacks during the initial synchronization.

I chose a total difficulty of 500k block height, header first sync will take about 1-2 minutes to reach this position, then start downloading the blocks.

### DIFF
--- a/spec/src/consensus.rs
+++ b/spec/src/consensus.rs
@@ -1,8 +1,8 @@
 #![allow(clippy::inconsistent_digit_grouping)]
 
 use crate::{
-    calculate_block_reward, OUTPUT_INDEX_DAO, OUTPUT_INDEX_SECP256K1_BLAKE160_MULTISIG_ALL,
-    OUTPUT_INDEX_SECP256K1_BLAKE160_SIGHASH_ALL,
+    calculate_block_reward, ChainSpec, OUTPUT_INDEX_DAO,
+    OUTPUT_INDEX_SECP256K1_BLAKE160_MULTISIG_ALL, OUTPUT_INDEX_SECP256K1_BLAKE160_SIGHASH_ALL,
 };
 use ckb_dao_utils::genesis_dao_data_with_satoshi_gift;
 use ckb_pow::{Pow, PowEngine};
@@ -82,7 +82,10 @@ pub struct ProposalWindow(pub BlockNumber, pub BlockNumber);
 
 // "TYPE_ID" in hex
 pub const TYPE_ID_CODE_HASH: H256 = h256!("0x545950455f4944");
-static IS_MAINNET: once_cell::sync::OnceCell<bool> = once_cell::sync::OnceCell::new();
+
+// 500_000 total difficulty
+const MIN_CHAIN_WORK_500K: U256 = u256!("0x3314412053c82802a7");
+// const MIN_CHAIN_WORK_1000K: U256 = u256!("0x6f1e2846acc0c9807d");
 
 /// Two protocol parameters w_close and w_far define the closest
 /// and farthest on-chain distance between a transaction's proposal
@@ -246,6 +249,7 @@ impl ConsensusBuilder {
                 primary_epoch_reward_halving_interval:
                     DEFAULT_PRIMARY_EPOCH_REWARD_HALVING_INTERVAL,
                 permanent_difficulty_in_dummy: false,
+                min_chain_work: u256!("0x0"),
             },
         }
     }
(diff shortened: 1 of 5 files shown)


spec/src/consensus.rs
```diff
@@ -778,6 +779,14 @@ impl Consensus {
             self.primary_epoch_reward(epoch.number() + 1)
         }
     }
+
+    pub fn is_mainnet(&self) -> bool {
+        *IS_MAINNET.get_or_init(|| {
+            let mainnet_genesis_hash: H256 =
+                h256!("0x92b197aa1fba0f63633922c61c92375c9c074a93e85963554f5499fe1450d0e5");
```
quake, line 786: suggest to use bundled resouce file and compare genesis hash:
```
spec = ChainSpec::load_from(Resource::bundled("specs/mainnet.toml"))
self. genesis_hash  == spec.genesis
```

quake: we may add a `mainnet` bool field to `Consensus` struct and move this comparison logic code to `ConsensusBuilder#build`, it will remove the `once_cell` deps.

sync/src/synchronizer/mod.rs
```diff
@@ -43,6 +44,10 @@ const SYNC_NOTIFY_INTERVAL: Duration = Duration::from_millis(200);
 const IBD_BLOCK_FETCH_INTERVAL: Duration = Duration::from_millis(40);
 const NOT_IBD_BLOCK_FETCH_INTERVAL: Duration = Duration::from_millis(200);
 
+// 500_000 total difficulty
+const MIN_CHAIN_WORK_500K: U256 = u256!("0x3314412053c82802a7");
```
quake, line 48: is it possible move this value to `SyncConfig` as a hidden configuration option and set this default value to it?

quake (changes requested):

spec/src/consensus.rs
doitian: I recommend adding it to chain spec as trivial options. I define trivial options as the ones does not affect consensus, thus they can be ignored when computing hash in #2255

util/app-config/src/configs/network.rs
```diff
@@ -40,10 +44,21 @@ pub struct Config {
...
     pub header_map: HeaderMapConfig,
+    #[serde(default)]
+    pub min_chain_work: U256,
+}
+
+impl Default for SyncConfig {
+    fn default() -> Self {
+        SyncConfig {
+            header_map: HeaderMapConfig::default(),
+            min_chain_work: MIN_CHAIN_WORK_500K,
```
doitian, line 59: 0 is more reasonable to be the default

sync/src/synchronizer/mod.rs
```diff
@@ -51,27 +52,60 @@ struct BlockFetchCMD {
...
                     }
                 }
             }
         }
     }
 
+    fn can_start(&mut self) -> bool {
+        if self.can_start {
+            true
+        } else if self.sync.shared.consensus().is_mainnet {
```
doitian, line 97: Use `min_chain_work > 0`. Checking whether the chain is mainnet can be done when setting the configurations during boot.

sync/src/synchronizer/mod.rs
```diff
@@ -51,27 +52,60 @@ struct BlockFetchCMD {
...
+                                    BlockFetchCMD::send_getblocks(item, &self.p2p_control, peer);
+                                }
                             }
                         }
+                    } else {
+                        let best_known = self.sync.shared.state().shared_best_header_ref();
+                        let number = best_known.number();
+                        if number != self.number && number % 10000 == 0 {
+                            self.number = number;
+                            info!(
```
doitian, line 79: Does it worth a metric record?

doitian (changes requested):

sync/src/synchronizer/mod.rs
doitian: This is my main reason to block the merge.

sync/src/synchronizer/mod.rs
driftluo: This only appears when synchronizing data from the beginning, and will not happen again, I think it should not be necessary to use metric @yangby-cryptape @keroro520

quake (dismissed):

sync/src/synchronizer/mod.rs
```diff
@@ -51,27 +52,59 @@ struct BlockFetchCMD {
...
+                                "best known header number: {}, difficulty: {:#x}, \
+                                 require min header number on 500_000, min difficulty: {:#x}, \
```
quake, line 81:
```suggestion
                                "best known header number: {}, total_difficulty: {:#x}, \
                                 require min header number on 500_000, min total_difficulty: {:#x}, \
```

doitian (approved):

quake (approved):

driftluo:
bors r=doitian,quake

driftluo:
bors retry

sync/src/synchronizer/mod.rs
keroro520: I agree with @driftluo.

yangby-cryptape: The floor of memory limit is increased when synchronizing data from the beginning.

PR #2268, opened 2020-09-09
refactor: add some transaction checking utils

keroro520:
Add [`test/src/util/check.rs`](https://github.com/nervosnetwork/ckb/pull/2268/files#diff-56a7c1b37b9c48a7b20036328fdb8125) some transaction checking utils.


test/src/specs/relay/transaction_relay_low_fee_rate.rs
```diff
@@ -22,16 +22,12 @@ impl Spec for TransactionRelayLowFeeRate {
...
         // confirm tx
         node1.generate_blocks(20);
-        let ret = wait_until(10, || {
-            node1
-                .rpc_client()
-                .get_transaction(hash.clone())
-                .map(|tx_result| tx_result.tx_status.status == Status::Committed)
-                .unwrap_or(false)
-        });
+        let ret = wait_until(10, || is_transaction_committed(node1, &tx));
```
chuijiaolianying, line 30: There seems doesn't need `wait_until`?

chuijiaolianying:
Please fix CI check error

test/src/specs/relay/transaction_relay_low_fee_rate.rs
keroro520: Yes, I think, `wait_until` is unnecessary here. However, I would rather leave it the same, since I only want to replace it with the new checking suites, but not improve the test codes.

chuijiaolianying (dismissed):

quake (approved):

quake:
bors r=quake,chuijiaolianying

doitian:
bors r=quake,chuijiaolianying

chuijiaolianying (approved):

quake (approved):

quake:
bors r=quake,chuijiaolianying

PR #2271, opened 2020-09-10
refactor: add some mining utils

keroro520:
Add some mining utils that replace `Node::generate_block`, `Node::generate_blocks`, and other several sugars.


test/src/specs/tx_pool/valid_since.rs
```diff
@@ -236,11 +237,11 @@ impl ValidSince {
             absolute_block
         );
         (0..DEFAULT_TX_PROPOSAL_WINDOW.0).for_each(|_| {
-            node.generate_block();
+            mine(&node, 1);
```
quake, line 240: suggest to replace all `(0..DEFAULT_TX_PROPOSAL_WINDOW.0).for_each(|_| mine(node, 10))` with:
```suggestion
            mine(&node, DEFAULT_TX_PROPOSAL_WINDOW.0)
```

keroro520: Updated, thanks for your suggestion. Please take a look.

test/src/specs/consensus/insufficient_reward.rs
```diff
@@ -75,7 +75,7 @@ impl Spec for InsufficientReward {
     // export data
     // fn run(&self, net: &mut Net) {
     //     let node = &mut net.nodes[0];
-    //     let hashes = node.generate_blocks(100);
+    //     let hashes = mine(node, 100);
```
chuijiaolianying, line 78: I'm not sure about this comment block, is there any necessary keep this and event update the method here? or we can just delete this part if we don't run it anymore.

test/src/util/sugar.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        let tip_number = node0.get_tip_block_number();
+        let range = 1..tip_number + 1;
+        for node in nodes.iter().skip(1) {
+            forward_main_blocks(node0, node, range.clone());
+        }
+    }
+}
+
+pub fn out_bootstrap_period(nodes: &[Node]) {
```
chuijiaolianying, line 20: I think `sugar` is not a clear enough name, move `out_bootstrap_period` into `mine.rs` and move `out_ibd_mode` into `node.rs` maybe a better way. If I want some function, I will find it in the relate file first, and maybe I'll never touch the sugar if I don't know what it has.

chuijiaolianying:
Please fix the CI error

test/src/specs/consensus/insufficient_reward.rs
keroro520: Out of this PR.

test/src/util/sugar.rs
keroro520: Agree! I will move them two into `mine.rs`, which should be treated as mining-related suites.

keroro520:
> Please fix the CI error

Fixed CI error, and merged `sugar.rs` into `mining.rs`. PTAL

chuijiaolianying (approved):

test/src/specs/relay/block_relay.rs
```diff
@@ -1,29 +1,11 @@
-use crate::node::{connect_all, exit_ibd_mode};
+use crate::util::mining::mine;
+use crate::util::mining::out_ibd_mode;
 use crate::utils::{now_ms, sleep, wait_until};
 use crate::{Node, Spec};
 use ckb_types::prelude::*;
 use log::info;
 use std::time::Duration;
 
-pub struct BlockRelayBasic;
```
quake, line 8: why was it removed?

keroro520: This case is good for the early stage. However, I think it is too simple for now, so I removed it when I resolve PR conflicts.

chuijiaolianying (dismissed):

quake (approved):

zhangsoledad (approved):

zhangsoledad:
bors r=quake,zhangsoledad

PR #2281, opened 2020-09-16
feat: add tx subscription

quake:
This PR added a `new_transaction` topic to subscription rpc, user will get notified when new transaction is submitted to pool.

tcp tx subscription example:
```
telnet localhost 18114
> {"id": 2, "jsonrpc": "2.0", "method": "subscribe", "params": ["new_transaction"]}
< {"jsonrpc":"2.0","result":"0x3","id":2}
< {"jsonrpc":"2.0","method":"subscribe","params":{"result":"... pool transaction entry json string...","subscription":"0x3"}}
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -2553,6 +2553,9 @@ Subscribe to a topic, if successful it returns the subscription id. For each eve
 #### Parameters
 
 * topic - Subscription topic (enum: new_tip_header | new_tip_block | new_transaction)
+    * new_tip_header: subscriber will get notified when new block is added to best chain tip
+    * new_tip_block: subscriber will get notified when new block is added to best chain tip
+    * new_transaction: subscriber will get notified when new transaction is submitted to pool
 #### Returns
 
 * id - Subscription id
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -2552,7 +2552,7 @@ Subscribe to a topic, if successful it returns the subscription id. For each eve
 
 #### Parameters
 
-* topic - Subscription topic (enum: new_tip_header | new_tip_block)
+* topic - Subscription topic (enum: new_tip_header | new_tip_block | new_transaction)
 #### Returns
 
 * id - Subscription id
--- a/rpc/src/module/subscription.rs
+++ b/rpc/src/module/subscription.rs
@@ -45,6 +45,7 @@ impl PubSubMetadata for SubscriptionSession {
 pub enum Topic {
     NewTipHeader,
     NewTipBlock,
+    NewTransaction,
 }
 
 #[allow(clippy::needless_return)]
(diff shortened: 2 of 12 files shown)


rpc/src/module/subscription.rs
```diff
@@ -126,9 +127,13 @@ impl SubscriptionRpc for SubscriptionRpcImpl {
 }
 
 impl SubscriptionRpcImpl {
+    // remove `allow` tag when https://github.com/crossbeam-rs/crossbeam/issues/404 is solved
+    #[allow(clippy::zero_ptr, clippy::drop_copy)]
     pub fn new<S: ToString>(notify_controller: NotifyController, thread_name: Option<S>) -> Self {
         let new_block_receiver =
             notify_controller.subscribe_new_block(thread_name.as_ref().unwrap().to_string());
+        let new_transaction_receiver =
+            notify_controller.subscribe_new_transaction(thread_name.as_ref().unwrap().to_string());
```
keroro520, line 136: An out-of-scope suggestion: the parameter `thread_name` must be `Some(_)` because the `unwrap` here. Hence `thread_name: Option<S>` can be changed to `thread_name: S`.

quake: updated, please review again, thanks.

rpc/README.md
```diff
@@ -2552,7 +2552,7 @@ Subscribe to a topic, if successful it returns the subscription id. For each eve
 
 #### Parameters
 
-* topic - Subscription topic (enum: new_tip_header | new_tip_block)
+* topic - Subscription topic (enum: new_tip_header | new_tip_block | new_transaction)
```
keroro520, line 2555: It's better to provide more information: what kinds of transactions will be sent if I subscribe "new_transaction"? Because it determines whether this interface is suitable for users to their scenario.
  - transactions submitted into the pending pool via RPC
  - transactions relayed from peers
  - transactions committed within blocks
  - transactions stay in the orphan_pool
  - maybe others

If it only subscribes to the first two, then I prefer renaming "new_pool_transaction".

keroro520 (changes requested):

rpc/README.md
quake: updated the rpc doc, please review again, thanks

quake:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 497
  - CKB Version: cbc023ec4da9cddee6998214c804ffcd33f23d97
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

driftluo (approved):

keroro520 (approved):

zhangsoledad (approved):

yangby-cryptape (approved):

quake:
bors r=keroro520,zhangsoledad,driftluo,yangby-cryptape

PR #2282, opened 2020-09-17
feat: create log monitor for integration test check status between nodes

chuijiaolianying:


test/src/util/log_monitor.rs
```diff
@@ -0,0 +1,58 @@
...
+    timeout: u64,
+    expected: &str,
+) -> Receiver<bool> {
+    let predicate = |file_reader: BufReader<&File>| {
+        for line in file_reader.lines() {
+            if line.unwrap().contains(expected) {
+                return true;
+            }
+        }
+        false
```
quake, line 22:
```suggestion
        file_reader.lines().any(|line| line.unwrap().contains(expected))
```

test/src/util/log_monitor.rs
```diff
@@ -0,0 +1,58 @@
...
+where
+    P: Fn(BufReader<&File>) -> bool,
+{
+    let (log_sender, log_notifier) = bounded(1);
+    let timeout = tweaked_duration(timeout);
+    let start = Instant::now();
+    let filename = node.log_file();
+    let mut file = File::open(&filename).unwrap();
+    loop {
+        let file_size = metadata(&filename).unwrap().len();
```
quake, line 42:
```suggestion
        let file_size = node.log_size();
```

test/src/node.rs
```diff
@@ -94,6 +94,15 @@ impl Node {
         &self.working_dir
     }
 
+    pub fn log_file(&self) -> String {
```
quake, line 97: node_log helper fn is enough, better to return a Path or PathBuf here if you want to add this helper fn to Node struct

quake (changes requested):

chuijiaolianying:
@quake updated, please take another look

quake (approved):

test/src/util/log_monitor.rs
```diff
@@ -0,0 +1,55 @@
...
+    let filename = node_log(node.working_dir());
+    let mut file = File::open(&filename).unwrap();
+    loop {
+        let file_size = node.log_size();
+        if seek_from != file_size {
+            file.seek(SeekFrom::Start(seek_from)).unwrap();
+            let file_reader = BufReader::new(&file);
+            if predicate(file_reader) {
+                let _ = log_sender.send(true);
+                return log_notifier;
```
keroro520, line 45: Why return a `log_notifier` but not just return `true`?

chuijiaolianying: Good question. I thought about this and I think we don't even need the channel in this condition, so I removed crossbeam and return an `Option<String>`. And return the `String` because sometime we may want to extract something from log.

quake (dismissed):

test/src/util/log_monitor.rs
keroro520: Nice. However, after your changing to return `Option<S>`, the function name `*_bool`(`monitor_log_until_bool`) becomes weird.

chuijiaolianying: yeah, updated, pls review again, thx

keroro520 (approved):

quake (approved):

quake:
bors r=quake,keroro520

doitian:
bors r=quake,keroro520

PR #2278, opened 2020-09-14
chore: integration test - put case log together

chuijiaolianying:
When debug integration case in local, we may run a single case lots time, then there is a big problem how to find the logs under an execution.

Old log structure:
```
├── ckb-it-get_blocks_timeout-net-G65fZ4
├── ckb-it-get_blocks_timeout-node0-LZRUJE
├── ckb-it-get_blocks_timeout-node1-11uL3C
```
New log structure:
```
├── ckb-it-get_blocks_timeout-UccnLI
│   ├── net
│   ├── node0
│   └── node1
```

Another change is updating `node.working_dir` and `net.working_dir` from `String` to `PathBuf`

PR #2286, opened 2020-09-18
feat: redesign the testing framework

keroro520:
Redesign the interfaces and workflow of Spec:

  * Change interface `fn before_run()` => `fn before_run() -> Vec<Node>`
  * Change interface `fn run(&mut Net)` => `fn run(&mut Vec<Node>)`
  * Change interfaces `fn modify_ckb_config` and `modify_chain_spec`
  * Initialize context within `before_run`, which is originally at worker.rs
  * Use struct name as spec name
  * Remove `protocols` and `connect_all` from `Setup` options
  * Remove `TestProtocol`. `ckb_network::SupportProtocols` is enough.

And some refactor to `Node` and `Net`.


test/src/net.rs
```diff
@@ -1,111 +1,91 @@
...
 
     pub fn working_dir(&self) -> &str {
         &self.working_dir
     }
 
-    pub fn vendor_dir(&self) -> &PathBuf {
-        &self.vendor_dir
+    pub fn vendor_dir(&self) -> PathBuf {
+        let vendor_path = VENDOR_PATH.lock();
+        (*vendor_path).clone()
```
chuijiaolianying, line 50: how about use `global::vendor()`?

chuijiaolianying:
Please resolve the conflict.

test/src/node.rs
```diff
@@ -49,163 +37,196 @@ impl Drop for ProcessGuard {
     }
 }
 
+pub struct Node {
+    working_dir: String,
```
quake, line 41: please use `PathBuf` here since we refactored it in https://github.com/nervosnetwork/ckb/pull/2278

test/src/net.rs
```diff
@@ -1,111 +1,91 @@
...
 pub type NetMessage = (PeerIndex, ProtocolId, Bytes);
 
 pub struct Net {
-    pub nodes: Vec<Node>,
-    controller: Option<(NetworkController, Receiver<NetMessage>)>,
+    consensus: Consensus,
+    protocols: Vec<SupportProtocols>,
     p2p_port: u16,
-    setup: Setup,
     working_dir: String,
```
quake, line 23: please rebase with develop branch since we refactored it to `PathBuf` in https://github.com/nervosnetwork/ckb/pull/2278

test/src/specs/alert/alert_propagation.rs
```diff
@@ -25,15 +26,16 @@ impl Default for AlertPropagation {
...
 
 impl Spec for AlertPropagation {
-    crate::name!("alert_propagation");
     crate::setup!(num_nodes: 3);
+
     // Case: alert propagation in p2p network
     //    1. create and send alert via node0; all nodes should receive the alert;
     //    2. cancel previous alert via node0; all nodes should receive the alert;
     //    3. resend the first alert, all nodes should ignore the alert.
+    fn run(&self, nodes: &mut Vec<Node>) {
```
quake, line 35: it's more common to pass slice as arg rather than vec since we just want to provide write access to individual elements (not whole vector)
```suggestion
    fn run(&self, nodes: &mut [Node]) {
```

quake (changes requested):

test/src/net.rs
keroro520: The function `fn Net::vendor()` is useless, so I have removed it. Thanks.

test/src/net.rs
keroro520: Updated.

I have resolved conflicts and pushed 2 commits: 
  * https://github.com/nervosnetwork/ckb/pull/2286/commits/6855343ccf592fbd573c7da4f49cb8c036f4b3af changes all paths into `PathBuf`

  * https://github.com/nervosnetwork/ckb/pull/2286/commits/dc97f66f239a551fe449cfe4ebbc15540cc9ef82 remove `log_size` out from `Node`. I think `log_size` is a rarely-use function, it is unnecessary to place inside `Node`.

cc @chuijiaolianying

test/src/specs/alert/alert_propagation.rs
keroro520: There are some code change the vector, e.g. https://github.com/nervosnetwork/ckb/blob/da59d97cd0d43954565adaab97ee8fff2e5e2218/test/src/specs/p2p/whitelist.rs#L24-L28

image

Maybe we can optimize all the code below, but I don't like to do it right now.

quake: Can you confirm that a simple search and replace will work?
```diff
- let node1 = nodes.pop().unwrap();
+ let node1 = nodes[0]
```
If yes, I insist on making changes in this PR since we change interface `fn run(&mut Net)` => `fn run(&mut Vec<Node>)`, otherwise we need to change all test case again with `fn run(&mut Vec<Node>)` => `fn run(&mut [Node])` in other PR

keroro520: I try to replace `nodes.pop().unwrap()` to `&nodes[x]` / `&mut nodes[x]` just now. But all failed because of the Rust mutable requirement.
  - `&nodes[x]`: In that cases, `Node` need mutable access to involve `Node::start()` and `Node::stop()`.
  - `&mut nodes[x]`: If I borrow a mutable element, then the later borrow would error:
    ```shell
    67 |         let node0 = &mut nodes[0];
       |                          ----- mutable borrow occurs here
    68 |         let node1 = &nodes[1];
    69 |         exit_ibd_mode(nodes);
       |                       ^^^^^ immutable borrow occurs here
    ```

quake: i didn't realize the borrow checking here, OK, let's keep the interface to Vec

quake (approved):

quake (approved):

chuijiaolianying (approved):

chuijiaolianying:
bors r=quake,chuijiaolianying

PR #2292, opened 2020-09-21
chore: reduce lock on protocols received

driftluo:
1. Reduce the performance cost of frequently acquiring global locks by removing other protocols registered on received
2. Reduce ping received calculation

### DIFF
--- a/network/src/protocols/mod.rs
+++ b/network/src/protocols/mod.rs
@@ -22,7 +22,7 @@ use std::{
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
-    time::{Duration, Instant},
+    time::Duration,
 };
 use tokio_util::codec::length_delimited;
 
@@ -300,12 +300,6 @@ impl ServiceProtocol for CKBHandler {
     }
 
     fn received(&mut self, context: ProtocolContextMutRef, data: Bytes) {
-        self.network_state.with_peer_registry_mut(|reg| {
-            if let Some(peer) = reg.get_peer_mut(context.session.id) {
-                peer.last_message_time = Some(Instant::now());
-            }
-        });
-
         if !self.network_state.is_active() {
             return;
         }
--- a/network/src/protocols/ping.rs
+++ b/network/src/protocols/ping.rs
@@ -19,8 +19,7 @@ use std::{
     str,
     sync::Arc,
     task::{Context, Poll},
-    time::Instant,
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::{Duration, Instant},
 };
 
 const SEND_PING_TOKEN: u64 = 0;
@@ -37,6 +36,7 @@ pub struct PingHandler {
     connected_session_ids: HashMap<SessionId, PingStatus>,
     network_state: Arc<NetworkState>,
     control_receiver: Receiver<()>,
+    start_time: Instant,
 }
 
 impl PingHandler {
@@ -46,7 +46,7 @@ impl PingHandler {
         network_state: Arc<NetworkState>,
     ) -> (PingHandler, Sender<()>) {
         let (control_sender, control_receiver) = channel(CONTROL_CHANNEL_BUFFER_SIZE);
-
+        let now = Instant::now();
         (
             PingHandler {
                 interval,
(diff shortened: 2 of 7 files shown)


network/src/protocols/ping.rs
```diff
@@ -37,6 +37,7 @@ pub struct PingHandler {
     connected_session_ids: HashMap<SessionId, PingStatus>,
     network_state: Arc<NetworkState>,
     control_receiver: Receiver<()>,
+    unix_epoch: Instant,
```
doitian, line 40: Misleading name

```suggestion
    start_time: Instant,
```

network/src/protocols/ping.rs
```diff
@@ -46,26 +47,35 @@ impl PingHandler {
...
-    fn ping_received(&self, id: SessionId) {
+    fn ping_received(&mut self, id: SessionId) {
         trace!("received ping from: {:?}", id);
-        self.mark_time(id, None);
+        self.mark_time(id, None, Instant::now());
     }
 
     fn ping_peers(&mut self, context: &ProtocolContext) {
-        let now = SystemTime::now();
+        let mut now = None;
```
doitian, line 76: This really looks ugly. What we want to do here is cache the time, so we can do it straightforwardly. 

* Add a field `now: Instant` in `PingHandler`
* Bump `now` to current time only in `ping_peers` and `received`
* Use `self.now` to get the current time.

driftluo: This is because if there is no task sent, there is no need to get the time here

The update method you said can't do this

network/src/protocols/mod.rs
```diff
@@ -268,12 +268,6 @@ impl ServiceProtocol for CKBHandler {
     }
 
     fn received(&mut self, context: ProtocolContextMutRef, data: Bytes) {
-        self.network_state.with_peer_registry_mut(|reg| {
-            if let Some(peer) = reg.get_peer_mut(context.session.id) {
-                peer.last_message_time = Some(Instant::now());
```
jjyr, line 273: Technically, this field represents the time of the last message we received, not just for the ping message. So I am not sure if you can safely remove it.

jjyr: Instead of removing here, actually, the updating on the Pong message is redundant.

driftluo: The only effect of this value is to serve as one of the reference conditions for eviction when the upper limit of the connection is reached. In other words, we do not need to update it so frequently, the cost is too high and the benefit is low

jjyr: The problem is you break the feature of this field:

1. the meaning of this field is `last_message_time`, but the implementation is a different thing if another piece of code relies on this field they may get an unexpected result.
2. The old eviction will consider all messages time, now we only consider the last ping time, a ping message is sent every 15 seconds, now an honest node will get higher probabilities to get evicted during the gap, even the node relyed more transactions or blocks than others.

I am doubtful about the performance benefit trade-off here, these two lines try to acquire a lock then update a field, itself is a very fast operation. If you find some performance bottleneck in here via some measurement, which means some other pieces of code holds the lock, I think the better solution is to digger into that lock and reduce the lock acquiring time.

driftluo: Any protocol implemented based on the handle will be forced to update this field. The more protocols, the more severe the conflict, that is, the processing message is unilaterally changed to a blocked state

doitian: Rename it to `last_ping_time`?

network/src/protocols/ping.rs
doitian: It's a tradeoff. In `received`, you already called `Instant::now()`, and in `ping_peers`, it has a very high possibility that you will call `Instance::now()`. Also `ping_peers` is only called once every 2  minutes.

doitian: Have you tried set `let now = Instance::now()` in `ping_peers` directly. I think the performance difference is trivial but the code will be much more simpler.

driftluo: The original logic is to take it directly. It is indeed simpler. Let me change it back. It is not worthwhile to trade complex logic for almost invisible improvement.

doitian: There's also a merge conflict.

network/src/protocols/ping.rs
```diff
@@ -46,26 +46,28 @@ impl PingHandler {
...
+    fn ping_received(&mut self, id: SessionId) {
         trace!("received ping from: {:?}", id);
-        self.mark_time(id, None);
+        self.mark_time(id, None, Instant::now());
     }
 
     fn ping_peers(&mut self, context: &ProtocolContext) {
-        let now = SystemTime::now();
+        let now = Instant::now();
+        let send_nonce = nonce(&now, self.start_time);
```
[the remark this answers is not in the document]
driftluo: `elapsed` need to get time again, and the current writing method does not need

network/src/protocols/ping.rs
```diff
@@ -92,22 +95,20 @@ impl PingHandler {
         }
     }
 
-    fn mark_time(&self, id: SessionId, ping_time: Option<Duration>) {
+    fn mark_time(&mut self, id: SessionId, ping_time: Option<Duration>, now: Instant) {
```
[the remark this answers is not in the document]
driftluo: Please pay attention to why I wrote this, the purpose is to share a time get

quake (dismissed):

quake:
I have to say that after this PR, the code about ping time is not easy to understand, especially the `mark_time` fn, suggest to remove this fn, split it to two fn:
```
    fn ping_received(&mut self, id: SessionId) {
        self.network_state.with_peer_registry_mut(|reg| {
            if let Some(peer) = reg.get_peer_mut(id) {
                peer.last_ping_pong_time = Some(Instant::now());
            }
        });
    }

    fn pong_received(&mut self, id: SessionId) -> bool {
        ... some status update code ...
        
        self.network_state.with_peer_registry_mut(|reg| {
            if let Some(peer) = reg.get_peer_mut(session.id) {
                peer.ping = Some(ping_time);
                peer.last_ping_pong_time = Some(now);
            }
        });
    }
```

and since `PingHandler#start_time` is only used as a nonce generator now, suggest to change it to AtomicU64.

doitian (approved):

quake (approved):

quake:
bors r=quake,doitian

driftluo:
bors retry

driftluo:
bors retry

PR #2296, opened 2020-09-23
chore: upgrade rust toolchain to 1.46

driftluo:
1.46 stabilizes some APIs that need to be used and relaxes most of the limitations of const functions


doitian:
docker builder: https://github.com/nervosnetwork/ckb-docker-builder/releases/tag/rust-1.46.0

network/src/protocols/mod.rs
```diff
@@ -148,6 +148,8 @@ impl CKBProtocol {
             protocol_name: format!("/ckb/{}", protocol_name),
             supported_versions: {
                 let mut versions: Vec<_> = versions.to_vec();
+                // https://github.com/rust-lang/rust-clippy/issues/6001
```
doitian, line 151:
```suggestion
                // TODO: https://github.com/rust-lang/rust-clippy/issues/6001
```

network/src/protocols/mod.rs
```diff
@@ -148,6 +148,8 @@ impl CKBProtocol {
             protocol_name: format!("/ckb/{}", protocol_name),
             supported_versions: {
                 let mut versions: Vec<_> = versions.to_vec();
+                // TODO: https://github.com/rust-lang/rust-clippy/issues/6001
+                #[allow(clippy::unnecessary_sort_by)]
```
quake, line 152: I think this lint can by removed by:
```
versions.sort_by_key(|v| std::cmp::Reverse(v));
```

driftluo: I have tried this way of writing, but it needs clone

yangby-cryptape (approved):

network/src/protocols/mod.rs
quake: aha, I didn't realize that `sort_by_key` returns an owned instead of ref, after re-read original code, I think we may use sort and reverse, which avoids cloning twice in `versions.to_vec()`:
```
                let mut versions: Vec<_> = versions.to_vec();
                versions.sort();
                versions.reverse();
                versions
```

driftluo: This can avoid `allow(clippy::unnecessary_sort_by)`, but actually do one more traversal. Although performance is not the most critical during startup, we should not change the correct code because of the wrong prompt

doitian (changes requested):
Hold until ckb 0.38.0-pre

doitian (approved):

doitian:

PR #2297, opened 2020-09-25
chain freezer

zhangsoledad:
## Description

Introduce chain freezer, Inspired by  [[Splitting the data directory]](https://en.bitcoin.it/wiki/Splitting_the_data_directory) and [[geth-v1-9-0]](https://blog.ethereum.org/2019/07/10/geth-v1-9-0/#freezer)

Now, separated database into two parts, recent block and ancient history.
if your data directory is located on a magnetic disk, you can link db to an SSD drive to improve performance.
If your data directory is on an SSD: you can link ancient to an HDD drive to save space.

The `ancient` path can be specified by config file, but you will need to copy the existing ancient folder manually and then start CKB with `ancient` set to the correct path.

When you update from an older version, CKB will automatically migrate data from the rocksdb database into the freezer.

---

* [x] Code review @doitian @quake 
* [x] Acceptance test @chuijiaolianying 
* [x] Rebase and add documentations @zhangsoledad


quake:
LGTM, one suggestion for error handling, there are 30+ `.map_err(internal_error)?` in freezer code, I think most of them should be replaced by `std::io::Error` directly, for example:
```
fn open_append<P: AsRef<Path>>(&self, path: P) -> Result<(File, u64), Error> 
=>
fn open_append<P: AsRef<Path>>(&self, path: P) -> Result<(File, u64), std::io::Error> 
```

That would reduce a lot of `IoError` => `InternalError` convention.

quake (dismissed):

quake (approved):

store/src/store.rs
```diff
@@ -83,6 +92,18 @@ pub trait ChainStore<'a>: Send + Sync + Sized {
         .collect()
     }
 
+    fn get_unfrozen_block(&'a self, h: &packed::Byte32) -> Option<BlockView> {
```
driftluo, line 95: Maybe need to write a panic warning here? Or turn any panic into None and return?

freezer/src/freezer_files.rs
```diff
@@ -0,0 +1,511 @@
...
+        helper::truncate_file(&mut self.head.file, new_index.offset)?;
+        self.head.bytes = new_index.offset;
+        self.number.store(item + 1, Ordering::SeqCst);
+        Ok(())
+    }
+
+    pub fn preopen(&mut self) -> Result<(), IoError> {
+        self.release_after(0);
+
+        for id in self.tail_id..self.head_id {
```
driftluo, line 291: This opens all read-only files that have reached the max size, do you need to consider the lazy part to reduce the number of fd fetching?

doitian (approved):

chuijiaolianying (approved):
I test freezer manually in my local, it looks good to me. And I will create a pr with those test cases in ckb-internal later.

store/src/store.rs
zhangsoledad: This copy from 
```
    fn get_block(&'a self, h: &packed::Byte32) -> Option<BlockView> {
        self.get_block_header(h).map(|header| {
            let body = self.get_block_body(h);
            let uncles = self
                .get_block_uncles(h)
                .expect("block uncles must be stored");
            let proposals = self
                .get_block_proposal_txs_ids(h)
                .expect("block proposal_ids must be stored");
            BlockView::new_unchecked(header, uncles, body, proposals)
        })
```
It fail-fast on inconsistency when header exists but uncles or proposals not.

zhangsoledad: I'm not going to change this in this PR

freezer/src/freezer_files.rs
zhangsoledad: https://github.com/nervosnetwork/ckb/pull/2393

driftluo (approved):

doitian (approved):

yangby-cryptape (approved):

yangby-cryptape:

doitian:
bors r=doitian,yangby-cryptape,quake,chuijiaolianying

PR #2315, opened 2020-10-12
chore: log shows error

driftluo:
When the local data is lower than minwork, the log display is incorrect


quake (approved):

chuijiaolianying (approved):

quake:
bors r=quake,chuijiaolianying

PR #2348, opened 2020-11-06
chore: change inflight request tx reach limit status to 1xx

driftluo:
This state is a normal state, which means that when the transaction is in the proposal phase, the node is not involved in it, and the transaction needs to be pulled remotely during compact block message processing, and the current state is pulling


quake (dismissed):

keroro520 (approved):

yangby-cryptape (approved):

doitian (approved):

doitian:

PR #2349, opened 2020-11-06
feat: metricize some indicators

keroro520:
Update the metrics to adjust to newly designed dashboards.


keroro520:
Post shapshots here:
* https://snapshot.raintank.io/dashboard/snapshot/OQTHVFABgvLfjkuz7jGU8oWSdYVIkvw8
* https://snapshot.raintank.io/dashboard/snapshot/SLQJkf25TKknzY0VT1ltwFLqCFSjIa87
* https://snapshot.raintank.io/dashboard/snapshot/39KpWpUi4hXOdNyO0YTgk22N8r5Xceui
* https://snapshot.raintank.io/dashboard/snapshot/ypg6Q8WhTmE9NG3w46gRAIonUFjEaWKo
* https://snapshot.raintank.io/dashboard/snapshot/ESr93vjQNKzdEoQUE9Yhprt7TKrwaSXI

chain/src/chain.rs
```diff
@@ -278,18 +278,20 @@ impl ChainService {
...
-                block.header().number(),
-                block.header().hash()
-            );
+            warn!("receive 0 number block: 0-{}", block_hash);
         }
-        self.insert_block(block, switch).map(|ret| {
-            debug!("finish processing block");
-            ret
-        })
+        let is_better = self.insert_block(block, switch)?;
```
chuijiaolianying, line 289: `is_better` seems used for checking the block is tip or uncle block, so I think `is_tip` or `is_best` will be more meaningful

chain/src/chain.rs
```diff
@@ -398,8 +399,9 @@ impl ChainService {
             );
             self.find_fork(&mut fork, current_tip_header.number(), &block, ext);
             if !fork.detached_blocks.is_empty() {
-                metrics!(gauge, "ckb-chain.reorg", fork.attached_blocks.len() as i64, "type" => "attached");
-                metrics!(gauge, "ckb-chain.reorg", fork.detached_blocks.len() as i64, "type" => "detached");
+                // We care about "the size of reorg", then the length of attached blocks covers it.
+                let attached_blocks = fork.attached_blocks().len();
+                metrics!(counter, "ckb.reorg_attached_blocks", attached_blocks as u64);
```
chuijiaolianying, line 404: I think detached blocks length also a useful metric here, or it just always being `attached_blocks + 1`?

sync/src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+
+pub(crate) fn send_message<Message: Entity>(
+    protocol_id: ProtocolId,
+    nc: &dyn CKBProtocolContext,
+    peer_index: PeerIndex,
+    message: &Message,
+) -> Status {
+    let name = message_name(protocol_id, message);
+    if let Err(err) = nc.send_message(protocol_id, peer_index, message.as_bytes()) {
+        return StatusCode::Network.with_context(format!("Send {}: {:?}", name, err));
```
chuijiaolianying, line 15: can we add a `debug!` log here? then we don't need check status every time after call this function

chain/src/chain.rs
keroro520: What we actually care about are:
  - The frequency of reorg happens
  - The scale of reorg

Back to the discussion. I agree that the length of the detached is meaningful. However, most of the time, detached-length is very close to attached-length(`detached + 1 == attached`). We can estimate the scale of reorg based attached-length only, that's enough for us.

sync/src/utils.rs
keroro520: Thanks. It should add a debug log here as you say. I will update it later.

sync/src/synchronizer/mod.rs
```diff
@@ -106,20 +108,23 @@ impl BlockFetchCMD {
...
             SupportProtocols::Sync.protocol_id(),
             message.as_bytes(),
         ) {
-            debug!("synchronizer send GetBlocks error: {:?}", err);
+            debug!("synchronizer send_getblocks error: {:?}", err);
         }
-        crate::synchronizer::metrics_counter_send(message.to_enum().item_name());
+
+        let bytes = message.as_bytes().len() as u64;
+        metrics!(counter, "ckb.messages_total", 1, "direction" => "out", "name" => "GetBlocks");
```
chuijiaolianying, line 126: if `messages_total` always being `1`, why we need this metric?

keroro520: It's a counter. A global counter named "ckb.messages_total" will increase by 1 when call `metrics!(counter, "ckb.messages_total", 1, "direction" => "out", "name" => "GetBlocks");`

keroro520:
@chuijiaolianying updated, PTAL again

chuijiaolianying (approved):

chuijiaolianying (approved):

robcxyz:
@keroro520 - Are any of these grafana dashboards public or are you able to share their json configs? Thanks!

keroro520:
> @keroro520 - Are any of these grafana dashboards public or are you able to share their json configs? Thanks!

Not public Grafana instance yet. I just put the dashboard json configs into [ckb-analyzer dashboards](https://github.com/keroro520/ckb-analyzer/tree/main/dashboards/sourced-promethreus), please take a look.

robcxyz:
@keroro520 This is perfect. Thank you.

doitian:
@quake please help to review

keroro520:
Split into #2507 #2508 #2509

PR #2356, opened 2020-11-12
fix: chain spec hash backward compatibility

zhangsoledad:
make chain spec params skip serializing if option is none


quake (approved):

doitian (approved):

doitian:
bors r=quake,doitian

PR #2351, opened 2020-11-09
feat: add with_sentry feature

quake:
move sentry to optional dependency, reduce dependency libs on other target (wasm32, etc)

PR #2381, opened 2020-12-01
chore(integration): add sentry section into ckb.toml in integration test

chuijiaolianying:
After https://github.com/nervosnetwork/ckb/pull/2351, integration test cannot run for ckb which builds with `--features with_sentry`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,7 @@ integration: submodule-init setup-ckb-test ## Run integration tests in "test" di
 	RUST_BACKTRACE=1 RUST_LOG=${INTEGRATION_RUST_LOG} test/run.sh -- --bin ../target/debug/ckb ${CKB_TEST_ARGS}
 
 .PHONY: integration-release
-integration-release: submodule-init setup-ckb-test
-	cargo build --release --features deadlock_detection
+integration-release: submodule-init setup-ckb-test prod
 	RUST_BACKTRACE=1 RUST_LOG=${INTEGRATION_RUST_LOG} test/run.sh --release -- --bin ../target/release/ckb ${CKB_TEST_ARGS}
 
 ##@ Document
--- a/util/app-config/src/app_config.rs
+++ b/util/app-config/src/app_config.rs
@@ -40,6 +40,7 @@ pub struct CKBAppConfig {
     pub logger: LogConfig,
     /// TODO(doc): @doitian
     #[cfg(feature = "with_sentry")]
+    #[serde(default)]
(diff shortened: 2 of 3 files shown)


quake:
I'm confused because we don't use sentry in the integration test, why sentry configuration is required here?

chuijiaolianying:
> I'm confused because we don't use sentry in the integration test, why sentry configuration is required here?

If ckb builds without `--features with_sentry`, sentry configuration is not required and everything works well. 

But if ckb builds with `--features with_sentry` (such as release version), sentry configuration is required, or we will meet a config error: `Config Error: Error { inner: ErrorInner { kind: Custom, line: Some(88), col: 0, at: Some(1738), message: "missing field `sentry`", key: [] } }`

Besides, I think in CI we should use the same command when build ckb like build release version as far as possible.

test/src/node.rs
```diff
@@ -89,6 +90,14 @@ impl Node {
...
         fs::write(&app_config_path, toml::to_string(&app_config).unwrap()).unwrap();
 
+        // write sentry section into ckb configuration file
+        // so that we can run integration test for ckb which builds with or without sentry feature
+        let mut file = fs::OpenOptions::new()
+            .append(true)
+            .open(&app_config_path)
+            .unwrap();
+        writeln!(file, "\n[sentry]\ndsn = \"\"").unwrap();
+
```
keroro520, line 100: https://github.com/nervosnetwork/ckb/blob/develop/test/template/ckb.toml#L18-L23 the test-used `ckb.toml` has configured `[sentry]`, why attach a new one here?

test/src/node.rs
```diff
@@ -18,6 +18,7 @@ use std::borrow::Borrow;
 use std::collections::HashSet;
 use std::convert::Into;
 use std::fs;
+use std::io::prelude::*;
```
keroro520, line 21: Why we need the import of `std::io::preclude::*`? Try to remove it or import what we require only.

keroro520 (changes requested):

chuijiaolianying:
@quake  @keroro520  please take another look, thx

quake (approved):

keroro520 (approved):

quake:
bors r=quake,keroro520

doitian:
bors retry

doitian:
bors retry

doitian (approved):

doitian:
bors retry

chuijiaolianying:
bors retry

PR #2386, opened 2020-12-02
refactor: replace `failure` by `thiserror` and `anyhow`

yangby-cryptape:
### Reasons
- [RUSTSEC-2020-0036: `failure`: `failure` is officially deprecated/unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0036.html)
- Most errors in `CKB` are not implement `::std::error::Error`.

### Commits
- [refactor: replace `failure` by `thiserror` and `anyhow` for `ckb-error` and crates which depend on it](https://github.com/nervosnetwork/ckb/pull/2386/commits/1694aec4f879cce5401b46a6b3dfd3188dc5aea8)
  This commit is the key commit for this PR, other commits are just equivalent transformations.
- [refactor: replace `failure` by `thiserror` and `anyhow` for all other ckb crates](https://github.com/nervosnetwork/ckb/pull/2386/commits/2e067a7c5b02eed6a9b565b5c2c5a73197bb227f)
  Just equivalent transformations.
- [feat: let `AnyError` be cloneable](https://github.com/nervosnetwork/ckb/pull/2386/commits/24a90e833ac94b997a9f965388adb393c5a96d8f)

~~p.s. This PR is base on an unmerged PR: [CKB PR-2387: ci: skip RUSTSEC-2020-0077 temporarily](https://github.com/nervosnetwork/ckb/pull/2387).~~

### Tips for Reviewers

- The field name `source` has special function in `thiserror`, so I rename the fields which were named `source` to `inner`.

  > The Error trait's source() method is implemented to return whichever field has a #[source] attribute or is named source, if any.

- Replaced `FailureError` by `AnyError(Arc<anyhow::Error>)`

- Replaced `pub struct $error { kind: failure::Context<$error_kind}> }` by `pub struct $error { kind: $error_kind, inner: AnyError }`

- Re-export `thiserror` in `ckb-error`.

- The old errors which has an error kind had different implementations of `::std::fmt::Display`.

  For example:
  - Some checked `alternate` but others didn't.
  - Some are `{kind}: {reason}`, some are `{kind}({reason})` and others are `{kind}`.

  Since I use a macro to unify all errors which has an error kind, all implementations of `::std::fmt::Display` are same now.
  And few implementations are not as same as the old implementations.

  For example, `Internal(CapacityOverflow)` became `Internal(CapacityOverflow(OccupiedCapacity: overflow))`.

- I only add documentations for new items which were introduced in this PR.

  I left those already existed `TODO(doc)`, some of them are integrated into the macro `def_error_base_on_kind`, call the macro with one more argument can add those documentations.

### DIFF
--- a/test/src/specs/sync/chain_forks.rs
+++ b/test/src/specs/sync/chain_forks.rs
@@ -10,9 +10,9 @@ use ckb_types::{
     prelude::*,
     H256,
 };
-use failure::_core::time::Duration;
 use log::info;
 use std::thread::sleep;
+use std::time::Duration;
 
 pub struct ChainFork1;
 
--- a/tx-pool/src/error.rs
+++ b/tx-pool/src/error.rs
@@ -1,69 +1,68 @@
 //! The error type for Tx-pool operations
 
-use ckb_error::{Error, ErrorKind};
+use ckb_error::{
+    impl_error_conversion_with_adaptor, impl_error_conversion_with_kind, prelude::*, Error,
+    ErrorKind, InternalError, InternalErrorKind, OtherError,
+};
 use ckb_fee_estimator::FeeRate;
 use ckb_types::packed::Byte32;
-use failure::Fail;
-use tokio::sync::mpsc::error::TrySendError as TokioTrySendError;
+use tokio::sync::{mpsc::error::TrySendError, oneshot::error::RecvError};
 
 /// TX reject message
-#[derive(Debug, PartialEq, Clone, Eq, Fail)]
+#[derive(Error, Debug, PartialEq, Clone, Eq)]
 pub enum Reject {
     /// Transaction fee lower than config
-    #[fail(
-        display = "The min fee rate is {} shannons/KB, so the transaction fee should be {} shannons at least, but only got {}",
-        _0, _1, _2
-    )]
+    #[error("The min fee rate is {0} shannons/KB, so the transaction fee should be {1} shannons at least, but only got {2}")]
     LowFeeRate(FeeRate, u64, u64),
 
     /// Transaction exceeded maximum ancestors count limit
-    #[fail(display = "Transaction exceeded maximum ancestors count limit, try send it later")]
+    #[error("Transaction exceeded maximum ancestors count limit, try send it later")]
     ExceededMaximumAncestorsCount,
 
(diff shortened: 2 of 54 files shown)


yangby-cryptape:
A bad night for me!
- First push, I forget to `git add` a new file.
- Second push, I got an error from the test `tests::block_assembler::test_prepare_uncles` which seems has randomness.
- Third push, [RUSTSEC-2020-0077: memmap: memmap is unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0077.html) was published.

test/src/specs/sync/chain_forks.rs
```diff
@@ -10,9 +10,9 @@ use ckb_types::{
     prelude::*,
     H256,
 };
-use failure::_core::time::Duration;
```
yangby-cryptape, line 13: [According to the `failure`](https://github.com/rust-lang-nursery/failure/blob/135e2a3b9af422d9a9dc37ce7c69354c9b36e94b/src/lib.rs#L31-L34):

```rust
// Re-export libcore using an alias so that the macros can work without
// requiring `extern crate core` downstream.
#[doc(hidden)]
pub extern crate core as _core;
```

tx-pool/src/error.rs
```diff
@@ -1,69 +1,68 @@
...
+    #[error("Malformed {0} transaction")]
     Malformed(String),
 }
 
-impl From<Reject> for Error {
-    fn from(error: Reject) -> Self {
-        error.context(ErrorKind::SubmitTransaction).into()
-    }
-}
+impl_error_conversion_with_kind!(Reject, ErrorKind::SubmitTransaction, Error);
```
doitian, line 35: I strongly recommend that adding a brief description about what the macro does. This suggestion can be applied to all the macro invocations in this PR.

Example

```suggestion
// Implement conversion from `Reject` to `Error` with the error kind `SubmitTransaction`.
impl_error_conversion_with_kind!(Reject, ErrorKind::SubmitTransaction, Error);
```

doitian (approved):

driftluo (approved):

yangby-cryptape:
bors r=doitian,driftluo

PR #2390, opened 2020-12-05
fix: don't open db when disable indexer module, fix deprecated method response

driftluo:
1. deprecated method name change to "deprecated.xxx", need to remove this prefix
2. indexer module always open DB on default，and it will leave log/DB files on disk, these are not needed

### DIFF
--- a/indexer/src/lib.rs
+++ b/indexer/src/lib.rs
@@ -5,5 +5,5 @@ mod migrations;
 mod store;
 mod types;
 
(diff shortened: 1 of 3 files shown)


quake:
I think we can remove indexer related code completely since we have deprecated it in 0.36 and now is 0.39

doitian (approved):

yangby-cryptape (approved):

yangby-cryptape:

PR #2392, opened 2020-12-07
ci: skip RUSTSEC-2020-0082 temporarily since not affected

yangby-cryptape:
### About Security Issue "RUSTSEC-2020-0082"

- [`ordered_float:NotNan` may contain `NaN` after panic in assignment operators](https://github.com/RustSec/advisory-db/blob/dec05d79abc4e468cc225e36af6351d19e7b8063/crates/ordered-float/RUSTSEC-2020-0082.md)
- `ordered_float` released two patched versions: `1.1.1` and `2.0.1`.

### Why CKB Couldn't Update `ordered_float` to the Patched Version

- CKB doesn't use `ordered_float` directly, CKB introduces the dependency via `heim`.
- `heim < 0.1` uses [`ordered_float = "~1.0"`](https://github.com/heim-rs/heim/blob/v0.0.11/heim-process/src/sys/common.rs#L21); `heim >= 0.1` uses [`ordered_float = "~2.0"`](https://github.com/heim-rs/heim/blob/2a6922ead22ef6e095de0be94779c82b5d1d4ca1/heim-process/Cargo.toml#L24), but it doesn't release.

### Why CKB Could Just Skip RUSTSEC-2020-0082 Temporarily

CKB **are not affected** by this security issue: `heim` only uses the `NotNan` in [a private field](https://github.com/heim-rs/heim/blob/v0.0.11/heim-process/src/sys/common.rs#L21), and doesn't do any assignment operations.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -8,6 +8,8 @@ ignore = [
     "RUSTSEC-2020-0036", # TODO failure is officially deprecated/unmaintained, but still a lot of required crates are dependent on it
     "RUSTSEC-2020-0043", # TODO ws allows remote attacker to run the process out of memory, since it is no longer actively maintained, we couldn't fix it in the short term
     "RUSTSEC-2020-0056", # We did not use the `stdweb` library, only `wasm32-unknown-unknown` would use `getrandom` and `wasm-bindgen`, `stdweb` would only be used in cargo-web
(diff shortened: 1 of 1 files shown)


doitian (approved):

driftluo (approved):

yangby-cryptape:
bors r=doitian,driftluo

PR #2445, opened 2020-12-16
chore: filter self listen first

driftluo:
Filter own monitor first, otherwise port reuse will not be enabled

### DIFF
--- a/network/src/protocols/discovery/state.rs
+++ b/network/src/protocols/discovery/state.rs
@@ -11,7 +11,7 @@ use p2p::{
 use super::{
     addr::AddrKnown,
     protocol::{encode, DiscoveryMessage, Node, Nodes},
-    MAX_ADDR_TO_SEND,
+    AddressManager, MAX_ADDR_TO_SEND,
 };
 
 // FIXME: should be a more high level version number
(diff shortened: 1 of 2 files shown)


network/src/protocols/discovery/state.rs
```diff
@@ -35,15 +35,26 @@ pub struct SessionState {
...
+                    // Verify self is a public node first
+                    // if not, try to make public network nodes broadcast hole punching information
+                    if addr_manager.is_valid_addr(address) {
+                        multiaddr_to_socketaddr(address).map(|socket_addr| socket_addr.port())
+                    } else {
+                        None
+                    }
+                })
+                .next()
+                .flatten();
```
doitian, line 57: `.map().next().flatten()` is different from `.filter_map().next()`. The former always return the first address map result.

driftluo: Maybe I should use `flat_map().next()`, now next returns an `Option<Option<u16>>`, use `flatten` to flattens nested `Option`

doitian (approved):

zhangsoledad (approved):

zhangsoledad:
bors r=doitian,zhangsoledad

PR #2379, opened 2020-11-26
feat: let the consensus params `orphan_rate_target` to be configurable

yangby-cryptape:
### Features

- We can test different `orphan_rate_target` in our internal test networks or mock tests.

### Potential Broken Changes

Only break those customized chains which have `genesis_epoch_length % 40 != 0`.

PR #2448, opened 2020-12-20
fix: genesis hash rate estimate

zhangsoledad:


spec/src/consensus.rs
```diff
@@ -196,11 +196,10 @@ pub fn build_genesis_epoch_ext(
...
     let remainder_reward = Capacity::shannons(epoch_reward.as_u64() % genesis_epoch_length);
 
-    // let genesis_orphan_count = genesis_epoch_length * genesis_orphan_rate.0 / genesis_orphan_rate.1;
+    let genesis_orphan_count =
+        genesis_epoch_length * genesis_orphan_rate.0 as u64 / genesis_orphan_rate.1 as u64;
     let genesis_hash_rate = compact_to_difficulty(compact_target)
-        * genesis_epoch_length
-        * (genesis_orphan_rate.0 + genesis_orphan_rate.1)
-        / genesis_orphan_rate.1
+        * (genesis_epoch_length + genesis_orphan_count)
```
zhangsoledad, line 202: :fearful: :hammer:

doitian (approved):

spec/src/consensus.rs
doitian: Which versions does it affect?

yangby-cryptape: Ref: [PR-2379: feat: let the consensus params `orphan_rate_target` to be configurable](https://github.com/nervosnetwork/ckb/pull/2379#issue-527911739)

yangby-cryptape (approved):

spec/src/consensus.rs
yangby-cryptape: I still think this should be high precision computation according the description in [RFC](https://github.com/nervosnetwork/rfcs/blob/v2020.01.15/rfcs/0020-ckb-consensus-protocol/0020-ckb-consensus-protocol.md#computing-the-next-epochs-main-chain-block-number).
:tired_face: But we **CAN NOT** change it.

zhangsoledad:

spec/src/consensus.rs
doitian:
```
$ git tag --contains be268a3aa
v0.39.0-rc1
v0.39.0-rc2
```

Only v0.39.0-pre versions

PR #2458, opened 2020-12-22
ci: fix potential failure in integration test TransactionRelayLowFeeRate

yangby-cryptape:
### Description

- In the function `gen_spendable(..)`, 12 blocks are created.
  https://github.com/nervosnetwork/ckb/blob/aff994e68f6bb7935e2dd371c92a27be72936c97/test/src/specs/relay/transaction_relay_low_fee_rate.rs#L21
- Then, the cell which is created in the 12th block, will be used as a spendable cell.
  https://github.com/nervosnetwork/ckb/blob/aff994e68f6bb7935e2dd371c92a27be72936c97/test/src/specs/relay/transaction_relay_low_fee_rate.rs#L24-L29
- So the spendable cell will be broadcasted from `node-0` to `node-1`.
https://github.com/nervosnetwork/ckb/blob/aff994e68f6bb7935e2dd371c92a27be72936c97/test/src/specs/relay/transaction_relay_low_fee_rate.rs#L34-L37
- Then, we expect to get a low fee rate error.
  https://github.com/nervosnetwork/ckb/blob/aff994e68f6bb7935e2dd371c92a27be72936c97/test/src/specs/relay/transaction_relay_low_fee_rate.rs#L39-L44
- But, if the transaction which uses the spendable cell as input was broadcasted faster than 12th block synchronization. **We will get an unknown outpoint error.**

### DIFF
--- a/test/src/specs/relay/transaction_relay_low_fee_rate.rs
+++ b/test/src/specs/relay/transaction_relay_low_fee_rate.rs
@@ -1,4 +1,5 @@
 use crate::node::connect_all;
+use crate::node::waiting_for_sync;
 use crate::util::cell::{as_input, as_output, gen_spendable};
 use crate::util::log_monitor::monitor_log_until_expected_show;
 use crate::util::mining::out_ibd_mode;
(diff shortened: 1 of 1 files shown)


test/src/specs/relay/transaction_relay_low_fee_rate.rs
```diff
@@ -31,6 +32,17 @@ impl Spec for TransactionRelayLowFeeRate {
             .rpc_client()
             .dry_run_transaction(low_fee.data().into())
             .cycles;
+
+        let ret = wait_until(5, || {
```
quake, line 36:
```suggestion
        waiting_for_sync(nodes);
```

quake (approved):

chuijiaolianying (approved):

quake:
bors r=quake,chuijiaolianying

PR #2466, opened 2020-12-23
ci: fix potential failure in integration tests, such as BlockSyncRelayerCollaboration

yangby-cryptape:
### Description

- `GetBlocks` is sent by `BlockFetchCMD` in thread `BlockDownload`.
  https://github.com/nervosnetwork/ckb/blob/32ee42273620d15e20da2e664de00bb424eef3b3/sync/src/synchronizer/mod.rs#L631
  https://github.com/nervosnetwork/ckb/blob/32ee42273620d15e20da2e664de00bb424eef3b3/sync/src/synchronizer/mod.rs#L604-L607

- `HeadersProcess` is asynchronously running in global runtime.

- `GetBlocks` and `HeadersProcess` are parallel.

- Choose integration test `BlockSyncRelayerCollaboration` as the example to explain the issue:

  - After sent all headers to node0, the test check the next `SyncMessage` immediately.
    https://github.com/nervosnetwork/ckb/blob/32ee42273620d15e20da2e664de00bb424eef3b3/test/src/specs/sync/block_sync.rs#L283-L289
    https://github.com/nervosnetwork/ckb/blob/32ee42273620d15e20da2e664de00bb424eef3b3/test/src/specs/sync/block_sync.rs#L491-L502

    - Nearly all the time, `HeadersProcess` can finish all tasks, **but when not, the check will fail**.

      If `BlockFetchCMD` runs when `HeadersProcess` only finished processing 3 headers, a `GetBlocks(length = 3)` will be sent, after `HeadersProcess` finished processing all headers, a `GetBlocks(length = 13)` will be sent again.
      **The test should check the last `GetBlocks`, not the first.**

    - How to fix?
      - We could check if the summation of `GetBlocks` lengths are 16.
        https://github.com/nervosnetwork/ckb/blob/32ee42273620d15e20da2e664de00bb424eef3b3/sync/src/lib.rs#L30-L32
      - Or, still check the last hash. **I chose this.**

    - P.S. The description for integration test `BlockSyncRelayerCollaboration` is incorrect: the first block was mined in `out_ibd_mode(nodes);`.

### DIFF
--- a/sync/src/synchronizer/mod.rs
+++ b/sync/src/synchronizer/mod.rs
@@ -616,7 +616,7 @@ impl Synchronizer {
                         .expect("download thread can't start");
                 }
             },
-            _ => {
+            None => {
                 for peer in self.get_peers_to_fetch(ibd, &disconnect_list) {
                     if let Some(fetch) = self.get_blocks_to_fetch(peer, ibd) {
                         for item in fetch {
--- a/test/src/specs/sync/block_sync.rs
+++ b/test/src/specs/sync/block_sync.rs
@@ -244,12 +244,6 @@ impl Spec for BlockSyncOrphanBlocks {
     }
 }
 
-/// test case:
-/// 1. generate 1-17 block
-/// 2. sync 1-16 header to node
-/// 3. sync 2-16 block to node
-/// 4. sync 1 block to node
-/// 5. relay 17 block to node
 pub struct BlockSyncRelayerCollaboration;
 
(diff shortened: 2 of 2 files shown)


yangby-cryptape:
Hold since the description is not suitable for these integration tests, these integration tests have other reasons why fails randomly.
I will fix other integration tests then back to debug these.

yangby-cryptape:
I will reopen this pull request when I could reproduce this bug more conclusive.

PR #2365, opened 2020-11-18
tx pool callback

zhangsoledad:

PR #2470, opened 2020-12-24
docs: fix reject tx subscription RPC doc

doitian:
Fix RPC doc added in #2365


devtools/doc/rpc.py
```diff
@@ -501,6 +504,10 @@ def handle_data(self, data):
             self.variant_parser.handle_data(data)
 
     def write(self, file):
+        if self.name == 'PoolTransactionReject':
```
keroro520, line 507: Just curious: why `PoolTransactionReject` special?

yangby-cryptape (approved):

quake (approved):

devtools/doc/rpc.py
doitian: It's an enum with associated fields.

doitian:
bors r=quake,yangby-cryptape

PR #2471, opened 2020-12-25
chore: if the file in ".git/HEAD" isn't existed, ckb-bin will be re-linked every time when "cargo build"

yangby-cryptape:
### Issue

Each time when I run `make integration` or some other commands, about 15 seconds will be spent for waiting linking `ckb-bin`.

It wastes me a lot of times. (More than 40 times / 10 minutes each day.)

After this PR, when run `cargo build` or `cargo run` but nothing changed, `ckb` won't be linked again.

### How to reproduce?

There are a lots of reason why the file in ".git/HEAD" isn't existed. It's common. For example, calling `git gc`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,6 +9,7 @@ fn rerun_if_changed(path_str: &str) -> bool {
         || path.starts_with("docker")
         || path.starts_with("docs")
         || path.starts_with("test")
+        || path.starts_with(".github")
     {
         return false;
     }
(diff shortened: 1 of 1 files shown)


TheWaWaR (approved):

quake (approved):

quake:
bors r=quake,TheWaWaR

PR #2473, opened 2020-12-27
disable freezer by default

zhangsoledad:


quake (approved):

util/app-config/src/configs/store.rs
```diff
@@ -15,6 +15,13 @@ pub struct Config {
     pub block_uncles_cache_size: usize,
     /// TODO(doc): @doitian
     pub cellbase_cache_size: usize,
+    /// whether enable freezer
+    #[serde(default = "default_freezer_enable")]
+    pub freezer_enable: bool,
```
yangby-cryptape, line 20: Just an advice: if we may add more options to freezer, may be a standalone section is better.
For example:  `path`, `max_open_files_limit`, `max_freeze_limit`, `threshold_epoch` and so on.

yangby-cryptape (approved):

doitian:
bors r=quake,yangby-cryptape

PR #2475, opened 2020-12-28
chore(doc): fix doc about epoch in type `ChainInfo`

chuijiaolianying:
Chain info's epoch come from tip_header https://github.com/nervosnetwork/ckb/blob/c3133ad66de38c03685c218b97d9f7845ad83d1d/rpc/src/module/stats.rs#L102-L108

And according to  https://github.com/nervosnetwork/ckb/blob/c3133ad66de38c03685c218b97d9f7845ad83d1d/util/types/src/core/views.rs#L405
`epoch` in `HeaderView` is `EpochNumberWithFraction` instead of `EpochNumber`

PR #2476, opened 2020-12-28
chore: force explictly pointing out EpochNumberWithFraction

keroro520:
During reviewing https://github.com/nervosnetwork/ckb/pull/2475, I realize that the implicit `impl From<core::EpochNumberWithFraction> for JsonUint<u64>` is may be misleading.  Better force explicitly specifying `epoch.number()` or `epoch.full_value()`.

https://github.com/nervosnetwork/ckb/blob/80320313b3e81d83bb0569312fe300099defb494/util/jsonrpc-types/src/uints.rs#L185-L189

### DIFF
--- a/util/jsonrpc-types/src/uints.rs
+++ b/util/jsonrpc-types/src/uints.rs
@@ -182,12 +182,6 @@ impl From<core::Capacity> for JsonUint<u64> {
     }
 }
 
-impl From<core::EpochNumberWithFraction> for JsonUint<u64> {
-    fn from(value: core::EpochNumberWithFraction) -> Self {
-        JsonUint(value.full_value())
-    }
-}
-
 impl From<JsonUint<u64>> for core::Capacity {
     fn from(value: JsonUint<u64>) -> Self {
         core::Capacity::shannons(value.value())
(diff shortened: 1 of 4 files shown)


util/jsonrpc-types/src/uints.rs
```diff
@@ -182,12 +182,6 @@ impl From<core::Capacity> for JsonUint<u64> {
     }
 }
 
-impl From<core::EpochNumberWithFraction> for JsonUint<u64> {
-    fn from(value: core::EpochNumberWithFraction) -> Self {
-        JsonUint(value.full_value())
```
chuijiaolianying, line 187: I don't understand why this is dangerous?

keroro520:
```rust
// core_epoch: core::EpochNumberWithFraction

let epoch: jsonrpc_types::EpochNumber = core_epoch.into();
let epoch: jsonrpc_types::EpochNumberWithFraction = core_epoch.into();
```

I just worry that the user mix-uses transition from `core::EpochNumberWithFraction` to `EpochNumber` and `EpochNumberWithFraction`. Although lacking a strong reason, I shall close this PR. Thanks

keroro520:
Close with https://github.com/nervosnetwork/ckb/pull/2476#discussion_r549692536

PR #2455, opened 2020-12-22
fix: relay and sync should be order independent

yangby-cryptape:
Fix #2450.

Issue #2450, opened 2020-12-21
some integration tests fail randomly: ckb-relay doesn't work without ckb-sync


yangby-cryptape:
#### Description

- The `Peers` is stored in the shared state between sync and relay protocol.

  https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/relayer/mod.rs#L70-L72
  https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/types/mod.rs#L1178-L1180
  https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/types/mod.rs#L1466-L1478

- The way to add a peer into `Peers` is **only in sync protocol**.
https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/synchronizer/mod.rs#L354-L362
https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/types/mod.rs#L861-L869

- But a lot of messages of relay protocol are dependent on `Peers`.

- So, if a node receive relay messages and handle them before the sync protocol started, it will do nothing.

#### Affected

This issue causes a lot of integration tests failures, for example:

- [LastCommonHeaderForPeerWithWorseChain](https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/test/src/specs/sync/last_common_header.rs#L49)
  https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/relayer/compact_block_process.rs#L94-L95
- [TransactionRelayTimeout](https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/test/src/specs/relay/transaction_relay.rs#L100-L103) & [RelayInvalidTransaction](https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/test/src/specs/relay/transaction_relay.rs#L136)
  https://github.com/nervosnetwork/ckb/blob/555d245a977ac77d20f39fc0dd1d34456b2b453a/sync/src/relayer/transaction_hashes_process.rs#L80

PR #2484, opened 2020-12-30
fix: don't do sync before sync connected

yangby-cryptape:
#### Issue

- This issue was introduced since #2455.

  That issue fix the bug: if relay protocol connected but sync protocol not, then there will be no response for all received relay messages.

  More details about that issue can be found in issue #2450.

  Before the PR, only sync protocol can add peers, but after it, relay protocol can add peers, too.

- If there are any peers in `Peers`, the sync protocol could do `notify`, even if it wasn't connected.

  So, it could do `send_getheaders_to_peer`.

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/sync/src/synchronizer/mod.rs#L749-L751
  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/sync/src/synchronizer/mod.rs#L506-L507

- But, since sync protocol wasn't connected, the `send_getheaders_to_peer` will do nothing.

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/sync/src/types/mod.rs#L2016-L2022

- And the node will not do `send_getheaders_to_peer` again in a few seconds, even if it was connected.

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/sync/src/types/mod.rs#L1987-L1992

- So the synchronization will be stuck.

#### Reproduce

- Update the follow codes:

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/network/src/protocols/identify/mod.rs#L439-L442

  to:

  ```rust
  let protos = self.network_state.get_protocol_ids(|id| {
      id != SupportProtocols::Feeler.protocol_id()
          && id != SupportProtocols::Sync.protocol_id()
  });
  let _ = context
      .open_protocols(context.session.id, TargetProtocol::Multi(protos));
  // make sure that relay protocol is connected before sync protocol.
  ::std::thread::sleep(::std::time::Duration::from_millis(5000));
  let protos = self
      .network_state
      .get_protocol_ids(|id| id == SupportProtocols::Sync.protocol_id());
  ```

- Reduce the `SYNC_NOTIFY_INTERVAL` to increase the probability that doing sync before sync connected.

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/sync/src/synchronizer/mod.rs#L47

  to:

  ```rust
  const SYNC_NOTIFY_INTERVAL: Duration = Duration::from_millis(1);
  ```

#### Affected

This issue causes a lot of integration tests failures, for example:

- `BlockSyncWithUncle`

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/test/src/specs/sync/block_sync.rs#L78-L79

- `ForksContainSameTransactions`

  https://github.com/nervosnetwork/ckb/blob/4a5e8a2a0790192f57dbad2d6f1abc1be5ab84bf/test/src/specs/sync/chain_forks.rs#L522-L523

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -216,6 +216,7 @@ pub struct PeerState {
     pub sync_started: bool,
     pub headers_sync_controller: Option<HeadersSyncController>,
     pub peer_flags: PeerFlags,
+    sync_connected: bool,
     pub disconnect: bool,
     pub chain_sync: ChainSyncState,
     // The key is a `timeout`, means do not ask the tx before `timeout`.
@@ -237,6 +238,7 @@ impl PeerState {
             sync_started: false,
             headers_sync_controller: None,
             peer_flags,
+            sync_connected: false,
             disconnect: false,
             chain_sync: ChainSyncState::default(),
             tx_ask_for_map: BTreeMap::default(),
@@ -249,7 +251,8 @@ impl PeerState {
 
     pub fn can_sync(&self, now: u64, ibd: bool) -> bool {
         // only sync with protect/whitelist peer in IBD
-        ((self.peer_flags.is_protect || self.peer_flags.is_whitelist) || !ibd)
+        self.sync_connected
+            && ((self.peer_flags.is_protect || self.peer_flags.is_whitelist) || !ibd)
             && !self.sync_started
             && self
                 .chain_sync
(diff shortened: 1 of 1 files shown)


driftluo (dismissed):

sync/src/types/mod.rs
```diff
@@ -216,6 +216,7 @@ pub struct PeerState {
     pub sync_started: bool,
     pub headers_sync_controller: Option<HeadersSyncController>,
     pub peer_flags: PeerFlags,
+    sync_connected: bool,
```
keroro520, line 219: I understand the purpose of this field. As it is a bit implicit for normal readers, I suggest adding some comments.

keroro520 (approved):

driftluo (approved):

yangby-cryptape:
bors r=keroro520,driftluo

PR #2489, opened 2021-01-04
chore(integration): add --verbose argument for showing verbose log

chuijiaolianying:
- remove `--quiet` command line argument
- add `--verbose`, when use this argument will set `ckb_test` log level as `trace`(`info` as a default value). And there will add more verbose `debug/trace` logs into integration test util functions.

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -36,10 +36,6 @@ struct TestResult {
 #[allow(clippy::cognitive_complexity)]
 fn main() {
     env::set_var("RUST_BACKTRACE", "full");
-    let _ = {
-        let filter = env::var("RUST_LOG").unwrap_or_else(|_| "info".to_string());
-        env_logger::builder().parse_filters(&filter).try_init()
-    };
 
     let clap_app = clap_app();
     let matches = clap_app.get_matches();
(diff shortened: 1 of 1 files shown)


test/src/main.rs
```diff
@@ -55,8 +51,21 @@ fn main() {
...
+    let verbose = matches.is_present("verbose");
+
+    let _ = {
+        let filter = if !verbose {
+            env::var("RUST_LOG").unwrap_or("info".to_string())
+        } else {
+            format!(
+                "{},{}",
+                env::var("RUST_LOG").unwrap_or("info".to_string()),
+                "ckb_test=trace"
```
yangby-cryptape, line 64: - I prefer to use `RUST_LOG` directly.
  It's easy to create confusion by mixing `RUST_LOG` and `--verbose`.
- Don't hardcode the package name, it could be introduced from an environment variable in compile-time.

chuijiaolianying: The purpose of `--verbose` is offer an easy way to set `ckb_test` log level as `trace`, and it's also can be treated as a flag of this log level. `RUST_LOG` is a good way to do this, so I make `--verbose` as a default `false` option argument.

And also, `--verbose` only add `ckb_test=trace` into `RUST_LOG`, the only condition to create confusion is when set `ckb_test=xxx` in `RUST_LOG` at same time, but I don't think this is a question since you already want to show `verbose` log.

keroro520 (approved):
Although I prefer using `RUST_LOG` directly too, I accept this change as well.

quake (approved):

chuijiaolianying:
bors r=quake,keroro520

PR #2494, opened 2021-01-05
chore(integration): add cases RPC - Miner module

chuijiaolianying:
Add integration test cases for RPC Miner Module:

- get_block_template
- submit_block

PR #2495, opened 2021-01-05
chore: add impl for EpochNumberWithFraction which used for getting details of epoch

chuijiaolianying:
According to the suggestion of https://github.com/nervosnetwork/ckb/pull/2494#discussion_r551694750, add impl for `EpochNumberWithFraction`, this would be useful for get `epoch_number`, `epoch_index` and `epoch_length` from `EpochNumberWithFraction`

### DIFF
--- a/util/jsonrpc-types/src/primitive.rs
+++ b/util/jsonrpc-types/src/primitive.rs
@@ -1,4 +1,5 @@
 use crate::{Uint32, Uint64};
+use ckb_types::core::EpochNumberWithFraction as CkbEpochNumberWithFraction;
 
 /// Consecutive block number starting from 0.
 ///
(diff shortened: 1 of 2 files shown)


util/jsonrpc-types/src/primitive.rs
```diff
@@ -50,3 +50,20 @@ pub type Timestamp = Uint64;
 ///
 /// This is a 32-bit unsigned integer type encoded as the 0x-prefixed hex string in JSON. See examples of [Uint32](type.Uint32.html#examples).
 pub type Version = Uint32;
+
+impl EpochNumberWithFraction {
```
yangby-cryptape, line 54: #### Why I disagree with this PR?

- `EpochNumberWithFraction` is an `alias` not a `new-type`.

  https://github.com/nervosnetwork/ckb/blob/e171a8a0ea7c1506547cb563fd8d8f3d267013b2/util/jsonrpc-types/src/primitive.rs#L28

  So `impl EpochNumberWithFraction` is equal to `impl Uint64`.

  Why `epoch_number`, `epoch_index` and `epoch_length` can be get from all `Uint64`?
  For example, developers will found `Capacity` has getters `epoch_number`, `epoch_index` and `epoch_length`, they will be confused.

- Bad smells: Duplicated Code and Magic Numbers 

  A better example:
  https://github.com/nervosnetwork/ckb/blob/e171a8a0ea7c1506547cb563fd8d8f3d267013b2/util/types/src/core/extras.rs#L412-L414

#### Let's look at this in another way:

- You could `impl From<Uint64> for ckb_types::EpochNumberWithFraction;`.
- 
- (Reuse Code) Then, get `epoch_number`, `epoch_index` or `epoch_length` from `ckb_types::EpochNumberWithFraction`.

yangby-cryptape (changes requested):
I

yangby-cryptape (approved):

quake (approved):

quake:
bors r=quake,yangby-cryptape

PR #2503, opened 2021-01-07
feat: customize chain spec for dev chains and update few preset params

yangby-cryptape:
#### Features

- Set `permanent_difficulty_in_dummy` to `true` as default for dev chains.

- Allow users to create different dev chains by customizing genesis message.
  And they could also create same chain in different directories or machines by setting a same genesis message.

- Allow users to create different dev chains by customizing genesis timestamp.
  If no timestamp is provided, use current timestamp.

- Display genesis hash after CKB direcotry created.


keroro520 (approved):

util/app-config/src/lib.rs
```diff
@@ -234,6 +234,9 @@ impl Setup {
         let block_assembler_message = matches.value_of(cli::ARG_BA_MESSAGE).map(str::to_string);
 
         let import_spec = matches.value_of(cli::ARG_IMPORT_SPEC).map(str::to_string);
+        let genesis_message = matches
+            .value_of(cli::ARG_GENESIS_MESSAGE)
+            .map(str::to_string);
```
TheWaWaR, line 239: Use a random message (or even current timestamp) as default genesis message?

TheWaWaR (approved):

ckb-bin/src/subcommand/init.rs
```diff
@@ -184,7 +189,16 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
         }
     } else if args.chain == "dev" {
         println!("create {}", SPEC_DEV_FILE_NAME);
-        Resource::bundled(SPEC_DEV_FILE_NAME.to_string()).export(&context, &args.root_dir)?;
+        let bundled = Resource::bundled(SPEC_DEV_FILE_NAME.to_string());
+        if args.use_default_spec {
```
keroro520, line 193: I think the priority of `customize_spec` is greater than `use_default_spec`. When a user provides both `customize_spec` and `use_default_spec`, it should apply `customize_spec` or fail.

If we apply `customize_spec`:

```
if !args.customize_spec.is_unset() {
  ..
} else if args.use_default_spec {
  ..
}
```

keroro520 (changes requested):

ckb-bin/src/subcommand/init.rs
yangby-cryptape: The default mode is "customize", that is why the priority of "customize" is less than "default".

ckb-bin/src/subcommand/init.rs
```diff
@@ -24,6 +24,11 @@ pub fn init(args: InitArgs) -> Result<(), ExitCode> {
         return Ok(());
     }
 
+    if args.chain != "dev" && (args.use_default_spec || !args.customize_spec.is_unset()) {
+        eprintln!("Customizing consensus parameters for chain spec only works for dev chains.");
+        return Err(ExitCode::Failure);
```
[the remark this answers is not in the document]
yangby-cryptape: First, unifying the `ExitCode` isn't target of this PR.
I just follow the the usual practice under `subcommand` directory.

Second, I think `ExitCode::Config` is incorrect. It's not a configuration error.
And it's not `ExitCode::Cli` because the arguments are right.
For example, if the input timestamp couldn't parse to a number, the `ExitCode::Cli` will be returned.
https://github.com/nervosnetwork/ckb/blob/80a40a5c58f1dba47ea4fdbf28702784984bdf9b/util/app-config/src/lib.rs#L239-L250

Here is a runtime error.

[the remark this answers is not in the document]
keroro520: Got it.

keroro520 (dismissed):

util/app-config/src/args.rs
```diff
@@ -111,6 +112,23 @@ pub struct InitArgs {
     pub block_assembler_message: Option<String>,
     /// TODO(doc): @doitian
     pub import_spec: Option<String>,
+    /// Don't customize any parameters for chain spec, use the default parameters.
+    ///
+    /// Only works for dev chains.
+    pub use_default_spec: bool,
```
quake, line 118: I'm wondering why this parameter is needed, isn't it equivalent to `CustomizeSpec#is_unset`?

util/app-config/src/args.rs
```diff
@@ -156,3 +174,23 @@ pub struct MigrateArgs {
...
+    }
+
+    /// Generates a vector of key-value pairs.
+    pub fn key_value_pairs(&self) -> Vec<(&'static str, String)> {
+        let mut vec = Vec::new();
+        let genesis_timestamp = self
+            .genesis_timestamp
+            .unwrap_or_else(unix_time_as_millis)
+            .to_string();
+        let genesis_message = self.genesis_message.clone().unwrap_or_else(String::new);
```
quake, line 191: I want to initialize an identical dev chain in both directories, using the same command with the same `genesis_message` argument, but the resulting chain is different, which is surprising, so I don't think it's appropriate to use the current time as the default value.

quake (changes requested):

ckb-bin/src/subcommand/init.rs
doitian: If two options are conflicting, use a exclusive options group instead.

doitian: Already done in cli.rs

util/app-config/src/args.rs
doitian: I agree that we just need to random genesis message by default.

1. Use the timestamp as the genesis message for dev chain by default.
2. User can use `--genesis-message` to customize the message
3. The option `--use-default-spec` is not needed.
4. The option `--genesis-timestamp` is optional. If you really want to customize genesis timestamp, go ahead.

doitian:
> I'm wondering why this parameter is needed, isn't it equivalent to `CustomizeSpec#is_unset`? 

Same here.

quake (approved):

doitian (approved):

doitian:
bors r=quake,doitian

PR #2505, opened 2021-01-11
feat: provide `--overwrite-spec` to override the chain spec in storage

keroro520:
Ckb stores chain spec into the database and compare the configured chain spec on starting.

A user may want to overwrite the stored chain spec when it changes the configured chain spec and makes sure the change is acceptable. Therefore this PR adds `--overwrite-spec` command option to overwrite the stored chain spec.

---

* If you want to just skip the chain spec checking without overriding the stored one:

  ```shell
  ckb run --skip-spec-check ...
  ```

* If you want to overwrite the stored chain spec with the configured one:

  ```shell
  ckb run --overwrite-spec ...
  ```

### DIFF
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -12,8 +12,9 @@ use ckb_network_alert::alert_relayer::AlertRelayer;
 use ckb_resource::Resource;
 use ckb_rpc::{RpcServer, ServiceBuilder};
 use ckb_shared::shared::{Shared, SharedBuilder};
-use ckb_store::ChainStore;
+use ckb_store::{ChainDB, ChainStore};
 use ckb_sync::{NetTimeProtocol, Relayer, SyncShared, Synchronizer};
+use ckb_types::packed::Byte32;
 use ckb_types::{core::cell::setup_system_cell_cache, prelude::*};
 use ckb_verification::{GenesisVerifier, Verifier};
 use std::sync::Arc;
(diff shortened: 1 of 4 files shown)


ckb-bin/src/subcommand/run.rs
```diff
@@ -192,33 +193,38 @@ fn verify_genesis(shared: &Shared) -> Result<(), ExitCode> {
...
-        info_target!(
-            crate::LOG_TARGET_MAIN,
-            "Touch chain spec hash: {}",
-            args.chain_spec_hash
-        );
     }
     Ok(())
 }
 
+fn write_chain_spec_hash(store: &ChainDB, chain_spec_hash: &Byte32) -> Result<(), ExitCode> {
```
yangby-cryptape, line 215: I think there should have a log to tell the user whether the chain spec hash was overwritten or not (`chain-spec-hash-old == chain-spec-hash-new`).

ckb-bin/src/subcommand/run.rs
```diff
@@ -192,33 +193,38 @@ fn verify_genesis(shared: &Shared) -> Result<(), ExitCode> {
...
-                "chain_spec_hash mismatch Config({}) storage({}), pass command line argument --skip-spec-check if you are sure that the two different chains are compatible.",
-                args.chain_spec_hash, spec_hash
+                "chain_spec_hash mismatch Config({}) storage({}), pass command line argument \
+                --skip-spec-check if you are sure that the two different chains are compatible; \
+                or pass --override-spec to force overriding stored chain spec with configured chain spec",
+                args.chain_spec_hash, stored_spec_hash
             );
             return Err(ExitCode::Config);
         }
-    } else {
```
yangby-cryptape, line 203: Maybe u could just keep the old `if ... else ...` and change this `else` to `else if args.override_chain_spec`.

ckb-bin/src/subcommand/run.rs
keroro520: An informational log is worthy.
And I am going to rename "override" to "overwrite" after reading your comments.

ckb-bin/src/subcommand/run.rs
keroro520: When `args.chain_spec_hash != spec_hash`, it may enter the [`!args.skip_chain_spec_check branch`](https://github.com/nervosnetwork/ckb/pull/2505/files#diff-24827db4c0deb64085aac4223696a31814046c5cc185bc289124a77d8a9a01fbL196).

Simply changing `else` to `else if` is not enough.

I have updated the code, please help review again, thanks.

quake (dismissed):

yangby-cryptape (approved):

quake (approved):

quake:
bors r=quake,yangby-cryptape

PR #2509, opened 2021-01-12
feat: metricize network traffic

keroro520:
* Metricize network message traffic and labeled with message name and processing result
* Remove some staled metrics

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -5,7 +5,6 @@ use crate::utils::send_message_to;
 use crate::{attempt, Status, StatusCode};
 use ckb_chain_spec::consensus::Consensus;
 use ckb_logger::{self, debug_target};
-use ckb_metrics::metrics;
 use ckb_network::{CKBProtocolContext, PeerIndex};
 use ckb_traits::{BlockMedianTimeContext, HeaderProvider};
 use ckb_types::{core, packed, prelude::*};
@@ -261,12 +260,6 @@ impl<'a> CompactBlockProcess<'a> {
         } else {
             StatusCode::CompactBlockRequiresFreshTransactions.with_context(&block_hash)
         };
-        if !missing_transactions.is_empty() {
-            metrics!(value, "ckb-net.fresh", missing_transactions.len() as u64, "type" => "transactions", "status" => status.tag());
-        }
-        if !missing_uncles.is_empty() {
-            metrics!(value, "ckb-net.fresh", missing_uncles.len() as u64, "type" => "uncles", "status" => status.tag());
-        }
 
         let content = packed::GetBlockTransactions::new_builder()
             .block_hash(block_hash)
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -267,7 +267,14 @@ impl<'a> CompactBlockProcess<'a> {
             .uncle_indexes(missing_uncles.pack())
             .build();
         let message = packed::RelayMessage::new_builder().set(content).build();
-        let _ignore = send_message_to(self.nc.as_ref(), self.peer, &message);
+        let sending = send_message_to(self.nc.as_ref(), self.peer, &message);
+        if !sending.is_ok() {
+            ckb_logger::warn_target!(
+                crate::LOG_TARGET_RELAY,
+                "ignore the sending message error, error: {}",
+                sending
+            );
+        }
 
         status
     }
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -1,6 +1,7 @@
 use crate::block_status::BlockStatus;
 use crate::relayer::compact_block_verifier::CompactBlockVerifier;
 use crate::relayer::{ReconstructionResult, Relayer};
+use crate::utils::send_message_to;
 use crate::{attempt, Status, StatusCode};
 use ckb_chain_spec::consensus::Consensus;
 use ckb_logger::{self, debug_target};
@@ -273,12 +274,7 @@ impl<'a> CompactBlockProcess<'a> {
             .uncle_indexes(missing_uncles.pack())
             .build();
         let message = packed::RelayMessage::new_builder().set(content).build();
-
-        if let Err(err) = self.nc.send_message_to(self.peer, message.as_bytes()) {
-            return StatusCode::Network
-                .with_context(format!("Send GetBlockTransactions error: {:?}", err));
-        }
-        crate::relayer::metrics_counter_send(message.to_enum().item_name());
+        let _ignore = send_message_to(self.nc.as_ref(), self.peer, &message);
 
         status
     }
--- a/sync/src/relayer/mod.rs
+++ b/sync/src/relayer/mod.rs
@@ -22,6 +22,7 @@ use self::transaction_hashes_process::TransactionHashesProcess;
 use self::transactions_process::TransactionsProcess;
 use crate::block_status::BlockStatus;
 use crate::types::{ActiveChain, SyncShared};
+use crate::utils::send_message_to;
 use crate::{Status, StatusCode};
 use ckb_chain::chain::ChainController;
 use ckb_constant::sync::BAD_MESSAGE_BAN_TIME;
@@ -245,18 +246,11 @@ impl Relayer {
                 .proposals(to_ask_proposals.clone().pack())
                 .build();
             let message = packed::RelayMessage::new_builder().set(content).build();
-
-            if let Err(err) = nc.send_message_to(peer, message.as_bytes()) {
-                debug_target!(
-                    crate::LOG_TARGET_RELAY,
-                    "relayer send GetBlockProposal error {:?}",
-                    err,
-                );
+            if !send_message_to(nc, peer, &message).is_ok() {
                 self.shared()
                     .state()
                     .remove_inflight_proposals(&to_ask_proposals);
             }
-            crate::relayer::metrics_counter_send(message.to_enum().item_name());
         }
     }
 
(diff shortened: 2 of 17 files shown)


sync/src/relayer/compact_block_process.rs
```diff
@@ -259,25 +259,14 @@ impl<'a> CompactBlockProcess<'a> {
...
             .uncle_indexes(missing_uncles.pack())
             .build();
         let message = packed::RelayMessage::new_builder().set(content).build();
-
-        if let Err(err) = self.nc.send_message_to(self.peer, message.as_bytes()) {
-            return StatusCode::Network
-                .with_context(format!("Send GetBlockTransactions error: {:?}", err));
-        }
-        crate::relayer::metrics_counter_send(message.to_enum().item_name());
+        let _ignore = send_message_to(self.nc.as_ref(), self.peer, &message);
```
doitian, line 269: It's better to add a log on error, and use the log message to explain why this error can be ignored.

doitian (approved):

sync/src/utils.rs
```diff
@@ -0,0 +1,62 @@
...
+) -> Status {
+    let protocol_id = nc.protocol_id();
+    send_message(protocol_id, nc, peer_index, message)
+}
+
+// As for `SyncMessage` and `RelayMessage`, returns the internal item name;
+// otherwise returns the entity name.
+fn message_name<Message: Entity>(protocol_id: ProtocolId, message: &Message) -> String {
+    if protocol_id == SupportProtocols::Sync.protocol_id() {
+        SyncMessage::from_slice(message.as_slice())
```
driftluo, line 48: `SyncMessageReader::new_unchecked` is better, no need to copy all data for log

keroro520: Thanks. This suggestion saves lots of verifying costs.

sync/src/relayer/compact_block_process.rs
keroro520: updated.

quake:
This PR is heavily coupled, can we move the metric related code to `DefaultCKBProtocolContext#send_message`?

keroro520:
> This PR is heavily coupled, can we move the metric related code to `DefaultCKBProtocolContext#send_message`?

* `DefaultCKBProtocolContext` is too underlying for collecting metrics. `DefaultCKBProtocolContext#send_message` accepts the raw message bytes as the parameter, it cannot know the actual message name from the raw message bytes. While the message names are important for metrics.

* Extracting the message names is trivial. Expect suggestions for that.

* Which specific parts do you consider are heavily coupled?

sync/src/utils.rs
```diff
@@ -0,0 +1,60 @@
...
+    peer_index: PeerIndex,
+    message: &Message,
+) -> Status {
+    let protocol_id = nc.protocol_id();
+    send_message(protocol_id, nc, peer_index, message)
+}
+
+// As for Sync protocol and Relay protocol, returns the internal item name;
+// otherwise returns the entity name.
+fn message_name<Message: Entity>(protocol_id: ProtocolId, message: &Message) -> String {
```
driftluo, line 46:
```
fn message_name<Message: Entity>(protocol_id: ProtocolId, message: &’a Message) -> &‘a str
```
It seems to be ok because the loggger will `format!`, the type return here does not need `str` to `string`

keroro520: I try and got the error "error[E0515]: cannot return value referencing temporary value", as the `SyncMessageReader` is a temporary object within the function.

driftluo: ok, to enum become a new temporary value, so the lifetime has expired

driftluo (dismissed):

quake:
> * `DefaultCKBProtocolContext` is too underlying for collecting metrics. `DefaultCKBProtocolContext#send_message` accepts the raw message bytes as the parameter, it cannot know the actual message name from the raw message bytes. While the message names are important for metrics.
> *

message union id is in the fixed position of serilzied slice, we can combine the protocol id and union id as an unique metric:
```
metrics!(counter, "ckb.messages_bytes", bytes, "direction" => "out", "protocol_id" => protocol_id, "union_id" => bytes[4..8]);
```

keroro520:
> > * `DefaultCKBProtocolContext` is too underlying for collecting metrics. `DefaultCKBProtocolContext#send_message` accepts the raw message bytes as the parameter, it cannot know the actual message name from the raw message bytes. While the message names are important for metrics.
> 
> message union id is in the fixed position of serilzied slice, we can combine the protocol id and union id as an unique metric:
> 
> ```
> metrics!(counter, "ckb.messages_bytes", bytes, "direction" => "out", "protocol_id" => protocol_id, "union_id" => bytes[4..8]);
> ```

`SyncMessage` and `RelayMessage` are union types, but others may not, such as `Alert`, `Time`, and other user-defined message types. Is there a way to check whether it is a union type?

doitian:
Please resolve conflicting files

keroro520:
@quake A new commit a909e68f6dedfd6b7b63dcee97b16e75907f3e0b removes the metric tag "message_name", and use "protocol_id" and "item_id" instead.

driftluo (dismissed):

sync/src/relayer/compact_block_process.rs
```diff
@@ -260,25 +260,21 @@ impl<'a> CompactBlockProcess<'a> {
...
             .indexes(missing_transactions.pack())
             .uncle_indexes(missing_uncles.pack())
             .build();
         let message = packed::RelayMessage::new_builder().set(content).build();
-
-        if let Err(err) = self.nc.send_message_to(self.peer, message.as_bytes()) {
-            return StatusCode::Network
-                .with_context(format!("Send GetBlockTransactions error: {:?}", err));
+        let sending = send_message_to(self.nc.as_ref(), self.peer, &message);
+        if !sending.is_ok() {
```
quake, line 271: can we remove this redundant log here since we have log the sending error in `send_message` fn already?

sync/src/relayer/mod.rs
```diff
@@ -173,12 +174,17 @@ impl Relayer {
...
     ) {
         let item_name = message.item_name();
+        let item_bytes = message.as_slice().len() as u64;
         let status = self.try_process(Arc::clone(&nc), peer, message);
 
-        metrics!(counter, "ckb-net.received", 1, "action" => "relay", "item" => item_name.to_owned());
-        if !status.is_ok() {
-            metrics!(counter, "ckb-net.status", 1, "action" => "relay", "status" => status.tag());
-        }
+        metrics!(
```
quake, line 180: I think this metrics should be moved to https://github.com/nervosnetwork/ckb/blob/a8829f4a54ec453c48b67808c5ec694a09cfae2a/sync/src/synchronizer/mod.rs#L673 

and use the same metrics parameters as direction out
```
    metrics!(
        counter,
        "ckb.messages_bytes", bytes,
        "direction" => "in",
        "protocol_id" => SupportProtocols::Sync::protocol_id(),
        "item_id" => msg.item_id()
    );
```

sync/src/synchronizer/mod.rs
```diff
@@ -260,12 +260,17 @@ impl Synchronizer {
...
     ) {
         let item_name = message.item_name();
+        let item_bytes = message.as_slice().len() as u64;
         let status = self.try_process(nc, peer, message);
 
-        metrics!(counter, "ckb-net.received", 1, "action" => "sync", "item" => item_name.to_owned());
-        if !status.is_ok() {
-            metrics!(counter, "ckb-net.status", 1, "action" => "sync", "status" => status.tag());
-        }
+        metrics!(
```
quake, line 266: I think this metrics should be moved to https://github.com/nervosnetwork/ckb/blob/a8829f4a54ec453c48b67808c5ec694a09cfae2a/sync/src/relayer/mod.rs#L660

and use the same metrics parameters as direction out
```
    metrics!(
        counter,
        "ckb.messages_bytes", msg.,
        "direction" => "in",
        "protocol_id" => SupportProtocols::Relay::protocol_id(),
        "item_id" => msg.item_id(),
    );
```

sync/src/relayer/mod.rs
```diff
@@ -529,15 +528,10 @@ impl Relayer {
...
             let message = packed::RelayMessage::new_builder().set(content).build();
-
-            if let Err(err) = nc.send_message_to(peer_index, message.as_bytes()) {
-                debug_target!(
-                    crate::LOG_TARGET_RELAY,
-                    "relayer send BlockProposal error: {:?}",
-                    err,
-                );
+            let status = send_message_to(nc, peer_index, &message);
+            if !status.is_ok() {
```
quake, line 532: same redundant log issue here

sync/src/relayer/mod.rs
```diff
@@ -570,15 +564,10 @@ impl Relayer {
...
                 let message = packed::RelayMessage::new_builder().set(content).build();
-
-                if let Err(err) = nc.send_message_to(*peer, message.as_bytes()) {
-                    debug_target!(
-                        crate::LOG_TARGET_RELAY,
-                        "relayer send Transaction error: {:?}",
-                        err,
-                    );
+                let status = send_message_to(nc, *peer, &message);
+                if !status.is_ok() {
```
quake, line 568: same redundant log issue here

quake (changes requested):

sync/src/relayer/compact_block_process.rs
keroro520: You are right that `fn send_message` logs when an error occurs. While I think that error log cannot provide us the context, where the sending behavior comes from, whether I should ignore it?
Therefore I leave warning logs so that users can get more context when an error occurs.
Yeah, it's a bit redundant, but it's worthy.

sync/src/relayer/mod.rs
keroro520: My fault.
Updated, thanks a lot.

quake (approved):

driftluo (approved):

keroro520:
bors r=quake,driftluo

Issue #2518, opened 2021-01-13
Given a block hash in the main chain, return the median time


doitian:
RPC: `get_block_median_time(block_hash)`

If the `block_hash` is in the chain, return the block median time, which is the median of timestamp of the consecutive 37 blocks where the given `block_hash` has the highest height.

Pay attention that the given block is included in the median time. Some consensus rules require the median time of the parent block.

PR #2520, opened 2021-01-13
feat: add RPC get_block_median_time

keroro520:
resolve https://github.com/nervosnetwork/ckb/issues/2518

### DIFF
--- a/rpc/src/module/chain.rs
+++ b/rpc/src/module/chain.rs
@@ -2,12 +2,13 @@ use crate::error::RPCError;
 use ckb_jsonrpc_types::{
     BlockEconomicState, BlockNumber, BlockReward, BlockView, CellWithStatus, Consensus,
     EpochNumber, EpochView, HeaderView, MerkleProof as JsonMerkleProof, OutPoint, ResponseFormat,
-    TransactionProof, TransactionWithStatus, Uint32,
+    Timestamp, TransactionProof, TransactionWithStatus, Uint32,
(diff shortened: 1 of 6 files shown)


rpc/src/module/chain.rs
```diff
@@ -1193,6 +1194,46 @@ pub trait ChainRpc {
     /// ```
     #[rpc(name = "get_consensus")]
     fn get_consensus(&self) -> Result<Consensus>;
+
+    /// Returns the past median time by block hash.
+    ///
+    /// ## Params
+    ///
+    /// * `block_hash` - the highest block hash
```
chuijiaolianying, line 1202: I think `the highest block hash` may create confusion about this param should be the tip block hash or block hash of any height in the chain

keroro520: Thanks. I agree with that. What about the below, or any suggestions?

> "A median time is calculated for a consecutive block sequence. `block_hash` indicates the highest block of the sequence."

chuijiaolianying: LGTM

keroro520: Updated document.

chuijiaolianying (approved):

chuijiaolianying (approved):

quake (approved):

yangby-cryptape:
bors r=chuijiaolianying,quake

PR #2523, opened 2021-01-14
fix: network should work without rpc

yangby-cryptape:
#### How to Reproduce?

Edit section `[rpc]` in `ckb.toml`:

```toml
modules = []
```

p.s. `NetworkController` contains all other `Controllers`, so we can only extend its lifetime to let all controllers work well.


quake (approved):

driftluo (approved):

keroro520 (approved):

keroro520:
bors r=quake,driftluo,keroro520

PR #2525, opened 2021-01-17
manually trigger compaction after freeze

zhangsoledad:
### Drawbacks
Even we use `DeleteRange` apply to delete the range of keys, seems Rocksdb still hasn't implemented the feature of using seek() to skip until the end of range delete end yet.
Rocksdb iter seek slows down dramatically when there are many deletes.
### Workaround
Manually call `CompactRange()` for the range to delete,  this approach can solve the problem.

### DIFF
--- a/freezer/src/freezer.rs
+++ b/freezer/src/freezer.rs
@@ -8,13 +8,13 @@ use ckb_types::{
 };
 use ckb_util::Mutex;
 use fs2::FileExt;
+use std::collections::BTreeMap;
 use std::fs::{File, OpenOptions};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
 
 const LOCKNAME: &str = "FLOCK";
-type FreezeResult = (BlockNumber, packed::Byte32, u32);
 
 struct Inner {
     pub(crate) files: FreezerFiles,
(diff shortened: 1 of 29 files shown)


freezer/src/freezer.rs
```diff
@@ -72,13 +72,13 @@ impl Freezer {
         &self,
         threshold: BlockNumber,
         get_block_by_number: F,
-    ) -> Result<Vec<FreezeResult>, Error>
+    ) -> Result<BTreeMap<packed::Byte32, (BlockNumber, u32)>, Error>
```
doitian, line 75: This is a public interface. Please add some document about the return type. What's returned in the map on success?

util/snapshot/src/lib.rs
```diff
@@ -22,8 +23,12 @@ use ckb_types::{
     packed::{Byte32, OutPoint, Script},
     U256,
 };
+use faketime::unix_time_as_millis;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+static IBD_FINISHED: AtomicBool = AtomicBool::new(false);
```
doitian, line 30: Global flag is not recommended. It makes test hard.

doitian (changes requested):

doitian (approved):

doitian (approved):

yangby-cryptape (approved):

doitian (approved):

doitian:

PR #2542, opened 2021-01-28
fix: resolve rocksdb cache size issue when using `default.db-options`

quake:
`cache_index_and_filter_blocks` is enabled in `default.db-options`, quote rocksdb doc:

```
By putting index, filter, and compression dictionary blocks in block cache, these blocks have to compete against data blocks for staying in cache
```

current rocksdb index size is around 32MB, and the default rocksdb block cache size is 8MB, which hurts the db read performance, this PR changed the default cache size to 128MB.

We should consider using [Partitioned Index](https://github.com/facebook/rocksdb/wiki/Partitioned-Index-Filters) later.

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -24,14 +24,23 @@ pub struct RocksDB {
     pub(crate) inner: Arc<OptimisticTransactionDB>,
 }
 
+const DEFAULT_CACHE_SIZE: usize = 128 << 20;
+
 impl RocksDB {
     pub(crate) fn open_with_check(config: &DBConfig, columns: u32) -> Result<Self> {
         let cf_names: Vec<_> = (0..columns).map(|c| c.to_string()).collect();
 
         let (mut opts, cf_descriptors) = if let Some(ref file) = config.options_file {
-            let mut full_opts = FullOptions::load_from_file(file, None, false).map_err(|err| {
-                internal_error(format!("failed to load the options file: {}", err))
-            })?;
+            let cache_size = match config.cache_size {
+                Some(0) => None,
+                Some(size) => Some(size),
+                None => Some(DEFAULT_CACHE_SIZE),
+            };
+
+            let mut full_opts =
+                FullOptions::load_from_file(file, cache_size, false).map_err(|err| {
+                    internal_error(format!("failed to load the options file: {}", err))
+                })?;
             let cf_names_str: Vec<&str> = cf_names.iter().map(|s| s.as_str()).collect();
             full_opts
                 .complete_column_families(&cf_names_str, false)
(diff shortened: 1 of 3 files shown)


db/src/db.rs
```diff
@@ -29,9 +29,12 @@ impl RocksDB {
         let cf_names: Vec<_> = (0..columns).map(|c| c.to_string()).collect();
 
         let (mut opts, cf_descriptors) = if let Some(ref file) = config.options_file {
-            let mut full_opts = FullOptions::load_from_file(file, None, false).map_err(|err| {
-                internal_error(format!("failed to load the options file: {}", err))
-            })?;
+            let mut full_opts = FullOptions::load_from_file(
+                file,
+                config.cache_size.or_else(|| Some(128 << 20)),
```
yangby-cryptape, line 34: How could I disable `cache_size`?

If I set `[db] cache_size = null`, the db still set `cache_size = 128 << 20`.

yangby-cryptape (changes requested):

db/src/db.rs
quake: We should not allow users to disable the rocksdb cache, user should lower this value (8MB for example), if he wants to reduce memory usage.

yangby-cryptape: Why don't allow users to disable the RocksDB cache?

yangby-cryptape: How about set it to `None` when `cache_size == 0`?

quake: user should modify the `default.db-options` by appending this line if he wants the full control of rocksdb block cache disable / enable:
```
[TableOptions/BlockBasedTable "default"]
no_block_cache=true
```

We chose to put the `cache size` in ckb.toml only because the `default.options` file does not support this parameter

yangby-cryptape: But the cache is allocated before the RocksDB load the options file, and couldn't free.

Ref:
- [Why in our `rust-rocksdb`, we couldn't free the cache.](https://github.com/nervosnetwork/rust-rocksdb/pull/18/files#r457146149)
- The [`C++` API `LoadOptionsFromFile(..)`](https://github.com/facebook/rocksdb/blob/0915c99f01b46f50af8e02da8b6528156f584b7c/include/rocksdb/utilities/options_util.h#L71-L79).

quake: Understood, I modified the code, and added more comment to the ckb configuration file, please review again, thanks.

yangby-cryptape (approved):

yangby-cryptape:
benchmark

yangby-cryptape (approved):

doitian (approved):

nervos-bot-user:
#### Benchmark Result
  - TPS: 485
  - CKB Version: 6e1092ec38f6d35bb6f77ba512613e6136ab8b7f
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

yangby-cryptape:

quake:

driftluo:
bors retry

PR #2544, opened 2021-01-28
ci: collect integration failures on azure via sentry

yangby-cryptape:
#### Why use `ckb-logger` instead of `env_logger` in integration tests?

Because it's difficult to capture output of `Invoke-Expression` into a output file.

No matter with `Out-File`, `Tee-Object` or just redirect the output, it's always blocked and it will output all logs only when the process stopped.

So, if your input  `Ctrl-C` before the process stopped, no logs will be appended in the output file.

#### Why disable `report` for the integration tests?

The sentry server could only save the last 100 lines. 

I would rather save the last 100 logs than the time elapsed report.

#### Preview

[This error](https://dev.azure.com/nervosnetwork/ckb/_build/results?buildId=15794&view=logs&j=133bd042-0fac-58b5-e6e7-01018e6dc4d4&t=b22eb6cd-ede2-5164-159f-eec2a8c1e51e&l=1808) can be found in our Sentry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,7 @@ matrix:
     - name: PR Integration
       if: 'tag IS NOT present AND branch != develop AND branch !~ /^rc\// AND (type = pull_request OR repo != nervosnetwork/ckb)'
       os: linux
-      script: make CKB_TEST_SEC_COEFFICIENT=5 CKB_TEST_ARGS="--max-time 1200 -c 4" integration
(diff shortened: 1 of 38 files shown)


doitian (approved):

chuijiaolianying (approved):

quake (approved):

yangby-cryptape:
bors r=doitian,chuijiaolianying,quake

PR #2563, opened 2021-02-10
chore: modify discovery announce interval

quake:
the timer interval of checking `DiscoveryMessage::Nodes` announce message should be 24 hours, I think it was incorrectly set to 7 seconds in old code, this PR resolved this issue and corrected some code comments

### DIFF
--- a/network/src/protocols/discovery/mod.rs
+++ b/network/src/protocols/discovery/mod.rs
@@ -31,7 +31,7 @@ mod addr;
 mod protocol;
 mod state;
 
-const CHECK_INTERVAL: Duration = Duration::from_secs(3);
+const ANNOUNCE_CHECK_INTERVAL: Duration = Duration::from_secs(60);
 const ANNOUNCE_THRESHOLD: usize = 10;
 // The maximum number of new addresses to accumulate before announcing.
 const MAX_ADDR_TO_SEND: usize = 1000;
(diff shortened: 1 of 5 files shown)


driftluo (approved):

network/src/protocols/discovery/mod.rs
```diff
@@ -42,18 +42,15 @@ const ANNOUNCE_INTERVAL: Duration = Duration::from_secs(3600 * 24);
...
-    check_interval: Option<Duration>,
 }
 
 impl<M: AddressManager> DiscoveryProtocol<M> {
     pub fn new(addr_mgr: M) -> DiscoveryProtocol<M> {
         DiscoveryProtocol {
             sessions: HashMap::default(),
-            dynamic_query_cycle: Some(Duration::from_secs(7)),
-            check_interval: None,
+            announce_internal: None,
```
driftluo, line 53: Maybe it can be changed to allow set `announce_internal` for testing

driftluo (dismissed):

zhangsoledad (approved):

network/src/protocols/discovery/mod.rs
quake: Currently our test code does not have a scenario where this period needs to be set, it can be added later when it is used

yangby-cryptape:
bors r=driftluo,zhangsoledad

driftluo:
bors cancel

doitian (changes requested):
Hold because ckb-test failes on spec Discovery randomly

network/src/protocols/discovery/mod.rs
quake: I found it's required in integration test, added a new config key, please review again, thanks

driftluo (dismissed):

network/src/protocols/discovery/mod.rs
```diff
@@ -230,7 +225,7 @@ impl<M: AddressManager> ServiceProtocol for DiscoveryProtocol<M> {
                 state.send_messages(context, *id);
                 // check timer
                 state
-                    .check_timer(now, dynamic_query_cycle)
+                    .check_timer(now, ANNOUNCE_INTERVAL)
```
doitian, line 228: Why `announce_check_interval` is not used here?

quake: `ANNOUNCE_INTERVAL` is setup as every 24 hours to send announce nodes message, it's a fixed value, and `announce_check_interval` is a configurable parameter to check whether 24 hours have elapsed.

doitian (dismissed):

driftluo (approved):

doitian (approved):

quake:
bors r=doitian,driftluo

PR #2565, opened 2021-02-15
chore: refactor AddrManager related code

quake:
In the previous code, the trait `MultiaddrExt` was used in many places to split out the `PeerId` from `Multiaddr` and merge back in where it was needed ((fn exclude_p2p / attach_p2p). This PR refactors the code related to AddrManager, removes the `PeerId` parameter from the public fn, avoiding these unnecessary exclude_p2p / attach_p2p operations


network/src/network.rs
```diff
@@ -341,12 +298,14 @@ impl NetworkState {
...
-    pub(crate) fn can_dial(
-        &self,
-        peer_id: &PeerId,
-        addr: &Multiaddr,
-        allow_dial_to_self: bool,
-    ) -> bool {
+    pub(crate) fn can_dial(&self, addr: &Multiaddr, allow_dial_to_self: bool) -> bool {
+        let peer_id = extract_peer_id(addr);
+        if peer_id.is_none() {
+            trace!("Do not dial addr without peer id");
```
jjyr, line 304: Theoretically, we should never dial an addr without peer_id, it is better to use `error!` to logging this behaviour.

network/src/peer_registry.rs
```diff
@@ -22,7 +22,7 @@ pub struct PeerRegistry {
     max_outbound: u32,
     // Only whitelist peers or allow all peers.
     whitelist_only: bool,
-    whitelist_peers: HashSet<PeerId>,
+    whitelist_peers: HashSet<Multiaddr>,
```
jjyr, line 25: The only usage of this struct is to find out whether an address is in the whitelist, so it is more appropriate to use a `HashSet<PeerId>` than hash set of address.

network/src/peer_registry.rs
```diff
@@ -81,11 +80,16 @@ impl PeerRegistry {
         if self.peers.contains_key(&session_id) {
             return Err(PeerError::SessionExists(session_id).into());
         }
+        let peer_id = extract_peer_id(&remote_addr).expect("opened session should have peer id");
         if self.get_key_by_peer_id(&peer_id).is_some() {
             return Err(PeerError::PeerIdExists(peer_id).into());
         }
 
-        let is_whitelist = self.whitelist_peers.contains(&peer_id);
+        let is_whitelist = self.whitelist_peers.iter().any(|whitelist_peer| {
```
jjyr, line 88: See the previous comment.

jjyr: I think we should also check whether an address has a peer_id in this function and reject all address which hasn't have a peer_id. Thus, we can keep the behavior consistent.

network/src/peer_registry.rs
```diff
@@ -218,20 +228,17 @@ impl PeerRegistry {
...
     /// Get session id by peer id
     pub fn get_key_by_peer_id(&self, peer_id: &PeerId) -> Option<SessionId> {
-        self.peers.values().find_map(|peer| {
-            if &peer.peer_id == peer_id {
-                Some(peer.session_id)
-            } else {
-                None
-            }
+        self.peers.iter().find_map(|(session_id, peer)| {
+            extract_peer_id(&peer.connected_addr).and_then(|pid| {
```
jjyr, line 232: A peer must have a peer_id, so we can using `expect` here, or add a peer_id method to the Peer structure.

network/src/peer_store/addr_manager.rs
```diff
@@ -1,13 +1,14 @@
...
+use p2p::{multiaddr::Multiaddr, utils::multiaddr_to_socketaddr};
 use rand::Rng;
 use std::collections::{HashMap, HashSet};
 
 /// Address manager
 #[derive(Default)]
 pub struct AddrManager {
     next_id: u64,
-    addr_to_id: HashMap<IpPort, u64>,
+    addr_to_id: HashMap<Multiaddr, u64>,
```
jjyr, line 11: We can append any protocol to the Multiaddr format in any order.

For example `ip/127.0.0.1/port/443/ws/127.0.0.2` is identical to `port/443/ip/127.0.0.1` in our system, that's the reason why we **can't use the string as key** here.

network/src/peer_store/addr_manager.rs
```diff
@@ -49,17 +50,21 @@ impl AddrManager {
...
-            // TODO remove this after fix the network tests.
-            let is_test_ip =
-                addr_info.ip_port.ip.is_unspecified() || addr_info.ip_port.ip.is_loopback();
-            if (is_test_ip || is_unique_ip) && !addr_info.is_terrible(now_ms) && filter(&addr_info)
-            {
-                addr_infos.push(addr_info);
-            }
-            if addr_infos.len() == count {
-                break;
+            if let Some(socket_addr) = multiaddr_to_socketaddr(&addr_info.addr) {
```
jjyr, line 53: Why do we need to assume there are invalid addresses in the manager? How to deal with these addresses? I think to check the format of the address in the entry of the peer store is better than just ignore them.

network/src/peer_store/ban_list.rs
```diff
@@ -66,10 +67,9 @@ impl BanList {
     /// Whether the address is banned
     pub fn is_addr_banned(&self, addr: &Multiaddr) -> bool {
         let now_ms = unix_time_as_millis();
-        if let Ok(ip_port) = addr.extract_ip_addr() {
-            return self.is_ip_banned_until(ip_port.ip, now_ms);
-        }
-        false
+        multiaddr_to_socketaddr(addr)
```
jjyr, line 70: If we can't check whether a address is banned we should ban the address or raise an error.

jjyr (changes requested):

network/src/peer_store/ban_list.rs
quake: I follow the original logic here: if we can not extract ip addr from the multiaddr, retur false:

```
  if let OK(ip_port) = addr.extract_ip_addr() {
 }
 false
```

network/src/peer_store/addr_manager.rs
quake: it's not assumption, we have checked the format of address in the entry of the peer store already, it's just a different style of code: we can use expect here, but I prefer `if let Some()`

network/src/peer_registry.rs
quake: it's just a different style of code, we can use expect here, but I prefer chain invoking here

network/src/peer_store/addr_manager.rs
quake: I don't understand the reason, could you explain more detail?

network/src/peer_store/addr_manager.rs
jjyr: `if let Some()` style often makes errors implicit. However, it depends on you, I do not insist on this.

network/src/peer_store/addr_manager.rs
jjyr: Consider the scenario:

1. We connected with address: `ip/127.0.0.1/port/443`.
2. Then we receive a new address `port/443/ip/127.0.0.1`.
3. The code after refactoring will determine we have two different addresses, it may cause inconsistent behavior.

quake: understood, I've updated the code, please review again, thanks.

network/src/benches/peer_store.rs
```diff
@@ -16,14 +16,18 @@ fn bench(c: &mut Criterion) {
         group.bench_with_input(BenchmarkId::new("add_addr", size), size, |b, i| {
             b.iter_batched(
                 || {
-                    let peer_ids = (0..*i).map(|_| PeerId::random()).collect::<Vec<_>>();
-                    let addr = "/ip4/255.0.0.1/tcp/42".parse::<Multiaddr>().unwrap();
-                    (peer_ids, addr)
+                    (0..*i)
+                        .map(|_| {
+                            format!("/ip4/255.0.0.1/tcp/42/p2p{}", PeerId::random().to_base58())
```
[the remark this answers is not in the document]
quake: resolved, please review again, thanks

driftluo (commented):
This change will cause all the historical data of the peer store to be cleared. Can we accept this impact? 

Most of the places where the interface is modified, the description of the peer id or IpPort is still reserved, which is the place to be modified.

network/src/peer_store/addr_manager.rs
driftluo: `port/443/ip/127.0.0.1` will be rejected by Tentacle, although it is a legal multiaddr

driftluo (approved):

jjyr (approved):

doitian (approved):

quake:
bors r=doitian,jjyr,driftluo

PR #2569, opened 2021-02-17
feat: add rpc generate_block_with_template to IntegrationTest rpc module

quake:
This PR adds `generate_block_with_template` rpc, so that dApps can get block template from `get_block_template` rpc, and then add or remove tx / proposal / uncle data in block template, and finally submit it via this rpc to control the newly generated block data.

btw, I think that we should add rpc doc to the `IntegrationTest` module, will create a new PR later.

### DIFF
--- a/rpc/src/module/test.rs
+++ b/rpc/src/module/test.rs
@@ -1,12 +1,23 @@
 use crate::error::RPCError;
 use ckb_app_config::BlockAssemblerConfig;
 use ckb_chain::chain::ChainController;
-use ckb_jsonrpc_types::{Block, Cycle, JsonBytes, Script, Transaction};
+use ckb_dao::DaoCalculator;
+use ckb_jsonrpc_types::{Block, BlockTemplate, Cycle, JsonBytes, Script, Transaction};
 use ckb_logger::error;
 use ckb_network::{NetworkController, SupportProtocols};
-use ckb_shared::shared::Shared;
+use ckb_shared::{shared::Shared, Snapshot};
 use ckb_store::ChainStore;
-use ckb_types::{core, packed, prelude::*, H256};
+use ckb_types::{
+    core::{
+        cell::{
+            resolve_transaction, OverlayCellProvider, ResolvedTransaction, TransactionsProvider,
+        },
+        BlockView,
+    },
+    packed,
+    prelude::*,
+    H256,
+};
 use ckb_verification_traits::Switch;
 use jsonrpc_core::Result;
 use jsonrpc_derive::rpc;
@@ -31,6 +42,9 @@ pub trait IntegrationTestRpc {
 
     #[rpc(name = "broadcast_transaction")]
     fn broadcast_transaction(&self, transaction: Transaction, cycles: Cycle) -> Result<H256>;
+
+    #[rpc(name = "generate_block_with_template")]
+    fn generate_block_with_template(&self, block_template: BlockTemplate) -> Result<H256>;
 }
 
 pub(crate) struct IntegrationTestRpcImpl {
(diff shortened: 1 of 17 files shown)


rpc/src/module/test.rs
```diff
@@ -165,4 +159,69 @@ impl IntegrationTestRpc for IntegrationTestRpcImpl {
...
+                    "resolve transactions error when generating block with block template, error: {:?}",
+                    err
+                );
+                err
+            }).ok()
+        }).collect();
+
+        let mut update_dao_template = block_template;
+        update_dao_template.dao = DaoCalculator::new(consensus, snapshot)
+            .dao_field(&rtxs, tip_header)
```
keroro520, line 198: assert `block_template.parent == tip_header.hash` before this line.

keroro520 (changes requested):

rpc/src/module/test.rs
quake: I expect this rpc to be used to submit uncle blocks as well, so I don't want to add this condition。

keroro520: `fn dao_field`'s second parameter is "parent header". And here hard-code `tip_header` as parent header, which means the block's parent is `tip_header`.

If block_template.parent != tip_header.hash, the `fn dao_field` return the wrong value.

quake: aha, copy / paste error, now fixed, please review again, thanks.

doitian (dismissed):

rpc/src/module/test.rs
```diff
@@ -165,4 +159,71 @@ impl IntegrationTestRpc for IntegrationTestRpcImpl {
...
+        let transactions_provider = TransactionsProvider::new(txs.as_slice().iter());
+        let overlay_cell_provider = OverlayCellProvider::new(&transactions_provider, snapshot);
+
+        let rtxs: Vec<_> = txs.iter().filter_map(|tx| {
+            resolve_transaction(
+                tx.clone(),
+                &mut seen_inputs,
+                &overlay_cell_provider,
+                snapshot,
+            ).map_err(|err| {
```
keroro520, line 189: The snapshot's alive cellset is based on `snapshot.tip` but not `parent`.
So I doubt this line has a bug.
For example, the `tx` spends an input cell that is alive for `parent` but not for `snapshot.tip`.

keroro520 (changes requested):

rpc/src/module/test.rs
quake: unresolvable tx will be ignored by `filter_map`, we cannot find a way to provide parent block's live cell set in current code.

keroro520: I see that the current code cannot provide the alive cellset for non-tip blocks.
Then I suggest that do not implicitly filter the unresolvable transactions but just explicitly throw an error to describe why this RPC cannot accept these transactions. As a user, a filtered block is not my expected.

quake: updated, please review again, thanks.

keroro520 (dismissed):

quake:
@doitian @keroro520 rebased with dev branch, please review again, thanks

keroro520:
The commit message is "wip"

quake:
> The commit message is "wip"

sorry, I forgot to squash the commits, now fixed

doitian (dismissed):

keroro520 (approved):

keroro520:
bors r=doitian,keroro520

keroro520 (approved):

doitian (approved):

doitian:
bors retry

PR #2570, opened 2021-02-18
chore: remove send_paused and use tentacle built-in max_send_buffer option

quake:


sync/src/synchronizer/get_blocks_process.rs
```diff
@@ -51,14 +51,6 @@ impl<'a> GetBlocksProcess<'a> {
                 continue;
             }
 
-            if self.nc.send_paused() {
```
doitian, line 54: How can we know that the send buffer is full now? Will `self.nc.send_message_to` return an error?

cc @driftluo

driftluo: An error will be returned on the service handle, and then the overloaded session will be forcibly closed

driftluo (approved):

doitian (approved):

sync/src/synchronizer/get_blocks_process.rs
doitian: Will this loop return early so we don't waste time to fetch the block from database and push it to the queue.

driftluo: The framework cannot control this behavior. If the session is killed here because exceed the upper limit of the buffer, sending any data is an invalid operation and will be discarded immediately, but it cannot interrupt the loop code

quake:
bors r=doitian,driftluo

PR #2571, opened 2021-02-19
feat: request the approval for database migrations

yangby-cryptape:
- docs: fix a typo
- chore: no migration version means the database is uninitialized
  At present, `migration_check` will return `true` if the database is uninitialized.
  But an uninitialized database doesn't have data required to migrate.
  This is important when do `migration_check` in `ckb run`.
- feat: request the approval for database migrations

### DIFF
--- a/ckb-bin/src/subcommand/migrate.rs
+++ b/ckb-bin/src/subcommand/migrate.rs
@@ -2,6 +2,8 @@ use ckb_app_config::{ExitCode, MigrateArgs};
 use ckb_async_runtime::Handle;
 use ckb_shared::shared::SharedBuilder;
 
+use crate::helper::prompt;
+
 pub fn migrate(args: MigrateArgs, async_handle: Handle) -> Result<(), ExitCode> {
     let builder = SharedBuilder::new(&args.config.db, None, async_handle);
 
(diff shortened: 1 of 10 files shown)


ckb-bin/src/subcommand/migrate.rs
```diff
@@ -13,6 +15,28 @@ pub fn migrate(args: MigrateArgs, async_handle: Handle) -> Result<(), ExitCode>
         }
     }
 
+    if !args.force {
+        if atty::is(atty::Stream::Stdin) && atty::is(atty::Stream::Stdout) {
```
doitian, line 19: No need to prompt if `builder.migration_check()` returns true.

doitian (changes requested):

doitian:
Some migrations may be very fast to execute, such as adding a single key into the database. We can let the migration tell whether it is time consuming, and only ask for confirmations in `ckb migration` and disable `ckb run` automatic migration when there’s at least a time consuming migration. However, it could be another PR.

doitian (approved):

quake (approved):

yangby-cryptape:
bors merge=doitian,quake

yangby-cryptape:
bors retry

PR #2572, opened 2021-02-19
perf: partial read tx hashes

quake:
According to the profiler result of sync test, deserializing all transactions in a block and reading only its hash each time is an a factor that affects performance. This PR changed it to using partial reads, theoretically able to improve sync performance by ~2%

### DIFF
--- a/store/src/store.rs
+++ b/store/src/store.rs
@@ -142,9 +142,9 @@ pub trait ChainStore<'a>: Send + Sync + Sized {
             )
             .take_while(|(key, _)| key.starts_with(prefix))
             .map(|(_key, value)| {
-                let reader =
-                    packed::TransactionViewReader::from_slice_should_be_ok(&value.as_ref());
-                reader.hash().to_entity()
+                packed::TransactionViewReader::new_unchecked(&value.as_ref())
+                    .hash()
+                    .to_entity()
             })
             .collect();
 


quake:
benchmark, sync-test

nervos-bot-user:
#### Benchmark Result
  - TPS: 493
  - CKB Version: e5b310d3d77259619cb4932b60fd53136913f0cc
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

store/src/store.rs
```diff
@@ -142,9 +142,8 @@ pub trait ChainStore<'a>: Send + Sync + Sized {
             )
             .take_while(|(key, _)| key.starts_with(prefix))
             .map(|(_key, value)| {
-                let reader =
-                    packed::TransactionViewReader::from_slice_should_be_ok(&value.as_ref());
-                reader.hash().to_entity()
+                // transaction hash is serialized as TransactionView's first field, we use partial reads here to avoid the extra cost of deserialization.
+                packed::Byte32::new_unchecked(value[16..48].to_owned().into())
```
yangby-cryptape, line 146: If the purpose of this PR is to skip all checks, the above code is equal to `packed::TransactionViewReader::new_unchecked(&value.as_ref()).hash().to_entity()`.

By the way, I think at least the length should be checked to avoid `panic`.
The deserialization only checks all the structure of data -- all length are right.
`Reader` doesn't do `clone`.

nervos-bot-user:
#### Sync Test Result
  - CKB Version: e5b310d3d77259619cb4932b60fd53136913f0cc
  - Instance Region: eu-west-2
  - Sync Start: 2021-02-19 03:32:58
  - Sync End: 2021-02-19 05:45:23
  - Sync Height: 3836145
  - Sync Speed: 482 blocks/s

quake:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: e5b310d3d77259619cb4932b60fd53136913f0cc
  - Instance Region: ap-southeast-1
  - Sync Start: 2021-02-19 07:15:36
  - Sync End: 2021-02-19 09:10:30
  - Sync Height: 3836936
  - Sync Speed: 556 blocks/s

store/src/store.rs
quake: resolved, please review again, and I created a new PR: https://github.com/nervosnetwork/ckb/pull/2579, if it's approved, we can close this PR.

doitian (approved):

nervos-bot-user:

doitian:
#2579 is closed because we have to discuss whether it's worth to skip the lengths verification for performance. If we don't verify the data form database, it may crash the app anywhere when we read the data.

But for this PR, the crash only happens in this function. The unchecked reader does not leak outside. So I approve this PR.

store/src/store.rs
```diff
@@ -142,9 +142,9 @@ pub trait ChainStore<'a>: Send + Sync + Sized {
             )
             .take_while(|(key, _)| key.starts_with(prefix))
             .map(|(_key, value)| {
-                let reader =
-                    packed::TransactionViewReader::from_slice_should_be_ok(&value.as_ref());
-                reader.hash().to_entity()
+                packed::TransactionViewReader::new_unchecked(&value.as_ref())
+                    .hash()
+                    .to_entity()
```
doitian, line 147:
```suggestion
                // DO NOT leak the unchecked reader to the callee.
                //
                // Using `new_unchecked` here can improve the sync performance by about 2%. If the data is dirty, the app crashes here or gets a wrong hash, which has the same guarantee with `from_slice_should_be_ok`.
                packed::TransactionViewReader::new_unchecked(&value.as_ref())
                    .hash()
                    .to_entity()
```

doitian (changes requested):

doitian:
After sync offline with @quake @yangby-cryptape @xxuejie , we decide to close this PR:

* The code is no longer in the hot path.
* It does not save the I/O to read the whole block body, but only skip some memory lengths check. The performance improvement is not worth removing the data check.

PR #2579, opened 2021-02-21
perf: skip molecule checking in release build

quake:
we should view the the slice as infallible, trusted entity in release build, skip length and bounds check.

### DIFF
--- a/util/types/src/prelude.rs
+++ b/util/types/src/prelude.rs
@@ -40,6 +40,7 @@ where
     R: Reader<'r>,
 {
     fn from_slice_should_be_ok(slice: &'r [u8]) -> Self {
+        #[cfg(debug_assertions)]
         match Self::from_slice(slice) {
             Ok(ret) => ret,
             Err(err) => panic!(
@@ -48,6 +49,9 @@ where
                 hex_string(slice)
             ),
         }
+        #[cfg(not(debug_assertions))]
+        // we should view the the slice as infallible, trusted entity in release build, skip length and bounds check.
+        Self::new_unchecked(slice)
     }
 }
 


quake:
benchmark, sync-test

nervos-bot-user:
#### Benchmark Result
  - TPS: 471
  - CKB Version: 621579f233ef139cfda565201fec665ba4c936f8
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

nervos-bot-user:
#### Sync Test Result
  - CKB Version: 621579f233ef139cfda565201fec665ba4c936f8
  - Instance Region: eu-west-2
  - Sync Start: 2021-02-21 02:34:07
  - Sync End: 2021-02-21 04:45:18
  - Sync Height: 3845008
  - Sync Speed: 488 blocks/s

quake:
benchmark, sync-test

nervos-bot-user:
#### Benchmark Result
  - TPS: 484
  - CKB Version: 621579f233ef139cfda565201fec665ba4c936f8
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

chuijiaolianying:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: 621579f233ef139cfda565201fec665ba4c936f8
  - Instance Region: ap-southeast-1
  - Sync Start: 2021-02-22 01:32:55
  - Sync End: 2021-02-22 03:21:06
  - Sync Height: 3849505
  - Sync Speed: 593 blocks/s

doitian (changes requested):
I'm against this PR. The function name suggests that it will panic on invalid data, it is surprising that it does not. It is very easy to be abused.

If we really want to skip the verification only in release mode, we should add a new function to do that.

quake:
> I'm against this PR. The function name suggests that it will panic on invalid data, it is surprising that it does not. It is very easy to be abused.
> 
> If we really want to skip the verification only in release mode, we should add a new function to do that.

how about change the name to `from_verified_slice_should_be_ok` and move it to store crate as a private trait?

quake:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: 621579f233ef139cfda565201fec665ba4c936f8
  - Instance Region: eu-west-2
  - Sync Start: 2021-02-26 07:32:03
  - Sync End: 2021-02-26 09:47:53
  - Sync Height: 3872051
  - Sync Speed: 475 blocks/s

doitian:
> how about change the name to `from_verified_slice_should_be_ok` and move it to store crate as a private trait?

What's the reason that we cannot use `new_unchecked` in the places that we are sure the slice is valid but need to change the behavior of this function?

yangby-cryptape:
First, just an additional explanation:
Before the mainnet launched and the first month after mainnet launched, there are several panics in our sentry were caused because the format of data from storage were malformed, some of those panics were sent by this function `from_slice_should_be_ok(..)`.

Do no checks means we can make sure that:
- We would never make mistakes **again** when update data structures.
- The RocksDB will be always work well.

The worst situation I imagined is: no panics caused at beginning, but only malformed data, then we found the bug several days later.

Speaking of this, I remembered that in Matt Damon's film "The Martian", their rocket exploded because the NASA skip few checks which only have very very very small probability to find bugs.

Second, there no **strong** evidences could approve that those checks cost too many resource, so that we have to remove them.

nervos-bot-user:

doitian:
Close since it does not gains enough approvals.

PR #2584, opened 2021-02-23
chore: expose the depended tokio from ckb-async-runtime

keroro520:
Runtimes may be incompatible in different versions of tokio. As ckb-async-runtime wraps a specified version tokio, it should publicly expose that version tokio so that the ckb-async-runtime users can use components of that specified version tokio, such as `tokio::time::sleep`.


yangby-cryptape (approved):

zhangsoledad (approved):

doitian (approved):

keroro520:
bors r=zhangsoledad,yangby-cryptape,doitian

PR #2594, opened 2021-02-26
chore: add EpochProvider trait and refactor dao calculation

quake:
This PR added a `EpochProvider` trait and refactor dao calculation related code, eliminated the `util/dao` crate's store dependency.

benchmark `next_epoch_ext` was deleted because it is benchmarking the performance of getting the same epoch information, I think it is unnecessary and meaningless.


doitian (approved):

zhangsoledad (approved):

zhangsoledad:
bors r=doitian,zhangsoledad

PR #2609, opened 2021-03-10
improvement: don't cache all data of block status map in memory during IBD

yangby-cryptape:
### Purpose

Use less memory during IBD.

### Description

Since #2147, CKB caches most data of `header map` into disk during IBD, but the `block status map` isn't cached into disk.

Now, blocks in mainnet has become almost twice than the day when #2147 merged, this issue cannot continue to be ignored.

So, I reuse the code in #2147 and let the disk-cached `HashMap` to be more abstract, and use the same RocksDB backend to cache `block status map`.

The first `byte` of the key is the prefix which specifies the data type, so we can reuse this RocksDB backend to cache other `HashMap` if we need in the future.
At present, the key of `header map` will be `[1] + [block-hash]` and the key of `block status map` will be `[2] + [block-hash]`.

### Commits

**Since I renamed "header_map" to "cached_hashmap" but GitHub didn't detect that, IMHO, review each commits of this PR will be simple that review the whole PR.**

- refactor: let the disk-cached HashMap to be more abstract

- improvement: don't cache all data of block status map in memory during IBD

- refactor: use same backend as disk cache for header map and block status map during IBD

### Effect

Deploy two nodes in the same environment:
- Ubuntu Server 18.04
- AWS t3.small; x86_64; 2vCPU; 2G RAM; 100 G g2 Disk; Up to 5 Gigabit; ap-southeast-1a
- Use the default `ckb.toml` (from `ckb init -C .`)

But `NodeA` use the latest commit in branch develop, `NodeB` use the last commit of this PR.

The max memory used in `NodeB` was decrease by about `350 MiB = 1350 MiB - 1000 MiB` than the max memory used in `NodeA` during IBD.

References:
- [CKB PR-2147 improvement: don't cache all data of header map in memory during IBD](https://github.com/nervosnetwork/ckb/pull/2147)


yangby-cryptape:
benchmark

nervos-bot-user:
#### Benchmark Result
  - TPS: 476
  - CKB Version: 971f298e9eacdd8bb2af79c4b6cd2bf74ad33b3a
  - Instance Type: c5.xlarge
  - Instances Count: 3
  - Bench Type: In2Out2

yangby-cryptape:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: 971f298e9eacdd8bb2af79c4b6cd2bf74ad33b3a
  - Instance Region: eu-west-2
  - Sync Start: 2021-03-11 02:57:41
  - Sync End: 2021-03-11 05:22:51
  - Sync Height: 3934411
  - Sync Speed: 451 blocks/s

yangby-cryptape:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: 971f298e9eacdd8bb2af79c4b6cd2bf74ad33b3a
  - Instance Region: eu-west-2
  - Sync Start: 2021-03-11 06:10:59
  - Sync End: 2021-03-11 08:22:24
  - Sync Height: 3934870
  - Sync Speed: 499 blocks/s

driftluo (approved):

quake:
LGTM, one suggestion for `cached_hashmap` naming, it's a lru cache with different storage tiers, we may refer to other common cache library naming: on-heap/off-heap/disk-store, etc ( https://www.ehcache.org/documentation/3.2/caching-concepts.html#storage-tiers )

quake (approved):

nervos-bot-user:

yangby-cryptape:
Re-open when all conflicts are resolved.

PR #2610, opened 2021-03-11
feat(rpc): add a controller to support reload configurations

yangby-cryptape:
### Commits

- chore: add a macro to construct `ExitCode=Failure` more efficient
- feat(rpc): add a controller to support reload configurations

### Known Issues

- Update `Shared` after the RPC server started will restart the RPC server, since the subscription RPC couldn't be hot-plugged.

  https://github.com/nervosnetwork/ckb/blob/1b776c23a2ea6af335afb6e14ae1725c9dcf3cef/rpc/src/controller.rs#L404-L419


nervos-bot-user:

PR #2611, opened 2021-03-11
improvement: smaller block status map during IBD

yangby-cryptape:
Not conflict with #2609:

- This PR to make block status map to be smaller.

  After this PR, the max size of block status map during IBD: `4*10^6 -> 5*10^4` (current mainnet has about `4*10^6` blocks, and **I only test once**).

- #2609 to cache block status map into disk if the size of block status map is bigger than a threshold (default: `10^6`, `1/4` of current mainchain).

  If this PR merged, #2609 will not work in almost all scenarios at present.

  But I don't close it because:
  - The max size of block status map will be increased if network is poor or too many forked blocks.
  - We can set it to a smaller value, #2609 provide an option.
  - That PR refactor the `HeaderMap` to `CachedHashMap`, so we can reuse the temporary storage for other `HashMap`s later.

Both of these two PRs could be merged.

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -1391,8 +1391,6 @@ impl SyncShared {
         );
         self.state.header_map.insert(header_view.clone());
         self.state
-            .insert_block_status(header.hash(), BlockStatus::HEADER_VALID);
-        self.state
             .peers()
             .may_set_best_known_header(peer, header_view.clone());
         self.state.may_set_shared_best_header(header_view);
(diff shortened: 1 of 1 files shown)


yangby-cryptape:
sync-test

nervos-bot-user:
#### Sync Test Result
  - CKB Version: a251023f546ac67b2ba0ac8405645d683418736e
  - Instance Region: eu-west-2
  - Sync Start: 2021-03-12 00:15:32
  - Sync End: 2021-03-12 02:23:13
  - Sync Height: 3938449
  - Sync Speed: 514 blocks/s

sync/src/types/mod.rs
```diff
@@ -1984,15 +1982,19 @@ impl ActiveChain {
...
-                let verified = self
-                    .snapshot
-                    .get_block_ext(block_hash)
-                    .map(|block_ext| block_ext.verified);
-                match verified {
-                    None => BlockStatus::UNKNOWN,
-                    Some(None) => BlockStatus::BLOCK_STORED,
-                    Some(Some(true)) => BlockStatus::BLOCK_VALID,
-                    Some(Some(false)) => BlockStatus::BLOCK_INVALID,
+                if self.state.header_map.contains_key(block_hash) {
```
driftluo, line 1985: It contains BLOCK_INVALID status inside, so you can’t write it like this

driftluo (approved):

keroro520 (approved):

quake (approved):

driftluo:
bors r=driftluo,keroro520,quake

PR #2612, opened 2021-03-12
test: no timeout if the tx-pool still works

yangby-cryptape:
Fix few panics in integration tests, when tx-pool doesn't reach the tip as chain in the timeout (10s).

Reset the timeout if the tx-pool still works. So the timeout is flexible but also bounded.

### DIFF
--- a/test/src/node.rs
+++ b/test/src/node.rs
@@ -382,13 +382,22 @@ impl Node {
         let rpc_client = self.rpc_client();
         let mut chain_tip = rpc_client.get_tip_header();
         let mut tx_pool_tip = rpc_client.tx_pool_info();
-        let instant = Instant::now();
+        if chain_tip.hash == tx_pool_tip.tip_hash {
+            return;
+        }
+        let mut instant = Instant::now();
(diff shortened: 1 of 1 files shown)


doitian (dismissed):

test/src/node.rs
```diff
@@ -382,13 +382,22 @@ impl Node {
         let rpc_client = self.rpc_client();
         let mut chain_tip = rpc_client.get_tip_header();
         let mut tx_pool_tip = rpc_client.tx_pool_info();
-        let instant = Instant::now();
+        let mut instant = Instant::now();
+        if chain_tip.hash == tx_pool_tip.tip_hash {
+            return;
+        }
```
quake, line 388:
```suggestion
        if chain_tip.hash == tx_pool_tip.tip_hash {
            return;
        }
        let mut instant = Instant::now();
```

quake (approved):

doitian (approved):

doitian:
bors r=quake,doitian

PR #2621, opened 2021-03-18
feat: cache tx-pool state into a file when it has been shutdown

yangby-cryptape:
#### Description

- ~At first, I want to save the structured state data of TxPool into RocksDB.~

  ~But the structure of TxPool is too complex, so I just serialize the whole state of TxPool and write it into a single binary.~

- Dump pending and proposed transactions into disk when CKB shutdown, and replay them when CKB started.

- Cache synchronization in real time will reduce performance.

  Therefore, I just write the cache before exit.

#### Known Issues:

- No cached tx-pool state file if forced exit.


tx-pool/src/process.rs
```diff
@@ -650,10 +652,50 @@ impl TxPoolService {
...
+                    "Failed to open the tx-pool cache file [{:?}]: {}",
+                    self.tx_pool_config.cache_file, err
+                );
+                OtherError::new(errmsg)
+            })?;
+        let pool_cache = {
+            let txs = TransactionVec::new_builder()
+                .extend(tx_pool.get_all_transactions().map(TransactionView::data))
+                .build();
+            TxPoolCache::new_builder()
```
quake, line 678: I don't see the need to introduce a new serialization structure, we can write the `Transaction` bytes to dump file line by line.

tx-pool/src/process.rs
```diff
@@ -650,10 +652,50 @@ impl TxPoolService {
 
     pub(crate) async fn clear_pool(&mut self, new_snapshot: Arc<Snapshot>) {
         let mut tx_pool = self.tx_pool.write().await;
-        let config = tx_pool.config;
+        let config = Arc::clone(&tx_pool.config);
         self.last_txs_updated_at = Arc::new(AtomicU64::new(0));
         *tx_pool = TxPool::new(config, new_snapshot, Arc::clone(&self.last_txs_updated_at));
     }
+
+    pub(crate) async fn cache_pool(&self) -> Result<(), AnyError> {
```
quake, line 660: This is not a caching behavior, suggest to rename the fn name to `dump_pool` and change the default dump file name to `tx_pool.dump`

tx-pool/src/service.rs
```diff
@@ -473,17 +493,75 @@ impl TxPoolServiceBuilder {
...
         let last_txs_updated_at = Arc::new(AtomicU64::new(0));
         let consensus = self.snapshot.cloned_consensus();
         let tx_pool = TxPool::new(
-            self.tx_pool_config,
+            Arc::new(self.tx_pool_config),
             self.snapshot,
             Arc::clone(&last_txs_updated_at),
         );
 
+        let pool_cache = if tx_pool.config.cache_file.exists() {
```
quake, line 505: suggest to extracting code to a `load_tx_pool` fn and use a background thread to add tx entry to the pool one by one.

quake (changes requested):

tx-pool/src/process.rs
yangby-cryptape: I will push a commit to fix this.

I remove part of codes when merge develop, so the name is not proper now.

tx-pool/src/process.rs
yangby-cryptape: I want to provide predictable compatibility, so may we want to cache more data or change the format of transactions.

tx-pool/src/service.rs
yangby-cryptape: I will extract code, **but I don't want to use a background thread**.

**In fact, I want to block all processes (but, not yet) until the tx pool is ready from the recovery.**
It's the simplest way to avoid few issues.
For example, I think the cached transactions should have the highest priority, but if network is working parallel, some other conflict transactions may be insert before these transactions.
And this node may request already-existed transactions through network.

tx-pool/src/service.rs
```diff
@@ -400,11 +402,48 @@ impl TxPoolController {
...
+        // a trick to commit transactions with the correct order
+        let mut remain_size = data.transactions().len();
+        let mut txs_next_turn = Vec::new();
+        for tx in data.transactions() {
+            let tx_view = tx.into_view();
+            if self.submit_local_tx(tx_view.clone())?.is_err() {
+                txs_next_turn.push(tx_view);
+            }
+        }
+        while !txs_next_turn.is_empty() && remain_size != txs_next_turn.len() {
```
quake, line 429: There is a potential dead loop here: because chain and pool processing is asynchronous, if the ckb is closed between the time of chain inserts a block and removes the committed txs from pool, the dumped transactions will never replay successfully, resulting in a dead loop. Suggest to add a loop count limit to replay txs.

tx-pool/src/process.rs
quake: We can use `TransactionReader::from_compatible_slice` to resolve the format compatibility issue of transaction, and since we are using the replay strategy, I don't see why changing the structure of the tx pool itself would cause the old dump file not work?

tx-pool/src/service.rs
yangby-cryptape: > There is a potential dead loop here: because chain and pool processing is asynchronous, if the ckb is closed between the time of chain inserts a block and removes the committed txs from pool, the dumped transactions will never replay successfully, resulting in a dead loop. Suggest to add a loop count limit to replay txs.

`remain_size != txs_next_turn.len()`

If in any turn there are no transaction was processed, then this loop will be broken.

tx-pool/src/process.rs
yangby-cryptape: Maybe we will cache more data, no only transactions.

yangby-cryptape: > I don't see why changing the structure of the tx pool itself would cause the old dump file not work?

I couldn't agree with you.

IMHO, keep a version field could handle some things in future **we couldn't imagine** at the present.
**I don't know** what could be happened in future.
**If I could or someone told me, I wouldn't left them to the future, I would done them in this PR.**

p.s. In the previous commit, I saved the whole tx pool state and it works well. After discussion, I changed to save transactions only. You never know what happens next.

tx-pool/src/service.rs
yangby-cryptape: > because chain and pool processing is asynchronous

Yes, I spent some hours on that.
Before tx-pool rewrite, I could drop network and rpc before tx-pool, also I could create tx-pool before network and rpc.

I really don‘t like current implementation, too.
I except that I could shutdown all other components before save tx-pool, and load tx-pool before start any other components.
But I didn't find a way to do that.

quake: aha, I didn't notice the `remain_size != txs_next_turn.len()` code here, however, is it easier to read by using `loop break`?
```
        let mut txs = data.transactions().into_iter().map(|tx| tx.into_view()).collect::<Vec<_>>();
        loop {
            let mut failed_txs = Vec::new();
            for tx in txs.iter() {
                if self.submit_local_tx(tx.clone())?.is_err() {
                    failed_txs.push(tx.clone());
                }
            }
            // return when all success or all failed
            if failed_txs.is_empty() || txs.len() == failed_txs.len() {
                break;
            }
            txs = failed_txs;
        }
        Ok(())
```

tx-pool/src/process.rs
quake: We can save the version in the first line of the dump file, and dump all the transaction data in the second line by reusing `TransactionVec` which has been defined in `blockchain.mol` already, so that we can keep the compatibility of data upgrade in the future, and we don't need to define the schema file repeatedly.

test/src/main.rs
```diff
@@ -380,6 +380,8 @@ fn all_specs() -> Vec<Box<dyn Spec>> {
         Box::new(TemplateSizeLimit),
         Box::new(PoolReconcile),
         Box::new(PoolResurrect),
+        #[cfg(target_os = "linux")]
+        Box::new(PoolCache),
```
quake, line 384:
```suggestion
        Box::new(PoolDump),
```

tx-pool/src/persisted/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+    fs::OpenOptions,
+    io::{Read as _, Write as _},
+    path::PathBuf,
+};
+
+mod generated;
+
+pub(crate) use generated::*;
+
+/// The version of the cache data.
```
quake, line 13:
```suggestion
/// The version of the dump data.
```

tx-pool/src/process.rs
yangby-cryptape: - How to define **"first line"** in a binary file?

  Maybe JSON is better than hex string, but store JSON string will introduce a lot of dependencies. 
  (JSON is my first design, it was simple but made the crate to be heavy.)

  If you insisted, I prefer to use JSON, at least it is readable; hex string has no advantages at all.

- Maybe it was not clear: there **only one (two, in fact) struct were new**, all others were imported from the `ckb-types`.

  I don't think define the schema file repeatedly is so bad.
  In fact, there are several projects copied the schema file and re-generate codes for **all** structs.
  Here, I provided an example how to use the schema file **as interfaces**, and use the rust code of these structs **without re-generate any duplicate codes**.
  I think it wasn't bad, not only for this feature, but also as a practical example..

  I had been asked several times how to add new structs with `ckb-types`.

  **If we didn't want to use it, even we had a proper scene we still tried to avoid to use it, how dare we encourage other developers to use it?** :dog:

nervos-bot-user:

PR #2628, opened 2021-03-24
feat: Fix download scheduler

driftluo:
1. disable penalty when download nodes are scarce
2. allow the protection node to be disconnected due to sync judgment

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -7,8 +7,8 @@ use ckb_chain_spec::consensus::Consensus;
 use ckb_constant::sync::{
     BLOCK_DOWNLOAD_TIMEOUT, HEADERS_DOWNLOAD_HEADERS_PER_SECOND, HEADERS_DOWNLOAD_INSPECT_WINDOW,
     HEADERS_DOWNLOAD_TOLERABLE_BIAS_FOR_SINGLE_SAMPLE, INIT_BLOCKS_IN_TRANSIT_PER_PEER,
-    MAX_BLOCKS_IN_TRANSIT_PER_PEER, MAX_HEADERS_LEN, POW_INTERVAL, RETRY_ASK_TX_TIMEOUT_INCREASE,
-    SUSPEND_SYNC_TIME,
+    MAX_BLOCKS_IN_TRANSIT_PER_PEER, MAX_HEADERS_LEN, MAX_OUTBOUND_PEERS_TO_PROTECT_FROM_DISCONNECT,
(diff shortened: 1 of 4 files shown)


test/src/specs/sync/get_blocks.rs
```diff
@@ -53,14 +53,14 @@ impl Spec for GetBlocksTimeout {
 
         let received = wait_get_blocks_point(&net, &node1, block_download_timeout_secs * 2, 1);
         assert!(
-            received.is_none(),
-            "Should not received GetBlocks anymore, the timeout could be any number."
+            received.is_some(),
```
quake, line 56: we shouldn't modify the assertion here, please setup up more dummy nodes and connect them to node1.

driftluo: Here, our test purpose has changed. The current test is that even if the timeout is triggered, the eviction will not occur if the number of connections is not enough. 

---

By the way, I think it is currently impossible to measure eviction scenarios in multiple node states. The reason is as follows：
1. There is no problem with the generation of multiple connections, but multiple connections are processing downloads at the same time, and this is considered uncertain 
2. Even if multiple connections are processing download tasks at the same time, but you need to respond to random download tasks on other connections, and do not respond on a specific node, this is not a well-controlled process, because the generation of download tasks is not caused by you control, but the node started by another process

quake: the assertion message of line 57 and line 66 should be modified accordingly if you want to change the test spec, "Should received GetBlocks anymore" looks weird...

driftluo: I rewrote the panic message, please review again

quake (approved):

sync/src/types/mod.rs
```diff
@@ -683,6 +685,16 @@ impl InflightBlocks {
...
         let mut disconnect_list = HashSet::new();
+        // Since statistics are currently disturbed by the processing block time, when the number
+        // of transactions increases, the node will be accidentally evicted.
+        //
+        // Especially on machines with poor CPU performance, the node connection will be frequently
+        // disconnected due to statistics.
+        //
+        // In order to protect the decentralization of the network and ensure the survival of low-performance
+        // nodes, the penalty mechanism will be closed when the number of download nodes is less than the number of protected nodes
+        let should_punish = self.download_schedulers.len() > self.protect_num;
```
doitian, line 696: Just a small suggestion: this can be extract as a small function `self.should_punish()`

doitian (approved):

driftluo:
bors r=quake,doitian

PR #2639, opened 2021-03-31
chore: dail observed address use another way

driftluo:
There is a modification conflict with #2565

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -39,6 +39,7 @@ use p2p::{
     yamux::config::Config as YamuxConfig,
     SessionId,
 };
+use rand::prelude::IteratorRandom;
 #[cfg(feature = "with_sentry")]
 use sentry::{capture_message, with_scope, Level};
 use std::sync::mpsc;
@@ -297,7 +298,7 @@ impl NetworkState {
             .collect::<Vec<_>>()
     }
 
-    pub(crate) fn can_dial(&self, addr: &Multiaddr, allow_dial_to_self: bool) -> bool {
+    pub(crate) fn can_dial(&self, addr: &Multiaddr) -> bool {
         let peer_id = extract_peer_id(addr);
         if peer_id.is_none() {
             error!("Do not dial addr without peer id, addr: {}", addr);
@@ -305,11 +306,11 @@ impl NetworkState {
         }
         let peer_id = peer_id.as_ref().unwrap();
 
-        if !allow_dial_to_self && self.local_peer_id() == peer_id {
+        if self.local_peer_id() == peer_id {
             trace!("Do not dial self: {:?}, {}", peer_id, addr);
             return false;
         }
-        if !allow_dial_to_self && self.public_addrs.read().contains(&addr) {
+        if self.public_addrs.read().contains(&addr) {
             trace!(
                 "Do not dial listened address(self): {:?}, {}",
                 peer_id,
@@ -379,9 +380,8 @@ impl NetworkState {
         p2p_control: &ServiceControl,
         addr: Multiaddr,
         target: TargetProtocol,
-        allow_dial_to_self: bool,
     ) -> Result<(), Error> {
-        if !self.can_dial(&addr, allow_dial_to_self) {
+        if !self.can_dial(&addr) {
             return Err(Error::Dial(format!("ignore dialing addr {}", addr)));
         }
 
@@ -400,7 +400,6 @@ impl NetworkState {
             p2p_control,
             addr,
             TargetProtocol::Single(SupportProtocols::Identify.protocol_id()),
-            false,
         ) {
             debug!("dial_identify error: {}", err);
         }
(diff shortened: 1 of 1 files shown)


network/src/network.rs
```diff
@@ -468,31 +469,46 @@ impl NetworkState {
...
                 p2p_control,
                 self.local_peer_id(),
-                addr,
+                addr.clone(),
                 TargetProtocol::Single(SupportProtocols::Identify.protocol_id()),
                 true,
             ) {
-                debug!("try_dial_observed_addrs error {}", err);
+                Err(err) => {
+                    debug!("try_dial_observed_addrs error {}", err);
```
yangby-cryptape, line 484: Just a thinly veiled criticism, the key word `error` in `debug` logs is terrible.
- It makes `error` to be worthless.
  When I enable `log=trace` to debug, then `grep -i error` is meaningless.
  I have to use a complex filter on a huge file, and wait more time.

- Cause panic to users who enable `log=debug`.
  Not all users could distinguish which `error` are insignificant and which are urgent. 
  Yes, I can filter the log records by log level, but I have to use a more complex logs filter, which is terrible with a huge file.

network/src/network.rs
```diff
@@ -591,7 +607,15 @@ impl<T: ExitHandler> ServiceHandle for EventHandler<T> {
...
                 {
                     debug!("dial observed address success: {:?}", address);
-                    public_addrs.insert(addr);
+                    match multiaddr_to_socketaddr(&addr) {
+                        Some(ip) => {
+                            if is_reachable(ip.ip()) {
+                                public_addrs.insert(addr);
+                            }
+                        }
+                        None => (),
```
yangby-cryptape, line 616: Maybe `if ... let ...` is better?

yangby-cryptape (approved):

yangby-cryptape (approved):

yangby-cryptape (approved):

network/src/network.rs
```diff
@@ -468,31 +469,43 @@ impl NetworkState {
     /// this method is intent to check observed addr by dial to self
     pub(crate) fn try_dial_observed_addrs(&self, p2p_control: &ServiceControl) {
         let mut pending_observed_addrs = self.pending_observed_addrs.write();
-        let public_addrs = { self.public_addrs.read().clone() };
-        for addr in pending_observed_addrs
-            .drain()
-            .chain(public_addrs.into_iter())
-        {
+        if pending_observed_addrs.is_empty() {
+            let addrs = self.public_addrs.write();
```
TheWaWaR, line 473: Why dial `self.public_addrs` in a `fn try_dial_observed_addrs()` ?

driftluo: the observed address is not always valid. For example, the address obtained through the reuse port extension may become invalid as the connection is disconnected. This part of the address information needs to be cleaned up.

network/src/network.rs
```diff
@@ -426,30 +427,43 @@ impl NetworkState {
...
+            let addr = { pending_observed_addrs.iter().next().unwrap().clone() };
+            trace!("try dial observed addr: {:?}", addr);
+            match self.dial_inner(
+                p2p_control,
+                addr.clone(),
+                TargetProtocol::Single(SupportProtocols::Identify.protocol_id()),
+                true,
+            ) {
+                Err(err) => {
+                    trace!("try_dial_observed_addrs fail {} on pending observed", err);
```
quake, line 454: should remove the addr from `pending_observed_addrs` when dial error?

driftluo: > should remove the addr from `pending_observed_addrs` when dial error?

I think this error is generally that the peerid already exists in the dial list, not tentacle dial queue full, so I think it’s not necessary to delete it here

driftluo: The latest commit deals with this problem in a more understandable way

network/src/network.rs
```diff
@@ -426,30 +424,37 @@ impl NetworkState {
...
+        if pending_observed_addrs.is_empty() {
+            let addrs = self.public_addrs.read();
+            if addrs.is_empty() {
+                return;
+            }
+            // random get addr
+            let index = rand::random::<usize>() % addrs.len();
+            let addr = addrs.iter().nth(index).unwrap();
+            if let Err(err) = p2p_control.dial(
+                addr.clone(),
```
quake, line 436:
```suggestion
            if let Some(addr) = addrs.iter().choose(&mut rand::thread_rng()) {
                if let Err(err) = p2p_control.dial(
                    addr.clone(),
                    TargetProtocol::Single(SupportProtocols::Identify.protocol_id()),
                ) {
                    trace!("try_dial_observed_addrs fail {} on public address", err)
                }
            }
```

quake (dismissed):

quake (approved):

nervos-bot-user:

quake (approved):

quake (approved):

doitian (approved):

driftluo:
bors r=quake,doitian

PR #2652, opened 2021-04-15
refactor: let verifiers know the real tip number and tx status

yangby-cryptape:
### Purpose

In the latest version, we hard-code `tip_number + 1` as the block number for verification.

For example:
https://github.com/nervosnetwork/ckb/blob/dd9040f207364a039fdbe5fa9a21510c26b67b8f/tx-pool/src/util.rs#L86-L93
https://github.com/nervosnetwork/ckb/blob/dd9040f207364a039fdbe5fa9a21510c26b67b8f/tx-pool/src/util.rs#L99-L106

There are several defects:

- The tip header hash doesn't correspond to the tip block number.

- The epoch number doesn't correspond to the tip block number.

- For hard fork at a particular epoch, the script verifier have to know the exact epoch number.

  But epoch number couldn't add a number directly, since a epoch length is depend on the whole previous epoch information.

  So, the best way, I think, is to pass the real epoch number to all verifiers and tell those verifiers the minimum block latency.

  Then, I think the block number should be checked with same logic.

### Further

The `MaturityVerifier` and `SinceVerifier` with epoch numbers could be optimized, too.

### DIFF
--- a/test/src/specs/tx_pool/valid_since.rs
+++ b/test/src/specs/tx_pool/valid_since.rs
@@ -1,4 +1,7 @@
-use crate::util::mining::{mine, mine_until_out_bootstrap_period};
+use crate::util::{
+    check,
+    mining::{mine, mine_until_out_bootstrap_period},
+};
 use crate::utils::{
     assert_send_transaction_fail, since_from_absolute_block_number, since_from_absolute_timestamp,
     since_from_relative_block_number, since_from_relative_timestamp,
@@ -6,7 +9,7 @@ use crate::utils::{
 use crate::{Node, Spec, DEFAULT_TX_PROPOSAL_WINDOW};
 
 use ckb_logger::info;
-use ckb_types::core::BlockNumber;
+use ckb_types::core::{BlockNumber, TransactionView};
 use std::thread::sleep;
 use std::time::Duration;
 
@@ -32,7 +35,8 @@ impl Spec for ValidSince {
 impl ValidSince {
     pub fn test_since_relative_block_number(&self, node: &Node) {
         mine_until_out_bootstrap_period(node);
-        let relative: BlockNumber = 5;
+        let started_tip_number = node.get_tip_block_number();
+        let relative: BlockNumber = 10;
         let since = since_from_relative_block_number(relative);
         let transaction = {
             let cellbase = node.get_tip_block().transactions()[0].clone();
@@ -40,7 +44,7 @@ impl ValidSince {
         };
 
         // Failed to send transaction since SinceImmaturity
-        for _ in 1..relative {
+        for _ in 1..=(relative - 3) {
             assert_send_transaction_fail(
                 node,
                 &transaction,
@@ -56,11 +60,13 @@ impl ValidSince {
                 .is_ok(),
             "transaction is ok, tip is equal to relative since block number",
         );
+
+        Self::check_committing_process(node, &transaction, started_tip_number + relative);
     }
 
     pub fn test_since_absolute_block_number(&self, node: &Node) {
         mine_until_out_bootstrap_period(node);
-        let absolute: BlockNumber = node.rpc_client().get_tip_block_number() + 5;
+        let absolute: BlockNumber = node.rpc_client().get_tip_block_number() + 10;
         let since = since_from_absolute_block_number(absolute);
         let transaction = {
             let cellbase = node.get_tip_block().transactions()[0].clone();
@@ -69,7 +75,7 @@ impl ValidSince {
 
         // Failed to send transaction since SinceImmaturity
         let tip_number = node.rpc_client().get_tip_block_number();
-        for _ in tip_number + 1..absolute {
+        for _ in tip_number + 1..=(absolute - 3) {
             assert_send_transaction_fail(
                 node,
                 &transaction,
@@ -85,6 +91,8 @@ impl ValidSince {
                 .is_ok(),
             "transaction is ok, tip is equal to absolute since block number",
         );
+
+        Self::check_committing_process(node, &transaction, absolute);
     }
 
     pub fn test_since_relative_median_time(&self, node: &Node) {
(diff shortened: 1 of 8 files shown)


verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
 use std::cell::RefCell;
 use std::collections::HashSet;
 
+/// The phase that transactions are in.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TransactionVerificationPhase {
+    /// The transaction has just been submitted.
+    ///
+    /// So the transaction will be:
```
keroro520, line 28: unless traffic jam?

keroro520: Confused with `Fresh`, `Submitted`, `Pending` ...

verification/src/transaction_verifier.rs
```diff
@@ -640,7 +702,7 @@ impl<'a, DL: HeaderProvider> SinceVerifier<'a, DL> {
         if since.is_absolute() {
             match since.extract_metric() {
                 Some(SinceMetric::BlockNumber(block_number)) => {
-                    if self.block_number < block_number {
+                    if self.tx_phase.earliest_committed_block(self.block_number) < block_number {
```
keroro520, line 705: Pay attention to this. You change the `block_number` from `tip_number + 1` to `tip_number` [outside](https://github.com/nervosnetwork/ckb/pull/2652/files#diff-f17291b75b5bafb128afde15b639dea00e71c718713a389060fe31cd9e53a914R103).

As a result, after the above change, block_number is 1 less than the old code.
I think here it should use `<=`.

keroro520 (changes requested):

verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
...
+    /// So the transaction will be committed after the `tip_number + 1` block.
+    Proposed,
+    /// The transaction is commit.
+    ///
+    /// So the transaction will be committed in current block.
+    Committed,
+}
+
+impl TransactionVerificationPhase {
+    /// Convert tip block number to earliest committed block number.
```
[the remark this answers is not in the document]
yangby-cryptape: The explanation of incremental blocks is already in the enumerated type definition. I think there are no need to make a copy.

verification/src/transaction_verifier.rs
yangby-cryptape: I think the integration test  `ValidSince` could prove that current comparison is correct.

test/src/specs/tx_pool/valid_since.rs
```diff
@@ -56,11 +60,29 @@ impl ValidSince {
                 .is_ok(),
             "transaction is ok, tip is equal to relative since block number",
         );
+
+        // Pending
+        node.assert_tx_pool_size(1, 0);
+        assert!(!is_transaction_committed(node, &transaction));
```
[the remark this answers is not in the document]
yangby-cryptape: DONE!

keroro520 (approved):

nervos-bot-user:

driftluo (dismissed):

verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
...
+    }
+
+    /// Convert tip epoch number to earliest committed epoch number.
+    pub fn earliest_committed_epoch(self, tip_epoch: EpochNumberWithFraction) -> EpochNumber {
+        let number = tip_epoch.number();
+        let length = tip_epoch.length();
+        let index = tip_epoch.index();
+        let index_added = match self {
+            Self::Submitted => 3,
+            Self::Gap => 2,
```
doitian, line 64: The gap is not a constant but is configurable in the consensus via the option `tx_proposal_window`

verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
...
+        }
+    }
+
+    /// Convert tip epoch number to earliest committed epoch number.
+    pub fn earliest_committed_epoch(self, tip_epoch: EpochNumberWithFraction) -> EpochNumber {
+        let number = tip_epoch.number();
+        let length = tip_epoch.length();
+        let index = tip_epoch.index();
+        let index_added = match self {
+            Self::Submitted => 3,
```
doitian, line 63: This is not a constant but is configurable in the consensus. It should be set to `tx_proposal_window.closest + 1`

verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
...
+    /// So the transaction will be committed in current block.
+    Committed,
+}
+
+impl TransactionVerificationPhase {
+    /// Convert tip block number to earliest committed block number.
+    pub fn earliest_committed_block(self, tip_block_number: BlockNumber) -> BlockNumber {
+        match self {
+            Self::Submitted => tip_block_number + 3,
+            Self::Gap => tip_block_number + 2,
```
doitian, line 51: The tx proposal window is configurable in the consensus, see comment below.

verification/src/transaction_verifier.rs
```diff
@@ -20,6 +20,59 @@ use lru::LruCache;
...
+    ///
+    /// So the transaction will be committed in current block.
+    Committed,
+}
+
+impl TransactionVerificationPhase {
+    /// Convert tip block number to earliest committed block number.
+    pub fn earliest_committed_block(self, tip_block_number: BlockNumber) -> BlockNumber {
+        match self {
+            Self::Submitted => tip_block_number + 3,
```
doitian, line 50: The tx proposal window is configurable in the consensus, see comment below.

doitian (changes requested):

verification/src/transaction_verifier.rs
yangby-cryptape: DONE!

verification/src/transaction_verifier.rs
```diff
@@ -589,29 +659,36 @@ impl Since {
...
     block_number: BlockNumber,
     epoch_number_with_fraction: EpochNumberWithFraction,
     parent_hash: Byte32,
+    tx_phase: TransactionVerificationPhase,
     median_block_count: usize,
     median_timestamps_cache: RefCell<LruCache<Byte32, u64>>,
 }
 
 impl<'a, DL: HeaderProvider> SinceVerifier<'a, DL> {
+    #[allow(clippy::too_many_arguments)]
```
yangby-cryptape, line 672: There are 4 verifiers have `#[allow(clippy::too_many_arguments)]`. All of them have 4 same arguments, three of them have 7 same arguments.
Could do refactor later, in another PR :dog:.

driftluo (approved):

yangby-cryptape:
We can NOT get the parent hash for a transaction which will committed after 2 blocks.

So this PR is not right.

PR #2656, opened 2021-04-20
feat: persistent tx-pool data into a file when it has been shutdown

quake:
base on https://github.com/nervosnetwork/ckb/pull/2621, using the `TransactionVec` struct to persistent tx-pool data directly without new schema defination.

---
This change is [Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/2656)


nervos-bot-user:

doitian (commented):
> Does user know where the file is and how to delete it? Consider add an option in ckb reset-data to help user delete the pool persistent file and ask user to use the command like ckb reset-data --pool to delete the corrupted file.

I modified the error message, it will print the file path:
```
TxPool Error: The tx-pool persisted data file ["/home/xxx/ckb/data/tx_pool_persisted_data.v1"] is broken, please delete it and restart, cause: TransactionVecReader total size doesn't match, expect 1717859076, actual 8
```

doitian:
Please also rebase

test/src/specs/tx_pool/pool_persisted.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        let tx_pool_info_reloaded = normalize_tx_pool_info(node0.get_tip_tx_pool_info());
+        info!("TxPool should be same as before");
+        assert_eq!(tx_pool_info_original, tx_pool_info_reloaded);
+
+        info!("Check the specific values of TxPool state");
+        node0.assert_tx_pool_size(txs_hash2.len() as u64, txs_hash1.len() as u64);
+        assert!(tx_pool_info_reloaded.total_tx_size.value() > 0);
+        assert!(tx_pool_info_reloaded.total_tx_cycles.value() > 0);
+    }
```
keroro520, line 60: I recommend adding more strict assertions:
1. Check the status of the transactions 
2. have you ever tested failed transactions(duplicated, committed, double-spending transactions)?

tx-pool/src/service.rs
```diff
@@ -400,11 +402,48 @@ impl TxPoolController {
...
+            for tx in txs.iter() {
+                if self.submit_local_tx(tx.clone())?.is_err() {
+                    failed_txs.push(tx.clone());
+                }
+            }
+            // return when all success or all failed
+            if failed_txs.is_empty() || txs.len() == failed_txs.len() {
+                info!(
+                    "Persisted tx-pool data is loaded and {} txs are failed to load",
+                    failed_txs.len()
```
keroro520, line 433: * check the transaction's status, if the transaction is already known, then it is considered successful.
* Please do not print this log if all is well, as "failed" indicates there is an error

keroro520 (changes requested):

keroro520:
When `systemctl stop ckb-node-8114`, it sends SIGTERM to stop the service, will SIGTERM trigger dumping the tx-pool?

tx-pool/src/pool.rs
```diff
@@ -456,4 +456,13 @@ impl TxPool {
 
         TxPoolEntryInfo { pending, proposed }
     }
+
+    /// Returns all pending or proposed transactions.
+    pub(crate) fn get_all_transactions(&self) -> impl Iterator<Item = &TransactionView> {
```
zhangsoledad, line 461: use `drain_all_transactions` replace.

quake:
> When systemctl stop ckb-node-8114, it sends SIGTERM to stop the service, will SIGTERM trigger dumping the tx-pool?
Yes

quake:
> What does `generate_transaction()` does? Why this transaction is special that we cannot create it in the loop?

the doc of `generate_transaction()`:
```
// generate a transaction which spend tip block's cellbase and send it to pool through rpc.
```

tx-pool/src/service.rs
quake: > check the transaction's status, if the transaction is already known, then it is considered successful.
I didn't understand the comment, could you be more specific?

> Please do not print this log if all is well, as "failed" indicates there is an error
resolved, please review again.

tx-pool/src/pool.rs
quake: resolved, please review again.

test/src/specs/tx_pool/pool_persisted.rs
quake: This PR simply calls the `tx_pool.submit_local_tx` method to re-add the tx one by one, the related check (duplicated, double-spending) should be guarded by `submit_local_tx` fn, I don't think we need to write these cases in this spec.

quake:
@doitian @keroro520 @zhangsoledad I have rebased this PR with develop branch and resolved issues accordingly, please review again, thanks.

keroro520 (approved):

keroro520 (approved):

keroro520 (dismissed):

doitian (changes requested):
Minor and Nit (Not important thing) mean the suggestions do not block the PR.

> Reviewed 7 of 28 files at r3, 22 of 22 files at r4.
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/2656)* status: all files reviewed, 10 unresolved discussions (waiting on @keroro520, @quake, and @zhangsoledad)

___
*[test/src/node.rs, line 30 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftdGYXEV82c3Oxdn4S:-MftdGYXEV82c3Oxdn4T:b-zx19k) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/test/src/node.rs#L30)):*
> ```XML
> 
> struct ProcessGuard {
>     pub child: Child,
> ```

Minor: Use `Option<Child>` 
___
*[tx-pool/src/pool.rs, line 184 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftbUbl3iNdHKn_Q0hN:-MftbUbl3iNdHKn_Q0hO:bgggi6b) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/pool.rs#L184)):*
> ```XML
>     }
> 
>     /// Add tx which proposed but still uncommittable to gap pool
> ```

Nit, add comment about the return value.
___
*[tx-pool/src/service.rs, line 407 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-Mfte_cM4apEPIJ_XUFV:-Mfte_cN0rjE8G4LODnc:b-7m5nxm) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L407)):*
> ```XML
>     }
> 
>     /// send suspend chunk process cmd
> ```

Minor

```diff
-   /// send suspend chunk process cmd
+   /// Sends suspend chunk process cmd 
```
___
*[tx-pool/src/service.rs, line 423 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftfCUv23yaqTqI4tO_:-MftfCUv23yaqTqI4tOa:b-pfzqd8) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L423)):*
Minor

> ```
>     /// Saves tx pool into disk.
> ```
___
*[tx-pool/src/service.rs, line 440 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftfseVArh3rIEoqrdU:-MftfseWEJOiYygqZLmo:bszrmyv) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L440)):*
> ```XML
> 
>     fn load_persisted_data(&self, mut txs: Vec<TransactionView>) -> Result<(), AnyError> {
>         if !txs.is_empty() {
> ```

Minor, early return reduces indentation.
___
*[tx-pool/src/service.rs, line 442 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-Mftg5VZ4MMk2QgJA91C:-Mftg5VZ4MMk2QgJA91D:bmpsba6) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L442)):*
> ```XML
>         if !txs.is_empty() {
>             info!("Loading persisted tx-pool data, total {} txs", txs.len());
>             // a trick to commit transactions with the correct order
> ```

Minor, explain what's the trick:

In each loop, try to add all the remaining transactions. Remove the successfully added transactions.

Stop the loop until there's no transactions to add in an iteration.
___
*[tx-pool/src/service.rs, line 444 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftiMBCCsA5eXTASkf5:-MftiMBDDbQGmgTDGKvB:b94v6gq) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L444)):*
> ```XML
>             // a trick to commit transactions with the correct order
>             loop {
>                 let mut failed_txs = Vec::new();
> ```

We can use `Vec::drain` to leave only the failed txs. Just remember the vec size before drain. If the size does not change, early return the loop. The loop also can use `while !txs.empty()` instead.
___
*[tx-pool/src/service.rs, line 461 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-Mftiq0a2aADeh5bp8yV:-Mftiq0a2aADeh5bp8yW:bew1kb9) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L461)):*
> ```XML
>                     break;
>                 }
>                 txs = failed_txs;
> ```

FYI, when I first review the code, it confused me because the assignment is in the last of loop. I don't know it when I reviewed the two `break` checks. Renaming the variable `txs` to `remaining_txs` can help, but using `while !txs.empty()` is better to give the hint that we are slowly draining the vector.

doitian (commented):
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/2656)* status: all files reviewed, 10 unresolved discussions (waiting on @keroro520, @quake, and @zhangsoledad)

___
*[tx-pool/src/service.rs, line 423 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftfCUv23yaqTqI4tO_:-MftkrCT-8YWwQjOQaD0:bwuus8v) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L423)):*
<details><summary><i>Previously, doitian (ian) wrote…</i></summary><blockquote>

Minor

> ```
>     /// Saves tx pool into disk.
> ```

</blockquote></details>

Keep consistent: saves -> Saves
___
*[tx-pool/src/service.rs, line 444 at r4](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MftiMBCCsA5eXTASkf5:-Mftkws72QzvZshWf9b-:bsqn3q8) ([raw file](https://github.com/nervosnetwork/ckb/blob/eb259becab1f74b9c63fbe208ab4c27d548da76f/tx-pool/src/service.rs#L444)):*
<details><summary><i>Previously, doitian (ian) wrote…</i></summary><blockquote>

We can use `Vec::drain` to leave only the failed txs. Just remember the vec size before drain. If the size does not change, early return the loop. The loop also can use `while !txs.empty()` instead.
</blockquote></details>

Sorry, it should be`Vec::retain`

tx-pool/src/service.rs
```diff
@@ -516,8 +565,10 @@ impl TxPoolServiceBuilder {
             Arc::clone(&last_txs_updated_at),
         );
 
+        let txs = tx_pool.load_from_file()?;
```
zhangsoledad, line 568: I think the start should not block by load file failed. if this failed under some circumstances，just clean the file，do not fail to start.

quake:
@doitian @zhangsoledad updated accordingly to your comments, please review again, thanks.

doitian:
Will the dump be saved as JSON or Molecule binary?

quake:
> Will the dump be saved as JSON or Molecule binary?

molecule binary

doitian (approved):
> Reviewed 13 of 13 files at r5.
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/2656)* status: all files reviewed, 8 unresolved discussions (waiting on @keroro520, @quake, and @zhangsoledad)

___
*[tx-pool/src/service.rs, line 443 at r5](https://reviewable.io/reviews/nervosnetwork/ckb/2656#-MgQ7_Af5BL-Lvhjwol_:-MgQ7_AgEHSLSMQg-E4_:bo020u) ([raw file](https://github.com/nervosnetwork/ckb/blob/b98ababf46544db785e7707d54e7585aced7c28c/tx-pool/src/service.rs#L443)):*
> ```XML
>         if !txs.is_empty() {
>             info!("Loading persisted tx-pool data, total {} txs", txs.len());
>             let mut failed_txs = 0;
> ```

Minor suggestion: `Iter::all` can simplify the code here.

doitian:
@zhangsoledad @keroro520 please review

doitian (changes requested):
Hold to merge it locally.

Still waiting for another reviewer.

zhangsoledad (approved):

doitian:
Merged in cc665b3f31c1e7261833a8237cb8fae141f759d3

PR #2677, opened 2021-05-10
chore: export reuse to config

driftluo:


nervos-bot-user:

quake (dismissed):

quake (dismissed):

resource/ckb.toml
```diff
@@ -76,6 +76,8 @@ bootnodes = [] # {{
 # whitelist_only = false
 ### Whitelist peers connecting from the given IP addresses
 # whitelist_peers = []
+### enable reuse port, default enable on Linux, not supported on other OS yet
+# reuse = false
```
doitian, line 80:
```suggestion
### Try to enable port reuse when the hint is true. Currently, port reuse will only be enabled on Linux.
# reuse_hint = false
```

util/app-config/src/configs/network.rs
```diff
@@ -269,5 +269,10 @@ impl Config {
 /// By default, using reuse port can make any outbound connection of the node become a potential
 /// listen address, which will help the robustness of our network
 const fn default_reuse() -> bool {
```
doitian, line 271: Let default `reuse_hint` be false.

ckb-bin/src/subcommand/run.rs
```diff
@@ -43,6 +43,8 @@ pub fn run(args: RunArgs, version: Version, async_handle: Handle) -> Result<(),
 
     launcher.check_assume_valid_target(&shared);
 
+    check_reuse_port(launcher.args.config.network.reuse)?;
```
doitian, line 46: Add two methods in `config.network`:

* `is_reuse_supported`: true for linux, false otherwise
* `reuse_enabled()`: return `reuse_hint && is_reuse_supported()`

doitian (changes requested):

resource/ckb.toml
doitian: If it's system specific, let user configure it separately.

```suggestion
### Enable to reuse port on Linux
# reuse_port_on_linux = false
```

doitian (changes requested):

resource/ckb.toml
keroro520: The term "reuse" is ambiguous. I suggest using the full name "reuseport", "enabled_reuseport" or other clear names.

Also, the comment should include information about `SO_REUSEPORT`.

keroro520 (changes requested):

yangby-cryptape (approved):

doitian (dismissed):

doitian:
waiting on @keroro520

yangby-cryptape:
```
  --> benches/benches/benchmarks/overall.rs:62:9
   |
62 |         reuse: true,
   |         ^^^^^ `NetworkConfig` does not have this field
   |
   = note: available fields are: `whitelist_only`, `max_peers`, `max_outbound_peers`, `path`, `dns_seeds` ... and 15 others
```

doitian (approved):

yangby-cryptape (approved):

doitian:

PR #2678, opened 2021-05-11
fix: load cell data hash by input

zhangsoledad:
Ensure load_cell_data_hash is consistent with the load_cell_data behavior


nervos-bot-user:

PR #2679, opened 2021-05-13
feat: add github action "terraform"

keroro520:
Action desctiption: [action-terraform](https://github.com/keroro520/ckb/blob/action-terraform/.github/actions/terraform/README.md)

CI tests "benchmark" and "sync-mainnet" apply AWS instances on-demand. I created a docker action using the Github Action `post` keyword to ensure that the applied instances must be destroyed.

PR #2680, opened 2021-05-13
feat: add workflow "benchmark"

keroro520:
This workflow is triggered by a review comment contains "#benchmark".

This PR is based on https://github.com/nervosnetwork/ckb/pull/2679

---

TODO:

* [ ] Add secrets @doitian @keroro520 
* [ ] Add Github Actions self-hosted runner @doitian @keroro520


nervos-bot-user:

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,135 @@
...
+  pull_request_review:
+    types: [ submitted ]
+
+jobs:
+  build_modified_ckb:
+    name: Build Modified CKB
+    runs-on: [ self-hosted, Linux, X64 ]
+    if: |
+      contains(github.event.review.body, '#benchmark') &&
+      contains(fromJson('[ "janx", "doitian", "quake", "xxuejie", "zhangsoledad", "jjyr", "TheWaWaR", "driftluo", "keroro520", "yangby-cryptape", "liya2017" ]'), github.actor)
```
yangby-cryptape, line 13: ### Is there a way to use a group name?

All actions have a same condition, and the condition has a lot of names, and we have to update those name in different action scripts every time when folks in or folks out.

#### An ugly alternative if no better way to check users permissions

Put the condition in a bash script, then
- Read group from a file, so we could only update one file.
- Or fetch group via GitHub APIs with a token, so we don't have to update any files.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,135 @@
...
+
+    env:
+      CARGO_TARGET_DIR: "${{ github.workspace }}/../target"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Modify Consensus Paarameters And Build CKB
```
yangby-cryptape, line 19:
```suggestion
      - name: Modify Consensus Parameters And Build CKB
```

yangby-cryptape (changes requested):

.github/workflows/benchmark.yml
keroro520: > Is there a way to use a group name?

There is no "group name", just Github user name.

> Read group from a file, so we could only update one file.

There are only [very limited functions we can call in `if`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions). And cannot call shell scripts or read files

keroro520: The information that we can get: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

The functions that we can use: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions

it's very limited

yangby-cryptape: If someone can do this, I think we could do this too.
An example of checking user permissions: [Has Permission Action](https://github.com/marketplace/actions/has-permission).

I didn't mean put everything in `if` statement, you can put that in a step.

keroro520: Yeah, this can be done in `steps`.

But when the process moves forward into `steps`, the job is already triggered:

  - it has entered the self-hosted runner
  - the job has started
  - if you want to exit the current job, you have to select an exit status, succeeded, failed, or canceled

As a result, I think filtering should occur at the `if` phase. It's ugly, but it's safe.

keroro520: PS: once a job is started, it will act as a Github check run for the target PR.

yangby-cryptape: > Yeah, this can be done in `steps`.
> 
> But when the process moves forward into `steps`, the job is already triggered:

A bit lack of imagination. :dog:

A pseudo workflow (without tests):
- Run a GitHub-hosted runner to check user permissions.
  - If passed, call the action in a self-hosted runner. 
  - if failed, do nothing, just returns done.

Ref: [workflow_dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,135 @@
+name: Benchmark
+
+on:
+  pull_request_review:
```
quake, line 4: should be `pull_request_review_comment`?
```suggestion
  pull_request_review_comment:
```

.github/workflows/benchmark.yml
keroro520: Oh, it seems that `workflow_dispatch` can do some combination. That's really cool! I didn't think of it.
Thanks. I marked it and will do some experiments when I have time.

.github/workflows/benchmark.yml
keroro520: During the review discussion, many comments will be made.
For example, several `pull_request_review_comment ` events will be triggered during https://github.com/nervosnetwork/ckb/pull/2680#discussion_r632384482. It would be too noisy.

We can comment on `pull_request_review`. `pull_request_review` is enough.

quake: we have checked comment body here: `contains(github.event.review.body, '#benchmark')` I think it won't be triggered by other comments, right?

quake: by using `pull_request_review`, I think it will create a skipped action always:
WX20210514-181650@2x

keroro520: It will be triggered but skipped.

> by using pull_request_review, I think it will create a skipped action always

Yeah, both `pull_request_review` and `pull_request_review_comment` trigger it.
Despite my desire to reduce the trigger events, pull_request_review versus pull_request_review_comment, the former wins.

If we don't want "Benchmark" to appear at PR status, we should choose `issue_comment` instead of `pull_request_*`

keroro520: pull_request_review: skip once

pull_request_review_comment: skip many times

liya2017, line 4: Do we need a daily cron job for benchmark?

quake:
#benchmark

quake (dismissed):

.github/workflows/benchmark.yml
keroro520: Sure, it would be better to set up a daily benchmark.

While there are also other tasks that must be scheduled, e.g. sync-mainnet, integration-test. Therefore, I believe we should discuss these daily tasks together before writing them down. That's out of this PR.

quake (approved):

.github/workflows/benchmark.yml
doitian: So, how to trigger `pull_request_review`? Could you add a brief doc here?

.github/workflows/benchmark.yml
doitian: I think `if` accepts variables, so we can define the name list in a variable.

keroro520: declaring variables is not supported

.github/workflows/benchmark.yml
keroro520: Sure. I add one line comment in last commit 7d14ef2662877c11b20913012fa7d68e90e6bb6d: 

```
# Submit your review with a comment body containing "#benchmark"
```

keroro520:
It will be split into several PRs, which will make it easier for reviewers.

PR #2734, opened 2021-05-28
chore: fix dependabot false alert

quake:
ref: https://github.com/dependabot/dependabot-core/issues/2041#issuecomment-747799414


yangby-cryptape (approved):

doitian:
😱

doitian (approved):

PR #2735, opened 2021-05-28
chore: remove unnecessary field from `PeerState`

quake:
`PeerState#disconnect` is only used in `Synchronizer#eviction`, it's unnecessary to use this field since we have added it to the eviction vec already.


driftluo (approved):

zhangsoledad (approved):

quake:
bors r=zhangsoledad,driftluo

PR #2736, opened 2021-05-29
refactor: move state flag to `HeadersSyncState` enum

quake:
We are using 3 fields `sync_started` / `sync_connected` / `not_sync_until` in the headers sync process, this PR refactored them to a state machine enum `HeadersSyncState`

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -76,7 +76,48 @@ pub struct ChainSyncState {
     pub work_header: Option<core::HeaderView>,
     pub total_difficulty: Option<U256>,
     pub sent_getheaders: bool,
-    pub not_sync_until: Option<u64>,
+    headers_sync_state: HeadersSyncState,
+}
+
+impl ChainSyncState {
+    fn can_start_sync(&self, now: u64) -> bool {
+        match self.headers_sync_state {
+            HeadersSyncState::Initialized => false,
+            HeadersSyncState::SyncProtocolConnected => true,
+            HeadersSyncState::Started => false,
+            HeadersSyncState::Suspend(until) => until < now,
+        }
+    }
+
+    fn connected(&mut self) {
+        self.headers_sync_state = HeadersSyncState::SyncProtocolConnected;
+    }
+
+    fn start(&mut self) {
+        self.headers_sync_state = HeadersSyncState::Started
+    }
+
+    fn suspend(&mut self, until: u64) {
+        self.headers_sync_state = HeadersSyncState::Suspend(until)
+    }
+
+    fn started(&self) -> bool {
+        matches!(self.headers_sync_state, HeadersSyncState::Started)
+    }
+}
+
+#[derive(Clone, Debug)]
+enum HeadersSyncState {
+    Initialized,
+    SyncProtocolConnected,
+    Started,
+    Suspend(u64), // suspend headers sync until this timestamp (milliseconds since unix epoch)
+}
+
+impl Default for HeadersSyncState {
+    fn default() -> Self {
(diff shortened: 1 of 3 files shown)


sync/src/types/mod.rs
```diff
@@ -69,13 +69,60 @@ const SHRINK_THRESHOLD: usize = 300;
 //     If their best known block is still behind when that new timeout is
 //     reached, disconnect.
 
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
```
doitian, line 72: You can derive `Default` for `HeadersSyncState` so `ChainSyncState` can continue to use auto derive.

sync/src/types/mod.rs
```diff
@@ -69,13 +69,60 @@ const SHRINK_THRESHOLD: usize = 300;
...
+            work_header: None,
+            total_difficulty: None,
+            sent_getheaders: false,
+            headers_sync_state: HeadersSyncState::Initialized,
+        }
+    }
+}
+
+impl ChainSyncState {
+    fn can_sync(&self, now: u64) -> bool {
```
doitian, line 94: The name is confusing. The name `can_start_sync` is a better choice. However, since `HeadersSyncState` already has the function `can_sync`, I'm OK that we continue to use this name in this PR.

sync/src/types/mod.rs
```diff
@@ -69,13 +69,60 @@ const SHRINK_THRESHOLD: usize = 300;
...
+
+    fn connected(&mut self) {
+        self.headers_sync_state = HeadersSyncState::SyncProtocolConnected;
+    }
+
+    fn start(&mut self) {
+        self.headers_sync_state = HeadersSyncState::Started
+    }
+
+    fn suspend(&mut self, timestamp: u64) {
```
doitian, line 111:
```suggestion
    fn suspend(&mut self, suspend_until: u64) {
```

sync/src/types/mod.rs
```diff
@@ -69,13 +69,60 @@ const SHRINK_THRESHOLD: usize = 300;
...
+    }
+}
+
+impl ChainSyncState {
+    fn can_sync(&self, now: u64) -> bool {
+        match self.headers_sync_state {
+            HeadersSyncState::Initialized => false,
+            HeadersSyncState::SyncProtocolConnected => true,
+            HeadersSyncState::Started => false,
+            HeadersSyncState::Suspend(timestamp) => timestamp < now,
```
doitian, line 99:
```suggestion
            HeadersSyncState::Suspend(suspend_until) => suspend_until < now,
```

sync/src/types/mod.rs
```diff
@@ -69,13 +69,60 @@ const SHRINK_THRESHOLD: usize = 300;
...
+        matches!(self.headers_sync_state, HeadersSyncState::Started)
+    }
+}
+
+#[derive(Clone, Debug)]
+enum HeadersSyncState {
+    Initialized,
+    SyncProtocolConnected,
+    Started,
+    Suspend(u64),
```
doitian, line 125: Better to add some comment about the `u64` field. What does it mean and what unit it uses.

sync/src/types/mod.rs
doitian: Better to add comment about what unit `suspend_until` uses.

sync/src/types/mod.rs
quake: @doitian updated accordingly, please review again, thanks.

driftluo (approved):

doitian (approved):

doitian:
bors r=doitian,driftluo

PR #2765, opened 2021-06-17
feat(hardfork): Net fork

driftluo:
The modification on the network is relatively simple, but it is closely related to the timing of the VM switch

1. All protocols support both versions 1, 2 at the same time, all protocols except `relay` are simple upgrades 
2. There will be two versions of the relay protocol for runtime switching, and they are mutually exclusive, in the work of relay tx, they will only have one working at the same time 
3. During the switching process, there is the possibility of losing transactions, which is almost an unavoidable problem 
4. Transaction execution in different versions of VM may have inconsistent cycle consumption. In order to avoid this behavior splitting the network, we must do some complicated things on the relay protocol 
5. Fortunately, these seemingly unfriendly network processing codes will be deleted in the next version after Fork is completed. Old clients that do not support the Fork feature will no longer be able to connect to the network, and the first one that supports the Fork feature The client will be compatible with the two networks for a period of time after the fork is completed, it will refuse the connection of the old client 

test：
1. Can communicate normally with old clients before the fork 
2. Reject old client connections after the fork
3. Expulse old client connections during the fork
3. The client after fork allows nodes that support the fork feature to sync from the beginning 
4. During the switching process, there will be no bans due to inconsistent relay tx


quake:
Please change all variables and comments in this PR that contain the word `fork` to `edition_2021`, it is more accurate, and less likely to be confused with normal block fork switching for reading the code.

keroro520 (changes requested):

tx-pool/src/process.rs
```diff
@@ -663,6 +684,20 @@ impl TxPoolService {
...
         }
+
+        // update network fork switch each block
+        {
+            if !self.network.load_ckb2021() {
+                let epoch = self.snapshot().tip_header().epoch().number();
+                if self
+                    .consensus
+                    .hardfork_switch
+                    .is_p2p_network_switch_enabled(epoch)
```
keroro520, line 695: Would it be possible that `is_p2p_network_switch_enabled(epoch) returns false` but `is_vm_version_1_and_syscalls_2_enabled(epoch) returns true`?

From [this document](https://hackmd.io/6GCpCIFnSmaYXdgG1lM-zA#%E5%AF%B9%E8%A1%8C%E4%B8%BA%E5%9C%A8-fork-%E5%89%8D%E5%90%8E%E4%BC%9A%E5%8F%91%E7%94%9F%E5%86%B2%E7%AA%81%E7%9A%84%E5%8D%8F%E8%AE%AE), ` is_p2p_network_switch_enabled` and `is_vm_version_1_and_syscalls_2_enabled` should return same value.

util/constant/src/sync.rs
```diff
@@ -10,6 +10,8 @@ pub const MAX_HEADERS_LEN: usize = 2_000;
 /// The default number of download blocks that can be requested at one time
 /* About Download Scheduler */
 pub const INIT_BLOCKS_IN_TRANSIT_PER_PEER: usize = 16;
+/// ckb2021 fork new limit
+pub const NEW_INIT_BLOCKS_IN_TRANSIT_PER_PEER: usize = 32;
```
keroro520, line 14: https://github.com/nervosnetwork/ckb/pull/2765#issuecomment-864634204

sync/src/relayer/mod.rs
```diff
@@ -638,6 +670,9 @@ impl Relayer {
 
 impl CKBProtocolHandler for Relayer {
     fn init(&mut self, nc: Arc<dyn CKBProtocolContext + Sync>) {
+        if let NetFork::ForkInV1 = is_ckb2021(&nc, self.v2) {
```
keroro520, line 673: Would it be possible?
Why return directly?

sync/src/relayer/mod.rs
```diff
@@ -803,3 +864,26 @@ impl CKBProtocolHandler for Relayer {
...
+            NetFork::ForkInV2
+        } else {
+            NetFork::NoForkInV2
+        }
+    } else if ckb2021 {
+        NetFork::ForkInV1
+    } else {
+        NetFork::NoForkInV1
+    }
+}
```
keroro520, line 889:
```suggestion
#[derive(Copy, Clone, Debug)]
enum RelaySwitch {
    Ckb2019RelayV1,
    Ckb2019RelayV2,
    Ckb2021RelayV1,
    Ckb2021RelayV2,
}

impl RelaySwitch {
    fn new(ckb2021: bool, is_relay_v2: bool) -> Self {
        match (ckb2021, is_relay_v2) {
            (true, true) => Self::Ckb2021RelayV2,
            (true, false) => Self::Ckb2021RelayV1,
            (false, true) => Self::Ckb2019RelayV2,
            (false, false) => Self::Ckb2019RelayV1,
        }
    }
    
    fn is_switching(&self) -> bool {
        match self {
            Self::Ckb2019RelayV2 | Self::Ckb2021RelayV1 => true,
            Self::Ckb2019RelayV1 | Self::Ckb2021RelayV2 => false,
        }
    }
}
```

keroro520 (changes requested):

tx-pool/src/process.rs
driftluo: I think we can ignore this situation, there are two possibilities here:

1. Set epoch error（impossible）
2. Inconsistency caused by the timing of snapshot acquisition （maybe）

Either way, its biggest impact is the loss of some transactions, and the impact on the entire network is negligible

sync/src/relayer/mod.rs
driftluo: This is possible. When the client stops and restarts after completing the fork, the current state is the state after the fork. At this time, the v1 version of the relayer should be silent

keroro520: Gotcha.
Although I think it's better to put this check before initializing the protocol, [in launcher](https://github.com/nervosnetwork/ckb/pull/2765/files#diff-3d835c3d61b1399a2a6920bb7fa9d79484d35293d9c41eeb75ad388c49b58d3eR337-R341): if is_p2p_network_switch_enabled, do not init relay v1.

keroro520 (changes requested):

sync/src/relayer/mod.rs
```diff
@@ -779,6 +814,32 @@ impl CKBProtocolHandler for Relayer {
             return;
         }
 
+        match is_ckb2021(&nc, self.v2) {
+            NetFork::NoForkInV2 => return,
+            NetFork::ForkInV1 => {
+                if nc.remove_notify(TX_PROPOSAL_TOKEN).is_err() {
```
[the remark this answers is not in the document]
driftluo: maybe never，channel size is 4096 yet

sync/src/types/mod.rs
```diff
@@ -209,6 +209,18 @@ impl HeadersSyncController {
     }
 }
 
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum Version {
+    Old,
+    New,
+}
```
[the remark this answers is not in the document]
driftluo: These fields are really needed. I explain the necessity of existence one by one, and they cannot be integrated with each other.

- Version: It is used to mark whether the client is a new client between the sync and relay protocols. According to this mark, the connected old client is closed when the fork occurs, and it is also used to control the length of the block sync limit(delete after the fork)
- relay‘s v2: Mark the version of the relay protocol(delete after the fork)
- netfork: Mark the dynamic state of the relay protocol, fork, and its own version(delete after the fork)
- legacy protocol version string：Used to externally respond to all protocol versions supported by this node(Only keep version 2 protocol support )

sync/src/synchronizer/get_blocks_process.rs
```diff
@@ -41,7 +44,19 @@ impl<'a> GetBlocksProcess<'a> {
         }
         let active_chain = self.synchronizer.shared.active_chain();
 
-        for block_hash in block_hashes.iter().take(INIT_BLOCKS_IN_TRANSIT_PER_PEER) {
+        let iter = match active_chain
+            .shared()
+            .state()
+            .peers()
+            .get_version(self.peer)
+            .unwrap_or(Version::Old)
```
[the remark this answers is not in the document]
driftluo: I think there is no problem here, the new version supports 32 requests, the old version only supports 16.
This version number will be set after the sync protocol is connected, In theory, there must be a value here, if not, the most conservative 16 will be used. 

Maybe 32 would be better?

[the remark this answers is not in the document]
driftluo: change to default 32

tx-pool/src/process.rs
yangby-cryptape: I think we could use the same `switch`, since they have to equal to each other.

keroro520: > I think we could use the same `switch`, since they have to equal to each other.

Yes, I agree

PR #2768, opened 2021-06-21
resume/suspend verification

zhangsoledad:


script/src/verify.rs
```diff
@@ -455,6 +467,267 @@ impl<'a, DL: CellDataProvider + HeaderProvider> TransactionScriptsVerifier<'a, D
...
+    fn make_snapshot(
+        &self,
+        mut vm: Machine<'a>,
+        group: &ScriptGroup,
+        current: (ScriptGroupType, Byte32),
+        remain: Vec<(ScriptGroupType, Byte32)>,
+        current_cycles: Cycle,
+        limit_cycles: Cycle,
+    ) -> Result<TransactionSnapshot, Error> {
+        let snap = make_snapshot(&mut vm.machine).map_err::<Error, _>(|e| {
```
driftluo, line 479: Communicate with the vm team, if there is a problem here, it should crash directly, it must be a bug

script/src/verify.rs
```diff
@@ -455,6 +467,267 @@ impl<'a, DL: CellDataProvider + HeaderProvider> TransactionScriptsVerifier<'a, D
...
+                ScriptError::VMInternalError(format!(
+                    "expect invalid cycles {} {}",
+                    limit_cycles, cycles
+                ))
+                .source(group)
+            })?;
+
+            match self.verify_group_with_chunk(&group, step_cycle, None) {
+                Ok(ChunkState::Completed(cycle)) => {
+                    cycles = wrapping_cycles_add(cycles, cycle, limit_cycles, group)?;
```
driftluo, line 521: I personally think that the cycles add/reduce check here is unnecessary, because vm uses checked add when calculating cycles, and it must not be greater than limit

script/src/verify.rs
```diff
@@ -455,6 +467,267 @@ impl<'a, DL: CellDataProvider + HeaderProvider> TransactionScriptsVerifier<'a, D
...
+        let current_group = self
+            .find_script_group(snap.current.0, &snap.current.1)
+            .ok_or_else(|| {
+                ScriptError::VMInternalError(format!("snapshot group missing {:?}", snap.current))
+                    .unknown_source()
+            })?;
+
+        if limit_cycles < snap.current_cycles {
+            return Err(ScriptError::ExceededMaximumCycles(limit_cycles)
+                .source(current_group)
```
driftluo, line 577: Here I think we need to use unknow group, it is not necessarily the problem of the current group

tx-pool/src/process.rs
```diff
@@ -682,7 +729,7 @@ impl TxPoolService {
                     let tip_header = snapshot.tip_header();
                     let tx_env = status.with_env(tip_header);
                     if let Ok(verified) =
-                        verify_rtx(snapshot, &rtx, &tx_env, verify_cache, max_cycles)
+                        verify_rtx(snapshot, &rtx, &tx_env, &verify_cache, max_cycles)
```
driftluo, line 732: The big transactions that are thrown out due to conflicts may be completely lost here. I think this kind of transaction should be thrown into the big transaction queue.

tx-pool/src/chunk_process.rs
```diff
@@ -0,0 +1,220 @@
...
+                    _ => (),
+                }
+            }
+
+            let step_cycle = cmp::min(MIN_STEP_CYCLE, max_cycles - current_cycles);
+
+            let ret = if let Some(ref snap) = tx_snap {
+                script_verifier.resume(snap, step_cycle)
+            } else {
+                script_verifier.resumable_verify(step_cycle)
```
driftluo, line 188: Every time you build a snapshot is almost equivalent to copy a 4mb memory. If you don’t need to verify the block, I don’t think you should use the resumable_verify function here. You only need to take a snapshot when you are sure to receive the command to verify the block.

tx-pool/src/process.rs
driftluo: all tx verify will hit the cache, so don't care lost tx here

tx-pool/src/chunk_process.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    async fn remove_front(&self) {
+        let mut guard = self.service.chunk.write().await;
+        guard.clean_front();
+    }
+
+    fn process(&mut self, entry: Entry) -> Stop {
+        self.process_inner(entry.clone()).unwrap_or_else(|e| {
+            self.handle
+                .block_on(self.service.after_process(entry.tx, entry.remote, &Err(e)));
```
driftluo, line 103: if process return error should call `remove_front` here

zhangsoledad: https://github.com/nervosnetwork/ckb/compare/a3655fd9cec4ba037c3aa220c76ccd9b9123128d..73dc50811fc669bca01d1a63ff78d570975add3a

yangby-cryptape:
Since the CI was blocked too many times, so I ran the integration tests in my laptop to find the reason.

I tested commit `73dc508`.
It will be blocked at the integration test "InsufficientReward" every time, and all other integration tests will be passed.

And I canceled the CI for commit `73dc508`.

driftluo (approved):

ckb-bin/src/subcommand/import.rs
```diff
@@ -16,6 +16,9 @@ pub fn import(args: ImportArgs, async_handle: Handle) -> Result<(), ExitCode> {
     let chain_service = ChainService::new(shared, pack.take_proposal_table());
     let chain_controller = chain_service.start::<&str>(Some("ImportChainService"));
 
+    pack.take_tx_pool_builder();
```
doitian, line 19: What these two lines do during importing?

PR #2771, opened 2021-06-21
chore: remove lru cache from SinceVerifier

quake:
I found `median_timestamps_cache` of `SinceVerifier` to be almost useless when counting lru cache hits / miss ratio: it pre-allocates an inputs-sized cache for each transaction, but for most transactions without the `since` field, this cache is useless and is quickly deallocated. For the few transactions that have the `since` field, the header fetch is already cached in store layer, so I don't see the need to add another cache here.

### DIFF
--- a/verification/src/transaction_verifier.rs
+++ b/verification/src/transaction_verifier.rs
@@ -15,8 +15,6 @@ use ckb_types::{
     packed::Byte32,
     prelude::*,
 };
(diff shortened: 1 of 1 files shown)


driftluo (approved):

zhangsoledad (approved):

quake:
bors r=zhangsoledad,driftluo

PR #2774, opened 2021-06-22
chore: remove lru cache from get_cellbase fn

quake:
The cellbase's cache hit ratio was found to be zero in testing the cache efficiency of store layer (running an ibd sync on mainnet), it's only used in `RewardVerifier` and `BlockAssembler`.

For `RewardVerifier`, it will only be called once for each block, it can never hit the cache, while for `BlockAssembler`, since there is a block template cache already, there is no need to overlay this cache, and the `get_cellbase` actually hits rocksdb's own cache quite well, so I don't see the need to introduce additional overhead.

please note that `cellbase_cache_size` is removed from configuration file, however, old version configuration file is compatible with this change since we ignore unknown conf fields.

### DIFF
--- a/util/app-config/src/configs/store.rs
+++ b/util/app-config/src/configs/store.rs
@@ -16,8 +16,6 @@ pub struct Config {
     /// The maximum number of blocks which extension section is cached.
     #[serde(default = "default_block_extensions_cache_size")]
     pub block_extensions_cache_size: usize,
-    /// The maximum number of blocks which cellbase transaction is cached.
-    pub cellbase_cache_size: usize,
     /// whether enable freezer
     #[serde(default = "default_freezer_enable")]
     pub freezer_enable: bool,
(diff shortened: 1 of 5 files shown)


util/app-config/src/configs/store.rs
```diff
@@ -40,7 +38,6 @@ impl Default for Config {
             block_tx_hashes_cache_size: 30,
             block_uncles_cache_size: 30,
             block_extensions_cache_size: default_block_extensions_cache_size(),
-            cellbase_cache_size: 30,
             freezer_enable: false,
```
zhangsoledad, line 41: This config can be ignored, but can't delete directly, it hurt backward compatibility.

zhangsoledad (dismissed):
backward compatibility

zhangsoledad (approved):

driftluo (approved):

quake:
bors r=zhangsoledad,driftluo

quake:
bors r=zhangsoledad,driftluo

PR #2773, opened 2021-06-21
chore: remove several duplicate or outdated dependencies, such as hyper, futures, tokio and so on

yangby-cryptape:
- Upgrade `hyper` for `ckb-miner` to remove a duplicate version of `hyper`, and upgrade all related dependencies.
  Remove `536 - 510 = 26` dependencies in total.
- Also fix [RUSTSEC-2020-0070], some lock_api lock guard objects can cause data races by upgrading `parking_lot`.

[RUSTSEC-2020-0070]: https://rustsec.org/advisories/RUSTSEC-2020-0070

PR #2778, opened 2021-06-22
fix: miner panics since the future created by hyper 0.13 need tokio 0.2 runtime

yangby-cryptape:
The bug is that miner always panics, because the future created by hyper 0.13 needs tokio 0.2 runtime.

This bug is introduced since #2773 yesterday.
That PR removed the duplicate version of "hyper" and only kept one version of "hyper".

In fact, I had tested the miner for each commit in that PR, and everything was OK when I did tests.
To be honest, I don't know why this 100%-panic bug wasn't appeared yesterday.
I guess if it's because I didn't clear the "target" directory when I did tests.

This time, I cleared the "target" directory and re-compiled the executable file.
And I ran a dev-chain with 1 node and 4 miners in different speeds for 1 hours, no panics anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -895,6 +895,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio 1.7.0",
+ "tokio-compat-02",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


doitian (approved):

driftluo (approved):

doitian:
bors r=doitian,driftluo

PR #2779, opened 2021-06-22
feat: give an unique id to each global runtime thread

yangby-cryptape:
- Improve the experience for reading our logs, to distinguish logs between different threads.

- Unify naming style of thread names.

  Since a long thread name will cut to 15 characters.
  No doubt about it, `CamelCase` is better than `snake_case`, since the name in `CamelCase` is more short.

  This limit is created by kernel.

  ```c
  /* Task command name length: */
  #define TASK_COMM_LEN     16
  ```

  Ref: [The definition of `TASK_COMM_LEN` in latest stable kernel.](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/sched.h?h=v5.10.12#n213)

  p.s. The default name of the tokio runtime is "tokio-runtime-worker", we could only see "tokio-runtime-w" in system tools, such as `top`, `htop` and so on.

### DIFF
--- a/util/runtime/src/lib.rs
+++ b/util/runtime/src/lib.rs
@@ -4,7 +4,10 @@ use ckb_logger::debug;
 use ckb_spawn::Spawn;
 use ckb_stop_handler::{SignalSender, StopHandler};
 use core::future::Future;
-use std::thread;
+use std::{
+    sync::atomic::{AtomicU32, Ordering},
+    thread,
+};
 use tokio::runtime::Builder;
 use tokio::runtime::Handle as TokioHandle;
 use tokio::sync::oneshot;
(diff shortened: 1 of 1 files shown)


doitian (dismissed):

util/runtime/src/lib.rs
```diff
@@ -54,15 +57,20 @@ impl Handle {
 pub fn new_global_runtime() -> (Handle, StopHandler<()>) {
     let runtime = Builder::new_multi_thread()
         .enable_all()
-        .thread_name("ckb-global-runtime")
+        .thread_name("GlobalRt")
+        .thread_name_fn(|| {
+            static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+            let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
```
driftluo, line 63: use `fetch_update` instead, our worker may up and down every time because we use `block in place` on lots of time

yangby-cryptape: > `pub fn fetch_add(&self, val: usize, order: Ordering) -> usize`
>
> [This operation wraps around on overflow.](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add)

But I also will update this code, because if id more than 1_000_000, we still missing the unique id, due to the length limit.

driftluo (approved):

util/runtime/src/lib.rs
```diff
@@ -54,15 +57,40 @@ impl Handle {
 pub fn new_global_runtime() -> (Handle, StopHandler<()>) {
     let runtime = Builder::new_multi_thread()
         .enable_all()
-        .thread_name("ckb-global-runtime")
+        .thread_name("GlobalRt")
+        .thread_name_fn(|| {
```
yangby-cryptape, line 61: [The official sample code.](https://docs.rs/tokio/1.7.1/tokio/runtime/struct.Builder.html#method.thread_name_fn)

quake (approved):

yangby-cryptape:
bors r=driftluo,quake

PR #2791, opened 2021-06-25
refactor: verify the epoch in block headers explicitly

yangby-cryptape:
### Issue

The data of epoch in bytes is not same as the `EpochNumberWithFraction`.

https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/util/types/src/conversion/blockchain.rs#L105-L110
https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/util/types/src/core/extras.rs#L452-L459

It cause a few unintended consequences.
For example, the RPC return `"0x"` as the epoch of genesis block, but the integration tests checks if its length is `1`.

And we couldn't convert it back with the same value.
```rust
let original = Uint64::default();
let epoch: EpochNumberWithFraction = original.unpack();
let result: Uint64 = epoch.pack();
assert_eq!(result, original); // failed!
```

### Changes

- When convert a series of bytes to an `EpochNumberWithFraction`, don't change its value.

- Add an `EpochVerifier` to `HeaderVerifier`:
  - Check if the epoch is well formed.
  - Check if the epoch is continuous.
  So we can reject those blocks which have invalid epochs earlier.

- If an `EpochNumberWithFraction`'s full value is `0`, returns `RationalU256::zero()` when calling `to_rational()`.
  Because the full value of a genesis block' epoch is `0`.

### DIFF
--- a/chain/src/tests/block_assembler.rs
+++ b/chain/src/tests/block_assembler.rs
@@ -101,7 +101,7 @@ fn gen_block(parent_header: &HeaderView, nonce: u128, epoch: &EpochExt) -> Block
         .parent_hash(parent_header.hash())
         .timestamp((parent_header.timestamp() + 10).pack())
         .number(number.pack())
-        .epoch(epoch.number().pack())
+        .epoch(epoch.number_with_fraction(number).pack())
         .compact_target(epoch.compact_target().pack())
         .nonce(nonce.pack())
         .dao(dao)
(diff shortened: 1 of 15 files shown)


quake (approved):

driftluo (approved):

yangby-cryptape:
bors r=quake,driftluo

PR #2792, opened 2021-06-25
refactor(hardfork): verify the epoch in since more strictly

yangby-cryptape:
### Before Hardfork

- When the length of since absolute (or relative) epoch is zero:
  - The `epoch_index` will be considered as `0`.
  - The `epoch_length` will be considered as `1`.

  https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/verification/src/transaction_verifier.rs#L544-L547
  https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/util/types/src/core/extras.rs#L452-L459

- Since relative epoch uses rational number operations but since absolute epoch doesn't.

  https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/verification/src/transaction_verifier.rs#L657-L662
  https://github.com/nervosnetwork/ckb/blob/85bf46065cc91e75980c9ee82925af77c524964d/verification/src/transaction_verifier.rs#L610-L612

### After Hardfork

- A transaction with since absolute (or relative) epoch is valid only if `epoch_index` is less than `epoch_length` or both `epoch_index` and `epoch_length` are zero.
- Using rational number operations for both since absolute epoch and since relative epoch.

### DIFF
--- a/test/src/specs/hardfork/v2021/since.rs
+++ b/test/src/specs/hardfork/v2021/since.rs
@@ -34,8 +34,20 @@ impl Spec for CheckAbsoluteEpochSince {
             assert_send_transaction_fail(node, &tx, ERROR_IMMATURE);
         }
         {
-            info!("CKB v2019, since absolute epoch failed");
+            info!("CKB v2019, since absolute epoch ok");
             let tx = create_tx_since_absolute_epoch(node, 1, 2);
+            let res = node.rpc_client().send_transaction_result(tx.data().into());
+            assert!(res.is_ok(), "result: {:?}", res.unwrap_err());
+        }
+        {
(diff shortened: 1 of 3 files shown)


yangby-cryptape:
[The Related RFC Proposal](https://github.com/nervosnetwork/rfcs/pull/223#issue-567310677) said:
> This RFC proposes to add a rule to verify that `i <= l` or `i` and `l` are both zero.

keroro520 (approved):

quake (approved):

yangby-cryptape:
bors r=keroro520,quake

yangby-cryptape:
bors r=keroro520,quake

PR #2796, opened 2021-07-01
feat(hardfork): net hardfork

driftluo:
Modifications on the network are relatively simple but are closely related to the timing of the VM switch

1. All protocols support both versions 1 and 2, and all protocols are simple upgrades except for relay.
2. Relay protocols for runtime switching will have two mutually exclusive versions, and only one of them will work in conjunction whilst the relay Tx is in operation.
3. During switching, a loss of transactions is possible, which is almost an unavoidable problem.
4. Transactions that are executed in different versions of the VM may have inconsistent cycle consumption. To avoid this split network behaviour, some complex stuff needs to be done on the relay protocol
5. These unfriendly-looking network processing codes will, fortunately, be removed in the next release once the Fork feature is completed. Previous clients not supporting the Fork feature will be unable to connect to the network anymore, yet the first Fork-enabled client will be compatible with both networks for some time after the Fork is completed and will reject connections from older clients.

Test:
1. Communicates normally with old clients before the fork
2. Rejects connections from old clients after the fork
3. Evicts connections from old clients during the fork
4. Client after the fork allows nodes that support the fork feature to synchronize from scratch
5. No banning for inconsistent relay tx during the switching process


util/types/src/core/hardfork.rs
```diff
@@ -209,6 +216,14 @@ define_methods!(
     "RFC PR 0237"
 );
 
+define_methods!(
+    rfc_pr_0234,
+    p2p_network_switch,
+    is_p2p_network_switch_enabled,
```
yangby-cryptape, line 222: I think we could use the same switch, since they have to equal to each other.

keroro520: I agree. https://github.com/nervosnetwork/ckb/pull/2765#discussion_r661295725

sync/src/types/mod.rs
```diff
@@ -249,11 +249,24 @@ impl HeadersSyncController {
     }
 }
 
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum Version {
+    Old,
+    New,
+}
```
keroro520, line 256: > A public type called Version is easy to misunderstand, protocol version, implementation version, binary version, or else?

https://github.com/nervosnetwork/ckb/pull/2765#discussion_r660292829

util/types/src/core/hardfork.rs
driftluo: Change 00237 to 234 + 237, so it will share the same value

sync/src/types/mod.rs
driftluo: change to binary version

keroro520 (approved):
LGTM, expect the discovery protocol. I know it will change later.

sync/src/types/mod.rs
```diff
@@ -249,11 +249,24 @@ impl HeadersSyncController {
     }
 }
 
+#[derive(Clone, Copy, Eq, PartialEq, Debug)]
+pub enum BinaryVersion {
```
quake, line 253: I don't think we need a new enum here, just add a new field to struct `PeerFlags`, which can reduce a large number of fn params changes:
```
pub struct PeerFlags {
    pub is_outbound: bool,
    pub is_protect: bool,
    pub is_whitelist: bool,
    pub is_2021edition: bool,
}
```

quake (changes requested):

sync/src/types/mod.rs
driftluo: Modification is complete, please review again

quake (dismissed):

doitian (approved):
Lgtm. Except one suggestion: avoid too many `v2: bool`, use an enum instead.

quake (approved):

driftluo:
bors r=quake,doitian

PR #2797, opened 2021-07-01
feat(hardfork): reject vm1 lock script before hardfork started to keep compatible with old clients

yangby-cryptape:
### Changes

- refactor: merge time relative verifiers in ContextualTransactionVerifier

  Before this commit:
  - `TimeRelativeTransactionVerifier = MaturityVerifier + SinceVerifier`
  - `ContextualTransactionVerifier = MaturityVerifier + SinceVerifier + OtherVerifiers`

  After this commit:
  - `TimeRelativeTransactionVerifier = MaturityVerifier + SinceVerifier`
  - `ContextualTransactionVerifier = TimeRelativeTransactionVerifier + OtherVerifiers`

- feat(hardfork): reject vm1 lock script before hardfork started to keep compatible with old clients

  Add `CompatibleVerifier` to `ContextualTransactionVerifier` to check compatible between different versions CKB clients.

  When a new client with hardfork features released, before the hardfork started, the old CKB clients will still be able to work.
  So, the new CKB client have to add several necessary checks to avoid fork attacks.

  After hardfork, the old clients will be no longer available. Then we can delete all code in that verifier until next hardfork.

  In current version, `CompatibleVerifier` only has one feature:
  - Reject vm1 lock script before hardfork started to keep compatible with old clients.

### DIFF
--- a/verification/src/transaction_verifier.rs
+++ b/verification/src/transaction_verifier.rs
@@ -1,7 +1,7 @@
 use crate::cache::CacheEntry;
 use crate::error::TransactionErrorSource;
 use crate::{TransactionError, TxVerifyEnv};
-use ckb_chain_spec::consensus::{Consensus, ConsensusProvider};
+use ckb_chain_spec::consensus::Consensus;
 use ckb_dao::DaoCalculator;
 use ckb_error::Error;
 use ckb_metrics::{metrics, Timer};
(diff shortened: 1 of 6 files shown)


verification/src/transaction_verifier.rs
```diff
@@ -16,29 +16,38 @@ use ckb_types::{
...
+use std::convert::TryInto;
 
 /// The time-related TX verification
 ///
 /// Contains:
+/// [`CompatibleVerifier`](./struct.CompatibleVerifier.html)
 /// [`MaturityVerifier`](./struct.MaturityVerifier.html)
 /// [`SinceVerifier`](./struct.SinceVerifier.html)
 pub struct TimeRelativeTransactionVerifier<'a, M> {
+    pub(crate) compatible: CompatibleVerifier<'a>,
```
doitian, line 28: It's a bit weird to put compatible verifier in `TimeRelativeTransactionVerifier`, although you can interpret that time causes incompatibilities.

yangby-cryptape: Maybe `CompatibleCountdownVerifier` is a better name?

doitian: I prefer remove it from `TimeRelativeTransactionVerifier`, unless we believe it is deeply coupled with the since and cell base immature verifier.

yangby-cryptape: 👌🏼 , and how about first commit? Can I keep it?

I will just refactor the second commit and force push current PR.

doitian: lgtm

doitian (approved):

quake (approved):

yangby-cryptape:
bors r=doitian,quake

PR #2798, opened 2021-07-01
feat(hardfork): remove the header deps immature rule

yangby-cryptape:
### Changes

- New hardfork Features

  -  [CKB-RFCs PR 240: RFC: Remove header deps immature rule](https://github.com/nervosnetwork/rfcs/pull/240)

- refactor: resolve transaction with bitflags


util/types/src/core/cell.rs
```diff
@@ -625,6 +632,7 @@ pub fn resolve_transaction<CP: CellProvider, HC: HeaderChecker, S: BuildHasher>(
     seen_inputs: &mut HashSet<OutPoint, S>,
     cell_provider: &CP,
     header_checker: &HC,
+    no_immature_header_deps: bool,
```
doitian, line 635: How about adding a bitsflags enum here?

```
TransactionResolvingOptions {
  Default = 0x0,
  ImmatureHeaderDeps = 0x1,
}

default => no flags
```

Then we can define two functions:

* `resolve_transaction`: resolve using the default options (no flags)
* `resolve_transaction_with_options`: add a new `TransactionResolvingOptions` parameter.

To resolve depending on the context, we can add a function which derive the correct TransactionResolvingOptions from the chain context.

yangby-cryptape: The function `resolve_transaction` without options is only used for tests.

quake (approved):

driftluo (approved):

yangby-cryptape:
bors r=quake,driftluo

PR #2827, opened 2021-07-14
fix: fix peer store evict

driftluo:
Originally, only the data in the largest group was considered, but now it is changed to traverse at least half of the groups


quake (approved):

yangby-cryptape (approved):

yangby-cryptape:

PR #2843, opened 2021-07-19
feat: dotenv store deployment variables

zhangsoledad:


tx-pool/Cargo.toml
```diff
@@ -32,6 +32,7 @@ ckb-network = { path = "../network", version = "= 0.100.0-pre" }
 ckb-channel = { path = "../util/channel", version = "= 0.100.0-pre" }
 ckb-traits = { path = "../traits", version = "= 0.100.0-pre" }
 sentry = { package = "ckb-sentry", version = "0.21.0", optional = true }
+dotenv = "0.15.0"
```
doitian, line 35: It's better to add `dotenv` only in `ckb-bin` and inject options only in that crate.

doitian: Please ignore .env in the git, and add a .env.example

tx-pool/src/chunk_process.rs
```diff
@@ -37,15 +37,21 @@ pub(crate) struct TxChunkProcess {
     handle: Handle,
     recv: Receiver<Command>,
     p_state: ProcessState,
+    step_cycles: Cycle,
 }
 
 impl TxChunkProcess {
     pub fn new(service: TxPoolService, handle: Handle, recv: Receiver<Command>) -> Self {
+        let step_cycles = dotenv::var("STEP_CYCLES")
```
doitian, line 45:
```suggestion
        let step_cycles = dotenv::var("CKB_STEP_CYCLES")
```

doitian (changes requested):

tx-pool/Cargo.toml
zhangsoledad: We will not be able to use this in other crates' unit tests If we only load the `dotenv` file in `ckb-bin`.

static START: Once = ONCE_INIT;

pub fn var<K: AsRef<OsStr>>(key: K) -> Result<String> {
    START.call_once(|| {
        dotenv().ok();
    });
    std::env::var(key)
}

tx-pool/src/chunk_process.rs
zhangsoledad: https://github.com/nervosnetwork/ckb/compare/bd201061d6e7b83f95f1b4f0769631b655c681a4..c8f6a50978b74a30e155f1ad98e3b10f4326f4fc

doitian (approved):

PR #2756, opened 2021-06-15
feat(hardfork): ckb2021 hardfork features (vm related part)

yangby-cryptape:
### Changes

- New hardfork Features

  -  [CKB-RFCs PR 236: RFC: ckb vm version1 changes](https://github.com/nervosnetwork/rfcs/pull/236)

  -  [CKB-RFCs PR 237: RFC: vm syscalls 2](https://github.com/nervosnetwork/rfcs/pull/237)

  -  [CKB-RFCs PR 238: CKB VM version selection](https://github.com/nervosnetwork/rfcs/pull/238)

- All changes are followed to the latest RFCs, except:

  - Add a "edition" field to all configuration files: CKB chain specification or app configurations.

#### CKB Chain Edition

- If no `edition` field, the default edition is `"2019"`.

-  After this PR, the default files created by `ckb init` will be edition `"2021"` files.

   But the public chains will still use the edition `"2019"` CKB chain specification as the built-in spec.

   At present, there are only to public chains: "mainnet" and "testnet".

- The edition `"2019"` doesn't support hardfork parameters.

- How to upgrade from edition `"2019"` to `"2021"`?

  Modify your spec file (default is `specs/dev.toml`), `ckb.toml` and `ckb-miner.toml`:

    - Update all `hash_type = "data"` to `hash_type.kind = "data"` and `hash_type.vm_version = 0`.

    - Update all `hash_type = "type"` to `hash_type.kind = "type"`.

    - Insert `edition = "2021"` into the head of the files.

  After upgrade the edition of chain specification,  the first time to run the CKB, you have to use `--overwrite-spec` parameter. The details of this parameter can be found with `ckb run --help`.

### BREAKING CHANGES

- The field `hash_type` in `Script` is changed from a `String` to an `Object` for all JSON RPC methods.

  The details can be found in the latest RFCs.

- The argument `--ba-hash-type` for `ckb init` is split into two arguments: `--ba-hash-type-kind` and `--ba-hash-type-vm-version`.

  More details can be found with `ckb-release init --help`.

PR #2844, opened 2021-07-19
chore: log failed transactions results

keroro520:
https://github.com/nervosnetwork/ckb/pull/2756#discussion_r662000531

### DIFF
--- a/tx-pool/src/process.rs
+++ b/tx-pool/src/process.rs
@@ -969,11 +969,15 @@ impl TxPoolService {
         let total = txs.len();
         let mut count = 0usize;
         for tx in txs {
-            if self._process_tx(tx, None).await.is_err() {
+            let tx_hash = tx.hash();
+            if let Err(err) = self._process_tx(tx, None).await {
+                error!("failed to process {:#x}, error: {:?}", tx_hash, err);
                 count += 1;
             }
         }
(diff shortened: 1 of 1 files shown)


tx-pool/src/process.rs
```diff
@@ -969,11 +969,15 @@ impl TxPoolService {
         let total = txs.len();
         let mut count = 0usize;
         for tx in txs {
-            if self._process_tx(tx, None).await.is_err() {
+            let tx_hash = tx.hash();
+            if let Err(err) = self._process_tx(tx, None).await {
+                error!("failed to process {:#x}, error: {:?}", tx_hash, err);
```
doitian, line 974: I think we'd better add a trace target to save the tx into a file. Example:

```
if log_enabled!(target: a_trace_traget_for_saveing_tx) {
  save_tx_into_file(&tx);
}
```

keroro520: I agree with it. We can save these errored transactions inside `fn _process_tx`.

In this PR, I just think it would be better to log the processing error outside of `self._process_tx()`.

doitian (approved):

yangby-cryptape (approved):

yangby-cryptape:
bors merge=doitian,yangby-cryptape

PR #2845, opened 2021-07-19
feat: add test-used rpc calculate_dao_field

keroro520:
Calculating `header.dao` is the most challenging step when creating testing blocks because it requires history blocks.

Now there is `generate_block_with_template`, but `generate_block_with_template` will insert the generated block: 

https://github.com/nervosnetwork/ckb/blob/0683c7d00222f9c0ae84121d821e282fabf75c48/rpc/src/module/test.rs#L236-L237

Sometimes test cases want to inject invalid blocks into the chain via [`process_block_without_verify`](https://github.com/nervosnetwork/ckb/blob/0683c7d00222f9c0ae84121d821e282fabf75c48/rpc/src/module/test.rs#L34-L35). So I added the test-used RPC `calculate_dao_field`.

### DIFF
--- a/rpc/src/module/test.rs
+++ b/rpc/src/module/test.rs
@@ -3,7 +3,7 @@ use ckb_app_config::BlockAssemblerConfig;
 use ckb_chain::chain::ChainController;
 use ckb_dao::DaoCalculator;
 use ckb_jsonrpc_types::{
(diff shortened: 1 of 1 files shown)


doitian (approved):

yangby-cryptape (approved):

keroro520:

PR #2849, opened 2021-07-20
feat: remove old version peer from peer store on fork

driftluo:
1. Remove the behavior of joining the peer store when the session is established, and now only operate the peer store after the identify or feeler is successfully established 
2. After the fork switch, the non-latest version of the node information will be slowly cleaned up, and no longer added to the peer store 
3. Remove unnecessary runtime judgment `RefCell`

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -670,7 +670,7 @@ impl<T: ExitHandler> ServiceHandle for EventHandler<T> {
                     match self.network_state.accept_peer(&session_context) {
                         Ok(Some(evicted_peer)) => {
                             debug!(
-                                "evict peer (disonnect it), {} => {}",
+                                "evict peer (disconnect it), {} => {}",
                                 evicted_peer.session_id, evicted_peer.connected_addr,
                             );
                             if let Err(err) = disconnect_with_message(
--- a/network/src/peer_registry.rs
+++ b/network/src/peer_registry.rs
@@ -108,7 +108,7 @@ impl PeerRegistry {
                 return Err(PeerError::ReachMaxOutboundLimit.into());
             }
         }
-        peer_store.add_connected_peer(remote_addr.clone(), session_type)?;
+        peer_store.add_connected_peer(remote_addr.clone(), session_type);
         let peer = Peer::new(session_id, session_type, remote_addr, is_whitelist);
         self.peers.insert(session_id, peer);
         Ok(evicted_peer)
--- a/network/src/peer_store/peer_store_impl.rs
+++ b/network/src/peer_store/peer_store_impl.rs
@@ -12,15 +12,14 @@ use crate::{
     PeerId, SessionType,
 };
 use ipnetwork::IpNetwork;
-use std::cell::{Ref, RefCell};
 use std::collections::{hash_map::Entry, HashMap};
 
 /// Peer store
 #[derive(Default)]
 pub struct PeerStore {
     addr_manager: AddrManager,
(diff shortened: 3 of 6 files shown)


keroro520:
Please add a test case to ensure that the staled peers are actually removed.

keroro520:
If we don't remove the old version peers, will there be security issues?

driftluo:
> Please add a test case to ensure that the staled peers are actually removed.

`check_purge` uses the same remove API, I don’t understand what you want to test

network/src/network.rs
```diff
@@ -217,6 +217,16 @@ impl NetworkState {
         }
     }
 
+    pub(crate) fn remove_peer_store(&self, session_id: SessionId) {
```
yangby-cryptape, line 220: Remove `info` in `store`, not remove `store`. I think the API name is not suitable.

keroro520:
> > Please add a test case to ensure that the staled peers are actually removed.
> 
> `check_purge` uses the same remove API, I don’t understand what you want to test
> 

Okay, I see. Why do we not remove old version peers inside `check_purge`?

driftluo:
> > > Please add a test case to ensure that the staled peers are actually removed.
> > 
> > 
> > `check_purge` uses the same remove API, I don’t understand what you want to test
> 
> Okay, I see. Why do we not remove old version peers inside `check_purge`?

`check_purge` only judges the connectivity, not the version. During the fork switch, connectivity cannot filter out the nodes of the old client

keroro520:
> > > > Please add a test case to ensure that the staled peers are actually removed.
> > > 
> > > 
> > > `check_purge` uses the same remove API, I don’t understand what you want to test
> > 
> > 
> > Okay, I see. Why do we not remove old version peers inside `check_purge`?
> 
> `check_purge` only judges the connectivity, not the version. During the fork switch, connectivity cannot filter out the nodes of the old client

Thanks, I do not have any additional questions about it.

keroro520 (approved):

doitian:
Not sure it is a good idea, since there's a very high probability that most the peers will switch to the fork.

driftluo:
> Not sure it is a good idea, since there's a very high probability that most the peers will switch to the fork.

This is just to clean up unnecessary node information so that it does not make meaningless broadcasts. There is no ban. If the node upgrades the client, it will be added back.

quake:
This PR title is easily misunderstood, suggest `remove old version peer from peer store on fork`

doitian:
> > Please add a test case to ensure that the staled peers are actually removed.
> 
> `check_purge` uses the same remove API, I don’t understand what you want to test
> 
> > If we don't remove the old version peers, will there be security issues?
> 
> There are no safety issues, there are efficiency issues

Could you expand this with more details? What's the performance impact if we do not purge the peer store on fork?

driftluo:
> Could you expand this with more details? What's the performance impact if we do not purge the peer store on fork?

The existing peer store rules are based on the detection of whether the session connection is successful. If it succeeds, the node information will be kept in the peer store.

In the fork version, the old and new networks can establish a session, but the corresponding protocol cannot be opened, and the network is isolated through the protocol version.

Formally because the session can be established, there will be a large number of invalid old version node information in the peer store. These data cannot be deleted by the current methods, and the dissemination of these invalid information in the network will result in a significant drop in the efficiency of node discovery

quake (approved):

yangby-cryptape (approved):

yangby-cryptape:

PR #2857, opened 2021-07-22
fix: fix the status marking problem of header sync

driftluo:
There has always been a problem with the header sync protocol. When only the sync protocol is enabled, there may be a relatively large delay in obtaining the latest block. This is caused by incorrect state marking. This PR introduces a new state. Try to fix this problem.

The action of header sync is not completed at one time. It needs to be updated in real-time during the entire connection process. It cannot be considered that this node is always at the highest just because it is synchronized to the highest one time. That is to say, the completed state of sync should be dynamically converted to need sync again. 

The `TipSynced` status indicates that this node has synchronized the headers and will try to synchronize again after 28s. At the same time, such a node can be used as the request object of `Getblocks`

`Suspend` and `TipSynced` are not exactly the same. Suspend thinks that the node may have some problems, such as being in the ibd state and suspending the sync request to it. Such a node cannot be used as the request object of the sync block, and it does not need to be based on the frequency of the block time to detection status

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -84,7 +84,7 @@ impl ChainSyncState {
             HeadersSyncState::Initialized => false,
             HeadersSyncState::SyncProtocolConnected => true,
             HeadersSyncState::Started => false,
-            HeadersSyncState::Suspend(until) => until < now,
+            HeadersSyncState::Suspend(until) | HeadersSyncState::TipSynced(until) => until < now,
         }
     }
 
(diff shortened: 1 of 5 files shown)


sync/src/types/mod.rs
```diff
@@ -111,6 +124,7 @@ enum HeadersSyncState {
     SyncProtocolConnected,
     Started,
     Suspend(u64), // suspend headers sync until this timestamp (milliseconds since unix epoch)
+    AlreadySync(u64), // already synced to the end, not as the sync target for the time being, until the pause time is exceeded
```
quake, line 127: suggest `TipSynced`

sync/src/types/mod.rs
```diff
@@ -100,9 +100,22 @@ impl ChainSyncState {
         self.headers_sync_state = HeadersSyncState::Suspend(until)
     }
 
+    fn already_sync(&mut self) {
+        let now = unix_time_as_millis();
+        // 28 seconds
```
quake, line 105: suggest to add comment `use avg block interval: (MAX_BLOCK_INTERVAL + MIN_BLOCK_INTERVAL) / 2 = 28`

sync/src/types/mod.rs
```diff
@@ -100,9 +100,22 @@ impl ChainSyncState {
...
+        self.headers_sync_state = HeadersSyncState::AlreadySync(now + 28000);
+    }
+
     fn started(&self) -> bool {
         matches!(self.headers_sync_state, HeadersSyncState::Started)
     }
+
+    fn synced(&self) -> bool {
```
quake, line 113: suggest `started_or_tip_synced`

quake (approved):

yangby-cryptape (approved):

driftluo:
bors r=quake,yangby-cryptape

PR #2864, opened 2021-07-23
chore(deps): upgrade several dependencies to remove tokio less than v1.0

yangby-cryptape:
### Purpose

Remove almost all tokio whose version is less than v1.0.

#### To Be Continued

This PR still left one dependency which is depended on tokio v0.2, because that upgrade will change a lot of code, I prefer to put that in another PR later.

### Changes

- Bump jsonrpc-* from 0.17.1 to 0.18.0.

- Bump reqwest from 0.10.9 to 0.11.4.

- Bump hyper from 0.13 to 0.14.

- Bump hyper-tls from 0.4 to 0.5.

- Bump trust-dns-resolver from 0.19 to 0.20.

- Bump ckb-sentry 0.21.0 to sentry 0.23.0.

  - We forked sentry to fix [RUSTSEC-2020-0041](https://rustsec.org/advisories/RUSTSEC-2020-0041.html), now that issue was fixed in upstream.

- Use `std::sync::mpsc::sync_channel(1)` instead of `tokio::sync::oneshot::channel()` in `tx_pool`.

  - To fix error "Cannot start a runtime from within a runtime." after upgrading RPC server's tokio.

    - JSON RPC server doesn't allow to use customize tokio executor.

      Ref: [`pub fn event_loop_executor(self, executor: tokio::Handle) -> Self;`](https://docs.rs/jsonrpc-http-server/18.0.0/jsonrpc_http_server/struct.ServerBuilder.html#method.event_loop_executor)

  - Use `block_in_place(|| response.recv())` instead of `self.handle.block_on(response)`.

    Without `block_in_place`, RPC service will be blocked very frequently.


driftluo (dismissed):

quake (approved):

doitian:
bors r=quake,driftluo

driftluo (approved):

doitian (approved):

doitian:
bors r=doitian,driftluo

PR #2879, opened 2021-07-28
feat(hardfork): add a new field "hardfork_features" to the return of RPC method "get_consensus"

yangby-cryptape:


rpc/README.md
```diff
@@ -74,6 +74,8 @@ The crate `ckb-rpc`'s minimum supported rustc version is 1.51.0.
         * [Method `get_raw_tx_pool`](#method-get_raw_tx_pool)
     * [Module Stats](#module-stats)
         * [Method `get_blockchain_info`](#method-get_blockchain_info)
+        * [Method `get_features`](#method-get_features)
```
doitian, line 77: Better add to module chain.

yangby-cryptape: 👌

doitian (approved):

quake:
I think these two rpc should be merged into `get_consensus` rpc.

quake (dismissed):

rpc/README.md
```diff
@@ -3512,6 +3523,19 @@ In JSONRPC, it is encoded as a 0x-prefixed hex string.
...
+
+The information about one hardfork feature.
+
+#### Fields
+
+`HardForkFeature` is a JSON object with the following fields.
+
+*   `rfc`: `string` - The related RFC ID.
+
+*   `epoch`: [`EpochNumber`](#type-epochnumber) `|` `null` - The first epoch when the feature is enabled.
```
doitian, line 3536: Please explain what does `null` mean.

doitian (changes requested):

rpc/README.md
keroro520: previously I misunderstood *epoch* as `EpochNumberWithFraction`, while it should actually be interpreted as `EpochNumber`. And sadly the underlying types of  `EpochNumberWithFraction` and `EpochNumber` are both `u64`.

I would suggest changing the name to *epoch_number*, to reduce ambiguity.

rpc/README.md
```diff
@@ -1260,6 +1261,10 @@ The response looks like below when `verbosity` is 0.
 
 Return various consensus parameters.
 
+##### Returns
+
+If any hardfork feature has `epoch=null`, it means the feature wouldn't be enabled forever.
```
doitian, line 1266:
```suggestion
If any hardfork feature has `epoch=null`, it means the feature will never be activated.
```

keroro520 (approved):

quake (approved):

doitian (approved):

doitian:
bors r=quake,doitian,keroro520

PR #2882, opened 2021-07-28
test(hardfork): test if the B extension feature is correct

yangby-cryptape:
### Purpose

At present, there is no tests for B extension.

Limited by the lack of tools, the test coverage for the whole B extension is dependent on `ckb-vm`.

In this PR, only add very simple tests with a very simple B extension instruction [`cpop`](https://github.com/riscv/riscv-bitmanip/blob/1.0.0/bitmanip/_popcount.adoc), to test if the B extension feature is correct:

```
|     Run Script with B-Ext Insts      |
|--------------------------------------|
|         |       |    B-Ext Insts     |
|         |       |---------+----------|
| Edition | Type  |         | Executed |
|         |       | Skipped |----------|
|         |       |         | OK | Err |
|---------+-------|---------+----+-----|
| ckb2019 | Data  |    O    | R  |  R  |
|         | Type  |    O    | R  |  R  |
|---------+-------+---------+----+-----|
|         | Data  |    O    | R  |  R  |
| ckb2021 | Type  |    O    | O  |  X  |
|         | Data1 |    O    | O  |  X  |
```

Description of the table as above:
- Skipped: the script has B extension instructions but not executed.
- Executed: the script has B extension instructions but executed.
  - OK: if all instructions are executed as them defined, the script should be passed.
  - Err: if all instructions are executed as them defined, the script should be failed.
- Result:
  - `O`: Passed.
  - `X`: Failed, the VM returns `ValidationFailure`.
  - `R`: Reject, the VM returns `InvalidInstruction`.


doitian:
Your table is confusing. What does O and X means in Err column?

yangby-cryptape:
> Your table is confusing. What does O and X means in Err column?

`O` is Passed, `X` is Failed.

But the reason for the failure is not same, I will add a `[R]eject` to show the result in table.

quake (approved):

driftluo (approved):

doitian:
bors r=quake,driftluo

PR #2894, opened 2021-08-02
fix: there may be competition between header sync and eviction

driftluo:
relation：#2857

The last PR left a problem. There may be a race condition between the two actions. Before reducing `n_sync_started`, we need to determine whether it is in the `Started` state. 

This PR will backport to 0.43


yangby-cryptape (approved):

quake (approved):

driftluo:
bors r=yangby-cryptape, quake

PR #2898, opened 2021-08-03
test: freeze block with extension

zhangsoledad:


store/src/db.rs
```diff
@@ -262,4 +263,38 @@ mod tests {
 
         assert_eq!(block.header(), store.get_tip_header().unwrap());
     }
+
+    #[test]
+    fn freeze_with_extension() {
```
doitian, line 268: We must cover another scenario:

> The block is saved without extension using the old schema, and we must load it successfully from frozen files using the new version schema.

doitian (changes requested):

yangby-cryptape (dismissed):

doitian (approved):

yangby-cryptape (approved):

zhangsoledad:
bors r=doitian, yangby-cryptape

PR #2899, opened 2021-08-03
feat: record recent reject

zhangsoledad:


doitian:
Related RFC: https://talk.nervos.org/t/rfc-ckb-better-support-to-query-transaction-status/6043

PR #2905, opened 2021-08-04
test: add cases that mining during sync

keroro520:


keroro520:
https://github.com/nervosnetwork/ckb/pull/2906

PR #2906, opened 2021-08-04
fix: try traverse all unknown parent hash

driftluo:
Try traverse all unknown parent hash

### DIFF
--- a/sync/src/orphan_block_pool.rs
+++ b/sync/src/orphan_block_pool.rs
@@ -1,6 +1,6 @@
 use ckb_types::{core, packed};
 use ckb_util::shrink_to_fit;
-use dashmap::DashMap;
+use dashmap::{DashMap, DashSet};
 use std::collections::{HashMap, VecDeque};
 
 pub type ParentHash = packed::Byte32;
@@ -10,8 +10,14 @@ const SHRINK_THRESHOLD: usize = 100;
 // orphan_block_pool and block_status_map, but `LruCache` would prune old items implicitly.
 #[derive(Default)]
 pub struct OrphanBlockPool {
+    // Group by blocks in the pool by the parent hash.
     blocks: DashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
+    // The map tells the parent hash when given the hash of a block in the pool.
+    //
+    // The block is in the orphan pool if and only if the block hash exists as a key in this map.
     parents: DashMap<packed::Byte32, ParentHash>,
+    // Leaders are blocks not in the orphan pool but having at least a child in the pool.
+    leaders: DashSet<ParentHash>,
 }
 
 impl OrphanBlockPool {
@@ -19,6 +25,7 @@ impl OrphanBlockPool {
         OrphanBlockPool {
             blocks: DashMap::with_capacity(capacity),
             parents: DashMap::new(),
+            leaders: DashSet::new(),
         }
     }
 
@@ -30,12 +37,27 @@ impl OrphanBlockPool {
             .entry(parent_hash.clone())
             .or_insert_with(HashMap::default)
             .insert(hash.clone(), block);
+        // Out-of-order insertion needs to be deduplicated
+        self.leaders.remove(&hash);
+        // It is a possible optimization to make the judgment in advance,
+        // because the parent of the block must not be equal to its own hash,
+        // so we can judge first, which may reduce one arc clone
+        if !self.parents.contains_key(&parent_hash) {
+            // Block referenced by `parent_hash` is not in the pool,
+            // and it has at least one child, the new inserted block, so add it to leaders.
+            self.leaders.insert(parent_hash.clone());
+        }
         self.parents.insert(hash, parent_hash);
     }
 
-    pub fn remove_blocks_by_parent(&self, hash: &packed::Byte32) -> Vec<core::BlockView> {
+    pub fn remove_blocks_by_parent(&self, parent_hash: &ParentHash) -> Vec<core::BlockView> {
+        // try remove leaders first
+        if self.leaders.remove(parent_hash).is_none() {
+            return Vec::new();
+        }
+
         let mut queue: VecDeque<packed::Byte32> = VecDeque::new();
-        queue.push_back(hash.to_owned());
+        queue.push_back(parent_hash.to_owned());
 
         let mut removed: Vec<core::BlockView> = Vec::new();
         while let Some(parent_hash) = queue.pop_front() {
@@ -51,6 +73,7 @@ impl OrphanBlockPool {
 
         shrink_to_fit!(self.blocks, SHRINK_THRESHOLD);
         shrink_to_fit!(self.parents, SHRINK_THRESHOLD);
+        shrink_to_fit!(self.leaders, SHRINK_THRESHOLD);
         removed
     }
 
@@ -64,6 +87,14 @@ impl OrphanBlockPool {
 
     pub fn len(&self) -> usize {
         self.parents.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn clone_leaders(&self) -> Vec<ParentHash> {
+        self.leaders.iter().map(|r| r.clone()).collect::<Vec<_>>()
     }
 }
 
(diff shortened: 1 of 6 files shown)


sync/src/relayer/mod.rs
```diff
@@ -836,12 +836,24 @@ impl CKBProtocolHandler for Relayer {
...
+                            if self.shared.state().orphan_pool().is_empty() {
+                                break;
+                            }
+                            if self.shared.is_stored(&hash) {
+                                self.shared.try_search_orphan_pool(&self.chain, &hash);
+                            }
+                        }
+                    })
+                }
+            }
```
quake, line 856: suggest to extract these lines and move it to `shared.try_search_orphan_pool` fn, and this fn should be refactor to:
```
pub fn try_search_orphan_pool(&self, chain: &ChainController) {

}
```

driftluo: `insert_new_block` also use this api，It needs to keep the current API

driftluo:
sync-test

sync/src/types/mod.rs
```diff
@@ -1445,6 +1451,11 @@ impl SyncShared {
             .is_some()
     }
 
+    /// Check whether block has been inserted to chain store
+    pub fn is_stored(&self, hash: &packed::Byte32) -> bool {
+        self.store().get_block_header(hash).is_some()
+    }
```
zhangsoledad, line 1457:
```
    let status = active_chain.get_block_status(&block_hash);
```
Maybe should use `get_block_status` to keep consistent with other units.

quake (dismissed):

doitian (changes requested):
I'll review it tomorrow.

driftluo:
I change `self.leaders.retain(|key| !self.parents.contain(key))` to `self.leaders.remove(&hash)`, it will reduce the consumption of traversal and hash，please review again @quake

sync/src/orphan_block_pool.rs
```diff
@@ -12,13 +12,15 @@ const SHRINK_THRESHOLD: usize = 100;
 pub struct OrphanBlockPool {
     blocks: DashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
     parents: DashMap<packed::Byte32, ParentHash>,
+    leaders: DashSet<ParentHash>,
```
doitian, line 15:
```suggestion
    // Leaders are blocks not in the orphan pool but but having at least a child in the pool.
    leaders: DashSet<ParentHash>,
```

sync/src/orphan_block_pool.rs
```diff
@@ -12,13 +12,15 @@ const SHRINK_THRESHOLD: usize = 100;
 pub struct OrphanBlockPool {
     blocks: DashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
```
doitian, line 13: (Minor suggestion) Add a comment:

```suggestion
    // Group by blocks in the pool by the parent hash.
    blocks: DashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
```

sync/src/orphan_block_pool.rs
```diff
@@ -12,13 +12,15 @@ const SHRINK_THRESHOLD: usize = 100;
...
     parents: DashMap<packed::Byte32, ParentHash>,
```
doitian, line 14: (Minor suggestion) Add a comment:

```suggestion
    // The map tells the parent hash when given the hash of a block in the pool.
    //
    // The block is in the orphan pool if and only if the block hash exists as a key in this map.
    parents: DashMap<packed::Byte32, ParentHash>,
```

sync/src/orphan_block_pool.rs
```diff
@@ -30,10 +32,20 @@ impl OrphanBlockPool {
             .entry(parent_hash.clone())
             .or_insert_with(HashMap::default)
             .insert(hash.clone(), block);
-        self.parents.insert(hash, parent_hash);
+        // Out-of-order insertion needs to be deduplicated
+        self.leaders.remove(&hash);
+        self.parents.insert(hash, parent_hash.clone());
+        if !self.parents.contains_key(&parent_hash) {
```
doitian, line 38:
```suggestion
        if !self.parents.contains_key(&parent_hash) {
            // Block referenced by `parent_hash` is not in the pool, and it has at least one child, the new inserted block, so add it to leaders. 
```

sync/src/orphan_block_pool.rs
```diff
@@ -30,10 +32,20 @@ impl OrphanBlockPool {
...
+            self.leaders.insert(parent_hash);
+        }
     }
 
     pub fn remove_blocks_by_parent(&self, hash: &packed::Byte32) -> Vec<core::BlockView> {
```
doitian, line 43: Since the hash intends to be a parent hash, it's better to make it explicitly

```suggestion
    pub fn remove_blocks_by_parent(&self, parent_hash: &ParentHash) -> Vec<core::BlockView> {
```
```

File: sync/src/orphan_block_pool.rs [hunk 8]
```diff
@@ -43,6 +55,7 @@ impl OrphanBlockPool {
                 let (hashes, blocks): (Vec<_>, Vec<_>) = orphaned.into_iter().unzip();
                 for hash in hashes.iter() {
                     self.parents.remove(hash);
+                    self.leaders.remove(hash);
```
doitian, line 58: In line 36, we should have already removed the block from leaders.

If we want to be protective, please add error logs if the hash does exist in leaders.

sync/src/orphan_block_pool.rs
```diff
@@ -65,6 +79,18 @@ impl OrphanBlockPool {
     pub fn len(&self) -> usize {
         self.parents.len()
     }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn leader_parents(&self) -> impl Iterator<Item = ParentHash> {
```
doitian, line 87: Confusing name. I would suggest `clone_leaders()` and return a `Vec`.

sync/src/orphan_block_pool.rs
```diff
@@ -133,4 +159,58 @@ mod tests {
...
+        assert_eq!(pool.len(), 3);
+        assert_eq!(pool.leaders.len(), 1);
+
+        pool.insert(blocks[15].clone());
+        assert_eq!(pool.len(), 4);
+        assert_eq!(pool.leaders.len(), 1);
+
+        let orphan_1 = pool.remove_blocks_by_parent(&blocks[14].hash());
+
+        let orphan: HashSet<BlockView> = orphan.into_iter().chain(orphan_1.into_iter()).collect();
```
doitian, line 210:
```suggestion
        let orphan_set: HashSet<BlockView> = orphan.into_iter().chain(orphan_1.into_iter()).collect();
```

sync/src/orphan_block_pool.rs
```diff
@@ -133,4 +159,58 @@ mod tests {
...
+        let block: HashSet<BlockView> = blocks.into_iter().collect();
```
doitian, line 211:
```suggestion
        let blocks_set: HashSet<BlockView> = blocks.into_iter().collect();
```

doitian: The singular name `block` makes no sense.

doitian (dismissed):
Some minor suggestions which should not block the merge.

doitian (approved):

doitian:
@zhangsoledad @quake please review

quake (approved):

driftluo:
bors r=doitian,quake

sync/src/orphan_block_pool.rs
```diff
@@ -65,6 +88,14 @@ impl OrphanBlockPool {
     pub fn len(&self) -> usize {
         self.parents.len()
     }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    pub fn clone_leaders(&self) -> Vec<ParentHash> {
```
zhangsoledad, line 96: Could use impl trait return an Iterator here?

driftluo: Yes, but https://github.com/nervosnetwork/ckb/pull/2906#discussion_r684604252, so I deleted it

doitian:
Blocked by https://github.com/nervosnetwork/ckb/pull/2913

driftluo:
bors r=doitian,quake

PR #2910, opened 2021-08-05
feat: adjust benchmark workflow

keroro520:


.github/workflows/benchmark.yml
```diff
@@ -6,6 +6,8 @@ on:
     types: [ submitted ]
   schedule:
     - cron: '0 0 * * *'
+  # Manurally trigger
+  workflow_dispatch:
```
keroro520, line 10: I add the `workflow_dispatch` trigger so that we can manually bench specified branches.

.github/workflows/benchmark.yml
```diff
@@ -14,11 +16,12 @@ jobs:
     if: |
       (
         github.event_name == 'pull_request_review' &&
-        contains(github.event.review.body, '#benchmark') &&
-        contains(fromJson('[ "janx", "doitian", "quake", "xxuejie", "zhangsoledad", "jjyr", "TheWaWaR", "driftluo", "keroro520", "yangby-cryptape", "liya2017" ]'), github.actor)
```
keroro520, line 18: As we are using github-hosted runner, we don't need to check `github.actor`

.github/workflows/benchmark/terraform/variables.tf
```diff
@@ -10,13 +10,13 @@ variable "secret_key" {
 
 variable "region" {
   type    = string
-  default = "ap-northeast-1"
+  default = "us-east-2"
```
keroro520, line 13: It may timeout for github-hosted runner to connect to ap-northeast-1 ec2, so I changed to us-east-2

.github/workflows/benchmark/terraform/variables.tf
```diff
@@ -10,13 +10,13 @@ variable "secret_key" {
...
   description = "AWS region"
 }
 
 variable "instances_count" {
   type    = number
-  default = 3
+  default = 1
```
keroro520, line 19: CKB has some issues with network synchronization at the moment. I'm just benching one node here.

keroro520 (commented):
#benchmark

keroro520 (commented):
#benchmark

quake (approved):

liya2017 (approved):

liya2017:
bors r=quake,liya2017

PR #2912, opened 2021-08-06
[perf] add dep resolve cache

chanhsu001:


doitian:
Reworking by @zhangsoledad

PR #2914, opened 2021-08-09
refactor: remove impl cell_provider on store

zhangsoledad:

In principle，we should only impl cell_provider on snapshot, not the store directly.

Remove impl cell_provider on store

### Release note 

```release-note
```


doitian (approved):

quake (approved):

quake:
bors r=quake,doitian

PR #2923, opened 2021-08-10
fix: callback potentially incorrect trigger on concurrent context

zhangsoledad:
https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process.rs#L471
https://github.com/nervosnetwork/ckb/blob/develop/tx-pool/src/process.rs#L707-L709 
duplicate checked before submitting, but the lock released later,  there is still potentially inclusion.


yangby-cryptape (approved):

quake (approved):

yangby-cryptape:
bors r=yangby-cryptape,quake

PR #2924, opened 2021-08-10
fix(test): make sure testnode graceful shutdown basic sync

### LINKED ISSUE #2927: Unit test failed because of error "pure virtual method called"
**Environment**

- **CKB version**: v0.100.0-pre
- **Chain**: N/A
- **Operating system**: Linux
- **Arch**: x64

**Describe the bug**

Unit test sometimes failed because of "pure virtual method called" recently, such as in this [ci check](https://github.com/nervosnetwork/ckb/pull/2922/checks?check_run_id=3288664116#step:3:1245).

As recommended by @xxuejie , I try to reproduce it using [rr](https://rr-project.org/).

- Find out the test binary file like /home/ian/codebase/nervos/ckb/target/debug/deps/ckb_sync-60892a517e6c0f11
- Run it and record using rr in a loop until the error occurs.

    ```
    while rr record /home/ian/codebase/nervos/ckb/target/debug/deps/ckb_sync-60892a517e6c0f11; do
        sleep 1
    done
    ```
- Remember the last test case before the test quits. Now replay it in the debugger.

    ```
    rr replay --debugger rust-gdb
    gdb> break ckb_sync::synchronizer::tests::test_process_new_block
    gdb> run
    gdb> next
    ```

After several `next`, the error occurs again, then check the backtrace (attached below).

From the backtrace, we can see that the root cause is RocksDB. The test quits without closing the database first, as mentioned in the comment https://github.com/facebook/rocksdb/issues/649#issuecomment-119764927 . The database is opened by [`TestNode`](https://github.com/nervosnetwork/ckb/blob/624d0372769873bf119944d10feb931091b757f2/sync/src/tests/mod.rs#L23). The test node does not exit gracefully, which leads to the error.

### Attachments

- Backtrace [gdb.txt](https://github.com/nervosnetwork/ckb/files/6961709/gdb.txt)

zhangsoledad:
tests::synchronizer::basic_sync do not graceful shutdown, may cause 
```
pure virtual method called                                                                                                                                                                   
terminate called without an active exception 
```
bump crossbeam-channel from 0.3.9 to 0.5.1 incidentally.

Closes #2927

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -538,7 +538,7 @@ dependencies = [
 name = "ckb-channel"
 version = "0.100.0-pre"
 dependencies = [
- "crossbeam-channel 0.3.9",
+ "crossbeam-channel",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


quake (approved):

yangby-cryptape (approved):

yangby-cryptape:
bors r=quake,yangby-cryptape

doitian (approved):

PR #2934, opened 2021-08-11
fix: fix stream body read

driftluo:
HTTP body is a stream, it can't be read all by once.

We need to use streaming to read it completely, otherwise, the miner will not be able to obtain the block template correctly when the transaction volume is large.

### DIFF
--- a/miner/src/client.rs
+++ b/miner/src/client.rs
@@ -9,7 +9,7 @@ use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_types::{packed::Block, H256};
 use futures::prelude::*;
 use hyper::{
-    body::{Bytes, HttpBody},
+    body::{to_bytes, Bytes},
     header::{HeaderValue, CONTENT_TYPE},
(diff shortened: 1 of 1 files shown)


miner/src/client.rs
```diff
@@ -70,11 +74,7 @@ impl Rpc {
                             .await
                             .map(|res|res.into_body())
                         {
-                            Ok(mut body) => body
-                                .data()
-                                .await
-                                .ok_or(RpcError::NoRespData)
-                                .and_then(|res| res.map_err(RpcError::Http)),
+                            Ok(body) => StreamBody::new(body).await,
```
quake, line 77: I think this should be fixed by [to_bytes](https://docs.rs/hyper/0.14.11/hyper/body/fn.to_bytes.html)

zhangsoledad (approved):

quake (approved):

driftluo:
bors r=zhangsoledad,quake

PR #2941, opened 2021-08-11
fix(hardfork): should always check the extension field when reconstruct blocks

yangby-cryptape:

Forget the extension fields when reconstruct blocks.

Only the relay protocol is broken, the sync protocol works well, so the chain works well.

The only things could be observed is that there are uncles in logs, but no real uncles in blocks.

I checked all `Block::new_builder(..)`, and found 2 locations.

### Release note

This bug is not in  `version < 0.100`, so no release note is needed for the next release (`v0.100.0`).

```release-note
None
```

### DIFF
--- a/util/types/src/extension/shortcuts.rs
+++ b/util/types/src/extension/shortcuts.rs
@@ -385,6 +385,35 @@ impl packed::CompactBlock {
             .filter(|index| !prefilled_indexes.contains(&index))
             .collect()
     }
+
+    /// Gets the i-th extra field if it exists; i started from 0.
+    pub fn extra_field(&self, index: usize) -> Option<bytes::Bytes> {
+        let count = self.count_extra_fields();
+        if count > index {
+            let slice = self.as_slice();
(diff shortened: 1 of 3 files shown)


driftluo (approved):

util/types/src/extension/shortcuts.rs
```diff
@@ -385,6 +385,35 @@ impl packed::CompactBlock {
...
+    }
+
+    /// Gets the extension field if it existed.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the first extra field exists but not a valid [`Bytes`](struct.Bytes.html).
+    pub fn extension(&self) -> Option<packed::Bytes> {
+        self.extra_field(0)
+            .map(|data| packed::Bytes::from_slice(&data).unwrap())
```
yangby-cryptape, line 415: #### Why it can panic here?

The only way to get a `CompactBlock` from the external is the `relay` protocols, and the relay messages are verified.

doitian:
The release note fence code block is for description to be included in the CHANGELOG.

```
    ```release-note
    ```
```

doitian (approved):

PR #2945, opened 2021-08-12
chore(hardfork): size of the block extension field should be included in block size limit

yangby-cryptape:
### What problem does this PR do?

- Add some code to count the size of the block extension field when package a block template.

  This code is meaningless currently, since there are no extension in block templates.
  The biggest significance of this code is to ensure that the logic is correct.

- Add tests for the feature: the size of the block extension field should be included in block size limit.

  See the discussion in https://github.com/nervosnetwork/rfcs/pull/224#issuecomment-896763326.

### Release note

```release-note
None
```


driftluo (approved):

quake (approved):

yangby-cryptape:
bors merge=driftluo,quake

PR #2948, opened 2021-08-14
refactor(ckb-resource): Remove dependency on tempfile in ckb-resource

chanhsu001:
write target file directly; append .tmp.$PID to the target file as the tmp file

### DIFF
--- a/resource/src/lib.rs
+++ b/resource/src/lib.rs
@@ -32,7 +32,6 @@ use std::fmt;
 use std::fs;
 use std::io::{self, BufReader, Read};
 use std::path::{Path, PathBuf};
-use tempfile::NamedTempFile;
 
 use ckb_system_scripts::BUNDLED_CELL;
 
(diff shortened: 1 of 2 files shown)


quake (changes requested):
please check the ci error, the changes of `Cargo.lock` file should be checked-in also.

resource/src/lib.rs
```diff
@@ -271,11 +269,10 @@ mod tests {
     use std::fs;
     use std::path::{Path, PathBuf};
 
-    fn mkdir() -> tempfile::TempDir {
-        tempfile::Builder::new()
-            .prefix("ckb_resource_test")
-            .tempdir()
-            .unwrap()
+    fn mkdir() -> PathBuf {
```
quake, line 272: Does `/tmp/ckb_resource_test` folder create successful on windows? I think we may keep the test code unchanged here, just move `tempfile` crate to `[dev-dependencies]` section.

chanhsu001: ok,  you are right, let's move tempfile crate into dev-dependencies and untouch test code.

quake (approved):

yangby-cryptape (approved):

yangby-cryptape:

PR #2960, opened 2021-08-20
refactor: redesign sync orphan_block_pool and add bench.

chanhsu001:
[sync]: refactor orphan block pool and add benchmark

resign data structure, reduce 3 maps to 2 maps, make interface not expose inner map.

Problem Summary:

redesign datastruct and invent two maps:   
trees  --  keep leader and its sub-node blocks
root_of --  keep all child-parent relationship

What's Changed:

### Related changes
sync/types/mod.rs  and relayer   -- where to use orphan block pool  

Tests 

- Unit test
- Benchmark

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/sync/src/orphan_block_pool.rs
+++ b/sync/src/orphan_block_pool.rs
@@ -1,151 +1,325 @@
-use ckb_logger::debug;
+use std::collections::{HashMap, VecDeque};
+use std::ops::Deref;
+
+use ckb_error::{Error, InternalErrorKind};
+use ckb_types::prelude::ShouldBeOk;
 use ckb_types::{core, packed};
-use ckb_util::{parking_lot::RwLock, shrink_to_fit};
-use std::collections::{HashMap, HashSet, VecDeque};
+use ckb_util::parking_lot::RwLock;
+use ckb_util::shrink_to_fit;
 
+/// alias name
 pub type ParentHash = packed::Byte32;
+/// alias name
+pub type Hash = packed::Byte32;
+
 const SHRINK_THRESHOLD: usize = 100;
+const MAX_SUB_NODES_UPDATE_IN_ROOTOF: usize = 16;
 
-#[derive(Default)]
-struct InnerPool {
-    // Group by blocks in the pool by the parent hash.
-    blocks: HashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
-    // The map tells the parent hash when given the hash of a block in the pool.
-    //
-    // The block is in the orphan pool if and only if the block hash exists as a key in this map.
-    parents: HashMap<packed::Byte32, ParentHash>,
-    // Leaders are blocks not in the orphan pool but having at least a child in the pool.
-    leaders: HashSet<ParentHash>,
-}
-
-impl InnerPool {
-    fn with_capacity(capacity: usize) -> Self {
-        InnerPool {
-            blocks: HashMap::with_capacity(capacity),
-            parents: HashMap::new(),
-            leaders: HashSet::new(),
-        }
-    }
-
-    fn insert(&mut self, block: core::BlockView) {
-        let hash = block.header().hash();
-        let parent_hash = block.data().header().raw().parent_hash();
-        self.blocks
-            .entry(parent_hash.clone())
-            .or_insert_with(HashMap::default)
-            .insert(hash.clone(), block);
-        // Out-of-order insertion needs to be deduplicated
-        self.leaders.remove(&hash);
-        // It is a possible optimization to make the judgment in advance,
-        // because the parent of the block must not be equal to its own hash,
-        // so we can judge first, which may reduce one arc clone
-        if !self.parents.contains_key(&parent_hash) {
-            // Block referenced by `parent_hash` is not in the pool,
-            // and it has at least one child, the new inserted block, so add it to leaders.
-            self.leaders.insert(parent_hash.clone());
-        }
-        self.parents.insert(hash, parent_hash);
-    }
-
-    pub fn remove_blocks_by_parent(&mut self, parent_hash: &ParentHash) -> Vec<core::BlockView> {
-        // try remove leaders first
-        if !self.leaders.remove(parent_hash) {
-            return Vec::new();
-        }
-
-        let mut queue: VecDeque<packed::Byte32> = VecDeque::new();
-        queue.push_back(parent_hash.to_owned());
-
-        let mut removed: Vec<core::BlockView> = Vec::new();
-        while let Some(parent_hash) = queue.pop_front() {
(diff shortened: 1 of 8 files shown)


sync/src/orphan_block_pool.rs
```diff
@@ -215,46 +328,54 @@ mod tests {
...
+
+        let result = pool.insert(blocks[0].clone());
+        assert!(result.is_ok());
+        assert_eq!(pool.root_of_len(), 18);
+        assert_eq!(pool.trees_len(), 2);
+
+        //block 0 already in pool, remove blocks by genesis hash will work
+        let result = pool.remove_blocks_by_parent(&consensus.genesis_block().hash());
+        assert_ne!(result, None);
+        // let vec = result.unwrap();
```
yangby-cryptape, line 362: Useless comments.

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,260 @@
...
+        InnerOrphanBlockPool {
+            trees: HashMap::with_capacity(capacity),
+            root_of: HashMap::new(),
         }
     }
 
-    /// Insert orphaned block, for which we have already requested its parent block
-    pub fn insert(&self, block: core::BlockView) {
-        self.inner.write().insert(block);
+    /// root_of table contains direct child-parent relationship
```
yangby-cryptape, line 76:
```suggestion
    /// root_of table contains direct child-parent relationship
```

Lots of trailing whitespaces, not only in this line.

yangby-cryptape (changes requested):
Please fix all clippy warnings and errors, then run rustfmt to format the code.

chanhsu001:
now the update pass ci, integration-test and integration-release.
this PR should be "proposal pull request".

new design reduce 3 maps to 2 maps(pros), but less Intuitive (cons) and a little performace loss(?).
benchmark part,  should simulate with real mainnet enviroment pattern, update later.

doitian:
Could you please add the benchmark result to support your point that we should use the new data strucutre?

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,326 @@
...
-    blocks: HashMap<ParentHash, HashMap<packed::Byte32, core::BlockView>>,
-    // The map tells the parent hash when given the hash of a block in the pool.
-    //
-    // The block is in the orphan pool if and only if the block hash exists as a key in this map.
-    parents: HashMap<packed::Byte32, ParentHash>,
-    // Leaders are blocks not in the orphan pool but having at least a child in the pool.
-    leaders: HashSet<ParentHash>,
+struct InnerOrphanBlockPool {
+    // store root and sub-node block
+    trees: HashMap<ParentHash, VecDeque<(Hash, core::BlockView)>>,
```
doitian, line 81: `BlockView` already cached the hash.

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,326 @@
...
+    /// params:
+    ///
+    /// key: which tree's sub-nodes needs be be upgrade (in trees)
+    ///
+    /// root: update root
+    fn update_root_in_rootof(&mut self, key: &Hash, root: &Hash) {
+        let tree = self.trees.get(key).should_be_ok();
+        if tree.iter().size_hint().0 < MAX_SUB_NODES_UPDATE_IN_ROOTOF {
+            for (hash, _) in tree.iter() {
+                *self.root_of.get_mut(hash).unwrap() = root.clone();
```
doitian, line 126: Just use `HashMap::insert`.

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,326 @@
...
+    ///
+    /// 4. update root
+    ///
+    /// params: block - block inserted
+    ///
+    /// Returns OK if none error occurs
+    ///
+    /// Returns Error or Panic if internal data not sync
+    fn insert(&mut self, block: core::BlockView) -> Result<(), Error> {
+        let hash = block.header().hash();
```
doitian, line 149: Use `block.hash()`.

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,326 @@
...
+                None
+            })
+        })
     }
 
-    pub fn clone_leaders(&self) -> Vec<ParentHash> {
-        self.inner.read().leaders.iter().cloned().collect()
+    /// get leaders(key) from trees
+    pub fn get_leaders(&self) -> Vec<Hash> {
+        let mut result = vec![];
```
doitian, line 292: `self.trees.keys().iter().cloned().collect()`

sync/src/orphan_block_pool.rs
```diff
@@ -1,152 +1,326 @@
...
+    pub fn remove_blocks_by_parent(&self, root: &Hash) -> Vec<core::BlockView> {
+        self.inner
+            .write()
+            .remove_blocks_by_parent(root)
+            .unwrap_or_default()
+    }
+    /// return block stored in pool
+    ///
+    /// return None if not found in pool
+    pub fn get_block(&self, hash: &Hash) -> Option<core::BlockView> {
```
doitian, line 50: Can we avoid this function? It has a worse performance in the new data strcuture.

chanhsu001: `get_block` function is already existed, and used in two places: one for test code(ok, omit it); the second is "sync/src/relay/mod.rs", function `reconstruct_block()`

get_block itself is simple(but worse performance ) and probably not improved by refactor,  so there is only way: NOT used it. 
i will check `reconstruct_block()` if possible

chanhsu001: in reconstruct_block(), the context of use `get_block()` :  
when the code goes to get block of 'uncle hash', it checks state.block_status_map to check if it received....
ok, pool already received it before, so let fetch it:
self.shared.state().get_orphan_block(&uncle_hash)  --> it will call orphan pool's `get_block()`

i **suspect** much the uncle block is  trees.values().front()  (1st node of every value of trees)
if it's true, we could change it into simple and fast implementation.
but i don't have enough knowledge to answer this question or confirm it, i need help.

chanhsu001: in new design, there are 2 tables:  
- root_of: keeps relationship(child -->father or ancestor)
- trees: keeps ancestor and all_connected_children block data

as above design, there is no shortcut, if we want to fetch any block.

driftluo: > i suspect much the uncle block is trees.values().front() (1st node of every value of trees)

This assumption is wrong; any block in the current epoch can be accepted as an uncle block

chanhsu001: so we have to use `get_block`...i have no clue to improve it, any suggestions?

PR #2964, opened 2021-08-23
refactor: refactor peer store

driftluo:

1. Peer store‘s detection and eviction efficiency is too low 
2. There is a certain conflict between the peer store obtaining available nodes and the nodes that need to be probed 
3. There is a problem with interval's choice of MissedTickBehavior
4.  When a node is disconnected, based on the peer store strategy, the node information may be lost 

1. Made some adjustments to the peer store's strategy of randomly obtaining node information
```
| -- choose to identify --| --- choose to feeler --- | --      delete     -- |
| 1      | 2     | 3      | 4    | 5    | 6   | 7    | More than seven days  |
```
2. Made some adjustments to the use of interval:
- Removed the immediate execution behavior of dump peer store startup 
- Adjust interval default MissedTickBehavior to delay or skip 

3. The detection behavior is adjusted to not conflict with the behavior of completing outbound and will be executed permanently 
4. When no valid completion node can be found continuously, try to connect to boot node randomly 
5. Remove ban list from peer store
6. Evict connectable nodes by network segment 
7. Remove nodes that have been connected within 15s 

### Check List 

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -75,7 +75,7 @@ pub struct NetworkState {
     pending_observed_addrs: RwLock<HashSet<Multiaddr>>,
     local_private_key: secio::SecioKeyPair,
     local_peer_id: PeerId,
-    bootnodes: Vec<Multiaddr>,
+    pub(crate) bootnodes: Vec<Multiaddr>,
     pub(crate) config: NetworkConfig,
     pub(crate) active: AtomicBool,
     /// Node supported protocols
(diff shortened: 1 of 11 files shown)


yangby-cryptape (approved):
I have read all related issues (I found two).
Two participants disappeared after several comments, I'm not sure they agreed or disagreed this PR.
I don't know whether our team have reached a consensus with this PR.

I only approve that the code is same as the issue described.

driftluo:
bors r=quake,yangby-cryptape

PR #2970, opened 2021-08-24
fix: chunk process cycle calculation

driftluo:

Tx chunk process cycles calculation error:
1. all suspended branches lack the cycle calculation process
https://github.com/nervosnetwork/ckb/blob/39c8cee51018931e4be8c90d776071faf42647c0/script/src/verify.rs#L508-L517
2. `resume_from_state` use incoming VM, It needs to avoid repeated calculation of cycles that already exists 
https://github.com/nervosnetwork/ckb/blob/39c8cee51018931e4be8c90d776071faf42647c0/script/src/verify.rs#L654-L655
3. `next_limit_cycles` on `TransactionSnapshot` or `TransactionState` must less than or equal to `step_cycles`
4. `resume_from_state` use incoming VM，reset max cycles must be greater than old max cycle
https://github.com/nervosnetwork/ckb/blob/39c8cee51018931e4be8c90d776071faf42647c0/script/src/verify.rs#L650

Rewrite the cycles calculation process in the entire verification process 

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/script/src/types.rs
+++ b/script/src/types.rs
@@ -116,6 +116,10 @@ impl<'a> ResumableMachine<'a> {
         self.machine.machine.cycles()
     }
 
+    pub(crate) fn max_cycles(&self) -> Cycle {
+        self.machine.machine.max_cycles()
+    }
+
     #[cfg(test)]
     pub(crate) fn set_cycles(&mut self, cycles: Cycle) {
         self.machine.machine.set_cycles(cycles)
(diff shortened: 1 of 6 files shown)


doitian:
Affected versions?

driftluo:
> Affected versions?

all 0.100.x

doitian:
Can you split it into two commits?

1. The first commit add test case to expose the bug, so the test will fail.
2. The second commit fix the bug.

doitian:
Since there's no corresponding bug report, please provide more details about the bug.

Hint: use the bug report template.

- Give steps to reproduce the bug or provide the minimal test case.
- Expected result.
- Real result.

driftluo:
resolve by  https://github.com/nervosnetwork/ckb/pull/2971, current PR repair method has potential problems

PR #2878, opened 2021-07-28
chore(hardfork): bump db migrate version for the new column family of block extension data

yangby-cryptape:
This PR is useless, unless the PR #2877 was backported to branch "rc/v0.43".

PR #3000, opened 2021-09-01
feat: add RPC to return median fee rate

chanhsu001:
add RPC to return median fee rate

Add new rpc interface to fetch transaction median fee rate
user could specify blocks_to_scan and min_transactions_in_block parameter to filter blocks or use default setting.

Problem Summary:

when new block is added,  also insert BlockTxStat, contains all transactions (exclude cellbase tx) cycle and tx_size information; 
we add new RPC to support scan number of latest blocks on chain, collect number of transactions; and calculate the median fee rate

What's Changed:

### Related changes

- db-schema

Tests 

- Unit test

Side effects

- db-schema change

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -18,7 +18,7 @@ use ckb_types::{
             ResolveOptions, ResolvedTransaction,
         },
         service::{Request, DEFAULT_CHANNEL_SIZE, SIGNAL_CHANNEL_SIZE},
-        BlockExt, BlockNumber, BlockView, HeaderView,
+        BlockExt, BlockNumber, BlockTxStat, BlockView, HeaderView,
     },
     packed::{Byte32, ProposalShortId},
     U256,
@@ -762,13 +762,34 @@ impl ChainService {
                             ) {
                                 Ok((cycles, cache_entries)) => {
                                     let txs_fees =
-                                        cache_entries.into_iter().map(|entry| entry.fee).collect();
+                                        cache_entries.iter().map(|entry| entry.fee).collect();
                                     txn.attach_block(b)?;
                                     attach_block_cell(txn, b)?;
                                     let mut mut_ext = ext.clone();
                                     mut_ext.verified = Some(true);
                                     mut_ext.txs_fees = txs_fees;
                                     txn.insert_block_ext(&b.header().hash(), &mut_ext)?;
+
+                                    // block tx statistics will not include cellbase tx
+                                    // entry.cycles already exclude cellbase
+                                    let tx_stat = BlockTxStat {
+                                        txs_cycles: cache_entries
+                                            .iter()
+                                            .map(|entry| entry.cycles)
+                                            .collect(),
+                                        txs_size: resolved
+                                            .iter()
+                                            .skip(1)
+                                            .map(|resolved_tx| {
+                                                resolved_tx
+                                                    .transaction
+                                                    .data()
+                                                    .serialized_size_in_block()
+                                                    as u64
+                                            })
+                                            .collect(),
+                                    };
+                                    txn.insert_block_tx_stat(&b.header().hash(), &tx_stat)?;
                                     if !switch.disable_script() && b.transactions().len() > 1 {
                                         info!(
                                             "[block_verifier] block number: {}, hash: {}, size:{}/{}, cycles: {}/{}",
--- a/db-schema/src/lib.rs
+++ b/db-schema/src/lib.rs
@@ -3,7 +3,7 @@
 /// Column families alias type
 pub type Col = &'static str;
 /// Total column number
-pub const COLUMNS: u32 = 16;
+pub const COLUMNS: u32 = 17;
 /// Column store chain index
 pub const COLUMN_INDEX: Col = "0";
 /// Column store block's header
@@ -38,6 +38,8 @@ pub const COLUMN_NUMBER_HASH: Col = "13";
 pub const COLUMN_CELL_DATA_HASH: Col = "14";
 /// Column store block extension data
 pub const COLUMN_BLOCK_EXTENSION: Col = "15";
+/// Column store block's transaction cycle and size
+pub const COLUMN_BLOCK_TXS_STAT: Col = "16";
 
 /// META_TIP_HEADER_KEY tracks the latest known best block header
 pub const META_TIP_HEADER_KEY: &[u8] = b"TIP_HEADER";
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -50,6 +50,7 @@ The crate `ckb-rpc`'s minimum supported rustc version is 1.51.0.
         * [Method `get_fork_block`](#method-get_fork_block)
         * [Method `get_consensus`](#method-get_consensus)
         * [Method `get_block_median_time`](#method-get_block_median_time)
+        * [Method `get_median_fee_rate`](#method-get_median_fee_rate)
     * [Module Experiment](#module-experiment)
         * [Method `dry_run_transaction`](#method-dry_run_transaction)
         * [Method `calculate_dao_maximum_withdraw`](#method-calculate_dao_maximum_withdraw)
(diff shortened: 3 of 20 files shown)


db-schema/src/lib.rs
```diff
@@ -38,6 +38,8 @@ pub const COLUMN_NUMBER_HASH: Col = "13";
 pub const COLUMN_CELL_DATA_HASH: Col = "14";
 /// Column store block extension data
 pub const COLUMN_BLOCK_EXTENSION: Col = "15";
+/// Column store block's transaction_cycle_size
```
quake, line 41:
```suggestion
/// Column store block's transactions cycle and size
```

db-schema/src/lib.rs
```diff
@@ -38,6 +38,8 @@ pub const COLUMN_NUMBER_HASH: Col = "13";
...
+pub const COLUMN_TX_STAT: Col = "16";
```
quake, line 42:
```suggestion
pub const COLUMN_BLOCK_TXS_STAT: Col = "16";
```

chain/src/chain.rs
```diff
@@ -763,13 +763,34 @@ impl ChainService {
...
+                                    // block tx statistics will not include cellbase tx
+                                    // entry.cycles already exclude cellbase
+                                    let tx_stat = BlockTxStat {
+                                        txs_cycles: cache_entries
+                                            .into_iter()
+                                            .map(|entry| entry.cycles)
+                                            .collect(),
+                                        txs_size: resolved
+                                            .iter()
+                                            .filter(|tx| !tx.is_cellbase())
```
quake, line 783: first tx of block is always cellbase, we may use skip:
```suggestion
                                            .skip(1)
```

chain/src/chain.rs
```diff
@@ -763,13 +763,34 @@ impl ChainService {
...
                                     let mut mut_ext = ext.clone();
                                     mut_ext.verified = Some(true);
                                     mut_ext.txs_fees = txs_fees;
                                     txn.insert_block_ext(&b.header().hash(), &mut_ext)?;
+
+                                    // block tx statistics will not include cellbase tx
+                                    // entry.cycles already exclude cellbase
+                                    let tx_stat = BlockTxStat {
+                                        txs_cycles: cache_entries
+                                            .into_iter()
```
quake, line 778:
```suggestion
                                            .iter()
```

rpc/README.md
```diff
@@ -1372,6 +1373,51 @@ Response
...
+* `get_median_fee_rate(blocks_to_scan, num_transactions)`
+    * `blocks_to_scan`: [`Uint32`](#type-uint32) `|` `null`
+    * `num_transactions`: [`Uint32`](#type-uint32) `|` `null`
+* result: [`Capacity`](#type-capacity)
+
+Returns the median fee rate in recent number of blocks specified by user.
+
+##### Params
+
+*   `blocks_scan` - the latest x (default and maximum blocks: 42) blocks scanned to calculate median fee rate.
```
quake, line 1386: wrong params name

```suggestion
*   `blocks_to_scan` - the latest x (default and maximum blocks: 42) blocks scanned to calculate median fee rate.
```

rpc/README.md
```diff
@@ -1372,6 +1373,51 @@ Response
 }
 ```
 
+#### Method `get_median_fee_rate`
+* `get_median_fee_rate(blocks_to_scan, num_transactions)`
+    * `blocks_to_scan`: [`Uint32`](#type-uint32) `|` `null`
+    * `num_transactions`: [`Uint32`](#type-uint32) `|` `null`
```
quake, line 1379: wrong params name?
```suggestion
    * min_transactions`: [`Uint32`](#type-uint32) `|` `null`
```

rpc/src/module/chain.rs
```diff
@@ -1645,4 +1692,124 @@ impl ChainRpc for ChainRpcImpl {
...
+        let tip = snapshot.tip_header().number();
+        let mut match_blocks = Vec::with_capacity(blocks_to_scan as usize);
+        let mut block_counter = 0;
+        let mut tx_counter = 0;
+
+        // get block hash and check if contains enough transaction
+        let mut hash = snapshot.get_block_hash(tip).ok_or_else(|| {
+            RPCError::custom(RPCError::ChainIndexIsInconsistent, "get_block_hash error")
+        })?;
+        while block_counter < blocks_to_scan {
```
quake, line 1727: please use block number to simplify the code, pseudocode:

```
for block_number in (tip_number..tip_number - blocks_to_scan) {
    let block_hash = snapshot.get_block_hash(block_number);
    let block_ext = snapshot.get_block_tx_stat(block_hash);
    ...
}
```

quake (changes requested):

chain/src/chain.rs
```diff
@@ -763,13 +763,34 @@ impl ChainService {
...
-                                        cache_entries.into_iter().map(|entry| entry.fee).collect();
+                                        cache_entries.iter().map(|entry| entry.fee).collect();
                                     txn.attach_block(b)?;
                                     attach_block_cell(txn, b)?;
                                     let mut mut_ext = ext.clone();
                                     mut_ext.verified = Some(true);
                                     mut_ext.txs_fees = txs_fees;
                                     txn.insert_block_ext(&b.header().hash(), &mut_ext)?;
+
+                                    // block tx statistics will not include cellbase tx
```
doitian, line 774: Please add a feature switch in the config file and disable it by default. Also mention the config option in the RPC document that if users want to use this RPC, they must enable the feature.

rpc/README.md
doitian: Missing a \` in the suggested change.

rpc/src/module/chain.rs
```diff
@@ -1645,4 +1692,124 @@ impl ChainRpc for ChainRpcImpl {
...
+                "chain hasn't enough transactions in {} recent blocks",
+                blocks_to_scan
+            )));
+        }
+
+        let mut fee_vec: Vec<u64> = Vec::with_capacity(blocks_to_scan as usize);
+        for hash in &match_blocks {
+            let stat = snapshot.get_block_tx_stat(hash).ok_or_else(|| {
+                RPCError::custom(
+                    RPCError::ChainIndexIsInconsistent,
```
doitian, line 1763: It's OK that database has no the stat for a block. Choose a new RPC error for this.

```
-1200: ChainBlockTxStatsAreNotFound
```

doitian (changes requested):

rpc/src/module/chain.rs
chanhsu001: ok, i will modify code as you and quake suggest, thanks!

rpc/README.md
chanhsu001: i change the param name to:  `blocks_to_scan` and `min_transactions`

db-schema/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 /// Column families alias type
 pub type Col = &'static str;
 /// Total column number
-pub const COLUMNS: u32 = 16;
+pub const COLUMNS: u32 = 17;
```
doitian, line 6: @zhangsoledad @yangby-cryptape Do we need migration to add a column now?

yangby-cryptape: > Do we need migration to add a column now?

Required.

> Or what happens if I open the database with 17 columns in the ckb version which knows only 16 columns?

Panic.

Ref: https://github.com/nervosnetwork/ckb/commit/f782409b7fb22c9d378200540ec5bae8a0f12858

chanhsu001: so i need add migration part, right?  and migration still need to done with hardfork? 
please tell me more about migration.

doitian: https://github.com/nervosnetwork/ckb/pull/2878

chanhsu001: already add  db migration commit

chanhsu001:
close temporarily, re-open if later needed

PR #2947, opened 2021-08-13
fix: Fix miner

driftluo:
### What problem does this PR solve?

1. In the case that the poll interval is small enough, the miner may work on unknown work, which is equivalent to doing useless work(work cache is 32, the channel is unbound, worker get new work just one loop one get)
2. Assuming that miner must be submitted successfully 
3. Miner will submit uncle block 
4. If `block_on_submit` is false, the miner will not submit the block

### Tests

- setup a dev node(test for Dummy)，use miner subcommand for it
- setup a staging node(test for Eaglesong)，use miner subcommand for it
- change `block_on_submit` to false，and `poll_interval` = 200, old miner can't work, the new one works well
- old miner may have uncle block submit, the new one doesn't

### Related changes

Rewritten part of the miner logic 

### Release note

```release-note
None
```

PR #3012, opened 2021-09-07
fix: fix dummy miner solve

driftluo:

It is a mix of several problems here:

1. Timed poll is turned on from off on dummy miner
2. There is a consistency problem with the deletion judgment because of the background poll, that is, when self start = false, it will also update current, resulting in the previous judgment becoming false when start becomes true
3. Poll new block template will not be updated in this case because of the same work id, i.e., it is not possible to remove the judgment error caused by 2 through poll template

The actual effect is that there is a thread in the miner that keeps on looping but does not work

### Check List

Tests

 init with dev chain, mining it, it will not be stuck

### Release note

```release-note
```

### DIFF
--- a/miner/src/worker/dummy.rs
+++ b/miner/src/worker/dummy.rs
@@ -105,17 +105,13 @@ impl Dummy {
 
 impl Worker for Dummy {
     fn run<G: FnMut() -> u128>(&mut self, mut rng: G, _progress_bar: ProgressBar) {
-        let mut current = self.pow_work.clone();
         loop {
             self.poll_worker_message();
-            if current.as_ref().map(|a| &a.0) != self.pow_work.as_ref().map(|a| &a.0) && self.start
-            {
+            if self.start {
                 if let Some((pow_hash, work)) = self.pow_work.clone() {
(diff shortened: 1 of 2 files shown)


yangby-cryptape (approved):

miner/src/worker/dummy.rs
```diff
@@ -105,17 +105,13 @@ impl Dummy {
...
             self.poll_worker_message();
-            if current.as_ref().map(|a| &a.0) != self.pow_work.as_ref().map(|a| &a.0) && self.start
-            {
+            if self.start {
                 if let Some((pow_hash, work)) = self.pow_work.clone() {
                     self.solve(pow_hash, work, rng());
                 }
             }
-
-            current = self.pow_work.clone();
```
doitian, line 118: I recommend move line 118 to 113 to fix the inconsistent issue.

doitian: The usage of the deleted guard is avoiding uncles. Since #2947 has resolved the issue, it's safe to remove the guard.

doitian (approved):

doitian:

PR #3015, opened 2021-09-07
test: add unit test for `get_block_template_cache` and `exceeded_maximum_ancestors_count`

quake:

Problem Summary: add two unit test:

1. add unit test for get_block_template cache expiration
2. add unit test for exceeded maximum ancestors count

What's Changed: only new unit test

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/3015)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,11 @@ GRCOV_EXCL_LINE = \s*(((log|ckg_logger)::)?(trace|debug|info|warn|error)|(debug_
 
 ##@ Testing
 .PHONY: test
-test: ## Run all tests.
+test: ## Run all tests, including some tests can be time-consuming to execute (tagged with [ignore])
+	cargo test ${VERBOSE} --all -- --nocapture --ignored
+
+.PHONY: quick-test
+quick-test: ## Run all tests, excluding some tests can be time-consuming to execute (tagged with [ignore])
 	cargo test ${VERBOSE} --all -- --nocapture
 
 .PHONY: cov-install-tools
--- a/error/src/util.rs
+++ b/error/src/util.rs
@@ -28,13 +28,13 @@ macro_rules! assert_error_eq {
 ///
 /// ## Examples
 ///
-/// ```ignore
+/// ```no_run
 /// impl_error_conversion_with_kind!(SourceType, error_kind, TargetType)
 /// ```
 ///
 /// the expanded code:
 ///
-/// ```ignore
+/// ```no_run
 /// impl From<SourceType> for TargetType {
 ///     fn from(source: SourceType) -> TargetType {
 ///         TargetType {
@@ -60,13 +60,13 @@ macro_rules! impl_error_conversion_with_kind {
 ///
 /// ## Examples
 ///
-/// ```ignore
+/// ```no_run
 /// impl_error_conversion_with_adaptor!(SourceType, AdaptorType, TargetType)
 /// ```
 ///
 /// the expanded code:
 ///
-/// ```ignore
+/// ```no_run
 /// impl From<SourceType> for TargetType {
 ///     fn from(source: SourceType) -> TargetType {
 ///         let adaptor: AdaptorType = source.into();
--- a/rpc/src/tests/examples.rs
+++ b/rpc/src/tests/examples.rs
@@ -1,23 +1,23 @@
-use crate::{RpcServer, ServiceBuilder};
+use crate::{
+    tests::{always_success_transaction, next_block},
+    RpcServer, ServiceBuilder,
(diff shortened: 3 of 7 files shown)


doitian (changes requested):
> Reviewed 4 of 4 files at r1, 2 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/3015)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @driftluo, @keroro520, and @quake)

___
*[rpc/src/tests/module/miner.rs, line 24 at r1](https://reviewable.io/reviews/nervosnetwork/ckb/3015#-MjCpKlC4qPXu7bqAOqu:-MjCpKlD3xdRN6D1-LGa:b-9o6jua) ([raw file](https://github.com/nervosnetwork/ckb/blob/384a0f49c64dbc7c0f0be8ae72588ec7759076c5/rpc/src/tests/module/miner.rs#L24)):*
> ```XML
> 
> #[test]
> fn test_get_block_template_cache() {
> ```

This test case takes at least 12 seconds. I think its too long for a unit test.
___
*[rpc/src/tests/module/miner.rs, line 73 at r1](https://reviewable.io/reviews/nervosnetwork/ckb/3015#-MjCoeHx-YYF6cUJjItK:-MjCoeHyEEUBeyl47B-O:b6yyv1z) ([raw file](https://github.com/nervosnetwork/ckb/blob/384a0f49c64dbc7c0f0be8ae72588ec7759076c5/rpc/src/tests/module/miner.rs#L73)):*
> ```
> block template cache will expire after 3 seconds
> ```

This and the next test case all sleeps 4 seconds. It can tell whether the cache is expired because of time out or that new blocks and txs are added.
___
*[rpc/src/tests/module/pool.rs, line 186 at r2](https://reviewable.io/reviews/nervosnetwork/ckb/3015#-MjCpw-05o1_ZQ79a-09:-MjCpw-05o1_ZQ79a-0A:bi9vlh4) ([raw file](https://github.com/nervosnetwork/ckb/blob/b274049649e7fb6deb1c5822b4b92367456858cb/rpc/src/tests/module/pool.rs#L186)):*
> ```XML
> 
>     // the default value of pool config `max_ancestors_count` is 25, only 25 txs will be added to committed list of the block template
>     sleep(Duration::from_secs(1));
> ```

This is fragile. I recommend polling the response every 300ms and giving up after 2 seconds.

doitian (commented):
> *[Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/3015)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @driftluo, @keroro520, and @quake)

___
*[rpc/src/tests/module/miner.rs, line 73 at r1](https://reviewable.io/reviews/nervosnetwork/ckb/3015#-MjCoeHx-YYF6cUJjItK:-MjCqUeuAzHXpRT38cbX:b-8ye3wp) ([raw file](https://github.com/nervosnetwork/ckb/blob/384a0f49c64dbc7c0f0be8ae72588ec7759076c5/rpc/src/tests/module/miner.rs#L73)):*
<details><summary><i>Previously, doitian (ian) wrote…</i></summary><blockquote>

> ```
> block template cache will expire after 3 seconds
> ```

This and the next test case all sleeps 4 seconds. It can tell whether the cache is expired because of time out or that new blocks and txs are added.

</blockquote></details>

typo: can -> cannot

quake:
> This and the next test case all sleeps 4 seconds. It cannot tell whether the cache is expired because of time out or that new blocks and txs are added.

there is a block template cache which only expires after 3 seconds: https://github.com/nervosnetwork/ckb/blob/05d530f5e7f51d77524fecdb552ec36820c66bb3/tx-pool/src/process.rs#L111

have to sleep more than 3 seconds to refresh the cache, and this unit test is trying to assert the refreshed block template should include new uncle and new txs.

I cannot find a better way to speedup this unit test.

doitian:
> I cannot find a better way to speedup this unit test.

I think we should add `#[ignore]` on slow tests and include them in CI via `cargo test -- --ignored`

quake:
> > I cannot find a better way to speedup this unit test.
> 
> I think we should add `#[ignore]` on slow tests and include them in CI via `cargo test -- --ignored`

resolved, please review again

keroro520 (approved):

driftluo (approved):

doitian (approved):

quake:

PR #3028, opened 2021-09-12
refactor(rpc): change 'connected_duration' duration unit to milliseconds

chanhsu001:

make duration unit consistent in `get_peers` RPC 

Problem Summary:

Proposal: in 'get_peer' RPC, connected_duration uses secs but last_ping_duration uses millis,  we need to change connected_duration uses millis.

What's Changed: 

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test

Side effects

- rpc client should change 'connected_duration' duration unit from 'second' to 'millisecond' when `get_peers` RPC is used.

### Release note 

```release-note
**BREAKING RPC**. Change field `connected_duration` time unit from seconds to **milliseconds** in `get_peers` RPC to make time unit  used  consistently in `get_peers` , RPC clients are suggested to modify time quantity correspond with this change.
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -4082,7 +4082,7 @@ A remote node connects to the local node via the P2P network. It is often called
 
     If the connection is established by the local node, `is_outbound` is true.
 
-*   `connected_duration`: [`Uint64`](#type-uint64) - Elapsed time in seconds since the remote node is connected.
+*   `connected_duration`: [`Uint64`](#type-uint64) - Elapsed time in milliseconds since the remote node is connected.
 
 *   `last_ping_duration`: [`Uint64`](#type-uint64) `|` `null` - Elapsed time in milliseconds since receiving the ping response from this remote node.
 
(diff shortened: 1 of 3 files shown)


doitian (changes requested):
This is a breaking change, please add the release note.

chanhsu001:
> This is a breaking change, please add the release note.
i already add a note at end of PR content, is it ok?

doitian:
> > This is a breaking change, please add the release note.
> 
> i already add a note at end of PR content, is it ok?

We'll use the content inside the release note to generate the CHANGELOG.

    fix(rpc): foo foo #3028

    ```release-note
    bar bar
    ```

Will generate an entry like this:

- #3028: fix(rpc): foo foo

    bar bar

doitian (approved):

chanhsu001:
bors merge

quake (approved):

quake:
bors r=quake,doitian

PR #3035, opened 2021-09-16
fix: update composite source to ckb repo

liya2017:

fix: update composite source to ckb repo

### Release note 

```release-note
```


quake (approved):

quake (approved):

yangby-cryptape (approved):
I don't agree with the `--release`. It will make CI slower.
But I think we should fix CI first, then we could try another better way for that.
So I approved.

quake:
bors r=quake,yangby-cryptape

yangby-cryptape:
bors r=quake,yangby-cryptape

PR #3041, opened 2021-09-19
feat(app_config): add new item "tx_block_stat_enable" in ckb.toml chain section

chanhsu001:
add new config item "tx_block_stat_enable" in ckb.toml [chain] section, as a feature to support tx
median fee rate.

Problem Summary: as a feature to support "tx median fee rate"(https://github.com/nervosnetwork/ckb/pull/3000)

What's Changed: 

- we add a config item"tx_block_stat_enable" in ckb.toml, default value is false;
- if user use legacy ckb.toml(without this item), also get default value(false) effect; 
- if user config it to true, "tx_median_fee_rate" rpc will work and transactions' info, such as tx_size, tx_cylces will be insert to db, and rpc "get_median_fee_rate" will return effective value.
- for implementation, we add this config item in Chain Config, and pass it as parameter to invented function "ChainService::new_with_config(...)",  in order to pass config information to ChainService. (we keep the orginal function "ChainService::new(...)" instead, avoid much code modification, especial test code) 

### Related changes

- PR to update `owner/repo`:

### Check List 

- User need to add "tx_block_stat_enable" item in ckb.toml and set it to true, before RPC "get_median_fee_rate" works. 

Tests 

- Unit test

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1,6 +1,7 @@
 //! CKB chain service.
 #![allow(missing_docs)]
 
+use ckb_app_config::ChainConfig;
 use ckb_channel::{self as channel, select, Sender};
 use ckb_error::{Error, InternalErrorKind};
 use ckb_logger::{self, debug, error, info, log_enabled, trace, warn};
@@ -185,6 +186,8 @@ impl GlobalIndex {
 pub struct ChainService {
     shared: Shared,
     proposal_table: ProposalTable,
+    #[allow(dead_code)]
+    chain_config: Option<ChainConfig>,
 }
 
 impl ChainService {
@@ -193,6 +196,19 @@ impl ChainService {
         ChainService {
             shared,
             proposal_table,
+            chain_config: None,
+        }
+    }
+
+    pub fn new_with_config(
+        shared: Shared,
+        proposal_table: ProposalTable,
+        chain_config: Option<ChainConfig>,
+    ) -> ChainService {
+        ChainService {
+            shared,
+            proposal_table,
+            chain_config,
         }
     }
 
@@ -687,6 +703,14 @@ impl ChainService {
         self.find_fork_until_latest_common(fork, &mut index);
 
         is_sorted_assert(fork);
+    }
+
+    #[allow(dead_code)]
+    fn is_tx_block_stat_enabled(&self) -> bool {
+        if let Some(chain_cfg) = &self.chain_config {
+            return chain_cfg.tx_block_stat_enable;
+        }
+        false
     }
 
     // we found new best_block
--- a/resource/ckb-miner.toml
+++ b/resource/ckb-miner.toml
@@ -8,6 +8,7 @@
 data_dir = "data"
 
 [chain]
+tx_block_stat_enable = false
 # Choose the kind of chains to run, possible values:
 # - { file = "specs/dev.toml" }
 # - { bundled = "specs/testnet.toml" }
--- a/test/template/ckb.toml
+++ b/test/template/ckb.toml
@@ -7,6 +7,7 @@ data_dir = "data"
 # - { file = "specs/dev.toml" }
 # - { bundled = "specs/testnet.toml" }
 # - { bundled = "specs/mainnet.toml" }
+tx_block_stat_enable = false
 spec = { file = "specs/integration.toml" }
 
 [logger]
--- a/util/app-config/src/app_config.rs
+++ b/util/app-config/src/app_config.rs
@@ -131,6 +131,9 @@ pub struct MinerAppConfig {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct ChainConfig {
+    #[serde(default)]
+    /// Specifies if transaction_block_statistics(tx_size, tx_cycles) feature is enabled
+    pub tx_block_stat_enable: bool,
     /// Specifies the chain spec.
     pub spec: Resource,
 }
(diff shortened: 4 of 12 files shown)


chanhsu001:
i need to know more about configration item in migration and confirm it worked as we suppose...
please help to provide detailed information, Thanks.

util/app-config/src/exit_code.rs
```diff
@@ -48,3 +49,10 @@ impl From<clap::Error> for ExitCode {
         ExitCode::Cli
     }
 }
+
+impl From<ParseBoolError> for ExitCode {
+    fn from(err: ParseBoolError) -> ExitCode {
+        eprintln!("Config Error: {:?}", err);
```
yangby-cryptape, line 55: A story before getting to the point:
I had a bad experience with [`capsule`].
When I ran [`capsule`] in my laptop, it always threw "file not found", but I never had a chance to know which file was not found.
Even I enabled `RUST_BACKTRACE=full`, I still didn't know where the error was thrown.
Ref: [How `capsule` handle the errors?](https://github.com/nervosnetwork/capsule/blob/56bf934d89824c03a0464f347d2f24f3558d9d6d/src/bin/capsule.rs#L406-L409)

So, if I was a ckb user, I would rather know the incorrect field than only know why has an error.
My suggestion here is: remove this `From` , construct the error at the location where throws the error and write the filed name into the error message.

[`capsule`]: https://github.com/nervosnetwork/capsule

util/app-config/src/legacy/chain.rs
```diff
@@ -0,0 +1,46 @@
+use serde::Deserialize;
```
yangby-cryptape, line 1: This file is useless, there is no deprecated (legacy) fields.

util/app-config/src/legacy/chain.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct ChainConfig {
+    #[serde(default = "default_tx_block_stat_enable")]
```
yangby-cryptape, line 6: Move this to the real `ChainConfig` after delete this file.

resource/ckb-miner.toml
```diff
@@ -8,6 +8,7 @@
 data_dir = "data"
 
 [chain]
+tx_block_stat_enable = false
```
yangby-cryptape, line 11: You should better comment this option which let the config file to be backward compatible.

You don't have to add it explicitly because there is a default value and the default value is same as the value you want to set here.

See the follow example which add a new option `use_binary_version_as_message_prefix`:
https://github.com/nervosnetwork/ckb/blob/d7a330fde8800ac242a2ccfe195bfe7ea4d50791/resource/ckb.toml#L183-L184

p.s. Same suggestion for the `resource/ckb.toml`.

test/template/ckb.toml
```diff
@@ -7,6 +7,7 @@ data_dir = "data"
 # - { file = "specs/dev.toml" }
 # - { bundled = "specs/testnet.toml" }
 # - { bundled = "specs/mainnet.toml" }
+tx_block_stat_enable = false
```
yangby-cryptape, line 10: It's OK to Just leave this file be same as before.

yangby-cryptape (changes requested):

chain/src/chain.rs
```diff
@@ -185,6 +186,8 @@ impl GlobalIndex {
 pub struct ChainService {
     shared: Shared,
     proposal_table: ProposalTable,
+    #[allow(dead_code)]
+    chain_config: Option<ChainConfig>,
```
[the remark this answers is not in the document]
chanhsu001: allow(dead_code)) will be removed when other PR(https://github.com/nervosnetwork/ckb/pull/3000) is merged...
sorry, but i thought the function contains too much change, so i divide into two PRs.

new_with_config(...), i also unsure about it.
but first of all, i choose Option<ChainConfig>, cause this parameter is not mandatory. 
specific code(eg. calculate median fee rate) will use ChainConfig, only if it need  ChainConfig and check chain_config is Some(...) 

invent new API - ChainService::new_with_config(),  i just don't want to change current codes(use ChainService::new(...)) but actually i'm unsure about it, cause lots of code, espcially test code using ChainService::new(...)
OK, if majority think it's better to keep only ONE API: ChainService:::new(), i will modify related code and remove Chain Service::new_with_config(...)

chain/src/chain.rs
```diff
@@ -689,6 +705,14 @@ impl ChainService {
         is_sorted_assert(fork);
     }
 
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
chanhsu001: "dead_code" will be removed after two PRs merged.

util/app-config/src/exit_code.rs
chanhsu001: good point!

util/app-config/src/app_config.rs
```diff
@@ -301,6 +304,7 @@ impl CKBAppConfig {
             touch(self.logger.log_dir.join(&self.logger.file))?;
         }
         self.chain.spec.absolutize(root_dir);
+        self.chain.tx_block_stat_enable = false;
```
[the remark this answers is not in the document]
chanhsu001: ok, i will check this part.

chanhsu001:
close temporarily, re-open it if later needed

PR #3052, opened 2021-09-24
fix: fix identify disconnect

driftluo:

fix https://github.com/nervosnetwork/ckb/runs/3673717980#step:4:1915 this ci fail

1. assert fail because the assert is wrong here, the `registered` behavior is before `received`, and the `discovery` in the test is broadcast every second. It is impossible to determine whether the node3 is just in the process of trying to connect when asserting. It can also be changed to `wait_connect_state(&node3, 1)`, but I don’t think it’s necessary, it’s no problem to delete it directly 
2. When identify is closed, we should first check whether the current node information is banned and then insert the peer store, otherwise, the banned address may be spread 

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/network/src/protocols/tests/mod.rs
+++ b/network/src/protocols/tests/mod.rs
@@ -312,8 +312,16 @@ fn test_identify_behavior() {
     wait_connect_state(&node1, 1);
     wait_connect_state(&node2, 0);
 
+    if !wait_until(10, || {
+        node1.network_state.with_peer_store_mut(|peer_store| {
+            peer_store.is_addr_banned(&node2.listen_addr)
+                && peer_store.addr_manager().get(&node2.listen_addr).is_none()
+        })
+    }) {
+        panic!("identify can't ban not same net")
+    }
+
     let sessions = node3.connected_sessions();
-    assert_eq!(sessions.len(), 1);
 
     if !wait_until(10, || node3.connected_protocols(sessions[0]).len() == 4) {
         panic!("identify can't open other protocols")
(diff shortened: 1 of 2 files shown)


network/src/protocols/tests/mod.rs
```diff
@@ -312,8 +312,16 @@ fn test_identify_behavior() {
...
+        node1.network_state.with_peer_store_mut(|peer_store| {
+            peer_store.is_addr_banned(&node2.listen_addr)
+                && peer_store.addr_manager().get(&node2.listen_addr).is_none()
+        })
+    }) {
+        panic!("identify can't ban not same net")
+    }
+
     let sessions = node3.connected_sessions();
-    assert_eq!(sessions.len(), 1);
```
keroro520, line 316: Why remove this assertion?

keroro520 (changes requested):

network/src/protocols/tests/mod.rs
driftluo: Because the assert is wrong here, the `registered` behavior is before `received`, and the `discovery` in the test is broadcast every second. It is impossible to determine whether the node3 is just in the process of trying to connect when asserting. It can also be changed to `wait_connect_state(&node3, 1)`, but I don’t think it’s necessary, it’s no problem to delete it directly

keroro520: According to the PR description, this PR fixes the failure of https://github.com/nervosnetwork/ckb/runs/3673717980#step:4:1915

As far as I know, this failure is due to https://github.com/nervosnetwork/ckb/pull/3052#discussion_r715374379

Then, it makes me confusing that I cannot figure out the relationship between this testing failure and "banned".

keroro520 (approved):

yangby-cryptape (approved):

driftluo:

driftluo:

yangby-cryptape:

PR #3060, opened 2021-09-28
chore: check if the corresponding tx hash has already been requested before processing the tx relay message

quake:

By design we should check that the corresponding tx hash has been requested before processing the relay transaction message, but the current code misses this check, this PR resolved this issue.

Check the `unknown_tx_hashes` structure before processing the message to see if the corresponding tx hash has already been requested.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

This PR removed `broadcast_transaction` rpc from the `IntegrationTest` module, other integration tests that rely on this rpc need to be adapted accordingly, please refer to the related change of test code in this PR.

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/module/test.rs
+++ b/rpc/src/module/test.rs
@@ -3,7 +3,7 @@ use ckb_app_config::BlockAssemblerConfig;
 use ckb_chain::chain::ChainController;
 use ckb_dao::DaoCalculator;
 use ckb_jsonrpc_types::{
-    AsEpochNumberWithFraction, Block, BlockTemplate, Byte32, Cycle, JsonBytes, Script, Transaction,
+    AsEpochNumberWithFraction, Block, BlockTemplate, Byte32, JsonBytes, Script, Transaction,
 };
 use ckb_logger::error;
 use ckb_network::{NetworkController, SupportProtocols};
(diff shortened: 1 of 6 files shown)


rpc/src/module/test.rs
```diff
@@ -150,44 +143,6 @@ impl IntegrationTestRpc for IntegrationTestRpcImpl {
...
-            Err(RPCError::custom_with_error(
-                RPCError::P2PFailedToBroadcast,
-                err,
-            ))
-        } else {
-            Ok(hash.unpack())
-        }
-    }
-
-    fn notify_transaction(&self, tx: Transaction) -> Result<H256> {
```
driftluo, line 179: This RPC does not directly push the transaction out. Its purpose is to use the `resumeble_process_tx` interface for testing. I don't think it needs to be deleted?

quake: I think I deleted this method by mistake, it shouldn't be impacted by this PR, I've reverted this change, please review it again, thanks.

driftluo (approved):

chanhsu001 (approved):

quake:
bors r=driftluo,chanhsu001

PR #3075, opened 2021-10-08
fix: clean self-runner space when disk space is less than threshold

liya2017:

fix: clean self-runner space when disk space is less than threshold

### Check List 

Tests

- Unit test

### Release note 

```release-note
```


quake:
bors r=quake,chanhsu001

PR #3058, opened 2021-09-27
fix: ignore proposal window on vm version selection

driftluo:
### What problem does this PR solve?

Transaction submission in the current epoch can only use the VM version based on the current epoch 

### Check List

Tests

- Integration test in other repo

### Release note

```release-note
Title Only: Include only the PR title in the release note.
```

PR #3090, opened 2021-10-12
fix: tx-pool refreshes caches with the incorrect VM version after hardfork

chaoticlonghair:

Tx-pool refreshes caches with the incorrect VM version after hardfork if current tip block is the last block before hardfork.

#### Problem Summary

After #3058, all scripts will be verified under the environment of current tip block.

So when tip block is the last block before hardfork, the hardfork features are not enabled, that means all scripts will be verified with old features.

But in next block, all scripts should be verified with new hardfork features.

#### PR Summary

- test(hardfork): add a script and a unit test to test the change of the script's cycles after upgrade vm0 to vm1

  For test propose, I need a **fixed-cycles** lock script to `assert` and it should has different cycles in vm0 and vm1.
  The cycles of secp256k1 lock is not stable.

  This unit test also ensures the VM version is correct:
  https://github.com/nervosnetwork/ckb/blob/e8f88fd73451c526f37e0879a5ed41272dbd3020/script/src/types.rs#L57-L58

- fix(test): fix `MockChain` that it always panics in `get_block_epoch(..)` when epoch changed

- fix(hardfork): fix tx-pool that it refreshes caches with the incorrect VM version if it re-org to the last block before hardfork

  When the tip is the last block before hardfork, the transactions in block template should be verified with new hardfork features.

  So, we should always verify the transactions which are not committed, with the features for next block.

- test(hardfork): add a unit test for the tx-pool caches (for #3079)

- fix(hardfork): network should enable ckb2021 features when tip is the last block of ckb2019

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/tx-pool/src/process.rs
+++ b/tx-pool/src/process.rs
@@ -688,7 +688,10 @@ impl TxPoolService {
         // The network protocol is switched after tx-pool confirms the cache,
         // there will be no problem with the current state as the choice of the broadcast protocol.
         let with_vm_2021 = {
-            let epoch = snapshot.tip_header().epoch().number();
+            let epoch = snapshot
+                .tip_header()
+                .epoch()
+                .minimum_epoch_number_after_n_blocks(1);
             self.consensus
                 .hardfork_switch
                 .is_vm_version_1_and_syscalls_2_enabled(epoch)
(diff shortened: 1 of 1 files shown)


quake:
bors r=quake,driftluo

PR #3095, opened 2021-10-15
feat: remove invalid header dep tx for reorg

zhangsoledad:

Previously, transaction is not wiped immediately after tip changed, header dep becomes invalid, this may lead header-dep transaction stuck in tx-pool, but to be rare in practice.

Remove invalid header-dep transaction after tip changed.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


zhangsoledad:
bors retry

PR #3110, opened 2021-10-19
fix: fix pending compact block memory bloat on abnormal flow

driftluo:

The cleanup of `pending_compact_blocks` is only in the process of the successful process([here](https://github.com/nervosnetwork/ckb/blob/92f1d8f69aa6b52e28a7164dc34dd62544531591/sync/src/relayer/block_transactions_process.rs#L122) or [here](https://github.com/nervosnetwork/ckb/blob/92f1d8f69aa6b52e28a7164dc34dd62544531591/sync/src/relayer/compact_block_process.rs#L207)). When the block is synchronized by the sync protocol first or the remote node does not respond, the cleanup process will not be executed, causing unexpected memory bloat.

The modification of this PR will depend on each successful compact block process. As long as there is one success, it will clean up all unexpected unsuccessful legacy issues. 

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -7,6 +7,7 @@ use ckb_logger::{self, debug_target};
 use ckb_network::{CKBProtocolContext, PeerIndex};
 use ckb_traits::HeaderProvider;
 use ckb_types::{core, packed, prelude::*};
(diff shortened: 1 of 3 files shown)


quake:
bors r=quake,yangby-cryptape

PR #3113, opened 2021-10-19
feat(fork): Activate ckb2021 in testnet since epoch 3113

doitian:

Proposal: Activate ckb2021 in testnet since epoch 3113, which is around 2021/10/24 3:15 UTC

### Related changes

After the activation, nodes running v0.101.0 and above are not compatible with nodes running old versions in the testnet.

### Check List 

Tests 

- Integration test

Side effects

- Breaking backward compatibility: Not compatible with old versions in the testnet after the activation.

### Release note 

```release-note
```


quake:
bors r=quake,yangby-cryptape

PR #3116, opened 2021-10-19
fix: fix pending compact retain

driftluo:

Issue introduced in #3110

Block number cannot be used as the judgment condition here, but epoch number should be used because any block of the current epoch can be accepted as uncle, and the practice of using block number as the judgment condition will not be able to broadcast and accept some uncle records, resulting in an abnormal uncle rate 

### Check List

Tests

- Unit test

### Release note

```release-note
```


doitian:

PR #3129, opened 2021-10-21
fix(hardfork): tx-pool clear statistical data before re-run all transactions

chaoticlonghair:

When tx-pool refreshes caches during hardfork, the statistical data should be cleared before re-run all transactions.

### Check List

Tests

- Unit test

### Release note

```release-note
```


quake:
bors r=quake,zhangsoledad,driftluo

PR #3141, opened 2021-10-23
feat: increase yamux window size

driftluo:

Increase the window size of yamux to improve bandwidth utilization.

By default, the window size of yamux is 256kb, but the maximum size of a block of ckb is 500k+, which means that the message needs to be sent and received twice + 2 window update messages to complete 

Side effects

- Memory usage may increase 

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -849,6 +849,7 @@ impl<T: ExitHandler> NetworkService<T> {
         let mut service_builder = ServiceBuilder::default();
         let yamux_config = YamuxConfig {
             max_stream_count: protocol_metas.len(),
+            max_stream_window_size: 1024 * 1024,
             ..Default::default()
         };
         for meta in protocol_metas.into_iter() {


chanhsu001:
> Side effects
Memory usage may increase

it would be better to provide more detailed evaluation of memory usage when a full node is running on "sync" stage and on "running" stage.

driftluo:
> > Side effects
> > Memory usage may increase
> 
> it would be better to provide more detailed evaluation of memory usage when a full node is running on "sync" stage and on "running" stage.

If you understand the concept of yamux window, you will not ask this question. This setting is to increase the upper limit of a single communication, which is the maximum memory usage of a protocal. It does not mean that so much is pre-allocated. It will only be allocated when it is used. if not used, this parameter is equivalent to not taking effect

keroro520 (approved):
Better add a code comment to explain that stream window size should be greater than the max block size.

driftluo:
bors r=keroro520,chanhsu001

PR #3156, opened 2021-11-01
chore: remove inflight block struct's compact's record

driftluo:

Currently, there is a record of `pending_compact_block` in `SyncState`, and a record of `pending_compact` in `InflightBlock`. The record in `InflightBlock` limits the `BlockTransaction` request to only two remote nodes at the same time. 

This is repetitive recording work, and because of the limitation, the construction work with n-2 data is completely wasted.

This change removes the internal records of `InflightBlock` and turns on the request limit of `BlockTransaction` .

The remaining question is, do we need to keep the limit on the number of requests? 

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -10,6 +10,7 @@ use ckb_types::{core, packed, prelude::*};
 use ckb_util::shrink_to_fit;
 use ckb_verification::{HeaderError, HeaderVerifier};
 use ckb_verification_traits::Verifier;
+use faketime::unix_time_as_millis;
 use std::collections::HashMap;
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


sync/src/relayer/compact_block_process.rs
```diff
@@ -122,6 +122,15 @@ impl<'a> CompactBlockProcess<'a> {
             .read_inflight_blocks()
             .inflight_compact_by_block(&block_hash)
         {
+            if peers.len() >= 2 {
+                debug_target!(
+                    crate::LOG_TARGET_RELAY,
+                    "BlockInFlight reach limit or had requested, peer: {}, block: {}",
```
yangby-cryptape, line 128: Bad grammar!

"reached or had" or "reach or have" or "reaches or has".

keroro520:
I think this change makes eclipse attacks more easily.

quake (approved):

zhangsoledad (approved):

quake:
bors r=quake,zhangsoledad

PR #3170, opened 2021-11-09
feat(rpc): add a method to remove a transaction from tx-pool

yangby-cryptape:

If there are some transactions in tx pool, but we want to delete them, at present, the only way is to clear the whole tx pool.

This PR add a new RPC to remove a transaction from tx pool via its hash.

### Check List

Tests

- Integration test

### Release note

```release-note
```


chanhsu001 (commented):
the RPC mainly serves "clear some transaction(big cycle transaction, etc.)" and "single node test"

yangby-cryptape:
> the RPC mainly serves "clear some transaction(big cycle transaction, etc.)" and "single node test"

This RPC only provides a new operation, how to use this RPC based on people's imagination.

There is no limit.

quake (approved):

tx-pool/src/pool.rs
```diff
@@ -342,6 +342,30 @@ impl TxPool {
...
+    pub(crate) fn remove_tx(&mut self, id: &ProposalShortId) -> bool {
+        let mut entries = self.proposed.remove_entry_and_descendants(id);
+        if !entries.is_empty() {
+            entries.sort_unstable_by_key(|entry| entry.ancestors_count);
+            for mut entry in entries {
+                entry.reset_ancestors_state();
+                self.update_statics_for_remove_tx(entry.size, entry.cycles);
+            }
+            return true;
+        }
```
zhangsoledad, line 354:
``` suggestion
        let entries = self.proposed.remove_entry_and_descendants(id);
        if !entries.is_empty() {
            for entry in entries {
                self.update_statics_for_remove_tx(entry.size, entry.cycles);
            }
            return true;
        }
```

zhangsoledad: sort and reset is useless for removing

zhangsoledad (approved):

zhangsoledad:
bors r= quake,zhangsoledad

yangby-cryptape:
bors r= quake,zhangsoledad

PR #3174, opened 2021-11-11
feat: delay txs during hardfork

zhangsoledad:

Proposal:  delay txs during hardfork

### Check List 

Tests 

- Integration test

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/process.rs
+++ b/tx-pool/src/process.rs
@@ -28,8 +28,8 @@ use ckb_types::{
             TransactionsChecker,
         },
         hardfork::HardForkSwitch,
-        BlockView, Capacity, Cycle, EpochExt, EpochNumber, HeaderView, ScriptHashType,
-        TransactionView, UncleBlockView, Version,
+        BlockView, Capacity, Cycle, EpochExt, HeaderView, ScriptHashType, TransactionView,
+        UncleBlockView, Version,
     },
     packed::{Byte32, Bytes, CellbaseWitness, OutPoint, ProposalShortId, Script},
     prelude::*,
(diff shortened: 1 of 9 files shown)


tx-pool/src/process.rs
```diff
@@ -1029,6 +1071,8 @@ impl TxPoolService {
             check_if_hardfork_during_blocks(&hardfork_switch, &detached_blocks);
         let hardfork_during_attach =
             check_if_hardfork_during_blocks(&hardfork_switch, &attached_blocks);
+
+        let new_tip_after_delay = after_delay_window(&snapshot);
```
yangby-cryptape, line 1075: I think it's OK, just mention it here.

The delayed transactions will not be processed as soon as it could.

Because this `snapshot` contains the old tip before reorg blocks.

So the delayed transactions will be processed in next reorg, not the turn when `new_tip.epoch_number == rfc_0032 && new_tip.block_index > proposal_window.farthest()`  is satisfied.

```
             hardfork-at
                 |
                 V      | no-delayed-txs
blocks ----(-----o-----)|------------|------------------------
           | delay-txs ||     N      | process-delayed-txs
               |   attached_blocks   |
                  
```

`N` is equal to or greater than `0`;  it depends on `attached_blocks`  which contains the first hardfork block, in `update_tx_pool_for_reorg`.

yangby-cryptape (approved):

yangby-cryptape (approved):

driftluo (approved):

zhangsoledad:
bors r= yangby-cryptape,driftluo

yangby-cryptape:
The test introduced since f5ef92c should be updated: sending the transaction should be more early.

zhangsoledad:
fixed and rebased @yangby-cryptape

yangby-cryptape (approved):

zhangsoledad:
bors r= yangby-cryptape,driftluo

PR #3175, opened 2021-11-11
feat: retain candidate uncle until next epoch

zhangsoledad:

Problem Summary:  Now, candidate uncles will be clear immediately once we checked it's invalid, but this is variable and temporary, we should keep candidate until next epoch.

Proposal: retain candidate uncle until next epoch.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```


driftluo (approved):

yangby-cryptape (approved):

zhangsoledad:
bors r=driftluo,yangby-cryptape

PR #3176, opened 2021-11-15
Fix: add dirty flag when load cell data as code

mohanson:
Solve the problem that 'load_data_cell_as_code' cannot work with 'ckb-vm chunk run'

ref: https://github.com/nervosnetwork/ckb-vm/pull/218


yangby-cryptape (approved):

quake (approved):

driftluo:
bors r=yangby-cryptape,quake

PR #3177, opened 2021-11-15
fix: upgrade ckb-vm to fix snapshot behavior

driftluo:

chunk run with the snapshot unable to support the binary replacement behavior of exec, which will cause abnormal behavior after resume.

The upgraded VM supports backing up the current code(executable pages)  to the snapshot to fix the above problems. https://github.com/nervosnetwork/ckb-vm/pull/218

and https://github.com/nervosnetwork/ckb/pull/3176 should keep its behavior until hardfork activated

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/script/src/syscalls/load_cell_data.rs
+++ b/script/src/syscalls/load_cell_data.rs
@@ -1,3 +1,5 @@
+use std::cell::RefCell;
+
 use crate::{
     cost_model::transferred_byte_cycles,
     syscalls::{
@@ -9,7 +11,7 @@ use crate::{
 use ckb_traits::CellDataProvider;
 use ckb_types::core::cell::CellMeta;
 use ckb_vm::{
-    memory::{Memory, FLAG_DIRTY, FLAG_EXECUTABLE, FLAG_FREEZED},
+    memory::{Memory, FLAG_EXECUTABLE, FLAG_FREEZED},
     registers::{A0, A1, A2, A3, A4, A5, A7},
     Error as VMError, Register, SupportMachine, Syscalls,
 };
@@ -21,6 +23,7 @@ pub struct LoadCellData<'a, DL> {
     resolved_cell_deps: &'a [CellMeta],
     group_inputs: &'a [usize],
     group_outputs: &'a [usize],
+    tracing_data_as_code_pages: &'a RefCell<Vec<(u64, u64)>>,
 }
 
 impl<'a, DL: CellDataProvider + 'a> LoadCellData<'a, DL> {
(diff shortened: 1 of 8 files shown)


script/src/syscalls/load_cell_data.rs
```diff
@@ -104,13 +107,15 @@ impl<'a, DL: CellDataProvider + 'a> LoadCellData<'a, DL> {
...
             .ok_or(VMError::Unexpected)?
             .slice((content_offset as usize)..(content_end as usize));
-        machine.memory_mut().init_pages(
-            addr,
-            memory_size,
-            FLAG_DIRTY | FLAG_EXECUTABLE | FLAG_FREEZED,
-            Some(data),
-            0,
-        )?;
+        // To be consistent with the mainnet, add this flag after hardfork
```
mohanson, line 110: I made a mistake before. Any write operation to the ckb-vm memory will be marked as dirty, and the correct logic should be:

```rs
machine.memory_mut().init_pages(
            addr,
            memory_size,
            FLAG_EXECUTABLE | FLAG_FREEZED,
            Some(data),
            0,
        )?;
        if !self.dirty_flag_enable {
            let mut current_addr = addr;
            while current_addr < addr + memory_size {
                let page = current_addr / RISCV_PAGESIZE as u64;
                machine.memory_mut().clear_flag(page, FLAG_DIRTY)?;
                current_addr += RISCV_PAGESIZE as u64;
            }
        }
```

mohanson (changes requested):

quake:
Suggest to add unit test to cover this PR.

yangby-cryptape:
I pushed #3178 to add more tests, I think that PR should be merged before this PR.

## Notice

#3178 doesn't fix the snapshot behavior.

## How to enable the test for resuming from snapshot?

This PR and my PR has conflicts, so resolved the conflicts at first.
- This PR add `true` for all `LoadCellData::new(..)` in tests, but I think test both `true` and `false` is better.
- Find `is_vm_version_1_and_syscalls_2_enabled` is a bit complicated, I think make a method is better than copied those code twice.

Then find the follow code in `script/src/verify/tests/ckb_latest/features_since_v2021.rs`:

```rust
    // TODO fix resume from snapshot
    let should_be_invalid_permission = script_version <= ScriptVersion::V1;
```

And change the above code as follow:

```diff
-    // TODO fix resume from snapshot
-    let should_be_invalid_permission = script_version <= ScriptVersion::V1;
+    let should_be_invalid_permission = script_version < ScriptVersion::V1;
```

Make sure all tests are passed at last.

yangby-cryptape (approved):

mohanson (approved):

zhangsoledad (approved):

zhangsoledad:
bors r= yangby-cryptape,mohanson,zhangsoledad

PR #3178, opened 2021-11-16
test: add more tests to ckb-script for resuming from snapshot

yangby-cryptape:

Currently, there is no test for resuming from snapshot.

Before we fix the snapshot feature, we need more tests to ensure the fix is correct.

### Check List

Tests

- Unit test

Side effects

- I left a TODO because this PR doesn't fix the snapshot feature.

### Release note

```release-note
```


script/src/verify/tests/ckb_latest/features_since_v2021.rs
```diff
@@ -685,3 +686,116 @@ fn check_typical_secp256k1_blake160_2_in_2_out_tx_with_complete() {
...
+            resolved_dep_groups: vec![],
+        }
+    };
+
+    let mut cycles = 0;
+    let cycles_step_1 = 100_000;
+    let max_cycles = Cycle::MAX;
+    let verifier = TransactionScriptsVerifierWithEnv::new();
+    // TODO fix resume from snapshot
+    let should_be_invalid_permission = script_version <= ScriptVersion::V1;
```
yangby-cryptape, line 767: Change to the follow code after fixing the snapshot:

```suggestion
    let should_be_invalid_permission = script_version < ScriptVersion::V1;
```

script/src/types.rs
```diff
@@ -92,6 +92,14 @@ impl ScriptVersion {
         let version = self.vm_version();
         CoreMachineType::new(isa, version, max_cycles)
     }
+
+    #[cfg(test)]
+    pub(crate) fn use_dirty_flag_when_load_data_as_code(self) -> bool {
```
yangby-cryptape, line 97: This is only for test.

We should run each tests in 3 scenes, but current test framework only supports 2.
- (Tested) `vm0`, `dirty-flag=false`
- (Skipped) `vm0`, `dirty-flag=true`
- (Tested)`vm1`, `dirty-flag=true`

This method use the same logic as the follow code (no `vm0` test after hardfork):
https://github.com/nervosnetwork/ckb/blob/3f405a0f5dc4ccbfb294d202ae9f4396691ca7b5/script/src/verify/tests/utils.rs#L152-L153

yangby-cryptape:
The last commit via force-push is used to remove the conflicts with #3177.

driftluo (approved):

zhangsoledad (approved):

yangby-cryptape:
bors merge=quake,driftluo,zhangsoledad

mohanson (approved):

PR #3187, opened 2021-11-17
test(script): add more complicated scripts to unit tests

yangby-cryptape:

Add more complicated scripts to test the crate `ckb-script`.

#### Changes

**Tip: It's better to review this PR by reviewing each commits.**

- Refactor tests.

  - test(script): convert duplicate codes into functions to simplify them (7d060aa)

    At present, to add one test has to copy a lot of duplicate codes, so do refactor first.

  - test(script): add a Makefile to make almost all test scripts to be reproducible (883b3d9)

    Before this commit, only the latest 3 scripts could be re-built to same binaries from source files.

    I add a `Makefile` and a `README.md` to record the compile environments for almost all test scripts.
    So we can re-build these scripts more easily.

    **TODO**: Still left 4 binaries couldn't be built from source files.

- Add new tests.

  - test(script): current cycles should be always monotonically increased (794c539)

  - test(script): resume from snapshot with loading codes many times (d987413)

  - test(script): vm_version should be correct after resuming from snapshot (bd24f17)

  - test(script): current cycles should be correct after resuming from snapshot (5fc8bed)

    Find a bug which resolved in #3188.

  - test(script): to load code into global is not allowed in vm0 (d13e1f5)

    Ref:
    - nervosnetwork/ckb-vm@189d097
    - nervosnetwork/ckb-vm#142

  - test(script): hint instructions are not allowed in vm0 (b244812)

    Ref:
    - nervosnetwork/ckb-vm@5ace65b
    - nervosnetwork/ckb-vm#188

  - test(script): flags of stack memory are not reset after they freed (a0c8927)

    This feature looks like a bug but not.

    If any function initializes a piece of stack memory, after the lifetime of that function, the stack memory will be freed.
    But the flag of that piece of stack memory will not reset.

    So, if we call a function which load code into a piece of stack memory, then we call another function, it could raise an error `InvalidPermission` if the second function tries to write the previous piece of stack memory.

    Ref: [`W^X`](https://en.wikipedia.org/wiki/W%5EX)

### Check List

Tests

- Unit test

### Release note

```release-note
```


driftluo (approved):

zhangsoledad (approved):

zhangsoledad:
bors r= driftluo, zhangsoledad

PR #3188, opened 2021-11-18
fix: fix current cycles syscall on chunk run with snapshot

driftluo:

The script group resume from the snapshot will lose the cycles consumed by the current execution, causing the value obtained by the current cycle syscall call to reset, causing potential inconsistencies.

This pr will fix the problem and only affects testnet  

### Check List

Tests

- Unit test

### Release note

```release-note
```

### DIFF
--- a/script/src/verify.rs
+++ b/script/src/verify.rs
@@ -556,6 +556,7 @@ impl<'a, DL: CellDataProvider + HeaderProvider> TransactionScriptsVerifier<'a, D
         snap: &TransactionSnapshot,
         limit_cycles: Cycle,
     ) -> Result<VerifyResult, Error> {
+        let current_group_used = snap.snap.as_ref().map(|s| s.1).unwrap_or_default();
         let mut cycles = snap.current_cycles;
         let mut current_used = 0;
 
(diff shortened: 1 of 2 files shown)


yangby-cryptape:
This BUG was found via the tests in #3187; see 4f12d9f for details.

yangby-cryptape (approved):
I tested this PR: this PR has passed all unit tests, included all unmerged tests.

script/src/verify.rs
```diff
@@ -569,7 +570,11 @@ impl<'a, DL: CellDataProvider + HeaderProvider> TransactionScriptsVerifier<'a, D
         // continue snapshot current script
         match self.verify_group_with_chunk(&current_group, limit_cycles, &snap.snap) {
             Ok(ChunkState::Completed(used_cycles)) => {
-                current_used = wrapping_cycles_add(current_used, used_cycles, &current_group)?;
+                current_used = wrapping_cycles_add(
+                    current_used,
+                    used_cycles - current_group_used,
```
yangby-cryptape, line 575: How about add a test to check this statement is overflow?

mohanson (approved):
LGTM

chanhsu001 (approved):

driftluo:
bors r=yangby-cryptape,mohanson,chanhsu001

PR #3195, opened 2021-11-25
chore: remove unnecessary timer

driftluo:

After more than two years of operation, I am sure that the stability of Tentacle’s notify can be guaranteed. This PR removes this unnecessary timer 

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/protocols/discovery/mod.rs
+++ b/network/src/protocols/discovery/mod.rs
@@ -146,7 +146,7 @@ impl<M: AddressManager> ServiceProtocol for DiscoveryProtocol<M> {
                                 state.remote_addr.change_to_listen();
                             }
 
-                            let max = ::std::cmp::max(MAX_ADDR_TO_SEND, count as usize);
+                            let max = ::std::cmp::min(MAX_ADDR_TO_SEND, count as usize);
                             if items.len() > max {
                                 items = items
                                     .choose_multiple(&mut rand::thread_rng(), max)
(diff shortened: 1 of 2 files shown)


yangby-cryptape (approved):

quake (approved):

driftluo:
bors r=yangby-cryptape,quake

chanhsu001:
There might be an issue with if condition, correct me if i'm wrong.

in network/src/protocols/discovery/mod.rs:
.......................
let mut items = self.addr_mgr.get_random(2500);

_let's assume we have enough remote multiaddr for self.addr_mgr to fetch, so items.len() == 2500_

_// MAX_ADDR_TO_SEND is 1000,  and "count" is number of  nodes requested in GetNodes Message.
 //guess here "max" change to "min" , would be better_ 
let max = ::std::cmp::min(MAX_ADDR_TO_SEND, count as usize);    
if items.len() > max {
   .....
   ....
}

if based on previous assumption, items.len() surely bigger than max.....

ok, if we have not that much remote multiaddr, we just return what we got, the number of multiadr is:
min(items.len(), min(MAX_ADDR_TO_SEND, count)),  we could know early and use random_fetch once.

driftluo:
> There might be an issue with if condition, correct me if i'm wrong.

This implementation is deliberate, first randomly 2500, and then reduced according to the real request

PR #2917, opened 2021-08-09
backport fix: try traverse all unknown parent hash

driftluo:
close #2916

PR #1556, opened 2019-09-06
fix: Data race between orphan_pool and storage

keroro520:
backport of https://github.com/nervosnetwork/ckb/pull/1554

PR #3227, opened 2021-12-12
chore: ignore received compact block

driftluo:

if block status is RECEIVED, it means block received and remain on orphan pool, the same block received afterwards does not need to be repeatedly inserted into the orphan pool 

### Check List 

Tests

- Unit test
- Integration test

### Release note

```release-note
```


quake (approved):

keroro520 (changes requested):
Please add tests to enforce this change.

keroro520 (approved):
This change reminds me of https://github.com/nervosnetwork/ckb/pull/1556. At that time, orphan_block_pool might maintain dirty blocks if there was a data race between Relay and Sync.


While https://github.com/nervosnetwork/ckb/pull/2917 fixed the orphan_block_pool issue. I think it is safe to apply this change now.

driftluo:
bors r=quake,keroro520

PR #3228, opened 2021-12-12
fix: fix block transaction process dead loop

driftluo:

fix block transaction process dead loop

Although the sync protocol will force such a dead loop to be broken, as a single protocol, this is a must-fix

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


quake (approved):

quake (approved):

zhangsoledad (approved):

driftluo:
bors r=quake,zhangsoledad

PR #3231, opened 2021-12-13
feat: observe listen port occupancy

driftluo:

Since the port reuse function of ckb is enabled by default on Linux, starting two ckb nodes will not report an error explicitly, after this PR will automatically exit the second ckb node 

### Check List

Tests

Manual test:
1. start 2 ckb nodes with the same net port
2. the second one will stop with an error
```
2021-12-13 17:21:31.585 +08:00 main INFO ckb_bin::helper  raise_fd_limit newly-increased limit: 524288                                 
2021-12-13 17:21:34.235 +08:00 main INFO ckb_bin::subcommand::run  ckb version: 0.102.0-pre (1595009-dirty 2021-12-13)                 
2021-12-13 17:21:34.236 +08:00 main INFO ckb_launcher  Miner is disabled, edit ckb.toml to enable it                                   
IO Error: Os { code: 98, kind: AddrInUse, message: "Address already in use" }                                                          
2021-12-13 17:21:34.236 +08:00 main INFO ckb_stop_handler  StopHandler(GT) send signal
```

### Release note 

```release-note
```

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -49,3 +49,39 @@ pub use tokio;
 
 /// Protocol version used by network protocol open
 pub type ProtocolVersion = String;
+
+/// Observe listen port occupancy
+pub async fn observe_listen_port_occupancy(
+    _addrs: &[multiaddr::MultiAddr],
+) -> Result<(), std::io::Error> {
+    #[cfg(target_os = "linux")]
+    {
+        use p2p::utils::dns::DnsResolver;
+        use std::net::{SocketAddr, TcpListener};
+
+        for raw_addr in _addrs {
+            let ip_addr: Option<SocketAddr> = match DnsResolver::new(raw_addr.clone()) {
+                Some(dns) => dns
+                    .await
+                    .ok()
+                    .as_ref()
+                    .map(multiaddr_to_socketaddr)
+                    .flatten(),
+                None => multiaddr_to_socketaddr(raw_addr),
+            };
+
+            if let Some(addr) = ip_addr {
+                if let Err(e) = TcpListener::bind(addr) {
+                    ckb_logger::error!(
+                        "addr {} can't use on your machines by error: {}, please check",
+                        raw_addr,
+                        e
+                    );
+                    return Err(e);
+                }
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/util/launcher/src/lib.rs
+++ b/util/launcher/src/lib.rs
@@ -19,8 +19,8 @@ use ckb_channel::Receiver;
 use ckb_jsonrpc_types::ScriptHashType;
 use ckb_logger::info;
 use ckb_network::{
-    CKBProtocol, DefaultExitHandler, NetworkController, NetworkService, NetworkState,
-    SupportProtocols,
+    observe_listen_port_occupancy, CKBProtocol, DefaultExitHandler, NetworkController,
+    NetworkService, NetworkState, SupportProtocols,
 };
 use ckb_network_alert::alert_relayer::AlertRelayer;
 use ckb_proposal_table::ProposalTable;
@@ -190,6 +190,10 @@ impl Launcher {
         &self,
         block_assembler_config: Option<BlockAssemblerConfig>,
     ) -> Result<(Shared, SharedPackage), ExitCode> {
+        self.async_handle.block_on(observe_listen_port_occupancy(
+            &self.args.config.network.listen_addresses,
+        ))?;
+
         let shared_builder = SharedBuilder::new(
             &self.args.config.bin_name,
             self.args.config.root_dir.as_path(),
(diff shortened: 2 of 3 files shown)


util/launcher/src/lib.rs
```diff
@@ -190,6 +190,8 @@ impl Launcher {
         &self,
         block_assembler_config: Option<BlockAssemblerConfig>,
     ) -> Result<(Shared, SharedPackage), ExitCode> {
+        observe_listen_port_occupancy(&self.args.config.network.listen_addresses)?;
```
keroro520, line 193: It should be done in [`NetworkService::new`](https://github.com/nervosnetwork/ckb/blob/78379cdb0a5476534c1dd701845acc08d4036b82/network/src/network.rs#L759-L769). It validates configurations inside.

keroro520 (changes requested):

network/src/lib.rs
```diff
@@ -49,3 +49,35 @@ pub use tokio;
...
+        for addr in _addrs {
+            let mut iter = addr.iter().peekable();
+
+            let mut ip_addr: Option<SocketAddr> = None;
+
+            while iter.peek().is_some() {
+                let proto = iter.next().unwrap();
+
+                if let Protocol::Tcp(port) = proto {
+                    ip_addr = Some(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port));
```
quake, line 71: I think we should match `Protocol::Ip4` and use the matched ip address / port to build the `SocketAddr`

driftluo: No need to do this because after any port of this machine is bound, this port cannot be bound on any IP combination, whether it is the loopback address, the network card address, the all-zero address

quake: user may use multiple NICs in vm and multiple ckb process listen on the different ip address with same port, using `0.0.0.0` to check port availability will breaks this use case.

driftluo: You are right, it has been modified

network/src/lib.rs
```diff
@@ -49,3 +49,76 @@ pub use tokio;
...
+
+                let proto1 = iter.next().unwrap();
+                let proto2 = if iter.peek().is_some() {
+                    iter.next().unwrap()
+                } else {
+                    break;
+                };
+
+                match (proto1, proto2) {
+                    (Protocol::Ip4(ip), Protocol::Tcp(port)) => {
```
quake, line 88: suggest to use `multiaddr_to_socketaddr` fn to extract the SocketAddr

driftluo: > suggest to use `multiaddr_to_socketaddr` fn to extract the SocketAddr

`multiaddr_to_socketaddr` cannot handle DNS

quake: is it possible to use `TcpTransport::new(timeout, None).listen(multiaddr)` directly in the network mod?

driftluo: `transports` is a private module, and it needs async runtime support，It is not currently available, but it seems to be possible to use DnsResolver directly

driftluo: resolve by https://github.com/nervosnetwork/ckb/pull/3231/commits/cbbb33b56a326be6c008e66529469216b767ae86

quake (approved):

chanhsu001:
sorry to highlight the result I've test:
2 ckb instances(different directories), one is A, using "**eth0_ip**:8114" in ckb.toml;  the other is B, using "**eth1_ip**:8114" in ckb.toml
but A and B cannot run at same time, error is "Address already in use"...
it seems DnsResolver not working well in this case.

chanhsu001:
I checked with 2 testcases:
pc with 2 network interface: eth0 and eth1, with different IP address.

1. ckb instance A with ckb.toml,  network listen_address=["/ipv4/eth0_ip/tcp/8115"]
    ckb instance B with ckb.toml,  network listen_address=["/ipv4/eth1_ip/tcp/8115"]
    both ckb instance start and work, without conflict.

2.  ckb instance A with ckb.toml,  network listen_address=["/ipv4/eth0_ip/tcp/8115"]
    ckb instance B with ckb.toml,  network listen_address=["/ipv4/0.0.0.0/tcp/8115"]
    only one instance could start,  the other will report "Address already in use" error and exit

It conform to the design.  
Any more running CKB instances will have to binding the explicit different IP address in their ckb.toml , network part.

network/src/lib.rs
```diff
@@ -49,3 +49,32 @@ pub use tokio;
...
+                    .await
+                    .ok()
+                    .as_ref()
+                    .map(multiaddr_to_socketaddr)
+                    .flatten(),
+                None => multiaddr_to_socketaddr(addr),
+            };
+
+            if let Some(addr) = ip_addr {
+                let _check = TcpListener::bind(addr)?;
```
doitian, line 74: It's better to show a more user friendly error message to users.

doitian (approved):

driftluo:
bors r=quake,doitian

PR #3242, opened 2021-12-15
fix: verification cache reset during hardfork

zhangsoledad:

Chain verification should reset cache before hardfork, and  make sure cache no dirty write.

Make sure tx-pool don't access verification cache during hardfork.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -187,7 +187,7 @@ impl ForkChanges {
         if blocks.is_empty() {
             false
         } else {
-            // This method assumes that the hardfork epochs are sorted and unique.
+            // This method assumes that the input blocks are sorted and unique.
             let hardfork_epochs = hardfork_switch.script_result_changed_at();
             if hardfork_epochs.is_empty() {
                 false
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -17,6 +17,7 @@ use ckb_types::{
             resolve_transaction_with_options, BlockCellProvider, OverlayCellProvider,
             ResolveOptions, ResolvedTransaction,
         },
+        hardfork::HardForkSwitch,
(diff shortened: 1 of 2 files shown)


yangby-cryptape (approved):

chain/src/chain.rs
```diff
@@ -168,6 +169,41 @@ impl ForkChanges {
...
+
+    fn check_if_hardfork_during_blocks(
+        &self,
+        hardfork_switch: &HardForkSwitch,
+        blocks: &VecDeque<BlockView>,
+    ) -> bool {
+        if blocks.is_empty() {
+            false
+        } else {
+            // This method assumes that the hardfork epochs are sorted and unique.
```
yangby-cryptape, line 190: This comment is not right.

It should be:
```suggestion
            // This method assumes that the input blocks are sorted and unique.
```

Focus on the follow code:
```
                let epoch_first = blocks.front().unwrap().epoch().number();
                let epoch_next = blocks
                    .back()
                    .unwrap()
                    .epoch()
                    .minimum_epoch_number_after_n_blocks(1);
                hardfork_epochs.into_iter().any(|hardfork_epoch| {
                    epoch_first < hardfork_epoch && hardfork_epoch <= epoch_next
                })
```

Suppose:
- `hardfork_epochs = [1]`.
- The epoch before `block-15` is `0`; and the epoch of `block-15` is `1`.
- The input blocks are not sorted, they are `[block-12, block-17, block-14]`.

We got:
- The epochs of input blocks are `[epoch-0, epoch-1, epoch-0]`.
- `epoch_first = 0` and `epoch_next = 0`.

Then this method will return `false`.

driftluo (approved):

zhangsoledad:
bors r= driftluo,yangby-cryptape

PR #3255, opened 2021-12-27
ci: Secrets cannot be used to condition job runs

liya2017:

Sync-mainnet test failed since secrets cannot be used to condition job runs

### Check List 

Tests 

### Release note 

```release-note
```

### DIFF
--- a/.github/workflows/sync-mainnet.yml
+++ b/.github/workflows/sync-mainnet.yml
@@ -16,11 +16,6 @@ jobs:
       AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
       AWS_SECRET_KEY: ${{ secrets.AWS_SECRET_KEY }}
       GITHUB_TOKEN: ${{ secrets.REPO_ACCESS_TOKEN }}
-      PGHOST: ${{ secrets.PGHOST }}
-      PGPORT: ${{ secrets.PGPORT }}
-      PGUSER: ${{ secrets.PGUSER }}
-      PGPASSWORD: ${{ secrets.PGPASSWORD }}
-      PGDATABASE: ckbtest
       GITHUB_BRANCH: ${{ github.ref_name }}
       GITHUB_REPOSITORY: ${{ github.repository }}
       GITHUB_EVENT_NAME: ${{ github.event_name }}
@@ -47,10 +42,16 @@ jobs:
         if: ${{ failure() }}
         run: echo "GITHUB_RUN_STATE=1" >> $GITHUB_ENV
       - name: insert report to postgres
-        if: ${{ always() && secrets.PGHOST != "" }}
+        if: ${{ always() }}
         env:
           JOB_ID: "sync-mainnet-${{ github.repository }}-${{ github.ref_name }}-${{ steps.date.outputs.date }}-in-10h"
-        run: ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
+          PGHOST: ${{ secrets.PGHOST }}
+          PGPORT: ${{ secrets.PGPORT }}
+          PGUSER: ${{ secrets.PGUSER }}
+          PGPASSWORD: ${{ secrets.PGPASSWORD }}
+          PGDATABASE: ckbtest
+        run: |
+            [ -z "${{ secrets.PGHOST }}" ] || ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
       - name: Clean Up
         if: ${{ always() }}
         env:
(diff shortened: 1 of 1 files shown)


keroro520 (approved):

yangby-cryptape (changes requested):
Have you tried any solutions which are in comments of actions/runner#520?

I haven't tried any of those solutions, but I think at least one of them should work well, since they have several thumbs up emoji. 

Put passwords into environment variables, just because secrets couldn't be used in condition expression.
Reduced security for convenience.
I don't think that's right.

yangby-cryptape (approved):

yangby-cryptape:
I still think it's strange:
- We copied almost all secrets into environments at top, so why we don't just put them into environment variables directly?
- How to distinguish the minimum requirements for each steps?

liya2017:
> Have you tried any solutions which are in comments of [actions/runner#520](https://github.com/actions/runner/issues/520)?
> 
> I haven't tried any of those solutions, but I think at least one of them should work well, since they have several thumbs up emoji.
> 
Yes, tried it.

yangby-cryptape:
> Yes, tried it.

Since the secret is already copied to environments, I think the solution in this PR is fine.

But, in fact, there still have many ways to avoid coping secrets into environments:

- I think this should work: https://github.com/actions/runner/issues/520#issuecomment-907770967.

- Another solution:

  ```diff
          - name: insert report to postgres
  -         if: ${{ always() && secrets.PGHOST != "" }}
            env:
              JOB_ID: "sync-mainnet-${{ github.repository }}-${{ github.ref_name }}-${{ steps.date.outputs.date }}-in-10h"
  -         run: ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
  +         run: [ -z "{ secrets.PGHOST }" ] || ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
  ```

  It doesn't cost too much since this is just a step, the host machine have already run.

liya2017:
> I still think it's strange:
> 
> * We copied almost all secrets into environments at top, so why we don't just put them into environment variables directly?

Those environments are used in another script in repo ckb-integration-test, include DB password and DB user. Do you mean just put host info  into environment variables directly?

> * How to distinguish the minimum requirements for each steps?
insert report to postgres should be always, don't do this step if PGHOST is null  based on @keroro520  comments, and I think it's right.

liya2017:
> * Another solution:
>   ```diff
>           - name: insert report to postgres
>   -         if: ${{ always() && secrets.PGHOST != "" }}
>             env:
>               JOB_ID: "sync-mainnet-${{ github.repository }}-${{ github.ref_name }}-${{ steps.date.outputs.date }}-in-10h"
>   -         run: ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
>   +         run: [ -z "{ secrets.PGHOST }" ] || ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
>   ```

This way is better, let me have a try.

.github/workflows/sync-mainnet.yml
```diff
@@ -47,10 +42,16 @@ jobs:
...
         env:
           JOB_ID: "sync-mainnet-${{ github.repository }}-${{ github.ref_name }}-${{ steps.date.outputs.date }}-in-10h"
-        run: ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
+          PGHOST: ${{ secrets.PGHOST }}
+          PGPORT: ${{ secrets.PGPORT }}
+          PGUSER: ${{ secrets.PGUSER }}
+          PGPASSWORD: ${{ secrets.PGPASSWORD }}
+          PGDATABASE: ckbtest
+        run: |
+            [ -z "${{ secrets.PGHOST }}" ] || ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
```
yangby-cryptape, line 54: Since `env: PGHOST: ${{ secrets.PGHOST }}`, the follow code is enough:

```suggestion
            [ -z "${PGHOST}" ] || ./ckb-sync-mainnet/script/sync-mainnet.sh insert_report_to_postgres
```

yangby-cryptape (approved):

.github/workflows/sync-mainnet.yml
liya2017: can not using ${PGHOST} directly in yaml file , just can change `${{ secrets.PGHOST }}` to `${{ env.PGHOST }}`

yangby-cryptape: [GitHub Docs: Example using Bash](https://docs.github.com/en/actions/security-guides/encrypted-secrets#example-using-bash):

```yaml
steps:
  - shell: bash
    env:
      SUPER_SECRET: ${{ secrets.SuperSecret }}
    run: |
      example-command "$SUPER_SECRET"
```

yangby-cryptape (approved):

liya2017:

PR #3261, opened 2021-12-29
feat: tuning the locator algorithm

driftluo:

The locator algorithm is designed to send a sample of local chains to the remote end, so that both sides can determine the same header of the maximum height of both sides as efficiently as possible, thus completing the header synchronization work faster

As time goes on, the current locator algorithm becomes more and more inapplicable as the chain height increases, and we need a locator algorithm that constantly adjusts itself to the height of the chain. It allows the entire chain to be sampled as much as possible during the sampling process, instead of the lowest sampling point being pulled up step by step as the height increases

This PR relaxes the length of the locator to 52

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -1959,6 +1959,7 @@ impl ActiveChain {
         let mut locator = Vec::with_capacity(32);
         let mut index = start.number();
         let mut base = start.hash();
+
         loop {
             let header_hash = self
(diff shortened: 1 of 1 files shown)


quake (approved):

zhangsoledad (approved):

zhangsoledad:
bors r= quake,zhangsoledad

PR #3284, opened 2022-01-25
ci: add ci monitoring

liya2017:

Add ci monirtoring

### Check List 

Tests 

### Release note 

```release-note
```


doitian:
It's not an elegant solution. I suggest rework on this using the workflow callback.

PR #3307, opened 2022-02-16
test: open IntegrationTestRpc and add comment

chanhsu001:

Open IntegrationTestRpc in rpc/module/test, add rpc comment document.

- open 6 functions,  add check Dummy PoW chain in integration test (not add separate check in each function) 
- add encoding_rs license.clarify in deny.toml( reqwest crate refer to encoding_rs)

### Check List

Tests 

### Release note 

```release-note
None: Exclude this PR from release note.
```

### DIFF
--- a/rpc/src/module/test.rs
+++ b/rpc/src/module/test.rs
@@ -28,15 +28,157 @@ use jsonrpc_derive::rpc;
 use std::collections::HashSet;
 use std::sync::Arc;
 
+/// RPC for Integration Test.
 #[rpc(server)]
-#[doc(hidden)]
 pub trait IntegrationTestRpc {
+    /// process block without any block verification.
+    ///
+    /// ## Params
+    ///
+    /// * `data` - block data(in binary).
+    ///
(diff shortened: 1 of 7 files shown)


rpc/src/module/test.rs
```diff
@@ -62,6 +471,9 @@ pub(crate) struct IntegrationTestRpcImpl {
 
 impl IntegrationTestRpc for IntegrationTestRpcImpl {
     fn process_block_without_verify(&self, data: Block, broadcast: bool) -> Result<Option<H256>> {
+        // check if on dev network
+        self.check_dev_chain()?;
```
zhangsoledad, line 475: These checks should only need to be done once at rpc server build time, not on every request.  (https://github.com/nervosnetwork/ckb/blob/develop/rpc/src/service_builder.rs#L145-L164)

zhangsoledad (changes requested):

zhangsoledad (approved):

PR #3319, opened 2022-02-25
fix(ckb-bin): fix replay command crash if profile time less than one second

chanhsu001:
fix subcommand replay crash due to divide profile_time(zero) if the time is less than one second

Problem Summary:
- fix subcommand replay crash due to divide profile_time(will be zero if the time is less than one second)
- set minimum profiling time to 5 seconds, if profiling time is less than 5 seconds, show prompt message to suggest increase block range, make profiling time meets minimum time.  
- minor change to output style.

What's Changed:
- fix crash
- add minimum profiling time limit
- minor output style refactor

### Related changes

- PR to update `owner/repo`:

Tests 

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/ckb-bin/src/subcommand/replay.rs
+++ b/ckb-bin/src/subcommand/replay.rs
@@ -9,6 +9,8 @@ use ckb_store::ChainStore;
 use ckb_verification_traits::Switch;
 use std::sync::Arc;
 
+const MIN_PROFILING_TIME: u64 = 5;
+
 pub fn replay(args: ReplayArgs, async_handle: Handle) -> Result<(), ExitCode> {
     let shared_builder = SharedBuilder::new(
         &args.config.bin_name,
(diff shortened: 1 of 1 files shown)


quake:
Short time profiling is inaccurate and non-referential, it is better to output logs and suggest users to increase the number of blocks, for example, output suggestion logs instead of tps when the duration is less than 5 seconds.

chanhsu001:
> Short time profiling is inaccurate and non-referential, it is better to output logs and suggest users to increase the number of blocks, for example, output suggestion logs instead of tps when the duration is less than 5 seconds.

Agreed.   
which one do you prefer?  if time is less than 5 seconds:
- "only suggestion(not show tps)“
- "tps and suggestion"

quake (approved):

zhangsoledad (approved):

doitian:
bors r=quake,zhangsoledad

PR #3326, opened 2022-03-02
chore: protect node in ibd should also disconnnected

driftluo:

1. `suspend_sync` has determined if it is in sync start state
2. Errors sent behavior in outbound should be disconnected unless they are whitelisted
3. The protect node is randomly generated and should be disconnected if it is in ibd state, otherwise the sync data source is inherently missing

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


doitian (approved):

quake (approved):

quake:
bors r=quake,doitian

PR #3330, opened 2022-03-07
perf(replay): disable info output, improve replay speed

chanhsu001:
raise default log level(Info) to Error when ckb replay profile, it'll reduce lots of messages output to
logfile and improve ckb replay profile speed, especially when replay with huge block range.

Problem Summary:

when ckb replay profile from 1 to tip(current 6 million height), disable normal Info output will reduce at least 10% time when replay takes; it also benefit after we add `ckb replay`  to daily SyncMainnet  procedure.
the code only temporarily dynamic raise logger level to Error when ckb replay run, that will not impact default ckb configuration.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

### Release note 

```release-note
```

### DIFF
--- a/ckb-bin/src/subcommand/replay.rs
+++ b/ckb-bin/src/subcommand/replay.rs
@@ -4,6 +4,7 @@ use ckb_chain::chain::ChainService;
 use ckb_chain_iter::ChainIterator;
 use ckb_instrument::{ProgressBar, ProgressStyle};
 use ckb_launcher::SharedBuilder;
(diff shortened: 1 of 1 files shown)


quake:
the logger filter can be override by env `CKB_LOG`

```
CKB_LOG=error ./ckb replay
```

chanhsu001:
@quake, Thanks. no need to change code, i withdraw this PR.

PR #3342, opened 2022-03-17
chore: add some logs

driftluo:

1. Add logging to abnormal disconnects
2. Align the highest version number of all protocols with a constant

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/protocols/support_protocols.rs
+++ b/network/src/protocols/support_protocols.rs
@@ -6,6 +6,8 @@ use p2p::{
 };
 use tokio_util::codec::length_delimited;
 
+const LASTEST_VERSION: &str = "2";
+
 /// All supported protocols
 ///
 /// The underlying network of CKB is flexible and complex. The flexibility lies in that it can support any number of protocols.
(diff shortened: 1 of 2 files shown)


doitian:
> Make sure the support version list is sorted with the code

Why?

network/src/protocols/support_protocols.rs
```diff
@@ -91,18 +93,23 @@ impl SupportProtocols {
...
-            SupportProtocols::Time => vec!["1".to_owned(), "2".to_owned()],
-            SupportProtocols::Alert => vec!["1".to_owned(), "2".to_owned()],
-            SupportProtocols::RelayV2 => vec!["2".to_owned()],
-        }
+            SupportProtocols::Time => vec!["1".to_owned(), LASTEST_VERSION.to_owned()],
+            SupportProtocols::Alert => vec!["1".to_owned(), LASTEST_VERSION.to_owned()],
+            SupportProtocols::RelayV2 => vec![LASTEST_VERSION.to_owned()],
+        };
+        // make sure version list is sorted
+        versions.sort();
```
quake, line 111: we should not use `sort` here since it's string vec:
```
["0.0.1", "2", "10"].sort => ["0.0.1", "10", "2"]
```

quake (changes requested):

network/src/protocols/support_protocols.rs
driftluo: I'll change it to add a comment, the order here will affect the handshake behavior

quake (approved):

zhangsoledad (approved):

quake:
bors r=quake,zhangsoledad

PR #3345, opened 2022-03-21
fix(ckb cli) add h256 validator for --assume-valid-target and ba-code-hash

chanhsu001:
fix(ckb cli) add h256 validator for --assume-valid-target and ba-code-hash

- add H256 validator for `ckb init --ba-code-hash` and `ckb run --assume-valid-target` replace naive is_hex check.
- add related 1 testcase

now cli report error at parsing phase, prompt invalid h256 message, instead of running error.

### Related changes

- PR to update `owner/repo`:

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/util/app-config/src/cli.rs
+++ b/util/app-config/src/cli.rs
@@ -180,7 +180,7 @@ fn run<'help>() -> Command<'help> {
         Arg::new(ARG_ASSUME_VALID_TARGET)
             .long(ARG_ASSUME_VALID_TARGET)
             .takes_value(true)
-            .validator(is_hex)
+            .validator(is_h256)
             .help("This parameter specifies the hash of a block. \
             When the height does not reach this block's height, the execution of the script will be disabled, \
             that is, skip verifying the script content. \
@@ -417,7 +417,7 @@ fn init<'help>() -> Command<'help> {
             Arg::new(ARG_BA_CODE_HASH)
                 .long(ARG_BA_CODE_HASH)
                 .value_name("code_hash")
-                .validator(is_hex)
+                .validator(is_h256)
                 .takes_value(true)
                 .help(
                     "Sets code_hash in [block_assembler] \
@@ -524,5 +524,13 @@ fn is_hex(hex: &str) -> Result<(), String> {
         Ok(())
     } else {
         Err("Must 0x-prefixed hexadecimal string".to_string())
+    }
+}
+
+fn is_h256(hex: &str) -> Result<(), String> {
+    if hex.len() != 66 {
+        Err("Must be 0x-prefixed hexadecimal string and string length is 66".to_owned())
+    } else {
+        is_hex(hex)
     }
 }
--- a/util/app-config/src/tests/cli.rs
+++ b/util/app-config/src/tests/cli.rs
@@ -37,7 +37,7 @@ fn ba_message_requires_ba_arg_or_ba_code_hash() {
         "--ba-arg",
         "0x00",
     ]);
-    let ok_ba_code_hash = basic_app().try_get_matches_from(&[
+    let ba_code_hash = basic_app().try_get_matches_from(&[
         BIN_NAME,
         "init",
         "--ba-message",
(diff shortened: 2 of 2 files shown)


util/app-config/src/cli.rs
```diff
@@ -526,3 +527,18 @@ fn is_hex(hex: &str) -> Result<(), String> {
...
+        if let Err(e) = faster_hex::hex_decode(&hex.as_bytes()[2..], &mut hd) {
+            return Err(e.to_string());
+        };
+        match H256::from_slice(&hd) {
+            Ok(_) => Ok(()),
+            _ => Err("input string is not valid H256 format".to_owned()),
+        }
+    } else {
+        Err("Must be a 0x-prefixed hexadecimal string".to_owned())
+    }
```
quake, line 543:
```suggestion
    is_hex(hex) && hex.len == 34
```

quake (changes requested):

util/app-config/src/cli.rs
chanhsu001: it seems it' not so good to check byte array length match 34 to decide h256 string.
here is a little issue:  in mainnet, height [6749464](https://explorer.nervos.org/block/6749464), block hash is `0x00d1b86f6824d33a91b72ec20e2118cf7788a5ffff656bd1ea1ea638c764cb5f`
we use `is_hex(hex) && hex.len == 34`, it works and works good.

but if we input  `0xd1b86f6824d33a91b72ec20e2118cf7788a5ffff656bd1ea1ea638c764cb5f`, the hex.len == 34 will fail, but it's hard to say the `0xd1b86f6824d33a91b72ec20e2118cf7788a5ffff656bd1ea1ea638c764cb5f` is incorrect.
actually, i did think about it for a while, and not sure about it.

quake: let's keep it simple stupid, user won’t copy a hex string and delete leading or trailing zeros manually

util/app-config/src/cli.rs
```diff
@@ -526,3 +527,18 @@ fn is_hex(hex: &str) -> Result<(), String> {
         Err("Must 0x-prefixed hexadecimal string".to_string())
     }
 }
+
+fn is_h256(hex: &str) -> Result<(), String> {
+    if hex.as_bytes()[..2] == b"0x"[..] {
+        let mut hd = [0u8; 32];
+        if let Err(e) = faster_hex::hex_decode(&hex.as_bytes()[2..], &mut hd) {
```
zhangsoledad, line 534: `faster_hex::hex_decode(&hex.as_bytes()[2..])`

util/app-config/src/cli.rs
chanhsu001: ok. decided for now.

util/app-config/src/cli.rs
chanhsu001: get it!....a little suggestion, we may need a line of  usage comment on hex_check().

chanhsu001: change to quake's suggestion....

util/app-config/src/cli.rs
```diff
@@ -526,3 +526,12 @@ fn is_hex(hex: &str) -> Result<(), String> {
...
     }
 }
+
+fn is_h256(hex: &str) -> Result<(), String> {
+    let tmp = hex.as_bytes();
+    if tmp[..2] == b"0x"[..] && tmp.len() == 66 {
+        is_hex(hex)
+    } else {
+        Err("Must be 0x-prefixed hexadecimal string and string length is 66".to_owned())
+    }
```
quake, line 536: `is_hex` fn has validated the input string already, only need to add an extra len checking here:

```suggestion
    if hex.len() != 66 {
        Err("Must be 0x-prefixed hexadecimal string and string length is 66".to_owned())
    } else {
        is_hex(hex)
    }
```

chanhsu001: done!

util/app-config/src/cli.rs
```diff
@@ -526,3 +526,12 @@ fn is_hex(hex: &str) -> Result<(), String> {
         Err("Must 0x-prefixed hexadecimal string".to_string())
     }
 }
+
+fn is_h256(hex: &str) -> Result<(), String> {
+    let tmp = hex.as_bytes();
```
quake, line 531: why convert to bytes here?

chanhsu001: oh, s... stumbled, my mistake.

quake (approved):

doitian:
bors d=chanhsu001,quake

quake:
bors r=quake

PR #3353, opened 2022-03-24
chore: remove unsafe reader to array unpack fn

quake:

Please refer to https://github.com/nervosnetwork/godwoken/pull/629, I think unpack a reader to array is quite rare, we should always unpack to new type, for example: `H256([u8; 32]) or ProposalId([u8;10])`. 

And in ckb current code, there are only one place to unpack the `ProposalIdReader` to [u8;10], which can be simply replaced by `try_into`, so I proposal this PR to remove these two unsafe fn.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

---
This change is [Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/3353)

### DIFF
--- a/util/jsonrpc-types/src/proposal_short_id.rs
+++ b/util/jsonrpc-types/src/proposal_short_id.rs
@@ -26,7 +26,11 @@ impl ProposalShortId {
 
 impl From<packed::ProposalShortId> for ProposalShortId {
     fn from(core: packed::ProposalShortId) -> ProposalShortId {
-        ProposalShortId::new(core.unpack())
+        ProposalShortId::new(
+            core.as_slice()
+                .try_into()
+                .expect("checked in packed::ProposalShortId"),
+        )
     }
 }
 
--- a/util/types/src/conversion/blockchain.rs
+++ b/util/types/src/conversion/blockchain.rs
@@ -51,28 +51,12 @@ impl Pack<packed::Byte32> for [u8; 32] {
     }
 }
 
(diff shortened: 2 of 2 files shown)


util/jsonrpc-types/src/proposal_short_id.rs
```diff
@@ -26,7 +26,11 @@ impl ProposalShortId {
 
 impl From<packed::ProposalShortId> for ProposalShortId {
     fn from(core: packed::ProposalShortId) -> ProposalShortId {
```
doitian, line 28: Should we inline this as recommended in https://github.com/nervosnetwork/godwoken/pull/629 ?

doitian:
```suggestion
    #[inline]
    fn from(core: packed::ProposalShortId) -> ProposalShortId {
```

quake: I don't understand this LTO feature, and if it proves to be a performance issue, it should be added in a separate PR.

chanhsu001: as far as i know, `inline` will provide more context for compiler/linker to know if  ` the length of the slice and just emit a simple load instruction most of the time`.
LTO likely will use `inline`, but if we clearly want to inline, inline it, only at cost of size(generate object size)

doitian (approved):

PR #3369, opened 2022-04-07
invoke sync to test replay step in ckb sync workflow

chanhsu001:

test replay fix...commit into sync-mainnet-staging to invoke sync mainnet flow
combined with ckb-integration-test's replay_fix branch(fix replay and limit  height to 12000 when mainnet sync)

only for test. if workflow result is correct, i will make a pr to ckb develop branch. 

### Related changes

- PR to update `owner/repo`:

### Release note 

```release-note
```


chanhsu001:
why not invoke `Mainnet Sync` Action for this commit to `sync-mainnet-staging`

PR #3386, opened 2022-04-21
feat(metrics): add metrics for ban and been_baned

chanhsu001:

Add metrics for network ban peer or  been banned, help to measure network partition.

What's Changed:
ban peer operation ocurred at sync and tx_pool module, but callee function is at network module, network.rs
we add ban_peer and reason info. in metrics.

- PR to update `owner/repo`:

### Check List 

Tests 

- test manually

### Release note 

```release-note
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -904,6 +904,7 @@ dependencies = [
  "ckb-app-config",
  "ckb-hash",
  "ckb-logger",
+ "ckb-metrics",
  "ckb-spawn",
  "ckb-stop-handler",
  "ckb-types",
--- a/network/Cargo.toml
+++ b/network/Cargo.toml
@@ -15,6 +15,7 @@ ckb-util = { path = "../util", version = "= 0.104.0-pre" }
 ckb-stop-handler = { path = "../util/stop-handler", version = "= 0.104.0-pre" }
 ckb-logger = { path = "../util/logger", version = "= 0.104.0-pre" }
 ckb-app-config = { path = "../util/app-config", version = "= 0.104.0-pre" }
+ckb-metrics = {path = "../util/metrics", version = "= 0.104.0-pre"}
 tokio = { version = "1", features = ["sync", "macros"] }
 tokio-util = { version = "0.7", features = ["codec"] }
 futures = "0.3"
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -20,6 +20,7 @@ use crate::services::{
 use crate::{Behaviour, CKBProtocol, Peer, PeerIndex, ProtocolId, ServiceControl};
 use ckb_app_config::{default_support_all_protocols, NetworkConfig, SupportProtocol};
 use ckb_logger::{debug, error, info, trace, warn};
+use ckb_metrics::metrics;
 use ckb_spawn::Spawn;
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_util::{Condvar, Mutex, RwLock};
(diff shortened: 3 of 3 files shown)


chanhsu001:
I just want to write a test but don't know how to setup promtheus backend...

network/src/protocols/disconnect_message.rs
```diff
@@ -27,6 +28,36 @@ impl ServiceProtocol for DisconnectMessageProtocol {
                 "Received disconnect message from peer={}: {}",
                 session_id, message
             );
+
+            // check if baned message and metric banned info
+            if message.starts_with("Ban for") && message.ends_with("seconds") {
```
doitian, line 33: Refer to the `NetworkState::ban_session` here to help code readers double check the message.

doitian: We can add ban reason in the message in `NetworkState::ban_session`. Here we should handle backward compatibility. It's OK when the ban message has no reason, but if it does, we can record it.

doitian (changes requested):

network/src/protocols/disconnect_message.rs
```diff
@@ -63,3 +79,33 @@ impl ServiceProtocol for DisconnectMessageProtocol {
...
+    // parse disconnect message and return peer_ip, ban seconds and reason(if exists)
+    // refer to NetworkState::ban_session() for message format
+    fn parse_disconnect_message(
+        &self,
+        session_id: SessionId,
+        message: &str,
+    ) -> (String, String, String) {
+        let addr = self.0.with_peer_registry(|reg| {
+            reg.get_peer(session_id)
+                .filter(|peer| !peer.is_whitelist)
```
quake, line 93: why filter out whitelist?

network/src/protocols/disconnect_message.rs
```diff
@@ -63,3 +79,33 @@ impl ServiceProtocol for DisconnectMessageProtocol {
...
+                .map(|peer| peer.connected_addr.clone())
+                .map_or(String::default(), |maddr| maddr.to_string())
```
quake, line 95:
```suggestion
                .map(|peer| peer.connected_addr.to_string())
                .unwrap_or_default()
```

network/src/protocols/disconnect_message.rs
```diff
@@ -63,3 +79,33 @@ impl ServiceProtocol for DisconnectMessageProtocol {
...
+        });
+
+        let mut ban_seconds = String::new();
+        if let Some(start) = message.find(char::is_numeric) {
```
quake, line 99: I don't think this code block is working for a ban message with any numeric char in the reason section, for example, with status code:
`Ban for 15 seconds, reason: HeadersIsInvalid(415)`

quake (dismissed):

network/src/protocols/disconnect_message.rs
chanhsu001: alright, i will correct it.

network/src/protocols/disconnect_message.rs
chanhsu001: filter of whitelist is unnecessary and removed...

chanhsu001:
After I write testcase to confirm metrics export to Prometheus, i will update result here.

doitian (approved):

network/src/protocols/disconnect_message.rs
quake: this review comment hasn't been resolved, please check again.

chanhsu001: i miss that part. thanks for your patience.

chanhsu001:
Close it now, ban/banned operation to inspect network partition may solve by other method.

PR #3387, opened 2022-04-25
fix(rpc): disconnect session of peers when invoke rpc set_ban

chanhsu001:

Problem Summary:

previous RPC `set_ban`(peer_ip, ... ) only adds peer into ckb inner ban_list, but doesn't disconnect the peer instantly.
the PR adds disconnect session operation when invoke RPC set_ban().

What's Changed:
we add `disconnect session` into the process, to disconnect peers instantly.
RPC `set_ban` support user input single IP address, and IP/Subnet form. PR's code traverse connecting peers to check if match IP or within Subnet.  

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test

Side effects

- IP address traversal match may be a concern of performance when node has lots of peer connection.

### Release note 

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -1230,6 +1230,7 @@ impl NetworkController {
 
     /// Ban an ip
     pub fn ban(&self, address: IpNetwork, ban_until: u64, ban_reason: String) {
+        self.disconnect_peers_in_ip_range(address, &ban_reason);
         self.network_state
             .peer_store
             .lock()
@@ -1259,6 +1260,23 @@ impl NetworkController {
     pub fn ban_peer(&self, peer_index: PeerIndex, duration: Duration, reason: String) {
         self.network_state
             .ban_session(&self.p2p_control, peer_index, duration, reason);
+    }
+
+    /// disconnect peers with matched peer_ip or peer_ip_network, eg: 192.168.0.2 or 192.168.0.0/24
+    fn disconnect_peers_in_ip_range(&self, address: IpNetwork, reason: &str) {
+        self.network_state.with_peer_registry(|reg| {
+            reg.peers().iter().for_each(|(peer_index, peer)| {
+                if let Some(addr) = multiaddr_to_socketaddr(&peer.connected_addr) {
+                    if address.contains(addr.ip()) {
+                        let _ = disconnect_with_message(
+                            &self.p2p_control,
+                            *peer_index,
+                            &format!("Ban peer {}, reason: {}", addr.ip(), reason),
+                        );
+                    }
+                }
+            })
+        });
     }
 
     fn try_broadcast(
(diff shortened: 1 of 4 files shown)


chanhsu001:
@quake ,  I just rewrite the code, please help to review, Thank you.

network/src/network.rs
```diff
@@ -1261,6 +1261,69 @@ impl NetworkController {
...
+                                    &self.p2p_control,
+                                    *peer_index,
+                                    &message,
+                                );
+                            }
+                        }
+                    }
+                });
+            }
+        }
```
quake, line 1324: please use some built-in fn:

```suggestion
       self.network_state.with_peer_registry(|reg| {
            reg.peers().iter().for_each(|(peer_index, peer)| {
                multiaddr_to_socketaddr(peer.connected_addr).map(|addr| {
                    if address.contains(addr.ip()) {
                        self.ban_peer(peer_index, duration, reason));
                    }
                });
            })
        });
```

quake (changes requested):

network/src/network.rs
chanhsu001: the built-in fn codes are awesome, I keep scratching head until meet it.

network/src/network.rs
```diff
@@ -1261,6 +1261,69 @@ impl NetworkController {
             .ban_session(&self.p2p_control, peer_index, duration, reason);
     }
 
+    /// disconnect peers with matched peer_ip or peer_ip_network, eg: 192.168.0.2 or 192.168.0.0/24
+    pub fn disconnect_peer_in_ip_rang(
```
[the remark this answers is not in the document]
chanhsu001: done!

network/src/network.rs
chanhsu001: done! 
- i attach message instead of  reason
- clippy suggest to use 
 `if Some(..) =  multiaddr_to_socketaddr(peer.connected_addr)` cause of this [reason](https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn)

network/src/network.rs
```diff
@@ -1261,6 +1262,28 @@ impl NetworkController {
...
+    fn disconnect_peers_in_ip_range(&self, address: IpNetwork, duration: Duration, reason: &str) {
+        self.network_state.with_peer_registry(|reg| {
+            reg.peers().iter().for_each(|(peer_index, peer)| {
+                if let Some(addr) = multiaddr_to_socketaddr(&peer.connected_addr) {
+                    if address.contains(addr.ip()) {
+                        let _ = disconnect_with_message(
+                            &self.p2p_control,
+                            *peer_index,
+                            &format!(
+                                "Ban peer {:?} for {} seconds, reason: {}",
```
quake, line 1275: this message is incorrectly, the params `ban_until` is a timestamp not duration.

chanhsu001: Keep it simple, i rewrite message as "Ban peer {}, reason: {}"

quake (approved):

PR #3391, opened 2022-04-28
improve sync header map

zhangsoledad:

* Optimize the implementation of header map to improve the speed of header synchronization from 11924/s to 14377/s, about 20% improvement.
* Human-friendly configuration

- Asynchronous batch write to hard disk when memory limit is exceeded.
- Replace the implementation of header map bankend. Rocksdb has some drawbacks in this scenario, Rocksdb cannot give feedback on whether the key is overwritten when writing, nor can it distinguish between successful or not found when deleting, it's inefficient and looks silly to retrieve a value just to check if it exists or not. 
```
RocksDBBackend::contains_key duration 4041 ns
RocksDBBackend::remove duration 7000 ns
RocksDBBackend::insert duration 10750 ns

Sometimes it deteriorates into the following, because of the inconvenience of implementation：
RocksDBBackend::contains_key duration 10750 ns
RocksDBBackend::remove duration 13000 ns
RocksDBBackend::insert duration 21583 ns

-----------------------------
SledBackend::contains_key duration 2666 ns
SledBackend::remove duration 2583 ns
SledBackend::insert_batch 35503541/10028 ns (3540 ns)
```

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


zhangsoledad:
https://gist.github.com/zhangsoledad/0605a6c98e58437d1796b7fe12638286

sync/src/types/mod.rs
```diff
@@ -958,25 +958,19 @@ impl HeaderView {
...
-            .and_then(|parent| {
-                parent.get_ancestor(
-                    tip_number,
-                    get_skip_height(self.number()),
-                    get_header_view,
-                    fast_scanner,
-                )
-            })
+        self.skip_hash = self
+            .clone()
```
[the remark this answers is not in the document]
zhangsoledad: There is no difference between this code and the original code in terms of function. The main change here is that since the header map introduce backend db, `get_header_view` may not just be a cheap memory operation, but  once IO operation, and the reason why the original code does not use get_ancestor directly, but gets the parent first, I guess it is just because get_ancestor is a method that consume self, so here just clone self also works , get the parent does not have any practical significance.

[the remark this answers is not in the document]
zhangsoledad: There is no difference between this code and the original code in terms of function. The main change here is that since the header map introduce backend db, `get_header_view` may not just be a cheap memory operation, but  once IO operation, and the reason why the original code does not use get_ancestor directly, but gets the parent first, I guess it is just because get_ancestor is a method that consume self, so here just clone self also works , get the parent does not have any practical significance.

sync/src/types/mod.rs
```diff
@@ -1586,10 +1580,10 @@ impl SyncState {
             return;
         }
 
-        assert!(
-            self.header_map.contains_key(&header.hash()),
-            "HeaderView must exists in header_map before set best header"
-        );
+        // assert!(
```
driftluo, line 1583: This comment can be removed

driftluo (approved):

sync/src/types/header_map/backend.rs
```diff
@@ -14,12 +14,11 @@ pub(crate) trait KeyValueBackend {
...
-
     fn contains_key(&self, key: &Byte32) -> bool;
     fn get(&self, key: &Byte32) -> Option<HeaderView>;
-    fn insert(&mut self, value: &HeaderView) -> Option<HeaderView>;
-    fn remove(&mut self, key: &Byte32) -> Option<HeaderView>;
+    fn insert(&self, value: &HeaderView) -> Option<()>;
+    fn insert_batch(&self, values: &[HeaderView]);
+    fn remove(&self, key: &Byte32) -> Option<HeaderView>;
+    fn remove_no_return(&self, key: &Byte32);
+    fn remove_batch(&self, keys: &[Byte32]);
```
quake, line 23: I cannot find the usage of this trait fn `remove_batch`, can we remove it?

sync/src/types/header_map/backend_sled.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn remove(&self, key: &Byte32) -> Option<HeaderView> {
+        let old_value = self
+            .db
+            .remove(key.as_slice())
+            .expect("failed to remove item from sled");
+        if old_value.is_some() {
+            self.count.fetch_sub(1, Ordering::SeqCst);
+        }
+        old_value.map(|slice| HeaderView::from_slice_should_be_ok(&slice))
```
quake, line 91:
```suggestion
        old_value.map(|slice| {
            self.count.fetch_sub(1, Ordering::SeqCst);
            HeaderView::from_slice_should_be_ok(&slice)
        })
```

sync/src/types/header_map/backend_sled.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut count = 0;
+        for key in keys {
+            let old_value = self
+                .db
+                .remove(key.as_slice())
+                .expect("failed to remove item from sled");
+            if old_value.is_some() {
+                count += 1;
+            }
+        }
```
quake, line 114:
```suggestion
        let count = keys.iter().filter(|key| {
            self.db.remove(key.as_slice()).expect("failed to remove item from sled").is_some()
        }).count();
```

sync/src/types/header_map/backend_sled.rs
```diff
@@ -0,0 +1,104 @@
...
+use crate::types::HeaderView;
+use ckb_types::{packed::Byte32, prelude::*};
+use sled::Db;
+use std::path;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+use tempfile::TempDir;
+
+pub(crate) struct SledBackend {
+    count: AtomicUsize,
```
doitian, line 11: Can we remove `len` and `is_empty` from `KeyValueBackedn`?

sync/src/types/header_map/backend_sled.rs
```diff
@@ -0,0 +1,104 @@
...
+            self.count.fetch_add(1, Ordering::SeqCst);
+        }
+        last_value.map(|_| ())
+    }
+
+    fn insert_batch(&self, values: &[HeaderView]) {
+        let mut count = 0;
+        for value in values {
+            let key = value.hash();
+            let last_value = self
```
doitian, line 72: Have you compared the performance difference between consecutive insertions with [`apply_batch`](https://docs.rs/sled/latest/sled/struct.Db.html#method.apply_batch)?

sync/src/types/header_map/backend_sled.rs
zhangsoledad: No, this is used for short-circuiting, when the count is 0 we can skip the db operation, this will be reflected in the profiling.
If you mean can we use `sled::len` or `sled::is_empty` , although I haven't actually tried it, but it shouldn't work either, performs a full O(n) scan under the hood.
```
   pub fn len(&self) -> usize {
        self.iter().count()
    }

    /// Returns `true` if the `Tree` contains no elements.
    pub fn is_empty(&self) -> bool {
        self.iter().next().is_none()
    }

   pub(crate) fn next_inner(&mut self) -> Option<<Self as Iterator>::Item> {
        let guard = pin();
        let (mut pid, mut node) = if let (true, Some((pid, node))) =
            (self.going_forward, self.cached_node.take())
        {
            (pid, node)
        } else {
            let view =
                iter_try!(self.tree.view_for_key(self.low_key(), &guard));
            (view.pid, view.deref().clone())
        };

        for _ in 0..MAX_LOOPS {
            if self.bounds_collapsed() {
                return None;
            }

            if !node.contains_upper_bound(&self.lo) {
                ...........
            } else if !node.contains_lower_bound(&self.lo, true) {
               ........
            } else {
                  ........
            }
        }
   
    }
```

sync/src/types/header_map/backend_sled.rs
zhangsoledad: [apply_batch](https://docs.rs/sled/latest/sled/struct.Db.html#method.apply_batch) has the same problem, it can't give feedback whether a key was overwritten or written directly，Here also tried to compare with [transaction](https://docs.rs/sled/latest/sled/transaction/index.html)，transaction performs worse because of the extra overhead required to guarantee ACID.

doitian:
Sled has the same memory usage after exiting sync, but has higher peak memory usage during sync.

CC @quake

quake (approved):

PR #3393, opened 2022-04-29
fix(test): fix PR3387's testcase CI running failure

chanhsu001:

fix CI test failure of testcase RpcSetBan of PR3387

Problem Summary:

CI test running enviroment differs from local host,  timming and lag will be a problem when write testcase and assert result.
here we should use `a more robust way to wait.`
we already check the result is positive in running test of PR3392.

### Related changes

- PR to update `owner/repo`:

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/test/src/specs/rpc/set_ban.rs
+++ b/test/src/specs/rpc/set_ban.rs
@@ -2,6 +2,7 @@ use crate::node::{connect_all, waiting_for_sync};
 use crate::specs::spec_name;
 use crate::util::mining::out_ibd_mode;
 use crate::utils::find_available_port;
+use crate::utils::wait_until;
 use crate::{Node, Spec};
 
 pub struct RpcSetBan;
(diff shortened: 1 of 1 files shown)


test/src/specs/rpc/set_ban.rs
```diff
@@ -27,6 +28,7 @@ impl Spec for RpcSetBan {
             None,
             Some("for_test".to_owned()),
         );
+        sleep(1);
```
doitian, line 31: Better add a wait loop. I think we already have such helpers.


```
CHECK_COUNT=3
while E is not true and CHECK_COUNT > 0:
  sleep(S)
  CHECK_COUNT-=1
```

doitian (dismissed):
Use a more robust way to wait.

test/src/specs/rpc/set_ban.rs
chanhsu001: Yes, let change it into "wait_until"

doitian (approved):

PR #3400, opened 2022-05-06
chore: upgrade secio

driftluo:
1. Upgrade secio
```
secio 0.5.4
Features

Use openssl repalce ring on unix and ossl110
```

### Check List

Tests 

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/util/memory-tracker/src/process.rs
+++ b/util/memory-tracker/src/process.rs
@@ -110,11 +110,11 @@ pub struct Memory {
     // Size of physical memory being used
     resident: u64,
     // Number of shared pages
-    shared: u64,
+    _shared: u64,
     // The size of executable virtual memory owned by the program
-    text: u64,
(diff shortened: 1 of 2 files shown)


zhangsoledad (approved):

util/memory-tracker/src/process.rs
```diff
@@ -110,11 +110,11 @@ pub struct Memory {
     // Size of physical memory being used
     resident: u64,
     // Number of shared pages
-    shared: u64,
+    _shared: u64,
```
doitian, line 113: Why adding the prefix `_`?

driftluo: Higher versions of rustc will report a warning that this data is not used.
I'm not sure if I'll need this data in the future, so I'll add `_` to disable the warning for now

PR #3443, opened 2022-06-10
feat(sync): print epoch statistics in_IBD instead of each block

chanhsu001:

Problem Summary:  Print 1st block in each epoch and previous epoch statistics information, when syncing node is in IBD

What's Changed:
- add statistics structure in ChainService to record epoch's info counters
- when syncing node is in IBD, only print 1st block of each epoch
- print previous epoch statistics
- make sure the adding code runs fast and minimal impact to performance

running output as:
```
2022-06-10 23:21:57.683 +08:00 ChainService INFO ckb_chain::chain  Epoch 156 Statis: blocks: 1800, blocks size: 1094335, txs: 1910, time: 10 secs
2022-06-10 23:21:57.683 +08:00 ChainService INFO ckb_chain::chain  block: 253182, hash: 0xe9e515bcb0e866dc987d03b737cbfef6ababc88515f2ca88049b773df54d3178, epoch: 157(0/1366), total_diff: 0x16665f819887d04f63, txs: 2
2022-06-10 23:22:05.138 +08:00 ChainService INFO ckb_chain::chain  Epoch 157 Statis: blocks: 1366, blocks size: 824881, txs: 1456, time: 7 secs
2022-06-10 23:22:05.138 +08:00 ChainService INFO ckb_chain::chain  block: 254548, hash: 0xaf531604ff01cdbbbf03325819edcc440a319cee1c9e5dad42bd7d0ff4c26611, epoch: 158(0/1800), total_diff: 0x168d77756fd5e476fe, txs: 1
2022-06-10 23:22:16.037 +08:00 ChainService INFO ckb_chain::chain  Epoch 158 Statis: blocks: 1800, blocks size: 1522519, txs: 1977, time: 10 secs
2022-06-10 23:22:16.037 +08:00 ChainService INFO ckb_chain::chain  block: 256348, hash: 0xee09df249f0d32404bf87524a45f8567f6ed674e8802e1fb8ced94e1a164a9ed, epoch: 159(0/1781), total_diff: 0x16b5c05c049bac55fd, txs: 1
```

- not adding tx's cycles sum in epoch statistics, cause don't find a easy way to fetch it(cycles info in reconcile_main_chain(), but use immutable ChainService).
- omit print epoch statistics information at point of OUT_OF_IBD, cause handling corner condition will add more checking in each block insert;  it will slow the running of ckb. (IMHP)  

### Related changes

- PR to update `owner/repo`:

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -14,6 +14,7 @@ use ckb_rust_unstable_port::IsSorted;
 use ckb_shared::shared::Shared;
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_store::{attach_block_cell, detach_block_cell, ChainStore, StoreTransaction};
+use ckb_types::core::{Cycle, EpochNumber, EpochNumberWithFraction};
 use ckb_types::{
     core::{
         cell::{
@@ -230,12 +231,36 @@ impl GlobalIndex {
     }
 }
 
+#[derive(Copy, Clone)]
+struct ChainStat {
+    block_cnt: u64,
+    block_size_cnt: u64,
+    tx_cnt: u64,
+    time_stamp: u64,
+    epoch_number: EpochNumber,
+    first_block_in_epoch: BlockNumber,
+    last_block_in_epoch: BlockNumber,
+}
+impl Default for ChainStat {
+    fn default() -> Self {
+        ChainStat {
+            block_cnt: 0,
+            block_size_cnt: 0,
+            tx_cnt: 0,
+            time_stamp: 0,
+            epoch_number: u64::MAX,
+            first_block_in_epoch: 1,
+            last_block_in_epoch: 0,
+        }
+    }
+}
 /// Chain background service
 ///
 /// The ChainService provides a single-threaded background executor.
 pub struct ChainService {
     shared: Shared,
     proposal_table: ProposalTable,
+    stat: ChainStat,
 }
 
 impl ChainService {
@@ -244,6 +269,7 @@ impl ChainService {
         ChainService {
             shared,
             proposal_table,
+            stat: Default::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


chain/src/chain.rs
```diff
@@ -494,14 +504,20 @@ impl ChainService {
...
+            } else {
+                info!(
+                    "block: {}, hash: {:#x}, epoch: {:#}, total_diff: {:#x}, txs: {}",
+                    tip_header.number(),
+                    tip_header.hash(),
+                    tip_header.epoch(),
+                    total_difficulty,
+                    block.transactions().len()
+                );
+            }
```
quake, line 520: suggest to move all log related code to a single fn, it can reduce duplicate code, for example  `info!("block: {}...`

chain/src/chain.rs
```diff
@@ -912,6 +930,55 @@ impl ChainService {
...
+    /// in IBD, print 1st block of the epoch and print statistics of previous epoch
+    fn update_print_statics(
+        &mut self,
+        block: &Arc<BlockView>,
+        tip_header: &HeaderView,
+        total_difficulty: U256,
+    ) {
+        let tip_epoch = tip_header.epoch();
+
+        if tip_epoch.index() == 0 {
```
quake, line 943: It is recommended to change the comparison logic to compare block number against epoch starting number plus epoch length, which will reduce the complexity of the code.

quake (changes requested):

chain/src/chain.rs
```diff
@@ -825,15 +841,17 @@ impl ChainService {
...
-                                        info!(
-                                            "[block_verifier] block number: {}, hash: {}, size:{}/{}, cycles: {}/{}",
-                                            b.number(),
-                                            b.hash(),
-                                            b.data().serialized_size_without_uncle_proposals(),
-                                            self.shared.consensus().max_block_bytes(),
-                                            cycles,
-                                            self.shared.consensus().max_block_cycles()
-                                        );
+                                        if !self.shared.is_initial_block_download() {
```
quake, line 844: why disable this log in ibd?

chain/src/chain.rs
chanhsu001: i just move related info_block call into standalone functions, but not all functions can be merged into single function, due to message format.

chain/src/chain.rs
chanhsu001: Good advice!
i refactor this part,  please help to check.
i'm not sure about if the code is the best way to tell first_block/last_block number of each epoch.

chain/src/chain.rs
chanhsu001: original block printing is as:
block #12 info...........
[block verifier] block #12 info... cycles: ....
 
my previous thoughts was omit both block and verifier print, but if you suggest to keep it, and i also think cycles information cannot be fetch from database, mabey keep it in log file is also usable...so i keep it.

PR #3459, opened 2022-06-21
feat: upgrade tentacle

driftluo:

introduce `async_trait` to ckb

this is a break change for ckb, it upgrade tentacle from 0.4.0-aphla to 0.4.0

ckb-network will also use `async_trait` to ckb protocol

[release note](https://github.com/nervosnetwork/tentacle/blob/master/CHANGELOG.md#040)

### Check List

Tests 

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/protocols/mod.rs
+++ b/network/src/protocols/mod.rs
@@ -11,10 +11,11 @@ mod tests;
 use ckb_logger::{debug, trace};
 use futures::{Future, FutureExt};
 use p2p::{
+    async_trait,
     builder::MetaBuilder,
     bytes::Bytes,
     context::{ProtocolContext, ProtocolContextMutRef},
-    service::{BlockingFlag, ProtocolHandle, ProtocolMeta, ServiceControl, TargetSession},
+    service::{ProtocolHandle, ProtocolMeta, ServiceAsyncControl, ServiceControl, TargetSession},
     traits::ServiceProtocol,
     ProtocolId, SessionId,
 };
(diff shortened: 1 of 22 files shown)


zhangsoledad (approved):

network/src/protocols/mod.rs
```diff
@@ -33,17 +34,53 @@ pub type BoxedFutureTask = Pin<Box<dyn Future<Output = ()> + 'static + Send>>;
...
+    async fn remove_notify(&self, token: u64) -> Result<(), Error>;
+    /// Send message through quick queue
+    async fn async_quick_send_message(
+        &self,
+        proto_id: ProtocolId,
+        peer_index: PeerIndex,
+        data: Bytes,
+    ) -> Result<(), Error>;
+    /// Send message through quick queue
+    async fn async_quick_send_message_to(
```
quake, line 57: I cannot find the usage of these `async_xxx` fn in this PR, please describe the specific use.

quake (changes requested):

network/src/protocols/mod.rs
driftluo: This is to give the context rewrapped with ckb-network the ability to use async-await. The current sync and relayer protocols are left as they are because of the complexity of locks and states, but that doesn't mean other dependent implementations can't use them.

Whether you can use it and whether you use it are two different things. In an environment where the context can use await, it would be strange not to provide a corresponding method.

There is a difference in the behavior of synchronous and asynchronous methods.
1. when a synchronous method encounters a full queue, it returns a would block error after trying to send and exits directly
2. When an asynchronous method encounters a full queue, it enters a pending state and waits for the next wake

quake (approved):

PR #3525, opened 2022-08-03
chore: upgrade gcs filter lib and rebuild data

quake:

upgrade gcs lib to 0.2.0 and rebuild the block filter data since the gcs params M has been changed from `784930` to `784931`

### Check List 

Tests 

- Unit test
- Integration test

### DIFF
--- a/block-filter/src/filter.rs
+++ b/block-filter/src/filter.rs
@@ -7,11 +7,13 @@ use std::thread;
 
 const THREAD_NAME: &str = "ckb-block-filter-thread";
 
+/// A block filter creation service
 pub struct BlockFilter {
     db: ChainDB,
 }
 
 impl BlockFilter {
+    /// Create a new block filter service
     pub fn new(db: ChainDB) -> Self {
         Self { db }
     }
--- a/block-filter/src/filter.rs
+++ b/block-filter/src/filter.rs
@@ -3,9 +3,10 @@ use ckb_logger::debug;
 use ckb_notify::NotifyController;
 use ckb_store::{ChainDB, ChainStore};
 use ckb_types::{core::HeaderView, prelude::*};
+use golomb_coded_set::{GCSFilterWriter, SipHasher24Builder, M, P};
 use std::thread;
 
-const THREAD_NAME: &str = "ckb-block-filter-thread";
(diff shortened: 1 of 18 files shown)


util/launcher/src/migrations/rebuild_block_filter.rs
```diff
@@ -0,0 +1,29 @@
+use ckb_db::{Result, RocksDB};
+use ckb_db_migration::{Migration, ProgressBar};
+use ckb_db_schema::{COLUMN_META, META_LATEST_BUILT_FILTER_DATA_KEY};
+use std::sync::Arc;
+
+pub struct RebuildBlockFilter;
+
+const VERSION: &str = "20220803143233";
```
yangby-cryptape, line 8: If I understand correctly, this file can be squashed (removed) later, since we haven't release any official `CKB` binaries.

block-filter/src/filter.rs
```diff
@@ -3,6 +3,7 @@ use ckb_logger::debug;
 use ckb_notify::NotifyController;
 use ckb_store::{ChainDB, ChainStore};
 use ckb_types::{core::HeaderView, prelude::*};
+use golomb_coded_set::{GCSFilterWriter, SipHasher24Builder, M, P};
 use std::thread;
 
 const THREAD_NAME: &str = "ckb-block-filter-thread";
```
yangby-cryptape, line 9: Only 16 bytes are supported for a thread name, and the last byte should be `NULL`.

Since a long thread name will cut to 15 characters, maybe `CamelCase` is better than `snake_case`.

Ref: This limit is created by kernel.

```c
/* Task command name length: */
#define TASK_COMM_LEN     16
```

Ref: [The definition of `TASK_COMM_LEN` in latest stable kernel.](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/sched.h?h=v5.10.12#n213)

yangby-cryptape (approved):

util/launcher/src/migrations/rebuild_block_filter.rs
quake: Yes, I've add a code comment

PR #3565, opened 2022-09-02
feat:ChainRpcImpl: support return raw molecule hex bytes for `get_transaction` rpc method

### LINKED ISSUE #3424: `get_trasaction` rpc needs to be able to support returning raw molecule hex bytes
## Feature Request

`get_transaction` rpc needs to be able to support returning raw molecule hex bytes

Previously, we used the verbosity parameter to control the mode of the extended `get_transaction`. verbosity=0 has been marked as deprecated and we can reuse verbosity=0 to indicate that raw hex bytes are returned. This has the advantage of being consistent with the behavior of `get_block` verbosity=0

#### Describe alternatives you've considered

New rpc `get_raw_transaction`.

eval-exec:
Signed-off-by: Eval EXEC <execvy@gmail.com>

Issue Number: close #3424 

Problem Summary: support return raw molecule hex bytes for `get_transaction` rpc method.

What's Changed:

- rewrite `verbosity=0` branch for `get_transaction` rpc method.
- delete `get_transaction_verbosity0()` 
- make `verbosity=0` 's behaviour as same as `verbosity=2`, except for when `verbosity=0`, it will return a 0x-prefixed hex encoded raw molecule `packed::Transaction` on `transaction` field;
- change `ChainRpc::get_transaction()`'s return type from `Result<Option<TransactionWithStatus>>` to `Result<Option<TransactionWithStatusResponse>>`
- move `json_types::TransactionWithStatus` to `tx_pool::TransactionWithStatus`
- change some specs tests assertion in `test/src/specs`
### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/util/jsonrpc-types/src/blockchain.rs
+++ b/util/jsonrpc-types/src/blockchain.rs
@@ -1,8 +1,9 @@
 use crate::bytes::JsonBytes;
 use crate::{
     BlockNumber, Byte32, Capacity, Cycle, EpochNumber, EpochNumberWithFraction, ProposalShortId,
-    Timestamp, Uint128, Uint32, Uint64, Version,
+    ResponseFormat, ResponseFormatInnerType, Timestamp, Uint128, Uint32, Uint64, Version,
 };
+use ckb_types::core::tx_pool;
 use ckb_types::{core, packed, prelude::*, H256};
 use serde::{Deserialize, Serialize};
 use std::fmt;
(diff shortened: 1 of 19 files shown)


zhangsoledad:
For the final output format, you can refer to https://github.com/nervosnetwork/ckb/discussions/3425#discussioncomment-2773449

zhangsoledad:
One of the problems we encountered in this PR is that we need to make `ResponseFormat` implement Deserialize, which is actually caused by some problems with the original `ResponseFormat` implementation itself, which does not follow the DIY principle and does not take advantage of serde's features, so I suggest that we first make some modifications to `ResponseFormat`. We already have `JsonBytes`, we just need to make ResponseFormat composable according to serde's design.

[ResponseFormat.patch view on github](https://github.com/nervosnetwork/ckb/compare/develop...zhangsoledad/composable-response-format)

util/jsonrpc-types/src/blockchain.rs
```diff
@@ -514,6 +514,43 @@ impl From<Transaction> for packed::Transaction {
...
+                )
+            }),
+            tx_status: t.tx_status,
+        }
+    }
+}
+
 /// The JSON view of a transaction as well as its status.
 #[derive(Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Debug)]
 pub struct TransactionWithStatus {
```
zhangsoledad, line 556: Here is also a legacy issue, in fact, the tx-pool should not use the JSON type directly.
Here we can move `TransactionWithStatus` to util/types/src/core/tx_pool.rs.
```
pub transaction: Option<TransactionView>,
```
This field is changed to Option<core::TransactionView>.
This avoids the problem of repeated serialization.

util/jsonrpc-types/src/blockchain.rs
```diff
@@ -515,67 +516,38 @@ impl From<Transaction> for packed::Transaction {
...
+        match inner_type {
+            ResponseFormatInnerType::Hex => TransactionWithStatusResponse {
+                transaction: t.transaction.map(|tx| {
+                    let packed_tx_view = packed::TransactionView::new_builder()
+                        .data(tx.data())
+                        .hash(tx.hash())
+                        .witness_hash(tx.witness_hash())
+                        .build();
+                    ResponseFormat::hex(packed_tx_view.as_bytes())
+                }),
```
zhangsoledad, line 540:
```suggestion
                transaction: t
                    .transaction
                    .map(|tx| ResponseFormat::hex(tx.data().as_bytes())),
```

eval-exec: > 
OK, I did the suggestion change: 
https://github.com/nervosnetwork/ckb/compare/183f773f1ab52efdd0727bbb403538730331d093..617a21f8e88fb8f42c7a15ce15c171150935ef3a

Now, when `verbosity=0`, It will return hex encoded `packed::Transaction` instead of `packed::TransactionView`

eval-exec:
[ci_unit_tests_macos](https://github.com/nervosnetwork/ckb/actions/runs/3239903574/jobs/5309865061#logs) has failed

zhangsoledad (approved):

PR #3566, opened 2022-09-06
fix: fix tcp reuse bind

driftluo:

fix https://github.com/nervosnetwork/ckb/discussions/2654#discussioncomment-3561768

Fix the panic caused by the configuration file

Reuse-bind cannot be performed between ipv4 and ipv6, and branch judgment is required

### Check List 

Tests

- Unit test
- Integration test

### Release note

```release-note
```


quake (approved):

quake:
bors r=quake

PR #3595, opened 2022-09-15
feat: Flags features

driftluo:

Expand the identify flags feature, and introduce some new concepts:
```rust
bitflags::bitflags! {
    pub struct Flags: u64 {
        /// Compatibility reserved
        const COMPATIBILITY = 0b1;
        /// Discovery protocol, which can provide peers data service
        const DISCOVERY = 0b10;
        /// Sync protocol can provide Block and Header download service
        const SYNC = 0b100;
        /// Relay protocol, which can provide CompactBlock and Transaction broadcast/forwarding services
        const RELAY = 0b1000;
        /// Light client protocol, which can provide Block / Transaction data and existence-proof services
        const LIGHT_CLIENT = 0b10000;
        /// Client-side block filter protocol can provide BlockFilter download service
        const BLOCK_FILTER = 0b100000;
    }
}
```
1. discovery peer address with flags
2. extended flags semantics
3. allow discovery messages to specify the address of the flags corresponding to the request
4. introduction of three concepts: `target_flags_filter`, `required_flags`, `identify_announce`, used to determine, respectively, the address to which this node will try to connect, the address to which this node will request from the remote end, and the attributes of this node to declare itself

Considering the emergence of light nodes, the above three concepts are not entirely consistent and must be configured externally. By default, this node is considered a full node, and there are not too many restrictions on discovery messages.

It is important to note that the discovery protocol does not verify the data given by the remote end, but stores it in the peer store and then checks the connection through the feeler. The advantage of this is that the peer store data is relatively even across the network, so that nodes with special needs can request some of the data, but cannot deny the data it does not need to spread across the network.

### Check List

Tests

- Unit test
- Integration test
- Manual test

### Release note 

```release-note
```

### DIFF
--- a/network/src/peer_store/types.rs
+++ b/network/src/peer_store/types.rs
@@ -46,6 +46,13 @@ pub struct AddrInfo {
     pub attempts_count: u32,
     /// Random id
     pub random_id_pos: usize,
+    /// Flags
+    #[serde(default = "default_flags")]
+    pub flags: u64,
+}
+
+fn default_flags() -> u64 {
+    0x1
 }
 
 impl AddrInfo {
@@ -58,6 +65,7 @@ impl AddrInfo {
             last_tried_at_ms: 0,
             attempts_count: 0,
             random_id_pos: 0,
+            flags: 0,
         }
     }
 
@@ -101,6 +109,11 @@ impl AddrInfo {
         self.last_connected_at_ms = connected_at_ms;
         // reset attempts
         self.attempts_count = 0;
+    }
+
+    /// Change address flags
+    pub fn flags(&mut self, flags: u64) {
+        self.flags = flags;
     }
 }
 
--- a/network/src/peer_store/types.rs
+++ b/network/src/peer_store/types.rs
@@ -1,5 +1,8 @@
 //! Type used on peer store
-use crate::peer_store::{Score, SessionType, ADDR_MAX_FAILURES, ADDR_MAX_RETRIES, ADDR_TIMEOUT_MS};
+use crate::{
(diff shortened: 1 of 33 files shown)


network/src/peer_store/types.rs
```diff
@@ -102,6 +110,11 @@ impl AddrInfo {
         // reset attempts
         self.attempts_count = 0;
     }
+
+    /// Change address flags
+    pub fn flags(&mut self, flags: u64) {
+        self.flags = flags;
+    }
```
quake, line 117:
```suggestion
    pub fn flags(&mut self, flags: Flags) {
        self.flags = flags.bits();
    }
```

network/src/protocols/discovery/mod.rs
```diff
@@ -356,20 +362,41 @@ impl AddressManager for DiscoveryAddressManager {
...
     }
 
-    fn get_random(&mut self, n: usize) -> Vec<Multiaddr> {
-        let fetch_random_addrs = self
-            .network_state
-            .with_peer_store_mut(|peer_store| peer_store.fetch_random_addrs(n));
+    fn get_random(&mut self, n: usize, flags: Flags) -> Vec<(Multiaddr, Flags)> {
+        let fetch_random_addrs = self.network_state.with_peer_store_mut(|peer_store| {
+            peer_store.fetch_random_addrs(n, |a| {
+                let f = unsafe { Flags::from_bits_unchecked(a) };
```
quake, line 368: suggest to use `Flags::from_bits_truncate` to remove drop other undefined bits

quake (approved):

PR #3598, opened 2022-09-20
merge indexer

zhangsoledad:

Early on, CKB had built-in indexer-related functionality, but because of implementation issues and driven by the micro-kernel design philosophy at the time, we decided to make the indexer independent.
As more and more developers gave us feedback about the convenience, the indexer service was an extra burden for development convenience and maintenance, so we rethought the relationship between indexer and ckb-core, and decided to merge indexer into ckb-core, so that developers can start the built-in indexer simply by command-line arguments or configuration
```bash
ckb run -C <path>  --indexer
```
or 
``` toml
[rpc]
modules = ["...", "Indexer"]
```
instead of downloading a binary and running a separate service.

Technically speaking, the built-in indexer also has some advantages. With rocksdb's support for second instances, the built-in indexer can directly synchronize data as a secondary_db instead of rpc, building indexes at a speed of about 39%.

### Some notes
* https://github.com/nervosnetwork/ckb/pull/3598/commits/f42ff4a7e2ef0ab33a3f4a1946ca282bbba994b8 The reason this refactoring is included in this PR is that the original notify uses crossbeam's blocking channel, and the indexer currently spawn into the tokio runtime, so if the indexer needs to subscribe to the transaction pool, it needs to use notify, but the blocking channel is not well handled in tokio's environment, and the spawn_blocking like tokio provides is not very applicable, spawn_blocking feature is intended for blocking operations that eventually finish, the async channel provided by Tokio can call outside of asynchronous contexts, but in turn the blocking channel is not really suitable for async, so here by the way the notify is completely replaced by async channel. 

* `util/indexer` is based on [ckb-indexer](https://github.com/nervosnetwork/ckb-indexer) v0.4.1, some types used for serialization in util/indexer/src/service.rs should be merged into ckb-json-type, and the PR after this one may do the related refactoring.

* To distinguish it from the original legacy configuration, the indexer configuration now has v2 as a suffix
```toml
[indexer_v2]
# Indexing the pending txs in the ckb tx-pool
index_tx_pool = false
```

* The changes to the Rpc interface compared to the original [ckb-indexer](https://github.com/nervosnetwork/ckb-indexer) are:
           *[get_tip](https://github.com/nervosnetwork/ckb-indexer#get_tip) was renamed to `get_indexer_tip`, to avoid ambiguity
           *[get_indexer_info](https://github.com/nervosnetwork/ckb-indexer#get_indexer_info) removed

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


util/indexer/src/service.rs
```diff
@@ -0,0 +1,1607 @@
...
+                            let longest_fork_number = tip_number.saturating_sub(keep_num);
+                            match self.get_block_by_number(longest_fork_number) {
+                                Some(block) => {
+                                    let stored_block_hash = indexer
+                                        .get_block_hash(longest_fork_number)
+                                        .expect("get block hash should be OK")
+                                        .expect("stored block header");
+                                    if block.hash() != stored_block_hash {
+                                        error!("long fork detected, ckb-indexer stored block {} => {:#x}, ckb node returns block {} => {:#x}, please check if ckb-indexer is connected to the same network ckb node.", longest_fork_number, stored_block_hash, longest_fork_number, block.hash());
+                                        sleep(self.poll_interval).await;
```
TheWaWaR, line 151: Can we just call `IndexerService.poll()` instantly when new block applied?

zhangsoledad:
Superseded by https://github.com/nervosnetwork/ckb/pull/3609

PR #3665, opened 2022-10-26
chore: remove block_filter_enable option

quake:

Two configuration options (block_filter_enable and support_protocols) are conflicting, one is off by default and one is on by default

Remove the `block_filter_enable` option.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/block-filter/src/filter.rs
+++ b/block-filter/src/filter.rs
@@ -1,5 +1,5 @@
 use ckb_async_runtime::tokio::{self, sync::oneshot, task::block_in_place};
-use ckb_logger::debug;
+use ckb_logger::{debug, warn};
 use ckb_shared::Shared;
 use ckb_stop_handler::{SignalSender, StopHandler};
 use ckb_store::ChainStore;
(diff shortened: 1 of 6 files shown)


driftluo (commented):
I think [this file](https://github.com/nervosnetwork/ckb/blob/develop/resource/ckb.toml#L61) needs to expose the support_protocols writeup(like rpc modules), now the user doesn't know how to change it.

Another question, do we need to turn off the filter protocol by default and let the user choose whether to turn it on or not?

quake:
> I think [this file](https://github.com/nervosnetwork/ckb/blob/develop/resource/ckb.toml#L61) needs to expose the `support_protocols` writeup(like rpc modules), now the user doesn't know how to change it.
> 

added in the new commit, please review again, thanks.

> Another question, do we need to turn off the filter protocol by default and let the user choose whether to turn it on or not?

I prefer turn on by default, this will result in more full nodes available to download the block filter for light client.

driftluo (approved):

zhangsoledad (approved):

PR #3674, opened 2022-10-28
fix: fix logging feature break

### LINKED ISSUE #3666: vm verify logging feature break
## Bug Report

after https://github.com/nervosnetwork/ckb/pull/2799 and https://github.com/nervosnetwork/ckb/pull/3083.

Transaction verification is divided into three processes:
- use `verify` during synchronization
- local RPC post, using `resume_verify`
- transactions propagated from other nodes, use `resume_verify` if the declared cycle is larger than the specified value, and `verify` if it is smaller. 

but `verify` has a logging feature like：
https://github.com/nervosnetwork/ckb/blob/b7f6ec7d1e9d997d092e5fa911fa9180028e124c/script/src/verify.rs#L425-L431

`resume_verify` miss this feature, this is an unexpected break change, I will re-add it back later

driftluo:

fix #3666 

resume verification also need logging to debug with some unexpected error

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


script/src/verify.rs
```diff
@@ -993,3 +991,11 @@ fn wrapping_cycles_sub(
     lhs.checked_sub(rhs)
         .ok_or_else(|| ScriptError::CyclesOverflow(lhs, rhs).source(group))
 }
+
+#[cfg(feature = "logging")]
+fn logging_feature(group: &Byte32, tx: &Byte32, error: &ScriptError) {
```
doitian, line 996: Not a good name.

Suggested changes:

``` rust
#[cfg(feature = "logging")]
mod logging {
fn log_script_error(...)
}
```

doitian (approved):

driftluo:
bors retry

PR #3029, opened 2021-09-12
feat: record recent reject

zhangsoledad:
### What is changed and how it works?

Proposal: https://talk.nervos.org/t/rfc-ckb-better-support-to-query-transaction-status/6043

### Check List 

Tests 

- Unit test
  db/src/tests/db_with_ttl.rs
  tx-pool/src/component/recent_reject.rs
- Integration test
  test/src/specs/tx_pool/different_txs_with_same_input.rs

Side effects

- Backward compatibility
  * The new configuration has default value if not provided, it‘s compatible.
  * No migration required.

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

---
This change is [Reviewable](https://reviewable.io/reviews/nervosnetwork/ckb/3029)

PR #3681, opened 2022-10-31
fix: set recent reject db log file limit to 10

driftluo:

In this [PR](https://github.com/nervosnetwork/ckb/pull/3029), we introduce the recent reject DB to record recently rejected transactions, which is implemented as rocksdb with TTL, using the default DB parameter configuration.

Since then, this db has been very heavily used on my machine until now:
```bash
dust -n 50 -d 1                                                                                                                      
4.0K   ┌── logs
8.0K   ├── ancient
212K   ├── tmp
2.8M   ├── network
164M   ├── indexer
1.5G   ├── tx_pool
 24G   ├── db
 26G ┌─┴ .        
```
However, the data inside is actually all `LOG.old.xxx` logs.

As a TTL database, its data storage is inherently time-limited, and it is obviously not reasonable to keep so many DB logs by default. 

By default, rocksdb keeps 1000 log files, which is too heavy for this DB, so this PR will change the default value to 10

### Check List 

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/db/src/db_with_ttl.rs
+++ b/db/src/db_with_ttl.rs
@@ -35,6 +35,7 @@ impl DBWithTTL {
         let mut opts = Options::default();
         opts.create_if_missing(true);
         opts.create_missing_column_families(true);
+        opts.set_keep_log_file_num(10);
 
         let cf_descriptors: Vec<_> = cf_names
             .into_iter()


doitian (commented):
The code looks good to me.

@zhangsoledad please look whether this is a reasonable change.

zhangsoledad:
[76dc5b9](https://github.com/nervosnetwork/ckb/pull/3681/commits/76dc5b92b67f327f91001131fa36019a7851a13e) is ok to me.

[3f049d3](https://github.com/nervosnetwork/ckb/pull/3681/commits/3f049d3e0edcb0105a30c109a359bc148d9ef589) 
The original option file we provided was trade-off for memory usage and speed of ckb synchronization, and there was no optimization for indexer, also I don't think these options are necessary for secondary_db, some options are not available for secondary_db. Given that there is no feedback from https://github.com/nervosnetwork/ckb-indexer about storage performance, I suggest to leave it unchanged and do special optimization if we find bottlenecks here in profiling later.

zhangsoledad (approved):

Issue #3673, opened 2022-10-28
Improved support for transaction fee estimation


zhangsoledad:
With the occurrence of ckb network congestion, a question has been more prominently exposed, how should ckb now go about setting transaction fees? Currently, the ckb nodes and the corresponding sdk do not have good support for transaction fee estimation, so this issue serves as a feature proposal to summarize and track the next actions.

- [x] Need an RPC that returns the cycles consumed by a transaction execution verification, although `dry_run_transaction` existed before, but its placement in the Experiment module needs some tweaking
    https://github.com/nervosnetwork/ckb/pull/3669

- [ ] Need an RPC to measure the fee_rate needed for a new transaction based on the fee_rate of the on-chain transaction. Here we need a solution that can be provided as soon as possible in the short term and a solution that is more optimal in the long term.
    *  In the short term, provide an RPC that returns the average or median of the transaction fee_rates over a historical interval on the chain, Users can use this value according to their needs and experience (e.g. estimate_fee_rate * 1.2 / 1.5 / 2) https://github.com/nervosnetwork/ckb/pull/3684
    
    * In the long term we need to provide better algorithms to measure the fee_rates required for new transactions to provide better guarantees, such as how much fee_rate can be used for new transactions to ensure that they can be confirmed within a certain time.

PR #3684, opened 2022-11-01
Improve support for cycles access

zhangsoledad:

Improve ckb's support for recording and accessing cycles as described in [3613](https://github.com/nervosnetwork/ckb/issues/3673), while implementing the fee_rate estimate short-term scheme.

* For cycles and tx_size, the appending of records is done in a compatible way, so that when a new version of the node is run, new data is written, cycles and tx_size are appended, and old data is read in a compatible way. RPC returns data that is not recorded as `null`.
* `get_block`/`get_block_by_number` adds a parameter `with_cycles` to choose whether to return cycles or not, mainly because some scenarios, such as indexer, do not need cycles and are performance sensitive, so an additional parameter is added to avoid the extra overhead needed to return cycles.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/component/mod.rs
+++ b/tx-pool/src/component/mod.rs
@@ -20,7 +20,7 @@ const DEFAULT_BYTES_PER_CYCLES: f64 = 0.000_170_571_4_f64;
 
 /// Virtual bytes(aka vbytes) is a concept to unify the size and cycles of a transaction,
 /// tx_pool use vbytes to estimate transaction fee rate.
-pub(crate) fn get_transaction_virtual_bytes(tx_size: usize, cycles: u64) -> u64 {
+pub fn get_transaction_virtual_bytes(tx_size: usize, cycles: u64) -> u64 {
     std::cmp::max(
         tx_size as u64,
         (cycles as f64 * DEFAULT_BYTES_PER_CYCLES) as u64,
--- a/tx-pool/src/component/mod.rs
+++ b/tx-pool/src/component/mod.rs
@@ -12,17 +12,3 @@ pub(crate) mod recent_reject;
 mod tests;
 
 pub use self::entry::TxEntry;
-
-/// Equal to MAX_BLOCK_BYTES / MAX_BLOCK_CYCLES, see ckb-chain-spec.
-/// The precision is set so that the difference between MAX_BLOCK_CYCLES * DEFAULT_BYTES_PER_CYCLES
-/// and MAX_BLOCK_BYTES is less than 1.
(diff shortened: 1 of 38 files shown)


tx-pool/src/component/mod.rs
```diff
@@ -20,7 +20,7 @@ const DEFAULT_BYTES_PER_CYCLES: f64 = 0.000_170_571_4_f64;
 
 /// Virtual bytes(aka vbytes) is a concept to unify the size and cycles of a transaction,
 /// tx_pool use vbytes to estimate transaction fee rate.
-pub(crate) fn get_transaction_virtual_bytes(tx_size: usize, cycles: u64) -> u64 {
```
yangby-cryptape, line 23: Could this function moved to `ckb-types`?

It's too heavy for third-party applications to call this function, they have to depend on `rocksdb`, `tokio` and so on.

driftluo (approved):

tx-pool/src/component/mod.rs
zhangsoledad: It's fine to expose it. I didn't want to expose this function, it's so simple that there's no need to import it from `ckb-types`, other non-rust projects can't use it either.

yangby-cryptape: If expose it, at least, users could find several documentation lines.
It's really hard to find it.
Last week, a former ckb-core member had asked me for how to calculate `vbytes`.

If I was correct, there is no RFC mention there is a concept of "vbytes" in CKB and the magic number `DEFAULT_BYTES_PER_CYCLES`.
From normal users' point, they may not know there is a concept of "vbyes" in CKB.

And, no standard (RFC) means there is no promise whether the number `DEFAULT_BYTES_PER_CYCLES` would be changed, so depend on an official API is more safe than hard-code.

**Maybe we should standardize this concept, better than force users to read our code to find the magic number and the cryptic formula.**

### References

- [Section "Paying for Transaction Fees" in CKB RFC-0015](https://github.com/nervosnetwork/rfcs/blob/af5c8f32ef6142abe520c6e6cc688049db125f08/rfcs/0015-ckb-cryptoeconomics/0015-ckb-cryptoeconomics.md#paying-for-transaction-fees)
- [Section "Computation Cost and Transaction Fees" in CKB RFC-0002](https://github.com/nervosnetwork/rfcs/blob/af5c8f32ef6142abe520c6e6cc688049db125f08/rfcs/0002-ckb/0002-ckb.md#52-computation-cost-and-transaction-fees)

yangby-cryptape (approved):

Issue #3668, opened 2022-10-28
Accelerated Sync attempt


driftluo:
As the main network runs longer, the time required for full Sync is gradually increasing, and a new round of Sync optimization begins, we need to find performance bottlenecks to make Sync speed up further and facilitate users to join the network.

Mission focus.
- Finding performance bottlenecks
- Download and commit block async
- Verification parallelization
- Provide a wiki for users to configure themselves to maximize download speed

@eval-exec 

Related PRs: https://github.com/nervosnetwork/ckb/pull/3629
Related wiki: 
- https://github.com/nervosnetwork/ckb/wiki/Synchronization-optimization
- https://github.com/nervosnetwork/ckb/wiki/Deployment-of-ckb

eval-exec:
- Performance bottlenecks
- script group verification parallelization:
    - [x] import rust-rocksdb which contains `get_pined`
    - [x] make `TransactionScriptsVerifier` thread safe
    - [ ] parallelize `TransactionScriptsVerifier::verify()`

kfulbright:
@eval-exec Did you ever get someone to review your PR?

eval-exec:
> @eval-exec Did you ever get someone to review your PR?

Yeah.

eval-exec:
# Analyze Results

On a AWS `c5.xlarge` (4C 8G) machine, I get CPU and IO usage by `pidstat -p $(pidof ckb) -u -p}` 
## CPU Usage
![tmp](https://user-images.githubusercontent.com/46400566/222641800-75cb3b2a-7226-4e10-b1ad-dc7fb7ccf28a.png)

As shown in the figure above, the average CPU usage is `20% ~ 30%`
## IO usage
TODO,

PR #3689, opened 2022-11-03
Enhance: Add blocks queue before block process to reduce IBD timecost

eval-exec:

Issue Number:  #3668 

When ckb is in IBD mode, newly downloaded blocks will pushed into a blocks queue, and another thread will consume the blocks queue. The queue will be dropped and the consume thread will exit after IBD mode end.

This change reduce the timecost of IBD sync progress by 5%.
tmp

### Todos
- [x] A comparison chart of sync progress by this PR
- [x] memory usage report
- [x] compare bounded and unbounded queue
- [x] determine the size of queue
- [x] fix CI failure
- [ ] add benchmark script
- [ ] add unit test for "block_queue will drop after IBD mode end".

What's Changed:

### Related changes

- Add `block_queue` field to `Synchronizer` 
- Add a thread to consume `block_queue`

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- The `block_queue` may use up to 45MB memory in IBD mode, but the `block_queue` will be dropped after IBD mode end.

### Release note 

```release-note
```

### DIFF
--- a/sync/src/synchronizer/block_process.rs
+++ b/sync/src/synchronizer/block_process.rs
@@ -38,7 +38,13 @@ impl<'a> BlockProcess<'a> {
             // determine should we drop the queue and consumer handle
 
             let mut queue_is_empty = false;
-            if let Some(queue) = self.synchronizer.block_queue.read().unwrap().as_ref() {
+            if let Some(queue) = self
+                .synchronizer
+                .block_queue
+                .read()
+                .expect("BlockProcess wants to acquire read lock on block_queue to check if block_queue is empty, but it has poisoned")
+                .as_ref()
+            {
                 if queue.is_empty() {
                     queue_is_empty = true;
                 }
@@ -46,13 +52,17 @@ impl<'a> BlockProcess<'a> {
 
             if queue_is_empty {
                 // drop block_queue and consumer handle
-                let _ = self.synchronizer.block_queue.write().unwrap().take();
-                let _ = self
+                let block_queue = self.synchronizer.block_queue.write().expect("BlockProcess wants to acquire write lock on block_queue to drop block_queue, but it has poisoned").take();
+                drop(block_queue);
+
+                let consumer_handle = self
                     .synchronizer
                     .block_queue_consumer_handle
                     .write()
-                    .unwrap()
+                    .expect("BlockProcess wants to acquire write lock on block_queue_consumer_handle to drop block_queue_consumer_handle , but it has poisoned")
                     .take();
+                drop(consumer_handle);
+
                 info!("both block queue and consumer handle are dropped");
             }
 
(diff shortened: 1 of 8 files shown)


sync/src/synchronizer/block_process.rs
```diff
@@ -32,14 +34,72 @@ impl<'a> BlockProcess<'a> {
         let shared = self.synchronizer.shared();
         let state = shared.state();
 
-        if state.new_block_received(&block) {
-            if let Err(err) = self.synchronizer.process_new_block(block.clone()) {
+        if !shared.active_chain().is_initial_block_download() {
+            // determine should we drop the queue and consumer handle
+
+            let mut queue_is_empty = false;
+            if let Some(queue) = self.synchronizer.block_queue.read().unwrap().as_ref() {
```
zhangsoledad, line 41: We have specified that unwrap is not allowed to appear anywhere other than in tests,

> Using expect instead of unwrap and providing good error messages can convey your intent and make tracking down the source of a panic easier. ——https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html

zhangsoledad (changes requested):

sync/src/synchronizer/block_process.rs
zhangsoledad: more https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used

sync/src/synchronizer/block_process.rs
```diff
@@ -32,14 +34,87 @@ impl<'a> BlockProcess<'a> {
...
+                    queue_op
+                        .as_ref()
+                        .map(|queue| queue.is_empty())
+                        .unwrap_or(false)
+                })
+                .unwrap_or(false);
+
+            if queue_is_empty {
+                // drop block_queue and consumer handle
+                let _ = self
```
zhangsoledad, line 54: I suggest using `expect` directly here, like `read().expect("block_queue lock poisoned on a panic")`
So when this happens unexpectedly, we will be able to know a panic occurs while a write operation is in progress.
let_underscore is also bad, although there is some let_underscore existing now.

zhangsoledad:
Is there a script available for reproduce the benchmark results, or steps description?

eval-exec:
> Is there a script available for reproduce the benchmark results, or steps description?

Ok, I will add it later.

eval-exec:
I just had a new idea: `thread::sleep()` in `BlockConsumer` thread may cause performance issues. Maybe in `BlockFetcher`, it would be better to wake up `BlockConsumer` thread when `block_queue` is 80% full. 

In `BlockConsumer` thread: call `thread::park()` after `BlockConsumer` consume all blocks in `block_queue`
In `BlockFetcher`: call `thread::unparK()` when the block_queue is almost 80% full.

I'm working on this.

driftluo:
> In BlockConsumer thread: call thread::park() after BlockConsumer consume all blocks in block_queue
In BlockProcess: call thread::unparK() when the block_queue is almost 80% full.

I think `thread::park_timeout()` and 80% `thread::unpark` it better

eval-exec:
Due to some design changes, this PR was blocked until the `TODO 1` .

Now there are 3 TODOs:
`TODO 1`. Move the `OrphanBlockPool` from `Synchronizer` to `ChainService`. 
`TODO 2`. Add an `block_queue` to `ChainService` to make "blocks download" and "blocks validation" asynchronous, and use `tokio::Task` to consume the `block_queue`.
`TODO 3`. A credible benchmark solution / test script.

PR #3675, opened 2022-10-28
fix(light-client): subtract overflow when try to get chain root for genesis block

yangby-cryptape:
### What problem does this PR solve?

Problem Summary:

The genesis block could NOT have a chain root, since chain root is a state for previous blocks of a chain.
Even for dev chains.

Hence, we have to skip fetching a chain root for the genesis block.

### Check List

Tests

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

### Release note

```release-note
Title Only: Include only the PR title in the release note.
```

PR #3697, opened 2022-11-09
fix(light-client): do not skip the genesis block for last state proofs

yangby-cryptape:

Should not skip the genesis block for last state proofs.

For example, when current tip header at 1, then the last state proof have to contain the genesis block.

Since #3675, the genesis block will use a default struct as its parent chain root, so we don't have to skip it anymore.

https://github.com/nervosnetwork/ckb/blob/5aa8b426d0aef48c5a854a74176b5eb9b85896fb/util/light-client-protocol-server/src/lib.rs#L125-L127

And, for more security, we should check if the parent chain root of a genesis block is a default struct.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/util/light-client-protocol-server/src/tests/components/get_last_state_proof.rs
+++ b/util/light-client-protocol-server/src/tests/components/get_last_state_proof.rs
@@ -1,0 +1,95 @@
+use ckb_merkle_mountain_range::{leaf_index_to_mmr_size, leaf_index_to_pos};
+use ckb_network::{CKBProtocolHandler, PeerIndex, SupportProtocols};
+use ckb_types::{
+    packed,
+    prelude::*,
+    utilities::merkle_mountain_range::{MMRProof, VerifiableHeader},
+};
+
+use crate::tests::{
+    prelude::*,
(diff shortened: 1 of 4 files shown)


util/light-client-protocol-server/src/tests/components/get_last_state_proof.rs
```diff
@@ -0,0 +1,99 @@
...
+                    let position = leaf_index_to_pos(index);
+                    let digest = header.digest();
+                    digest.verify()?;
+                    Ok((position, digest))
+                })
+                .collect::<Result<Vec<_>, String>>();
+            assert!(result.is_ok(), "failed since {}", result.unwrap_err());
+            result.unwrap()
+        };
+        let result = proof.verify(parent_chain_root, digests_with_positions);
```
yangby-cryptape, line 91: Without the fix, this line will throw the error "Corrupted proof".

TheWaWaR (approved):

quake (approved):

yangby-cryptape:
bors merge

PR #3698, opened 2022-11-10
fix(tx-pool): potential leaks when MaximumAncestors occurs

zhangsoledad:

Fix a potential issue where the transaction pool was not properly cleared of state when the MaximumAncestors limit was triggered

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


driftluo (approved):
If I understand correctly, the problem with the previous code was that the entry information was added to the deps before deciding whether the current entry could be inserted or not

quake (approved):

PR #3702, opened 2022-11-14
fix: unknown list should get header from headermap and db

driftluo:

Thanks to @eval-exec for pointing out these problems

ref: #3668

#### unknown list bug
`unknown_list` is the key to enabling multi-node parallel download in ibd phase. Its implementation is the get locator algorithm:
https://github.com/nervosnetwork/ckb/blob/1ce2b34cbad6c9573cb2dbb50af8de1dce4b027c/sync/src/types/mod.rs#L2032-L2035

Note this line in the algorithm, all results are added to the genesis hash again at the end

If `insert_peer_unknown_header_list` can find the corresponding value at the beginning of a new connection, this node can be used as one of the download sources:
https://github.com/nervosnetwork/ckb/blob/1ce2b34cbad6c9573cb2dbb50af8de1dce4b027c/sync/src/types/mod.rs#L1875-L1889

Otherwise, `try_update_best_known_with_unknown_header_list` will get stuck in a dead loop on the genesis hash and the node will not be able to become a download source again in the connected state:

https://github.com/nervosnetwork/ckb/blob/1ce2b34cbad6c9573cb2dbb50af8de1dce4b027c/sync/src/types/mod.rs#L1891-L1903

**Finally**, the `try_update_best_known_with_unknown_header_list` function is removed and `get_header_view` is used instead of just looking from the headermap.

Because if the block is inserted into the DB and the header map is cleaned up, the unknown list may get stuck in this situation where the hash cannot be found, resulting in reduce the download of the source. 

#### eviction bug

In the ibd process, a node can only request headers data from one peer at a time. For headers, multiple data sources will result in duplicate downloads.

When synchronization starts, the headers_sync_controller of the node will be initialized to the normal state:

https://github.com/nervosnetwork/ckb/blob/1ce2b34cbad6c9573cb2dbb50af8de1dce4b027c/sync/src/synchronizer/mod.rs#L370-L380

`eviction`'s remaining judgments will likely result in simultaneous requests for headers to multiple data sources, which is unnecessary in the ibd phase and should be disabled

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -81,10 +81,19 @@ impl<'a> BlockFetcher<'a> {
         // Update `best_known_header` based on `unknown_header_list`. It must be involved before
         // our acquiring the newest `best_known_header`.
         if let IBDState::In = self.ibd {
-            self.synchronizer
-                .shared
-                .state()
-                .try_update_best_known_with_unknown_header_list(self.peer)
+            let state = self.synchronizer.shared.state();
+            // unknown list is an ordered list, sorted from highest to lowest,
+            // when header hash unknown, break loop is ok
+            while let Some(hash) = state.peers().take_unknown_last(self.peer) {
+                // Here we need to first try search from headermap, if not, fallback to search from the db.
+                // if not search from db, it can stuck here when the headermap may have been removed just as the block was downloaded
+                if let Some(header) = self.synchronizer.shared.get_header_view(&hash, None) {
+                    state.peers().may_set_best_known_header(self.peer, header);
+                } else {
+                    state.peers().insert_unknown_header_hash(self.peer, hash);
+                    break;
+                }
+            }
         }
 
         // This peer has nothing interesting.
--- a/sync/src/synchronizer/mod.rs
+++ b/sync/src/synchronizer/mod.rs
@@ -379,6 +379,15 @@ impl Synchronizer {
                 }
             }
 
+            // On ibd, node should only have one peer to sync headers, and it's state can control by
+            // headers_sync_controller.
+            //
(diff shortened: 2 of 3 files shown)


eval-exec (approved):
LGTM.

sync/src/synchronizer/get_headers_process.rs
```diff
@@ -59,6 +59,10 @@ impl<'a> GetHeadersProcess<'a> {
             let state = self.synchronizer.shared.state();
             if let Some(flag) = state.peers().get_flag(self.peer) {
                 if flag.is_outbound || flag.is_whitelist || flag.is_protect {
+                    // Remove genesis hash on locators, because `try_update_best_known_with_unknown_header_list` can use
```
quake, line 62: fn `try_update_best_known_with_unknown_header_list` has been deleted in this PR, please reword the comment

sync/src/synchronizer/get_headers_process.rs
```diff
@@ -59,6 +59,10 @@ impl<'a> GetHeadersProcess<'a> {
...
+                    // genesis hash on headermap search, it will case some peers can not set best knonwn header and causes
+                    // the ibd download source to be lost
+                    block_locator_hashes.pop();
```
quake, line 65: I don't think we need to remove the genesis hash from the locator since we changed the code to loading from db first in this PR

quake (changes requested):

sync/src/synchronizer/get_headers_process.rs
driftluo: `get_header_view` will try searching from headermap first, if can't find it, failback to search DB.

quake (approved):

zhangsoledad (approved):

PR #3707, opened 2022-11-17
ckb-network: add try_from<ProtocolId> trait for SupportProtocols

LiRiu:

Enhance the readability of network module

add a try_from trait for SupportProtocols 

```rust
impl TryFrom<ProtocolId> for SupportProtocols {
    type Error = ();
    fn try_from(proto_id: ProtocolId) -> Result<Self, Self::Error> {
        match proto_id.value() {
            0 => Ok(SupportProtocols::Ping),
            1 => Ok(SupportProtocols::Discovery),
            2 => Ok(SupportProtocols::Identify),
            3 => Ok(SupportProtocols::Feeler),
            4 => Ok(SupportProtocols::DisconnectMessage),
            100 => Ok(SupportProtocols::Sync),
            101 => Ok(SupportProtocols::RelayV2),
            102 => Ok(SupportProtocols::Time),
            110 => Ok(SupportProtocols::Alert),
            120 => Ok(SupportProtocols::LightClient),
            121 => Ok(SupportProtocols::Filter),
            _ => Err(()),
        }
    }
}
```

so that we can not only log proto message as int, but also as protocol_name.
```rust
    async fn received(&mut self, context: ProtocolContextMutRef<'_>, data: bytes::Bytes) {
        let proto = SupportProtocols::try_from(context.proto_id).unwrap();
        println!(
            "Received new message: proto [{}] data {:?}",
            proto.name(),
            str::from_utf8(data.as_ref()).unwrap(),
        );
    }
```

### Release note
None

### DIFF
--- a/network/src/protocols/support_protocols.rs
+++ b/network/src/protocols/support_protocols.rs
@@ -158,3 +158,23 @@ impl From<SupportProtocols> for MetaBuilder {
             })
     }
 }
+
+impl TryFrom<ProtocolId> for SupportProtocols {
+    type Error = ();
+    fn try_from(proto_id: ProtocolId) -> Result<Self, Self::Error> {
+        match proto_id.value() {
+            0 => Ok(SupportProtocols::Ping),
+            1 => Ok(SupportProtocols::Discovery),
+            2 => Ok(SupportProtocols::Identify),
+            3 => Ok(SupportProtocols::Feeler),
+            4 => Ok(SupportProtocols::DisconnectMessage),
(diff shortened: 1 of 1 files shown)


quake:
the protocol name can be fetched from context's protocols map,  no need to add conversion code:
```
let protocol_name: &str = context.protocols().get(&self.proto_id).expect("checked protocol mapping").name.as_ref();
```

LiRiu:
感谢

PR #3709, opened 2022-11-17
Move OrphanBlockPool from Synchronizer to ChainService

eval-exec:

This PR is a pre-task for "make blocks download and blocks verification asynchronous". [tasks](https://github.com/nervosnetwork/ckb/pull/3689#issuecomment-1318365944)

The purpose of this PR is to move `OrphanBlockPool` from `Synchronizer` to `ChainService` so that `Synchronizer` will no longer be aware of `OrphanBlockPool`.

- move `header_map`, `block_status_map` and `assume_valid_target` from `Synchronizer::SyncShared::SyncState` to `ckb-shared::Shared`
- move `OrphanBlockPool` to `ChainService`
- move `HeaderMap`, `HeaderView` and `BLOCK_STATUS` to `ckb-shared` package, because `ckb-chain` and `ckb-sync` depend on them. 

### Related changes

- added a copy of `OrphanBlockPool` to `ChainController`, because `Relayer` and `NetRpc` need to access `OrphanBlockPool` via `ChainController` 
- protect `assume_valid_target` by `RwLock` instead of `Mutex`
- remove `Relayer::SEARCH_ORPHAN_POOL_TOKEN` timer job, and move `orphan_pool().clean_expired_blocks()` to `ChainService::insert_block()`

### Check List 

Tests 

- Unit test
- Integration test

Side effects

### Release note 

```release-note
```


eval-exec:
This PR is not going to be merged at the moment, because there is a higher priority work: Building a credible measurement method for measuring the timecost of the ckb synchronization process. 

Like: start several ckb nodes, use the `ckb import` subcommand to quickly increase the tip height to the specified target, and use the `netem` tool to increase the network delay or packet loss rate of some nodes, so as to simulate the real network environment.

# helpful links:
https://man7.org/linux/man-pages/man8/tc-netem.8.html

PR #3713, opened 2022-11-22
fix: check in-pool chidren for all newly added tx

zhangsoledad:

Check the in-pool chidren of all newly added transactions and no longer distinguish the entry source of newly added transactions.

Removed the separate `add_entry_from_detached ` interface, all new add transactions will be checked in-pool chidren

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


quake (approved):

quake:
bors r=quake

PR #3727, opened 2022-11-25
fix: blocktemplate dao potential inconsistent with transactions

zhangsoledad:

In generating the blocktemplate is probably due to asynchronous reasons, resulting in the calculation of DAO is to filter out some conflicting transactions, but in filling in the blocktemplate transaction field did not use the filtered transactions, so led to potential inconsistency problems.

Ensure that the transaction used in filling the blocktemplate field is the same as the used in calculating the DAO

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/block_assembler/mod.rs
+++ b/tx-pool/src/block_assembler/mod.rs
@@ -57,6 +57,34 @@ pub(crate) struct TemplateSize {
     pub(crate) total: usize,
 }
 
+impl TemplateSize {
+    pub(crate) fn calc_total_by_proposals(&self, new_proposals_size: usize) -> usize {
+        if new_proposals_size > self.proposals {
+            self.total
+                .saturating_add(new_proposals_size - self.proposals)
+        } else {
+            self.total
+                .saturating_sub(self.proposals - new_proposals_size)
+        }
+    }
+
+    pub(crate) fn calc_total_by_uncles(&self, new_uncles_size: usize) -> usize {
+        if new_uncles_size > self.uncles {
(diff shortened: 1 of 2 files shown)


tx-pool/src/block_assembler/mod.rs
```diff
@@ -549,20 +560,17 @@ impl BlockAssembler {
         current_epoch: &EpochExt,
         cellbase: TransactionView,
         entries: Vec<TxEntry>,
-    ) -> Result<Byte32, AnyError> {
+    ) -> Result<(Byte32, Vec<TxEntry>, usize), AnyError> {
```
quake, line 563: suggest to return `Result<(Byte32, Vec<TxEntry>)`, the total tx size can be calculated by `tx_entries.iter().map(entry.size).sum()` if needed

tx-pool/src/block_assembler/mod.rs
```diff
@@ -162,25 +163,25 @@ impl BlockAssembler {
...
                 tx_pool_reader.package_txs(max_block_cycles, txs_size_limit);
-            (proposals, txs, txs_size, basic_size)
+            (proposals, txs, basic_size)
         };
 
         let proposals_size = proposals.len() * ProposalShortId::serialized_size();
-        let total_size = basic_size + txs_size;
-
-        let dao = Self::calc_dao(
+        let (dao, checked, txs_size) = Self::calc_dao(
```
quake, line 172: suggest to rename `checked` to `checked_txs`

quake (approved):

PR #3750, opened 2022-12-09
fix: notify message blocking

zhangsoledad:

Notify service is essentially a message bus, which should push messages to subscribers in non-blocking form, and push to subscribers should be treated as a remote call, so the timeout setting should be added.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/notify/src/lib.rs
+++ b/notify/src/lib.rs
@@ -3,16 +3,19 @@ use ckb_app_config::NotifyConfig;
 use ckb_async_runtime::Handle;
 use ckb_logger::{debug, error, trace};
 use ckb_stop_handler::{SignalSender, StopHandler};
+use ckb_types::packed::Byte32;
 use ckb_types::{
     core::{tx_pool::Reject, BlockView},
     packed::Alert,
 };
-use std::collections::HashMap;
-use std::process::Command;
+use std::{collections::HashMap, time::Duration};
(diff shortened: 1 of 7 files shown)


notify/src/lib.rs
```diff
@@ -173,22 +238,43 @@ impl NotifyService {
...
         // notify all subscribers
-        for subscriber in self.new_block_subscribers.values() {
-            let _ = subscriber.send(block.clone()).await;
+        for subscriber in self.new_block_subscribers.values().cloned() {
+            let block = block.clone();
+            self.handle.spawn(async move {
+                if let Err(e) = subscriber.send(block).await {
+                    error!("notify new block error {}", e);
+                }
+            });
```
quake, line 251: Since rust lacks the sugar `clone_to_closure`, we can only clone var manually, but since we have already cloned the `block`, I suggest cloning the `subscriber` again, which will remove the `clippy::unnecessary_to_owned` hint also:

```suggestion
        for subscriber in self.new_block_subscribers.values() {
            let subscriber = subscriber.clone();
            let block = block.clone();
            self.handle.spawn(async move {
                if let Err(e) = subscriber.send(block).await {
                    error!("notify new block error {}", e);
                }
            });
        }
```

quake (approved):

PR #3757, opened 2022-12-15
fix: fee_rate statistics target limit

zhangsoledad:

The previous implementation had a bug that limited the maximum value of the target only when the input target was an even number

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```


quake (approved):

quake:
bors r=quake

PR #3759, opened 2022-12-15
Fix: block_filter: refresh_snapshot() in build_filter_data may race with ChainService

eval-exec:

Problem Summary:

`block-filter::build_filter_data` call `refresh_snapshot` on every `new_block_watcher.changed()` here:
https://github.com/nervosnetwork/ckb/blob/d3fc58fa24a8c26783c604122c643602dfb527fb/block-filter/src/filter.rs#L94
and `ChainService` may update `snaphsot` simultaneously in here:
https://github.com/nervosnetwork/ckb/blob/d3fc58fa24a8c26783c604122c643602dfb527fb/chain/src/chain.rs#L480-L486

For example: 
1. In `fn build_filter_data`, `refresh_snapshot` load a `Snapshot` with tip_header(100)  
2. `ChainService` received a new_tip_header(101), and store it into snapshot
3. In `fn build_filter_data`, `refresh_snapshot` store the `Snapshot` with tip_header(100), this cause `tip_header` goes backwards. 

What's Changed:

remove `refresh_snapshot()` in `block-filter::build_filter_data`

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/sync/src/filter/get_block_filter_hashes_process.rs
+++ b/sync/src/filter/get_block_filter_hashes_process.rs
@@ -2,7 +2,6 @@ use crate::filter::{block_filter_hash, BlockFilter};
 use crate::utils::send_message_to;
 use crate::{attempt, Status};
 use ckb_network::{CKBProtocolContext, PeerIndex};
-use ckb_store::ChainStore;
 use ckb_types::{core::BlockNumber, packed, prelude::*};
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


eval-exec:
The CI job was failed after I removed `refresh_snapshot()` in `block-filter::build_filter_data`.

When a new tip block comes, `ChainService` will get a new RocksDB's `StoreSnapshot` and store it to snapshot after commit the new tip block into RocksDB. Then `block-filter` get notified and build filter data for the new tip block.

In `build_filter_data_for_block`, it just only commit the block filter data into rocksdb via `StoreTransaction`, but the snapshot's  inner value(`RocksDBSnapshot`) doesn't contains the new tip block's block filter data.
https://github.com/nervosnetwork/ckb/blob/2ad82e6e8aa19c8ea5cd45e3807ccb69434b9c9d/block-filter/src/filter.rs#L144-L149

but in `ChainRpc::get_block_filter`, it will load a snapshot which doesn't contains the new tip blocks' block filter data, since I removed `refresh_snapshot()` from `block-filter::build_filter_data`. So the CI job failed.
https://github.com/nervosnetwork/ckb/blob/cd1e2f891c171320978a6a796184d189ae94e0fe/rpc/src/module/chain.rs#L1646-L1653

sync/src/filter/get_block_filter_hashes_process.rs
```diff
@@ -31,26 +30,27 @@ impl<'a> GetBlockFilterHashesProcess<'a> {
     }
 
     pub fn execute(self) -> Status {
-        let snapshot = self.filter.shared.shared().snapshot();
+        let active_chain = self.filter.shared.active_chain();
         let start_number: BlockNumber = self.message.to_entity().start_number().unpack();
-        let tip_number: BlockNumber = snapshot.get_tip_header().expect("tip stored").number();
+        let tip_number: BlockNumber = active_chain.tip_header().number();
```
quake, line 35:
```suggestion
        let tip_number: BlockNumber = active_chain.tip_number();
```

quake (approved):

driftluo (approved):

quake:
bors r=quake,driftluo

PR #3777, opened 2022-12-28
Remove nervosnetwork/faketime, replace cargo test by nextest-rs/nextest

### LINKED ISSUE #3729: Fix and simplify timestamp-related tests
Now ckb is using [faketime](https://github.com/nervosnetwork/faketime) to mock time and test for timestamps, faketime is slightly complicated and does not work correctly in some cases, for example, if you specify `--test-threads=1`, timestamp related tests will fail, so we need a new solution to fix these tests, the simplest way should be thread_local or global variable, combined with https://github.com/palfrey/serial_test/.

eval-exec:

Issue Number: close #3729 

Problem Summary:

This PR want to remove nervosnetwork/faketime, and simply use static atomics and nextest to organize ckb's unit tests. 

The `nervosnetwork/faketime` is complicated, and when we specified `--test-threads=1` flag to `cargo test`, these `thread_local` ( 
https://github.com/nervosnetwork/faketime/blob/1ead0aba139f0893d04aa95c35e0179c3b4482c7/src/faketime.rs#L109-L115 ) behave like global variables, then some tests will fail.

The `cargo test` run all unit test in a single process, and many unit tests may run on a same thread. And there exists some faketime related test want to change the whole process's faketime by set current process's environment like this:
https://github.com/nervosnetwork/ckb/blob/fafa37d979bd83f867358efb7aa176a0ec295bbd/chain/src/tests/block_assembler.rs#L135

this may affect other non-faketime related unit tests.

Fortunately, there is [nextest-rs/nextest](https://github.com/nextest-rs/nextest) which will run every unit function on single process, providing process-level isolation granularity.

What's Changed:

### Related changes

- remove nervosnetwork/faketime
- add `util/systemtime` to provide real systemtime, and provide faketime when `enable_faketime` feature is enabled
- use [nextest-rs/nextest](https://github.com/nextest-rs/nextest) to run all unit tests

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


eval-exec:
Try to make github ci action use `cargo-nextest`. 
But get an error:
https://github.com/nervosnetwork/ckb/actions/runs/3799433532/jobs/6461954529
```
error: failed to compile `cargo-nextest v0.9.47`, intermediate artifacts can be found at `C:\Users\ADMINI~1\AppData\Local\Temp\cargo-installaoztQa`

Caused by:
  package `cargo-nextest v0.9.47` cannot be built because it requires rustc 1.62 or newer, while the currently active rustc version is 1.61.0
Error: Process completed with exit code 1.
```

Trying to install an older version of cargo-nextest on ci_*_windows

tx-pool/src/component/orphan.rs
```diff
@@ -29,7 +29,7 @@ impl Entry {
             tx,
             peer,
             cycle,
-            expires_at: faketime::unix_time().as_secs() + ORPHAN_TX_EXPIRE_TIME,
```
zhangsoledad, line 32: Should provide function equivalent to the original `unix_time` here

util/systemtime/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+// Indicate whether faketime is enabled
+#[cfg(feature = "enable_faketime")]
+static FAKETIME_ENABLED: AtomicBool = AtomicBool::new(false);
+
+// Get real system's timestamp in millis
+fn system_time_as_millis() -> u64 {
+    let duration = std::time::SystemTime::now()
+        .duration_since(std::time::SystemTime::UNIX_EPOCH)
+        .expect("SystemTime before UNIX EPOCH!");
+    duration.as_secs() * 1000 + u64::from(duration.subsec_millis())
```
zhangsoledad, line 21: use https://doc.rust-lang.org/stable/std/time/struct.Duration.html#method.as_millis directly

tx-pool/src/component/orphan.rs
eval-exec: > Should provide function equivalent to the original `unix_time` here

Added `unix_time` here:

https://github.com/nervosnetwork/ckb/pull/3777/commits/c5d61f678bb5f19346d49a7120cc2f26c62ed3df#diff-e473f752faba52c41617830aeab8ef9e1437ce6b16e98d5089d2a391d60226f4R45-R50

PR #3783, opened 2023-01-05
feat: fee-rate estimator

zhangsoledad:

The algorithm of https://bitcoiner.live/?tab=info implements the fee-rate estimator to achieve the goal in https://github.com/nervosnetwork/ckb/issues/3673, and there are still some limitations, Weight-Units Flow algorithm is mainly for sampling and analyzing the transactions in tx-pool, so it will be available only after the node has been up for a warm-up period, if the estimator is accessed during the warm-up period it will fallback to the mean of the past historical data (21 blocks)

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


util/fee-rate-estimator/src/weight_flow.rs
```diff
@@ -0,0 +1,616 @@
...
+
+impl Ord for TxStatus {
+    fn cmp(&self, other: &Self) -> ::std::cmp::Ordering {
+        let order = self.fee_rate.cmp(&other.fee_rate);
+        if order == ::std::cmp::Ordering::Equal {
+            other.weight.cmp(&self.weight)
+        } else {
+            order
+        }
+    }
```
eval-exec, line 82: How about this:
```rust
impl Ord for TxStatus {
    fn cmp(&self, other: &Self) -> ::std::cmp::Ordering {
        self.fee_rate
            .cmp(&other.fee_rate)
            .then_with(|| other.weight.cmp(&self.weight))
    }
}
```

util/fee-rate-estimator/src/weight_flow.rs
```diff
@@ -0,0 +1,616 @@
...
+
+impl FeeEstimator {
+    fn submit_transaction(&mut self, tx: &types::Transaction) {
+        let current_dt = tx.seen_dt();
+        let expired_dt = current_dt - Self::historical_dur(self.max_target_dur);
+        let weight = get_transaction_weight(tx.size() as usize, tx.cycles());
+        let fee_rate = FeeRate::calculate(Capacity::shannons(tx.fee()), weight);
+        let new_tx = TxAdded::new(tx.hash(), weight, fee_rate, current_dt);
+        self.txs.add_transaction(new_tx);
+        self.txs.expire(expired_dt);
```
eval-exec, line 566: `self.txs.expire(expired_dt);` have a return value, but it's never used.

util/fee-rate-estimator/src/weight_flow.rs
```diff
@@ -0,0 +1,616 @@
...
+            .rev()
+            .skip_while(|tx| tx.added_dt < expired_dt)
+            .count();
+        let total = self.0.len();
+        if count > 0 && total >= count {
+            self.0.truncate(total - count);
+        }
+        count
+    }
+
```
eval-exec, line 117: How about this:
```rust
    fn expire(&mut self, expired_dt: Duration) -> usize {
        let original_len = self.0.len();
        self.0.retain(|tx| tx.added_dt >= expired_dt);
        original_len - self.0.len()
    }
```

PR #3801, opened 2023-01-16
fix: fix identify unregister

driftluo:

we introduced peer flags network-wide propagation in https://github.com/nervosnetwork/ckb/pull/3595

But there is a problem here, only the outbound connection time needs to be refreshed when disconnecting because the global state of connected peers may be deleted first and the flags obtained may be empty, which will reset the correct flags to the default state

### Check List

Tests

- Unit test
- Integration test

Side effects

- Performance regression

### Release note

```release-note
```

### DIFF
--- a/network/src/peer_store/peer_store_impl.rs
+++ b/network/src/peer_store/peer_store_impl.rs
@@ -85,6 +85,16 @@ impl PeerStore {
         ));
     }
 
+    /// Update outbound peer last connected ms
+    pub fn update_outbound_addr_last_connected_ms(&mut self, addr: Multiaddr) {
+        if self.ban_list.is_addr_banned(&addr) {
+            return;
+        }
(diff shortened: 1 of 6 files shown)


network/src/peer_store/peer_store_impl.rs
```diff
@@ -85,6 +85,16 @@ impl PeerStore {
         ));
     }
 
+    /// Change outbound peer last connected ms
+    pub fn change_outbound_addr_last_connected_ms(&mut self, addr: Multiaddr) {
```
quake, line 89: please follow the naming convention, rename the fn to `update_xxx`  when updating the struct field:
```
```suggestion
    pub fn update_outbound_addr_last_connected_ms(&mut self, addr: Multiaddr) {
```

quake (approved):

PR #3825, opened 2023-02-08
fix: resolve the error of block filter hash calculation

quake:

in current develop branch, the hash calculation of block filter is wrong, it should be `blake2b(parent_block_filter_hash | current_block_filter_data)`

this PR fixed the hash calculation related code, and saved the hash results to rocksdb to improve performance. a db migration is added to save the hash of existing block filter, it takes about 1.5 hours to run on my machine for testnet,

### Check List 

Tests 

- Unit test
- Integration test

Side effects

This PR break the backward compatibility of rpc `get_block_filter` and block filter protocol's message `BlockFilterHashes` and `BlockFilterCheckPoints`, however, this rpc and messages are not used in current light client implementation (v0.2.1), and considering that only the testnet has activated the light client protocol, I think this PR without compatibility code is OK

### Release note 

```release-note
```


util/launcher/src/migrations/rebuild_block_filter.rs
```diff
@@ -1,38 +0,0 @@
...
-use ckb_db_migration::{Migration, ProgressBar};
-use ckb_db_schema::{COLUMN_BLOCK_FILTER, COLUMN_META, META_LATEST_BUILT_FILTER_DATA_KEY};
-use ckb_error::Error;
-use std::sync::Arc;
-
-pub struct RebuildBlockFilter;
-
-const VERSION: &str = "20220803143236";
-
-/// TODO: this migration can be archived when release a new version of ckb
```
quake, line 11: this migration file is useless, should be deleted on v0.106

yangby-cryptape (approved):

quake:
bors r=yangby-cryptape

PR #3826, opened 2023-02-09
fix: Tests db paths without thread_local

zhangsoledad:

`thread_local` is not necessary and has caused some unexpected problems since the switch to nextest runner

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


quake (approved):

quake:
bors r=quake

zhangsoledad:
bors retry

PR #3838, opened 2023-02-15
feat: add exact search mode

quake:

backport https://github.com/nervosnetwork/ckb-indexer/pull/64

Problem Summary: Developer may need to match the exact script for the query in some cases, currently the rpc only has prefix matching, the developer can only get the cells and then filter them on the client side.

this PR adds the exact match mode.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```


doitian (approved):

benches/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ckb-benches"
-version = "0.108.0-pre"
+version = "0.108.0-rc1"
```
doitian, line 3: Please rebase

doitian (changes requested):

doitian:
Please create a new branch since I have resolved the conflicts in #3839 which has merged changes to the branch.

PR #3840, opened 2023-02-15
fix transaction rebroadcast by resubmitting

zhangsoledad:

When a transaction fails a double-spend check, it is still marked as known, making it impossible to rebroadcast by resubmitting

Ensure that if a transaction can be rebroadcast in case of verification failure, it should be removed from the known filter。

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -47,15 +47,15 @@ use crate::utils::send_message;
 use ckb_types::core::EpochNumber;
 pub use header_map::HeaderMap;
 
-const FILTER_SIZE: usize = 20000;
 const GET_HEADERS_CACHE_SIZE: usize = 10000;
 // TODO: Need discussed
 const GET_HEADERS_TIMEOUT: Duration = Duration::from_secs(15);
-const TX_FILTER_SIZE: usize = 50000;
+const FILTER_SIZE: usize = 50000;
 const ORPHAN_BLOCK_SIZE: usize = 1024;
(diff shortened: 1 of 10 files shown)


sync/src/types/mod.rs
```diff
@@ -329,34 +329,60 @@ impl PeerState {
...
+
+    /// Removes expired items.
+    pub fn remove_expired(&mut self) {
+        let now = ckb_systemtime::unix_time().as_secs();
+        let expired_keys: Vec<T> = self
+            .inner
+            .iter()
+            .filter_map(|(key, time)| {
+                println!("now {:?}", now);
+                println!("time + self.ttl {:?}", time + self.ttl);
```
quake, line 372: please remove println or use trace log

sync/src/types/mod.rs
```diff
@@ -329,34 +329,58 @@ impl PeerState {
...
+impl<T: Eq + Hash + Clone> TtlFilter<T> {
+    pub fn new(size: usize, ttl: u64) -> Self {
         Self {
             inner: LruCache::new(size),
+            ttl,
         }
     }
 
-    pub fn contains(&self, item: &T) -> bool {
+    pub fn contains(&mut self, item: &T) -> bool {
```
driftluo, line 351: `&mut` here is not necessary

driftluo (approved):

PR #3841, opened 2023-02-16
[backport] fix transaction rebroadcast by resubmitting

zhangsoledad:

When a transaction fails a double-spend check, it is still marked as known, making it impossible to rebroadcast by resubmitting

Ensure that if a transaction can be rebroadcast in case of verification failure, it should be removed from the known filter。

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -47,15 +47,15 @@ use crate::utils::send_message;
 use ckb_types::core::EpochNumber;
 pub use header_map::HeaderMap;
 
-const FILTER_SIZE: usize = 20000;
 const GET_HEADERS_CACHE_SIZE: usize = 10000;
 // TODO: Need discussed
 const GET_HEADERS_TIMEOUT: Duration = Duration::from_secs(15);
-const TX_FILTER_SIZE: usize = 50000;
+const FILTER_SIZE: usize = 50000;
 const ORPHAN_BLOCK_SIZE: usize = 1024;
 // 2 ** 13 < 6 * 1800 < 2 ** 14
 const ONE_DAY_BLOCK_NUMBER: u64 = 8192;
 const SHRINK_THRESHOLD: usize = 300;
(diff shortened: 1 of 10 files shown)


sync/src/types/mod.rs
```diff
@@ -329,34 +329,60 @@ impl PeerState {
...
+
+    /// Removes expired items.
+    pub fn remove_expired(&mut self) {
+        let now = ckb_systemtime::unix_time().as_secs();
+        let expired_keys: Vec<T> = self
+            .inner
+            .iter()
+            .filter_map(|(key, time)| {
+                println!("now {:?}", now);
+                println!("time + self.ttl {:?}", time + self.ttl);
```
eval-exec, line 372: I think it might be better to use `ckb_logger::info` here.

util/types/src/core/tx_pool.rs
```diff
@@ -79,6 +79,13 @@ impl Reject {
             _ => false,
         }
     }
+
+    /// Returns true if the reject reason is low_fee_rate.
+    pub fn is_allowed_relay(&self) -> bool {
+        // Declared Wrong Cycles should be allowed to relay again
+        matches!(self, Reject::DeclaredWrongCycles(..))
+            || (!matches!(self, Reject::LowFeeRate(..)) && !self.is_malformed_tx())
+    }
```
eval-exec, line 88: The method comment is "Returns true if the reject reason is low_fee_rate.", but there is a `!` in front of `matches!(self, Reject::LowFeeRate(..))`.

 so when `matches!(self, Reject::DeclaredWrongCycles(..))` is `false` and the reject reason is low_fee_rate, `fn is_allowed_replay` will return false.

doitian (approved):

eval-exec (approved):

PR #3842, opened 2023-02-16
feat: add exact search mode

quake:

backport https://github.com/nervosnetwork/ckb-indexer/pull/64

Problem Summary: Developer may need to match the exact script for the query in some cases, currently the rpc only has prefix matching, the developer can only get the cells and then filter them on the client side.

this PR adds the exact match mode.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```

### DIFF
--- a/util/jsonrpc-types/src/indexer.rs
+++ b/util/jsonrpc-types/src/indexer.rs
@@ -48,10 +48,12 @@ impl<T> IndexerPagination<T> {
 /// SearchKey represent indexer support params
 #[derive(Deserialize)]
 pub struct IndexerSearchKey {
-    /// Script, supports prefix search
+    /// Script
     pub script: Script,
     /// Script Type
     pub script_type: IndexerScriptType,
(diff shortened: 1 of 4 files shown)


eval-exec (approved):

zhangsoledad (approved):

util/jsonrpc-types/src/indexer.rs
```diff
@@ -48,10 +48,12 @@ impl<T> IndexerPagination<T> {
 /// SearchKey represent indexer support params
 #[derive(Deserialize)]
 pub struct IndexerSearchKey {
-    /// Script, supports prefix search
+    /// Script
     pub script: Script,
     /// Script Type
     pub script_type: IndexerScriptType,
+    /// Script search mode, optional default is `prefix`, means search script with prefix
+    pub script_search_mode: Option<IndexerScriptSearchMode>,
```
driftluo, line 56: There are only two states here, `null` == `prefix`, and I think `#[serde(default)]` is a better approach

PR #3869, opened 2023-03-06
Replace opentelemetry by tikv/rust-prometheus

### LINKED ISSUE #3866: metrics!(gauge) creates histogram metric in prometheus
### Current Behavior

The gauge metric added in 
https://github.com/nervosnetwork/ckb/blob/f9cf6db90bc302f840d19727f710c33966b503c1/chain/src/chain.rs#L506 creates the histogram metric in prometheus:

```
# HELP ckb_chain_tip ckb.chain_tip
# TYPE ckb_chain_tip histogram
ckb_chain_tip_bucket{le="0.5"} 0
ckb_chain_tip_bucket{le="0.9"} 0
ckb_chain_tip_bucket{le="0.99"} 0
ckb_chain_tip_bucket{le="+Inf"} 27
ckb_chain_tip_sum 240678864
ckb_chain_tip_count 27
```

### Expected Behavior
It should create a prometheus metric of type gauge.

### Environment

- **CKB version**: ckb 0.108.0-rc2 (f232f87 2023-02-27)
- **Chain**: mainnet
- **Operating system**: macOS 11.4
- **Arch**: x64
- **Installation**: GitHub Release

eval-exec:

Issue Number: close #3866 

Problem Summary: 

This PR want to drop `opentelemetry` and replace it by [rust-prometheus](https://github.com/tikv/rust-prometheus) on `util/metrics` 

What's Changed:

### Related changes

- Drop `opentelemetry`, and replace it by [rust-prometheus](https://github.com/tikv/rust-prometheus)
### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


network/src/network.rs
```diff
@@ -192,14 +191,13 @@ impl NetworkState {
...
-                "duration" => duration.as_secs().to_string(),
-                "reason" => reason.clone(),
-            );
+            ckb_metrics::CKB_NETWORK_BAN_PEER
+                .with_label_values(&[
+                    addr.to_string().as_str(),
+                    duration.as_secs().to_string().as_str(),
+                    &reason,
+                ])
+                .inc();
```
eval-exec, line 200: Do we really need to expose `addr`,`duration`, and `reason` to prometheus? 
I think it's a potential point of high memory consumption, it will cause OOM in the worst case.

doitian: No, we can search that in log.

util/metrics/src/lib.rs
```diff
@@ -1,81 +1,170 @@
...
+use prometheus::{
+    register_int_counter, register_int_gauge, register_int_gauge_vec, IntCounter, IntGauge,
+    IntGaugeVec,
+};
+use prometheus_static_metric::make_static_metric;
+
+lazy_static! {
+    /// CKB Chain Tip Header Number
+    pub static ref CKB_CHAIN_TIP: IntGauge =
+        prometheus::register_int_gauge!("ckb_chain_tip", "The CKB Chain Tip Number").unwrap();
```
doitian, line 19: :+1: for the metric description.

eval-exec: I should change the comment to `/// IntGauge metric for CKB Chain Tip Header Number`

network/src/network.rs
```diff
@@ -192,14 +191,7 @@ impl NetworkState {
...
             );
-            metrics!(
-                counter,
-                "ckb.network.ban_peer",
-                1,
-                "peer_addr" => addr.to_string(),
-                "duration" => duration.as_secs().to_string(),
-                "reason" => reason.clone(),
-            );
+            ckb_metrics::CKB_NETWORK_BAN_PEER.inc();
```
doitian, line 194: We may add reason as a label, but let's defer it until we really need this.

doitian (approved):

network/src/network.rs
eval-exec: We can expose `reason` to Prometheus, but it's preferable for `reason` to be an enumerated value known at compile time. This way, `reason` won't occupy too much memory space.

util/metrics/src/lib.rs
eval-exec: I updated the metrics's comments and help messages.

eval-exec:
In current code base, metrics are always collected, but these metrics won't be exposed unless we specific `[metrics.exporter.prometheus]` in `ckb.toml` config file.

We don't have a way to disable to collect metrics.

@doitian Do we need to add a `feature` for CKB to control weather to collect metrics on these points?

network/src/network.rs
doitian: It's easy to break, for example, if someone formats a message and adds the peer id into it.

doitian:
> In current code base, metrics are always collected, but these metrics won't be exposed unless we specific `[metrics.exporter.prometheus]` in `ckb.toml` config file.
> 
> We don't have a way to disable to collect metrics.
> 
> @doitian Do we need to add a `[feature] enable_metrics ` for CKB to control whether to collect metrics on these points?

We can use a macro wrapper to turn the code like

```
ckb_metrics::CKB_NETWORK_BAN_PEER.inc();
```

into

```
if (metrics_enabled!()) {
    ckb_metrics::CKB_NETWORK_BAN_PEER.inc();
}
```

eval-exec:
> We can use a macro wrapper to turn the code like


Agreed, I will go ahead and implement it.

eval-exec:
In the current implementation, the values of "labels" exposed in ckb_message_bytes are difficult to understand as they are represented by numbers. like:
```
ckb_message_bytes{direction="out",item_id="6",protocol="101",status="101"} 62
ckb_message_bytes{direction="out",item_id="7",protocol="101",status="101"} 11390
```

After communicating with @gpBlockchain, I should update the "labels" field with meaningful values, like:
```
ckb_message_bytes{direction="out",item_id="GetBlockProposal",protocol="RelayV2",status="Ignored"} 62
ckb_message_bytes{direction="out",item_id="BlockProposal",protocol="RelayV2",status="Ignored"} 11390
```

Cc. @doitian

sync/src/utils.rs
```diff
@@ -23,17 +24,57 @@ pub(crate) fn send_message<Message: Entity>(
...
+
+    if ckb_metrics::metrics_enabled() {
+        metric_ckb_message_bytes(
+            MetricDirection::Out,
+            &protocol_name,
+            &item_name,
+            None,
+            bytes,
+        );
+    }
```
eval-exec, line 42: I want to create a function to get protocol_name from protocol_id.
```rust
/// Convert protocol id to protocol name
/// Will return "Unknown" if the protocol id is not supported in SupportProtocols
pub fn protocol_name(protocol_id: ProtocolId) -> String {
    match protocol_id.value() {
        0 => SupportProtocols::Ping.name(),
        1 => SupportProtocols::Discovery.name(),
        2 => SupportProtocols::Identify.name(),
        3 => SupportProtocols::Feeler.name(),
        4 => SupportProtocols::DisconnectMessage.name(),
        100 => SupportProtocols::Sync.name(),
        101 => SupportProtocols::RelayV2.name(),
        102 => SupportProtocols::Time.name(),
        110 => SupportProtocols::Alert.name(),
        120 => SupportProtocols::LightClient.name(),
        121 => SupportProtocols::Filter.name(),
        _ => "unknown".to_string(),
    }
}
```
Do you think there is a simpler way to achieve this? @driftluo

driftluo: I think if there is an unknown, it is a bug

eval-exec: When `fn protocol_message_name()` found a `ProtocolId` which can't be mapped to `SupportedProtocols`, I will return "Unknown" and print a error log:
```rust
fn protocol_message_name<Message: Entity>(
    protocol_id: ProtocolId,
    message: &Message,
) -> (String, String) {
    if protocol_id == SupportProtocols::Sync.protocol_id() {
        (
            SupportProtocols::Sync.name(),
            SyncMessageReader::new_unchecked(message.as_slice())
                .to_enum()
                .item_name()
                .to_owned(),
        )
    } else if protocol_id == SupportProtocols::RelayV2.protocol_id() {
        (
            SupportProtocols::RelayV2.name(),
            RelayMessageReader::new_unchecked(message.as_slice())
                .to_enum()
                .item_name()
                .to_owned(),
        )
    } else {
        error!("send message got an Unknown protocol id: {}", protocol_id);
        ("Unknown".to_string(), Message::NAME.to_owned())
    }
}
```

chain/src/chain.rs
```diff
@@ -503,7 +502,9 @@ impl ChainService {
             if log_enabled!(ckb_logger::Level::Debug) {
                 self.print_chain(10);
             }
-            metrics!(gauge, "ckb.chain_tip", block.header().number() as i64);
+            if ckb_metrics::metrics_enabled() {
+                ckb_metrics::CKB_CHAIN_TIP.set(block.header().number() as i64);
```
doitian, line 506: Can we enforce that we must check before recording?

doitian: Solution 1: using macro, and disable the direct access to meters like `ckb_metrics::CKB_CHAINTIP`

Solution 2: use a proxy object to wrap meters like `ckb_metrics::CKB_CHAINTIP`. And the `set`, `inc` functions in the proxy objects will check whether metrics are enabled. However, if we don't use metrics, args are still evaluated.

util/metrics/src/lib.rs
```diff
@@ -1,81 +1,185 @@
...
+
+pub fn metrics_enabled() -> bool {
+    match ENABLE_METRICS_SERVICE.get() {
+        Some(&enabled) => enabled,
+        None => false,
+    }
+}
+
+/// Indicate whether the metrics service is enabled.
+pub static ENABLE_METRICS_SERVICE: once_cell::sync::OnceCell<bool> =
```
doitian, line 24: I'm not sure whether thread local is faster.

The thread local variable is a `Option<bool>`. When it is None, get the value from `OnceCell` and save the value. If it is some, we don't need to access OnceCell.

chain/src/chain.rs
eval-exec: I'd like to choose Solution 2.

util/metrics/src/lib.rs
eval-exec: If we declare a thread_local! variable, it should be mutable, then we need Mutex to protect the thread_local! variable.

eval-exec: /// Indicate whether the metrics service is enabled.
pub static ENABLE_METRICS_SERVICE: once_cell::sync::OnceCell<bool> =
    once_cell::sync::OnceCell::new();

thread_local! {
    static ENABLE_COLLECT_METRICS: Option<bool> = None;
}

pub fn metrics_enabled() -> bool {
    ENABLE_COLLECT_METRICS.with(|enabled| enabled.unwrap_or_else(||{
        // how to update the thread_local value?
        false
    }))
}

eval-exec: Or use RefCell?
```rust
use std::cell::RefCell;

/// Indicate whether the metrics service is enabled.
pub static ENABLE_METRICS_SERVICE: once_cell::sync::OnceCell<bool> =
    once_cell::sync::OnceCell::new();

thread_local! {
    static ENABLE_COLLECT_METRICS: RefCell<Option<bool>> = RefCell::default();
}

pub fn metrics_enabled() -> bool {
    ENABLE_COLLECT_METRICS.with(|cell| {
        cell.borrow()
            .unwrap_or_else(|| match ENABLE_METRICS_SERVICE.get() {
                Some(enabled) => {
                    cell.replace(Some(*enabled));
                    *enabled
                }
                None => false,
            })
    })
}
```

zhangsoledad:
> once_cell also has a Lazy<T> type, build on top of OnceCell which provides the same API as the lazy_static! macro
We don't need both once_cell and lazy_static, and we prefer once_cell when choosing between them, once_cell will be part of rust std in the future, lazy_static has some anti-convention design

util/metrics/src/lib.rs
zhangsoledad: https://docs.rs/thread_local/latest/thread_local/ 
try this one.

doitian: Since `Option<bool>` is `Copy`, `Cell<Option<bool>>` should work.

chain/src/chain.rs
doitian: Let's just keep the `if enabled? { do_metric() }` explicit in the code.

It's simple and we don't really care much the performance lose when we forget to add the `if guard`.

eval-exec: I have implement solution 2, but I forget push my commit yesterday. :sob:

doitian (approved):

eval-exec:
Considering use Histogram type to record ckb_message_bytes.
https://prometheus.io/docs/practices/histograms/#quantiles

doitian (approved):

util/metrics-service/Cargo.toml
```diff
@@ -10,9 +10,9 @@ repository = "https://github.com/nervosnetwork/ckb"
...
 ckb-metrics-config = { path = "../metrics-config", version = "= 0.109.0-pre" }
+ckb-metrics = { path = "../metrics", version = "= 0.109.0-pre" }
 ckb-logger = { path = "../logger", version = "= 0.109.0-pre" }
 ckb-async-runtime = { path = "../runtime", version = "= 0.109.0-pre" }
 ckb-util = { path = "..", version = "= 0.109.0-pre" }
-opentelemetry-prometheus = "0.8"
-prometheus = "0.12"
-hyper = { version = "0.14", features = ["http1", "tcp", "server"] }
+prometheus = "0.13.3"
+hyper = { version = "0.14", features = ["http1", "tcp", "server"] }
```
zhangsoledad, line 18: feed new line

zhangsoledad (approved):

PR #3894, opened 2023-03-21
fix tx-pool remove expired

zhangsoledad:

The tx-pool does not expire for all transactions

* Fix the expiration of tx-pool
* Tweak the default size limit of the tx-pool, the previous value was too small
* Remove the max cycles limit of the tx-pool, the significance of the size limit is to control the memory usage, while for the cycles limit is not very meaningful, so it is removed.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/tests/error.rs
+++ b/rpc/src/tests/error.rs
@@ -32,9 +32,9 @@ fn test_submit_transaction_error() {
             RPCError::from_submit_transaction_reject(&reject).message
         );
 
-    let reject = Reject::Full("size".to_owned(), 10);
+    let reject = Reject::Full("size".to_owned());
     assert_eq!(
-        "PoolIsFull: Transaction pool exceeded maximum size limit(10), try send it later",
+        "PoolIsFull: Transaction are replaced because the pool is full size",
         RPCError::from_submit_transaction_reject(&reject).message
     );
 
--- a/rpc/src/tests/error.rs
+++ b/rpc/src/tests/error.rs
@@ -22,7 +22,7 @@ fn test_submit_transaction_error() {
     let min_fee_rate = FeeRate::from_u64(500);
     let reject = Reject::LowFeeRate(min_fee_rate, 100, 50);
     assert_eq!(
-            "PoolRejectedTransactionByMinFeeRate: The min fee rate is 500 shannons/KB, so the transaction fee should be 100 shannons at least, but only got 50",
+            "PoolRejectedTransactionByMinFeeRate: The min fee rate is 500 shannons/KW, so the transaction fee should be 100 shannons at least, but only got 50",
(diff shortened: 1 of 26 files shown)


rpc/src/tests/error.rs
```diff
@@ -32,9 +32,9 @@ fn test_submit_transaction_error() {
             RPCError::from_submit_transaction_reject(&reject).message
         );
 
-    let reject = Reject::Full("size".to_owned(), 10);
+    let reject = Reject::Full("size".to_owned());
     assert_eq!(
-        "PoolIsFull: Transaction pool exceeded maximum size limit(10), try send it later",
+        "PoolIsFull: Transaction are replaced because the pool is full size",
```
eval-exec, line 37:
```suggestion
        "PoolIsFull: Transaction are replaced because the pool is full",
```

tx-pool/src/process.rs
```diff
@@ -953,15 +947,20 @@ fn _update_tx_pool_for_reorg(
...
                 callbacks.call_reject(tx_pool, &entry, e.clone());
             }
         }
     }
 
-    // remove expired transaction from pending
+    // Remove expired transaction from pending
     tx_pool.remove_expired(callbacks);
+
+    // Remove transactions from the pool until its size < size_limit.
```
eval-exec, line 962: I think `<=` instead of `<`

tx-pool/src/process.rs
```diff
@@ -953,15 +947,20 @@ fn _update_tx_pool_for_reorg(
...
+    tx_pool.limit_size(callbacks);
```
eval-exec, line 963: I think `<=` instead of `<`.

quake (approved):

eval-exec (approved):

doitian:
> > Suggest to add a new RPC method to get the minimal fee rate to enter the pool.
> 
> https://github.com/nervosnetwork/ckb/tree/develop/rpc#method-tx_pool_info With the RPC `tx_pool_info` to get the min_fee_rate, do you think we still need a separate rpc to provide min_fee_rate?

It's enough. But this RPC lacks of the max size cycles info.

zhangsoledad:
> > > Suggest to add a new RPC method to get the minimal fee rate to enter the pool.
> > 
> > 
> > https://github.com/nervosnetwork/ckb/tree/develop/rpc#method-tx_pool_info With the RPC `tx_pool_info` to get the min_fee_rate, do you think we still need a separate rpc to provide min_fee_rate?
> 
> It's enough. But this RPC lacks of the max size cycles info.

added

PR #3917, opened 2023-03-29
ckb 2023 edition

zhangsoledad:

The complete implementation of ckb2023 contains the following parts: 
1. hardfork deployment
2. vm upgrade
3. new syscall implementation
4. remove the reservation rule of header version field

Related rfcs:
* https://github.com/nervosnetwork/rfcs/pull/417
* https://github.com/nervosnetwork/rfcs/pull/416
* https://github.com/nervosnetwork/rfcs/pull/418
* https://github.com/nervosnetwork/rfcs/pull/419

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


network/src/protocols/identify/mod.rs
```diff
@@ -370,7 +370,18 @@ impl Callback for IdentifyCallback {
...
     fn unregister(&self, context: &ProtocolContextMutRef) {
-        if context.session.ty.is_outbound() {
+        let protocol_version_match = self
+            .network_state
+            .with_peer_registry(|reg| {
+                reg.get_peer(context.session.id)
+                    .map(|p| p.protocol_version(context.proto_id))
+            })
+            .flatten()
+            .map(|version| version != "3")
```
eval-exec, line 380: Is the meaning of '3' the same as `const LATEST_VERSION: &str = "3";` which is defined in `network/src/protocols/support_protocols.rs`?

driftluo: yes

script/src/types.rs
```diff
@@ -51,19 +51,22 @@ pub enum ScriptVersion {
     V0 = 0,
     /// CKB VM 1 with Syscall version 1 and version 2.
     V1 = 1,
+    /// /// CKB VM 1 with Syscall version 1, version 2 and version 3.
+    V2 = 2,
```
eval-exec, line 55: I guess it should be `/// CKB VM 2 ...`?

eval-exec (approved):

script/src/syscalls/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+    cost_model::estimate_cycles,
+    registers::{A0, A1, A2, A3, A4, A5, A7},
+    DefaultMachineBuilder, Error as VMError, Memory, Register, SupportMachine, Syscalls,
+};
+use std::sync::{Arc, Mutex};
+
+pub struct Spawn<DL> {
+    data_loader: DL,
+    group_inputs: Indices,
+    group_outputs: Indices,
```
quake, line 25: could you explain that why we add `group_inputs` and `group_outputs` fields since there is a `script_group` field already, aren't same with `script_group.group_inputs` and `script_group.group_outputs`?

script/src/syscalls/spawn.rs
```diff
@@ -0,0 +1,266 @@
...
+                if end > size {
+                    machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
+                    return Ok(true);
+                }
+                data.slice(offset..end)
+            }
+        };
+        // Build arguments.
+        let mut addr = argv_addr.to_u64();
+        let mut argv_vec = Vec::new();
```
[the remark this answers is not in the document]
mohanson: Be aware of argc may be an arbitrary value(u64::Max, e.g.) passed in by a evil user. By using `Vec::new()`, `load_c_string` will fail at some point.

script/src/syscalls/spawn.rs
mohanson: I think it is safe to remove these two parameters.

quake: > I think it is safe to remove these two parameters.

resolved by https://github.com/nervosnetwork/ckb/pull/3989 please help to review, thanks.

PR #3929, opened 2023-04-11
chore: remove deps cycle

chenyukang:

Remove deps cycle in Cargo.toml.

### Problem Summary:

`rust-analyzer` is blaming for it:

image

And if we remove the deps cycle, we could use `cargo deps` to visualize deps graph.

`devtools/ci/check-cyclic-dependencies.py` in CI is not working, do we need to enable it to check the dep order name?

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


devtools/ci/ci_main.sh
```diff
@@ -59,7 +59,7 @@ case $GITHUB_WORKFLOW in
     make check-whitespaces
     make check-dirty-rpc-doc
     make check-dirty-hashes-toml
-    devtools/ci/check-cyclic-dependencies.py
+    devtools/ci/check-cyclic-dependencies.py --dev
```
chenyukang, line 62: we didn't check the exit code of this script, should we fix it?

chenyukang:
ckb

A good way to understand the whole repo.

devtools/ci/ci_main.sh
zhangsoledad: Yes it should be fixed

chain/Cargo.toml
```diff
@@ -36,7 +36,6 @@ ckb-reward-calculator = { path = "../util/reward-calculator", version = "= 0.110
 ckb-tx-pool = { path = "../tx-pool", version = "= 0.110.0-pre", features = ["internal"] }
 ckb-jsonrpc-types = { path = "../util/jsonrpc-types", version = "= 0.110.0-pre" }
 ckb-network = { path = "../network", version = "= 0.110.0-pre" }
-ckb-launcher = { path = "../util/launcher", version = "= 0.110.0-pre" }
```
eval-exec, line 39: I don't think we should remove `ckb-launcher` from `[dev-dependencies]`, because there are some unit tests in `ckb-chain` need `ckb-launcher`.

quake (changes requested):
I think the deps cyclic is related to dev-dependencies only, some unit test will failed if you remove it. Could you try the unit test on your local env to verify the result?

chenyukang:
> I think the deps cyclic is related to dev-dependencies only, some unit test will failed if you remove it. Could you try the unit test on your local env to verify the result?

Yes, I can reproduce it in my local env.
Got it, seems some UI test are a little bit like integrated test, so we have rev-order deps.

chenyukang:
According to https://github.com/rust-lang/rust-analyzer/issues/14167
Even `rust-analyzer` has a warning on it, we could ignore it safely.
Closing this PR.

PR #3937, opened 2023-04-14
Fix: migrate --check should return exit with 0 when DB version is equal

chenyukang:

Problem Summary:

`ckb migrate --check` will exit with non-zero when the DB version is `Equal`:

```console
➜  test-ckb git:✗ ../target/debug/ckb migrate --check
path: "/Users/yukang/code/ckb/test-ckb/data/db"
read_only_db: false
db_status: Equal
➜  test-ckb git:✗ echo $?
64
```

What's Changed:

Fix the checking logic

### Check List 

Tests 

- Unit test
- Integration test

```console
➜  test-ckb git:✗ ../target/debug/ckb migrate --check
path: "/Users/yukang/code/ckb/test-ckb/data/db"
read_only_db: false
db_status: Equal
➜  test-ckb git:✗ echo $?
0
```

### Release note 

```release-note
```

### DIFF
--- a/ckb-bin/src/subcommand/migrate.rs
+++ b/ckb-bin/src/subcommand/migrate.rs
@@ -24,16 +24,10 @@ pub fn migrate(args: MigrateArgs) -> Result<(), ExitCode> {
                 return Err(ExitCode::Failure);
             }
 
-            if args.check {
-                if matches!(db_status, Ordering::Less) {
-                    return Ok(());
-                } else {
-                    return Err(ExitCode::Cli);
-                }
-            }
-
             if matches!(db_status, Ordering::Equal) {
                 return Ok(());
+            } else if args.check {
+                return Err(ExitCode::Cli);
             }
 
             if migrate.require_expensive(&db) && !args.force {


ckb-bin/src/subcommand/migrate.rs
```diff
@@ -24,16 +24,10 @@ pub fn migrate(args: MigrateArgs) -> Result<(), ExitCode> {
...
-                if matches!(db_status, Ordering::Less) {
-                    return Ok(());
-                } else {
-                    return Err(ExitCode::Cli);
-                }
-            }
-
             if matches!(db_status, Ordering::Equal) {
                 return Ok(());
+            } else if args.check {
```
eval-exec, line 29: The `ckb migration --check`'s description is "Perform database version check without migrating, if migration is in need ExitCode(0) is returned，otherwise ExitCode(64) is returned". 
So, when database version is equal, `ExitCode(64)` is expected.

doitian: It's for automation. For example:

```
if ckb migrate --check; then
  ckb migrate --force
fi
```

chenyukang: yes, got it in the document: 
```console
Add command `ckb migrate --check`. If migration is in need 0 will be return，otherwise 64.
```
thanks.

PR #3948, opened 2023-04-21
Fix bug in truncate RPC all and add limit for better error message

### LINKED ISSUE #3941: Fix rpc `truncate` leak
### Current Behavior
https://github.com/nervosnetwork/ckb/tree/develop/rpc#method-truncate
The truncate rpc is often used when testing, but truncate now has a defect, it will read all the selected blocks into memory at once, when truncate a long chain will be a memory leak and crash

### Expected Behavior

Truncate can be performed in segments

chenyukang:

Fixes #3941

Problem Summary:

- The RPC call has a timeout limitation, so we can not rollback too many blocks, with some testing I found 50000 is a proper limit.
- The iteration on `attached_blocks` is a typo, we can remove it.

We add a limit on the number of blocks, if we have a requirement for rollbacking many blocks in one time, we will implement it `cli` mode.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -279,7 +279,7 @@ impl ChainService {
     }
 
     // Truncate the main chain
-    // Use for testing only
+    // Use for testing only, can only truncate less than 50000 blocks each time
     pub(crate) fn truncate(&mut self, target_tip_hash: &Byte32) -> Result<(), Error> {
         let snapshot = Arc::clone(&self.shared.snapshot());
         assert!(snapshot.is_main_chain(target_tip_hash));
@@ -291,6 +291,19 @@ impl ChainService {
             .and_then(|index| snapshot.get_epoch_ext(&index))
             .expect("checked");
         let origin_proposals = snapshot.proposals();
+
+        let block_count = snapshot
+            .tip_header()
+            .number()
+            .saturating_sub(target_tip_header.number());
+
+        if block_count > 5_0000 {
+            let err = format!(
+                "trying to truncate too many blocks: {}, exceed 50000",
+                block_count
+            );
+            return Err(InternalErrorKind::Database.other(err).into());
+        }
         let mut fork = self.make_fork_for_truncate(&target_tip_header, snapshot.tip_header());
 
         let db_txn = self.shared.store().begin_transaction();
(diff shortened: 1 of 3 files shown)


chain/src/chain.rs
```diff
@@ -279,7 +279,7 @@ impl ChainService {
     }
 
     // Truncate the main chain
-    // Use for testing only
+    // Use for testing only, can only truncate less than 50000 blocks each time
     pub(crate) fn truncate(&mut self, target_tip_hash: &Byte32) -> Result<(), Error> {
         let snapshot = Arc::clone(&self.shared.snapshot());
         assert!(snapshot.is_main_chain(target_tip_hash));
```
eval-exec, line 285: I think we shouldn't `assert!` here, it may cause node panic.
If `target_tip_hash` is not in main chain, `fn truncate` should return `Err`.

chain/src/chain.rs
```diff
@@ -299,9 +312,9 @@ impl ChainService {
         db_txn.insert_tip_header(&target_tip_header)?;
         db_txn.insert_current_epoch_ext(&target_epoch_ext)?;
 
-        for blk in fork.attached_blocks() {
-            db_txn.delete_block(blk)?;
-        }
```
eval-exec, line 304: I think previous implementation is wrong? Because we expect `target_tip_hash` is on main_chain, so `fork.attached_blocks()` should be empty, is it right?

chain/src/chain.rs
chenyukang: I think here is a double-check, custom error message will be given before calling `chain`:

https://github.com/nervosnetwork/ckb/blob/6de945f4512c2781fc17763cbabf1b3e9b580e08/rpc/src/module/test.rs#L497

chain/src/chain.rs
chenyukang: yes.

If we want to delete blocks, we should use `for blk in fork.detached_blocks()`, but actually we don't need to delete blocks for testing usage, and even worse if we run `sync` at the same time and there are incoming new blocks, a crash will make the db corrupted.

So, I think it's Ok to remove this part.

doitian: Unrelated to this PR.

zhangsoledad (approved):

PR #3955, opened 2023-04-23
refactor: use BlockNumberAndHash as last_common_header

quake:

Same purpose as https://github.com/nervosnetwork/ckb/pull/3952, this PR changed `last_common_header` field to  `BlockNumberAndHash`

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/sync/src/types/mod.rs
+++ b/sync/src/types/mod.rs
@@ -275,7 +275,7 @@ pub struct PeerState {
     // The best known block we know this peer has announced
     pub best_known_header: Option<HeaderView>,
     // The last block we both stored
-    pub last_common_header: Option<core::HeaderView>,
+    pub last_common_header: Option<BlockNumberAndHash>,
     // use on ibd concurrent block download
     // save `get_headers` locator hashes here
     pub unknown_header_list: Vec<Byte32>,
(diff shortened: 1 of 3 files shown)


eval-exec (approved):

sync/src/types/mod.rs
```diff
@@ -442,6 +442,15 @@ impl From<&core::HeaderView> for BlockNumberAndHash {
...
 
+impl From<core::HeaderView> for BlockNumberAndHash {
+    fn from(header: core::HeaderView) -> Self {
+        Self {
+            number: header.number(),
+            hash: header.hash(),
+        }
+    }
+}
+
```
eval-exec, line 453: We already have `impl From<&core::HeaderView> for BlockNumberAndHash` on above.

How about use `&` to convert `HeaderView` to `BlockNumberAndHash`?

like:
```rust
        let m_right: BlockNumberAndHash =
            (&self.get_ancestor(&m_right.hash(), m_left.number())?).into();

```

eval-exec: Or 
```rust
        m_right = self.get_ancestor(&m_right.hash(), m_left.number()).as_ref()?.into();
```

quake: It's for usability, both the owned and reference versions of the `HeaderView` struct are commonly used and conversions between them are expected to be frequent in our codebase, implementing both `From` conversions can make the code more ergonomic and user-friendly. It allows our code to convert between the two versions using the same syntax and idioms, regardless of whether they have an owned or a reference version of the struct.

eval-exec (approved):

PR #3959, opened 2023-04-25
fix: use snapshot in light-client protocol

quake:

light-client protocol should not use the `ActiveChain` to find ancestor headers, because of `ActiveChain` finding headers from the sync `HeaderMap` which contains unverified block body.

Use snapshot to find ancestor headers.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/store/src/store.rs
+++ b/store/src/store.rs
@@ -575,6 +575,21 @@ pub trait ChainStore: Send + Sync + Sized {
                 reader.to_entity()
             })
     }
+
+    /// Gets ancestor block header by a base block hash and number
+    fn get_ancestor(&self, base: &packed::Byte32, number: BlockNumber) -> Option<HeaderView> {
+        let header = self.get_block_header(base)?;
+        if number > header.number() {
+            None
+        } else if number == header.number() {
+            Some(header)
+        } else if self.is_main_chain(base) {
+            self.get_block_hash(number)
+                .and_then(|hash| self.get_block_header(&hash))
+        } else {
+            self.get_ancestor(&header.parent_hash(), number)
+        }
+    }
 }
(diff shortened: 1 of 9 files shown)


store/src/store.rs
```diff
@@ -575,6 +575,21 @@ pub trait ChainStore: Send + Sync + Sized {
...
+    /// Gets ancestor block header by a base block hash and number
+    fn get_ancestor(&self, base: &packed::Byte32, number: BlockNumber) -> Option<HeaderView> {
+        let header = self.get_block_header(base)?;
+        if number > header.number() {
+            None
+        } else if number == header.number() {
+            Some(header)
+        } else if self.is_main_chain(base) {
+            self.get_block_hash(number)
+                .and_then(|hash| self.get_block_header(&hash))
```
eval-exec, line 588: I think `self.in_main_chain` and `self.get_block_hash` can combine. Because they read same RocksDB column family:
```rust
    /// Get block number by block header hash
    fn get_block_number(&self, hash: &packed::Byte32) -> Option<BlockNumber> {
        self.get(COLUMN_INDEX, hash.as_slice())
            .map(|raw| packed::Uint64Reader::from_slice_should_be_ok(raw.as_ref()).unpack())
    }

    /// TODO(doc): @quake
    fn is_main_chain(&self, hash: &packed::Byte32) -> bool {
        self.get(COLUMN_INDEX, hash.as_slice()).is_some()
    }
```

eval-exec (approved):

yangby-cryptape (approved):

zhangsoledad (approved):

Issue #3940, opened 2023-04-17
Memory tuning  on header synchronization 


chenyukang:
With the merge of the previous PR: https://github.com/nervosnetwork/ckb/pull/3391

We may get some memory issue because of https://github.com/spacejam/sled/issues/1388

As discussed with @zhangsoledad , a possible solution is:

1. Replace sled. For this scenario, a `btree` database such as `lmdb` is more suitable. We are currently considering `lmdb` as a candidate, we need more tests and research.

2. Add the original small memory mode back as a configurable option.

3. Provide multiple configuration options to adapt to different scenarios, such as:

- Stable operation under low memory conditions.
- Fastest synchronization but with high memory usage.
- Balanced performance.

PR #3970, opened 2023-05-06
refactor(perf): replace sync struct HeaderView with HeaderIndexView

quake:

resolve https://github.com/nervosnetwork/ckb/issues/3940

replace the struct `HeaderView` of sync mod with a more memory compact struct `HeaderIndexView`, and avoid holding headers messages in memory (remove bytes slice clone).

Compare with v0.109.0, this PR reduced the peak memory usage from 3.8G to 2.8G during IBD sync on my local PC. @jiangxianliang007 also did test on an aws c5.xlarge instance (4C8G), it reduced from 3.2G to 2.3G, sync performance has also improved, from 0 to 9 million, dropping from 19 hours and 25 minutes to 17 hours and 30 minutes. 

please not that this PR didn't replace the HeaderMap storage sled with other solution, there is another PR in testing

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -101,7 +101,7 @@ bootnode_mode = false
 support_protocols = ["Ping", "Discovery", "Identify", "Feeler", "DisconnectMessage", "Sync", "Relay", "Time", "Alert", "LightClient", "Filter"]
 
 # [network.sync.header_map]
-# memory_limit = "600MB"
+# memory_limit = "256MB"
 
 [rpc]
 # By default RPC only binds to localhost, thus it only allows accessing from the same machine.
--- a/sync/src/synchronizer/block_fetcher.rs
+++ b/sync/src/synchronizer/block_fetcher.rs
@@ -1,13 +1,13 @@
 use crate::block_status::BlockStatus;
 use crate::synchronizer::Synchronizer;
-use crate::types::{ActiveChain, BlockNumberAndHash, HeaderIndex, IBDState};
+use crate::types::{ActiveChain, BlockNumberAndHash, HeaderIndex, HeaderIndexView, IBDState};
 use ckb_constant::sync::{
     BLOCK_DOWNLOAD_WINDOW, CHECK_POINT_WINDOW, INIT_BLOCKS_IN_TRANSIT_PER_PEER,
 };
 use ckb_logger::{debug, trace};
 use ckb_network::PeerIndex;
 use ckb_systemtime::unix_time_as_millis;
-use ckb_types::{core, packed};
+use ckb_types::packed;
 use std::cmp::min;
 
 pub struct BlockFetcher<'a> {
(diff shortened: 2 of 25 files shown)


resource/ckb.toml
```diff
@@ -101,7 +101,7 @@ bootnode_mode = false
 support_protocols = ["Ping", "Discovery", "Identify", "Feeler", "DisconnectMessage", "Sync", "Relay", "Time", "Alert", "LightClient", "Filter"]
 
 # [network.sync.header_map]
-# memory_limit = "600MB"
+# memory_limit = "256MB"
```
quake, line 104: Changed the default memory limit to smaller one, it can keep about the same amount of header data in memory as before.

sync/src/synchronizer/block_fetcher.rs
```diff
@@ -171,11 +166,7 @@ impl<'a> BlockFetcher<'a> {
                 header = self
                     .synchronizer
                     .shared
-                    .get_header_view(
-                        &parent_hash,
-                        Some(status.contains(BlockStatus::BLOCK_STORED)),
-                    )?
-                    .into_inner();
+                    .get_header_index_view(&parent_hash, false)?;
```
quake, line 169: we have checked the status if contains `BLOCK_STORED` in line 148 and returned early, no need to check it again here.

sync/src/types/header_map/mod.rs
```diff
@@ -30,8 +31,8 @@ impl Drop for HeaderMap {
 }
 
 const INTERVAL: Duration = Duration::from_millis(500);
-// key + total_difficulty + skip_hash
-const ITEM_BYTES_SIZE: usize = packed::HeaderView::TOTAL_SIZE + 32 * 3;
+// HeaderIndexView size is 152 bytes
+const ITEM_BYTES_SIZE: usize = 152;
```
eval-exec, line 35: How about calculate `HeaderIndexView`'s size in compile time by this way instead of hardcode `152`:
```rust
    const ITEM_BYTES_SIZE: usize = std::mem::size_of::<HeaderIndexView>();
```

eval-exec (approved):

sync/src/types/header_map/backend_sled.rs
```diff
@@ -27,8 +27,15 @@ impl KeyValueBackend for SledBackend {
         }
         .expect("failed to create a tempdir to save header map into disk");
 
-        let db: Db = sled::open(tmpdir.path())
+        // use a smaller system page cache here since we are using sled as a temporary storage,
+        // most of the time we will only read header from memory.
+        let db: Db = Config::new()
+            .mode(Mode::HighThroughput)
+            .cache_capacity(64 * 1024 * 1024)
```
eval-exec, line 34: How about use `ByteUnit::Megabyte(64).as_u64()` here.

zhangsoledad (approved):

sync/src/types/header_map/mod.rs
quake: updated code, thanks

sync/src/types/header_map/backend_sled.rs
quake: I don't like to import a crate for this simple usage.

traits/src/header_provider.rs
```diff
@@ -1,33 +1,45 @@
...
-    fn timestamp_and_parent(&self, block_hash: &Byte32) -> (u64, BlockNumber, Byte32) {
-        let header = self.get_header(block_hash).expect("parent header exist");
-        (
-            header.timestamp(),
-            header.number(),
-            header.data().raw().parent_hash(),
-        )
-    }
+/// A compact representation of header fields, used for header verification and median time calculation
+pub struct HeaderFields {
```
doitian, line 13: Let's call it `HeaderIndexFields`?

quake: we have a struct named HeaderIndex already, which includes a `total_difficulty` field for comparing longest chain, there is no need for this comparison in `HeaderFields`, so I use a different name.
```
pub struct HeaderIndex {
    number: BlockNumber,
    hash: Byte32,
    total_difficulty: U256,
}
```

PR #3971, opened 2023-05-06
Improve tx pool index for different sorting strategies

### LINKED ISSUE #3942: Improve index of ckb tx-pool
Currently, ckb tx-pool only maintains ancestor-related information, but not descendants, which leads to the implementation of some functions that are not perfect. For example, when the transaction pool is full, we want to evict the least number of descendants to avoid evicting too many transactions at once.

In addition, considering the complexity of implementation, we may need to introduce something like https://www.boost.org/doc/libs/1_72_0/libs/multi_index/doc/index.html

chenyukang:

Issue Number: close #3942

Problem Summary:

What's Changed:

### Related changes

- Introduce multiple indexes in `SortedTxMap` for other sorting strategies.
- Code refactor.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

There is a performance penalty for `Add`, `Remove` entries in `SortedTxMap`: 

before:

```console
test component::tests::proposed::test_container_bench_add_remove ... bench:   5,718,670 ns/iter (+/- 174,237)
test component::tests::proposed::test_container_bench_sort       ... bench:   3,063,072 ns/iter (+/- 148,255)
```
 
after: 
```console
test component::tests::proposed::test_container_bench_add_remove ... bench:   6,640,891 ns/iter (+/- 163,363)
test component::tests::proposed::test_container_bench_sort       ... bench:   3,387,712 ns/iter (+/- 170,507)
```
  
No

### Release note 

```release-note

```

### DIFF
--- a/tx-pool/src/component/container.rs
+++ b/tx-pool/src/component/container.rs
@@ -6,10 +6,11 @@ use ckb_types::{
     core::Capacity,
     packed::{OutPoint, ProposalShortId},
 };
+use multi_index_map::MultiIndexMap;
 use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::collections::hash_map::Entry as HashMapEntry;
-use std::collections::{BTreeSet, HashMap, HashSet};
+use std::collections::{HashMap, HashSet};
 
 /// A struct to use as a sorted key
 #[derive(Eq, PartialEq, Clone, Debug)]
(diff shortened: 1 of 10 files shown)


tx-pool/src/component/entry.rs
```diff
@@ -90,7 +91,7 @@ impl TxEntry {
     }
 
     /// Returns a sorted_key
-    pub fn as_sorted_key(&self) -> AncestorsScoreSortKey {
+    pub fn as_score_key(&self) -> AncestorsScoreSortKey {
```
eval-exec, line 94: Do we need to change this method's comment?

chenyukang: I think `as_score_key` maybe a better name, since we may have other sort keys.

tx-pool/src/component/container.rs
```diff
@@ -401,18 +442,10 @@ impl SortedTxMap {
...
             // So we need update statistics state
-            for desc_id in &descendants {
-                if let Some(desc_entry) = self.entries.get_mut(desc_id) {
-                    let deleted = self.sorted_index.remove(&desc_entry.as_sorted_key());
-                    debug_assert!(deleted, "pool inconsistent");
-                    desc_entry.sub_entry_weight(&entry);
-                    self.sorted_index.insert(desc_entry.as_sorted_key());
-                }
-            }
+            self.update_descendants_index_key(&entry, EntryOp::Remove);
```
eval-exec, line 448: How about change `EntryOp::Remove` to `EntryOp::Sub`, since its purpose isn't "remove something" but "descrease weight".

eval-exec (approved):

PR #3973, opened 2023-05-08
ckb_types: deduplicate cell push and header push

liuck8080:
Transaction builder should not contain duplicate cell dep and header dep, but sometimes it may add several times, so this commit add dedup_ version push, extend, so no duplicate dependencies will be add

Problem Summary: Duplicate cell dep and header deps are not allowed in a transaction, but transaction builder do not provide such ability to do such deduplicate work.

What's Changed: add functions: dedup_cell_dep, dedup_cell_deps,  dedup_header_dep, dedup_header_deps, when add cell_dep and header_dep, check if according item is already in the list before actually add.

### Check List 

Tests 

- Unit test

Side effects

- Performance regression.
For small size of cell deps and header deps, the impact is minimum, for big ones, it may cause performance regression, since it search the list one by one.

### Release note 

```release-note
```

### DIFF
--- a/util/types/src/core/advanced_builders.rs
+++ b/util/types/src/core/advanced_builders.rs
@@ -187,6 +187,56 @@ macro_rules! def_setter_for_vector {
     };
 }
 
+macro_rules! def_dedup_setter_for_vector {
+    (
+        $prefix:ident, $field:ident, $type:ident,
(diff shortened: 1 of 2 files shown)


liuck8080:
ckb-sdk-rust should keep a self-maintain version TransactionBuilder

PR #3993, opened 2023-05-25
Tx pool rewrite with multi_index_map

chenyukang:

Problem Summary:

This PR unifies the `pending` and `gap` pool with `multi_index_map`, with these two benefits:

1. remove duplicated code logic between different pools.
2. we add the same sorting fields for all entries in `tx_pool` for future improvement
3. easy to implement future requirements such as replace-by-fee strategy.

### Check List 

Tests 

- Unit test
- Integration test
- No code ci-runs-only: [ quick_checks, linters ]

Side effects

- Performance regression

This PR may introduce performance regression for `tx_pool`, another change is we now do double-spend checking when a transaction is added to the pool.

I will do more testing for this change.

### Release note 

```release-note
```


tx-pool/src/pool.rs
```diff
@@ -621,27 +486,27 @@ impl TxPool {
...
-            Ok(recent_reject) => Some(recent_reject),
-            Err(err) => {
-                error!(
-                    "Failed to open recent reject database {:?} {}",
-                    config.recent_reject, err
-                );
-                None
+    fn build_recent_reject(config: &TxPoolConfig) -> Option<RecentReject> {
+        if !config.recent_reject.as_os_str().is_empty() {
+            let recent_reject_ttl = config.keep_rejected_tx_hashes_days as i32 * 24 * 60 * 60;
```
eval-exec, line 492: Do we need to check `config.keep_rejected_tx_hashes_days` is not `0`?

tx-pool/src/pool.rs
```diff
@@ -321,12 +214,20 @@ impl TxPool {
     // Remove transactions from the pool until total size < size_limit.
```
eval-exec, line 214:
```rust
    // Remove transactions from the pool until its size <= size_limit.
    tx_pool.limit_size(callbacks);
```
Code comments are not consistency. (I think it should be `<=` instead of `<`)

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,588 @@
...
+        &self,
+        limit: usize,
+        exclusion: &HashSet<ProposalShortId>,
+        proposals: &mut HashSet<ProposalShortId>,
+        status: &Status,
+    ) {
+        for entry in self.entries.get_by_status(status) {
+            if proposals.len() == limit {
+                break;
+            }
```
eval-exec, line 302: Should we break the for loop only when `proposals.len() == limit`?
I think `proposals.len() >= limit` is more safe?

tx-pool/src/process.rs
```diff
@@ -1003,27 +988,23 @@ fn _submit_entry(
     let tx_hash = entry.transaction().hash();
     match status {
         TxStatus::Fresh => {
-            if tx_pool.add_pending(entry.clone()) {
-                debug!("submit_entry pending {}", tx_hash);
+            if tx_pool.add_pending(entry.clone()).unwrap_or(false) {
```
eval-exec, line 991: If `tx_pool.add_pending()` is not `Ok`, the reason may not be `Reject::Duplicated`. It may be `Reject::ExceededMaximumAncestorsCount` throw by 
 `fn add_entry` => `self.record_entry_links`.

tx-pool/Cargo.toml
```diff
@@ -36,7 +36,7 @@ sentry = { version = "0.26.0", optional = true }
 serde_json = "1.0"
 rand = "0.8.4"
 hyper = { version = "0.14", features = ["http1", "client", "tcp"] }
-multi_index_map = "0.5.0"
+ckb_multi_index_map = "0.0.2" # ckb team fork crate
```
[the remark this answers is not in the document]
chenyukang: we're waiting for the changes are merged by upstream.
after it's finished, we will remove this fork crate.

tx-pool/src/component/pool_map.rs
chenyukang: it's added one-by-one, so it makes no difference? I think it's also good to change to `>=`.

tx-pool/src/service.rs
```diff
@@ -901,8 +914,8 @@ impl TxPoolService {
         TxPoolInfo {
             tip_hash: tip_header.hash(),
             tip_number: tip_header.number(),
-            pending_size: tx_pool.pending.size() + tx_pool.gap.size(),
-            proposed_size: tx_pool.proposed.size(),
+            pending_size: tx_pool.pool_map.pending_size(),
```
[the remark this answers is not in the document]
chenyukang: `pool_map.pending_size()` is also implemented as `pending + gap`.

tx-pool/src/process.rs
```diff
@@ -1003,27 +988,23 @@ fn _submit_entry(
...
             } else {
                 return Err(Reject::Duplicated(tx_hash));
             }
         }
         TxStatus::Proposed => {
             if tx_pool.add_proposed(entry.clone())? {
-                debug!("submit_entry proposed {}", tx_hash);
                 callbacks.call_proposed(tx_pool, &entry, true);
-            } else {
-                return Err(Reject::Duplicated(tx_hash));
```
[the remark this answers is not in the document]
chenyukang: yes, since we used `?` in `tx_pool.add_proposed(entry.clone())?`.
I think we should also change for `add_pending` and `add_gap`.

tx-pool/src/pool.rs
```diff
@@ -491,98 +346,109 @@ impl TxPool {
         let entry =
             TxEntry::new_with_timestamp(rtx, verified.cycles, verified.fee, size, timestamp);
         let tx_hash = entry.transaction().hash();
-        if self.add_proposed(entry)? {
-            Ok(CacheEntry::Completed(verified))
-        } else {
-            Err(Reject::Duplicated(tx_hash))
-        }
```
zhangsoledad, line 498: Here we can just modify the state instead of re-adding the entry, so we can change the function signature to just pass in the id, similar to 
```
   pub(crate) fn proposed_rtx(
        &mut self.
        ...
        Id: &ProposalShortId.
    ) -> Result<CacheEntry, Reject> {
```

tx-pool/src/process.rs
```diff
@@ -1053,46 +1028,41 @@ fn _update_tx_pool_for_reorg(
...
-                false
+        }
+
+        for entry in tx_pool.pool_map.entries.get_by_status(&Status::Pending) {
+            let e = &entry.inner;
+            let short_id = e.proposal_short_id();
+            if snapshot.proposals().contains_proposed(&short_id) {
+                proposals.push(e.clone());
+            } else if snapshot.proposals().contains_gap(&short_id) {
+                gaps.push(e.clone());
```
zhangsoledad, line 1048: As mentioned above, instead of passing the entry, you can just pass the short_id

tx-pool/src/pool.rs
zhangsoledad: It is better to have a minimum value `max(1, config.keep_rejected_tx_hashes_days)`.

tx-pool/src/process.rs
```diff
@@ -1032,43 +1032,40 @@ fn _update_tx_pool_for_reorg(
...
-            let tx_hash = entry.transaction().hash();
-            let cached = CacheEntry::completed(entry.cycles, entry.fee);
-            if let Err(e) =
-                tx_pool.gap_rtx(cached, entry.size, entry.timestamp, Arc::clone(&entry.rtx))
-            {
-                debug!("Failed to add tx to gap {}, reason: {}", tx_hash, e);
+        for id in gaps {
+            debug!("begin to gap: {:x}", id);
+            if let Err(e) = tx_pool.gap_rtx(&id) {
+                let entry = tx_pool.get(&id).unwrap().clone();
```
eval-exec, line 1063: I'm concern about the `unwrap()` here, how about define `proposals` and `gaps` as below:
```rust
        let mut proposals: Vec<(ProposalShortId, Byte32)> = Vec::new();
        let mut gaps: Vec<(ProposalShortId, Byte32)> = Vec::new();
```
...
```rust
        for (id,entry) in proposals {
            debug!("begin to proposed: {:x}", id);
            if let Err(e) = tx_pool.proposed_rtx(&id) {
                callbacks.call_reject(tx_pool, &entry, e);
            } else {
                callbacks.call_proposed(tx_pool, &entry, false)
            }
        }

```

chenyukang: I considered this here, the issue here is `tx_pool.proposed_rtx(&id)` is a `mut` reference here, then any later immutable borrow will be invalid.

All the `id` are got from `entries`, and if we re-query with `id` and get nothing, a `panic` is accepted since something weird and unexpected happened?

chenyukang: I also want to avoid the `clone`, but haven't got a good solution right now.

eval-exec: It's hard, how about this:
```rust
    if mine_mode {
        let mut proposals: Vec<(ProposalShortId, TxEntry)> = Vec::new();
        let mut gaps: Vec<(ProposalShortId, TxEntry)> = Vec::new();

        for entry in tx_pool.pool_map.entries.get_by_status(&Status::Gap) {
            let short_id = entry.inner.proposal_short_id();
            if snapshot.proposals().contains_proposed(&short_id) {
                proposals.push((short_id, entry.inner.to_owned()));
            }
        }

        for entry in tx_pool.pool_map.entries.get_by_status(&Status::Pending) {
            let short_id = entry.inner.proposal_short_id();
            if snapshot.proposals().contains_proposed(&short_id) {
                proposals.push((short_id, entry.inner.to_owned()));
            } else if snapshot.proposals().contains_gap(&short_id) {
                gaps.push((short_id, entry.inner.to_owned()));
            }
        }

        for (id, entry) in proposals {
            debug!("begin to proposed: {:x}", id);
            if let Err(e) = tx_pool.proposed_rtx(&id) {
                callbacks.call_reject(tx_pool, &entry, e);
            } else {
                callbacks.call_proposed(tx_pool, &entry, false)
            }
        }

        for (id, entry) in gaps {
            debug!("begin to gap: {:x}", id);
            if let Err(e) = tx_pool.gap_rtx(&id) {
                debug!(
                    "Failed to add tx to gap {}, reason: {}",
                    entry.transaction().hash(),
                    e
                );
                callbacks.call_reject(tx_pool, &entry, e.clone());
            }
        }
    }
```

chenyukang: I think it's ok, at least we can avoid another query and the `unwrap`.

chenyukang:
@wyjin

I have finished some benchmark testing for comparing `develop` and this branch, here are some data, FYI:

develop branch, code: https://github.com/chenyukang/ckb/tree/develop-bench
```console
test component::tests::bench::test_container_bench_add        ... bench:   2,144,760 ns/iter (+/- 3,794,550)
test component::tests::bench::test_container_bench_add_remove ... bench:   9,826,341 ns/iter (+/- 336,570)
test component::tests::bench::test_container_bench_sort       ... bench:   4,752,366 ns/iter (+/- 418,446)
```

current branch bench, code: https://github.com/chenyukang/ckb/tree/tx_pool_refactor_bench

```console
test component::tests::bench::test_container_bench_add        ... bench:   1,956,728 ns/iter (+/- 4,352,650)
test component::tests::bench::test_container_bench_add_remove ... bench:  10,976,779 ns/iter (+/- 174,750)
test component::tests::bench::test_container_bench_sort       ... bench:   3,896,260 ns/iter (+/- 102,732)
```

It seems `sort` has been improved.

wyjin:
> @wyjin
> 
> I have finished some benchmark testing for comparing `develop` and this branch, here are some data, FYI:
> 
> develop branch, code: https://github.com/chenyukang/ckb/tree/develop-bench
> 
> ```
> test component::tests::bench::test_container_bench_add        ... bench:   2,144,760 ns/iter (+/- 3,794,550)
> test component::tests::bench::test_container_bench_add_remove ... bench:   9,826,341 ns/iter (+/- 336,570)
> test component::tests::bench::test_container_bench_sort       ... bench:   4,752,366 ns/iter (+/- 418,446)
> ```
> 
> current branch bench, code: https://github.com/chenyukang/ckb/tree/tx_pool_refactor_bench
> 
> ```
> test component::tests::bench::test_container_bench_add        ... bench:   1,956,728 ns/iter (+/- 4,352,650)
> test component::tests::bench::test_container_bench_add_remove ... bench:  10,976,779 ns/iter (+/- 174,750)
> test component::tests::bench::test_container_bench_sort       ... bench:   3,896,260 ns/iter (+/- 102,732)
> ```
> 
> It seems `sort` has been improved.

Is this good enough? `remove` appears to be slower because size of the pool never exceeds 3. I'd expect to see a major improvement in `remove` when the pool is bigger and the number of duplications in non-unique fields is also reasonably big.

The `develop` branch uses the upstream multi_index_map crate. I am curious/worried about how the performance compares to not using multi index map at all. (Is using multi index map causing performance regression?) Seems multi index is used in the `ckb2023` branch so probably this test is already done?

chenyukang:
> Is this good enough?

Yes, upstream `multi_index_map` is used in develop branch, but only for `proposed` pool. So I want to have a benchmark compare develop with this branch.

From the above benchmark data, we can say the performance is on the same level with trivial difference. I guess this difference won't have too much effect on the whole CKB, since there are `rpc` operation involved in.

I have concern on the memory, today I write some testing code to verify it. I constantly add `parent-child` transactions to the pool(this is the worst case, since we need to keep the links).

The memory footprint is:
```console
2w 65.7 MB
3w 86.1 MB
4w 131.5 MB
5w 158.2 MB
6w 201.4 MB
```

But we have `max_ancestors_count` to limit the depth.

Anyway, let's wait for the result from acceptance, I talked with them today for the testing.

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,544 @@
...
+        for id in &removed_ids {
+            self.remove_entry_links(id);
+        }
+
+        for id in removed_ids {
+            if let Some(entry) = self.remove_entry(&id) {
+                removed.push(entry);
+            }
+        }
+        removed
```
quake, line 219:
```suggestion
removed_ids.iter().filter_map(|id| self.remove_entry(id)).collect()
```

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,544 @@
...
+            let mut parent = entry.inner.clone();
+            match op {
+                EntryOp::Remove => parent.sub_descendant_weight(child),
+                EntryOp::Add => parent.add_descendant_weight(child),
+            }
+            let short_id = parent.proposal_short_id();
+            self.entries.modify_by_id(&short_id, |e| {
+                e.evict_key = parent.as_evict_key();
+                e.inner = parent;
+            });
```
quake, line 355: can we use `modify_by_id` directly here?
```suggestion
            // update parent score
            self.entries.modify_by_id(anc_id, |e| {
                match op {
                    EntryOp::Remove => e.inner.sub_descendant_weight(child),
                    EntryOp::Add => e.inner.add_descendant_weight(child),
                };
                e.evict_key = e.inner.as_evict_key();
            });
```

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,544 @@
...
+        }
+    }
+
+    fn update_descendants_index_key(&mut self, parent: &TxEntry, op: EntryOp) {
+        let descendants: HashSet<ProposalShortId> =
+            self.links.calc_descendants(&parent.proposal_short_id());
+        for desc_id in &descendants {
+            // update child score
```
quake, line 363: same question as `update_ancestors_index_key` fn's review comment

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,544 @@
...
+        // update parents references
+        for ancestor_id in &ancestors {
+            let ancestor = self
+                .entries
+                .get_by_id(ancestor_id)
+                .expect("pool consistent");
+            entry.add_ancestor_weight(&ancestor.inner);
+        }
+        if entry.ancestors_count > self.max_ancestors_count {
+            eprintln!("debug: exceeded maximum ancestors count");
```
quake, line 470: suggest to use debug level logger

tx-pool/Cargo.toml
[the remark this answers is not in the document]
chenyukang: Yes, I updated the crate in the `RBF` branch:
https://github.com/nervosnetwork/ckb/pull/4079/commits/3a254aed749058df198e00520e54d877fc0bf80c

tx-pool/src/pool.rs
```diff
@@ -338,251 +249,183 @@ impl TxPool {
...
         limit: usize,
         exclusion: &HashSet<ProposalShortId>,
     ) -> HashSet<ProposalShortId> {
         let mut proposals = HashSet::with_capacity(limit);
-        self.pending
-            .fill_proposals(limit, exclusion, &mut proposals);
-        self.gap.fill_proposals(limit, exclusion, &mut proposals);
+        self.pool_map
+            .fill_proposals(limit, exclusion, &mut proposals, &Status::Pending);
+        self.pool_map
```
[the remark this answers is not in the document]
chenyukang: The logic here is same as before, I was also confused on this, the code looks a bit counterintuitive here.

The `Gap` status is a internal status of `tx-pool`, `Gap` means it's proposed but haven't been ready to be added to block, it's waiting for another block submitted.
So it's a status for internal transition, generally `Pending` and `Gap` are both considered to `Pending` for outside of pool.

We also have `Spec` test for this scenario:

https://github.com/nervosnetwork/ckb/blob/384583a9019dfddd7d82bfde587c45ab15d00d32/test/src/specs/tx_pool/descendant.rs#L132-L156

Per my understanding, the parent tx is in `Gap`, if we don't call `fill_proposals` for `Gap`, the `template.proposals.len()` won't get to 2:

```rust
        let block = node.new_block_with_blocking(|template| template.proposals.len() != 2);
```

We can only process `MAX_BLOCK_PROPOSALS_LIMIT` in a block, `Gap` is also used to achieve pagination 

https://github.com/nervosnetwork/ckb/blob/384583a9019dfddd7d82bfde587c45ab15d00d32/tx-pool/src/block_assembler/mod.rs#L345

@zhangsoledad have any more comments on this?

tx-pool/src/component/pool_map.rs
```diff
@@ -0,0 +1,530 @@
...
+    Remove,
+}
+
+#[derive(MultiIndexMap, Clone)]
+pub struct PoolEntry {
+    #[multi_index(hashed_unique)]
+    pub id: ProposalShortId,
+    #[multi_index(ordered_non_unique)]
+    pub score: AncestorsScoreSortKey,
+    #[multi_index(ordered_non_unique)]
```
quake, line 48: is it better to use `hashed_non_unique` for performance since we don't need to iterate the pool map by status?

quake (approved):

tx-pool/src/component/pool_map.rs
chenyukang: fixed, thanks.

PR #3996, opened 2023-05-31
feat: Suspend support in spawn syscalls

xxuejie:

This PR fixes a bug that when suspend happens when a child machine created by the spawn syscall is running, CKB cannot correctly suspend & resume the VM.

What's Changed: the bug relating spawn syscall and suspend / resume feature is fixed.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/script/src/syscalls/spawn.rs
+++ b/script/src/syscalls/spawn.rs
@@ -7,7 +7,7 @@ use crate::syscalls::{
     SPAWN_MAX_PEAK_MEMORY, SPAWN_MEMORY_PAGE_SIZE, SPAWN_WRONG_MEMORY_LIMIT, WRONG_FORMAT,
 };
 use crate::types::{
-    set_vm_max_cycles, CoreMachineType, Machine, MachineContext, ResumableMachine, ResumeData,
+    set_vm_max_cycles, CoreMachineType, Machine, MachineContext, ResumableMachine, SpawnData,
 };
 use crate::TransactionScriptsSyscallsGenerator;
 use crate::{ScriptGroup, ScriptVersion};
@@ -147,7 +147,7 @@ where
             return Ok(true);
         }
         // Build child machine.
-        let resume_data = ResumeData::Spawn {
+        let spawn_data = SpawnData {
             callee_peak_memory: self.peak_memory + memory_limit,
             callee_memory_limit: memory_limit,
             content: Arc::new(Mutex::new(Vec::<u8>::new())),
@@ -161,7 +161,7 @@ where
             self.script_version,
             &self.syscalls_generator,
             cycles_limit,
-            &resume_data,
+            &spawn_data,
             &self.context,
         )?;
         // Get binary.
@@ -225,7 +225,7 @@ where
         // Run the child machine and check result.
         match machine_child.run() {
             Ok(data) => {
-                update_caller_machine(machine, data, machine_child.machine.cycles(), &resume_data)?;
+                update_caller_machine(machine, data, machine_child.machine.cycles(), &spawn_data)?;
                 Ok(true)
             }
             Err(VMError::CyclesExceeded) => {
@@ -235,7 +235,7 @@ where
                     .map_err(|e| VMError::Unexpected(format!("Failed to acquire lock: {}", e)))?;
                 context
                     .suspended_machines
-                    .push(ResumableMachine::new(machine_child, resume_data));
+                    .push(ResumableMachine::spawn(machine_child, spawn_data));
                 Err(VMError::CyclesExceeded)
             }
             Err(err) => Err(err),
@@ -250,28 +250,16 @@ pub fn build_child_machine<
     script_version: ScriptVersion,
     syscalls_generator: &TransactionScriptsSyscallsGenerator<DL>,
     cycles_limit: u64,
-    resume_data: &ResumeData,
+    spawn_data: &SpawnData,
     context: &Arc<Mutex<MachineContext>>,
 ) -> Result<Machine, VMError> {
-    let (callee_peak_memory, callee_memory_limit, content, content_length) = match resume_data {
-        ResumeData::Spawn {
-            callee_peak_memory,
-            callee_memory_limit,
-            content,
-            content_length,
-            ..
-        } => (
-            *callee_peak_memory,
-            *callee_memory_limit,
-            content,
-            *content_length,
-        ),
-        _ => {
-            return Err(VMError::Unexpected(
-                "Building child machine requires Spawn variant of ResumeData!".to_string(),
-            ))
-        }
-    };
+    let SpawnData {
+        callee_peak_memory,
+        callee_memory_limit,
+        content,
+        content_length,
+        ..
+    } = spawn_data;
 
     let machine_isa = script_version.vm_isa();
     let machine_version = script_version.vm_version();
@@ -288,15 +276,15 @@ pub fn build_child_machine<
         .into_iter()
         .fold(machine_builder, |builder, syscall| builder.syscall(syscall));
     let machine_builder = machine_builder.syscall(Box::new(
-        syscalls_generator.build_get_memory_limit(callee_memory_limit),
+        syscalls_generator.build_get_memory_limit(*callee_memory_limit),
     ));
     let machine_builder = machine_builder.syscall(Box::new(
-        syscalls_generator.build_set_content(Arc::clone(content), content_length),
+        syscalls_generator.build_set_content(Arc::clone(content), *content_length),
     ));
     let machine_builder = machine_builder.syscall(Box::new(syscalls_generator.build_spawn(
         script_version,
         script_group,
-        callee_peak_memory,
+        *callee_peak_memory,
         Arc::clone(context),
     )));
     let mut machine_child = Machine::new(machine_builder.build());
@@ -308,35 +296,26 @@ pub fn update_caller_machine<Mac: SupportMachine>(
     caller: &mut Mac,
     callee_exit_code: i8,
     callee_cycles: u64,
-    resume_data: &ResumeData,
+    spawn_data: &SpawnData,
 ) -> Result<(), VMError> {
-    let (content, caller_exit_code_addr, caller_content_addr, caller_content_length_addr) =
-        match resume_data {
-            ResumeData::Spawn {
-                content,
-                caller_exit_code_addr,
-                caller_content_addr,
-                caller_content_length_addr,
-                ..
-            } => (
-                content,
-                *caller_exit_code_addr,
-                *caller_content_addr,
-                *caller_content_length_addr,
-            ),
-            _ => return Ok(()),
-        };
+    let SpawnData {
+        content,
+        caller_exit_code_addr,
+        caller_content_addr,
+        caller_content_length_addr,
+        ..
+    } = spawn_data;
 
     caller.set_register(A0, Mac::REG::from_u32(0));
     caller.memory_mut().store8(
-        &Mac::REG::from_u64(caller_exit_code_addr),
+        &Mac::REG::from_u64(*caller_exit_code_addr),
         &Mac::REG::from_i8(callee_exit_code),
     )?;
     caller
         .memory_mut()
         .store_bytes(caller_content_addr.to_u64(), &content.lock().unwrap())?;
     caller.memory_mut().store64(
-        &Mac::REG::from_u64(caller_content_length_addr),
+        &Mac::REG::from_u64(*caller_content_length_addr),
         &Mac::REG::from_u64(content.lock().unwrap().len() as u64),
     )?;
     caller.add_cycles_no_checking(callee_cycles)?;
--- a/script/src/syscalls/spawn.rs
+++ b/script/src/syscalls/spawn.rs
@@ -297,7 +297,7 @@ pub fn build_child_machine<
         script_version,
         script_group,
         callee_peak_memory,
-        context.clone(),
+        Arc::clone(context),
     )));
     let mut machine_child = Machine::new(machine_builder.build());
     set_vm_max_cycles(&mut machine_child, cycles_limit);
--- a/script/src/syscalls/spawn.rs
+++ b/script/src/syscalls/spawn.rs
@@ -206,6 +206,10 @@ where
             argv_vec.push(cstr);
             addr += 8;
         }
+        // Deduct cycles used to build the child machine
+        let extra_cycles =
+            SPAWN_EXTRA_CYCLES_BASE + memory_limit * SPAWN_EXTRA_CYCLES_PER_MEMORY_PAGE;
+        machine_child.machine.add_cycles_no_checking(extra_cycles)?;
         // Load program into child machine.
         match machine_child.load_program(&program, &argv_vec) {
             Ok(size) => {
@@ -218,10 +222,6 @@ where
                 return Ok(true);
             }
         }
-        // Deduct cycles used to build the child machine
-        let extra_cycles =
-            SPAWN_EXTRA_CYCLES_BASE + memory_limit * SPAWN_EXTRA_CYCLES_PER_MEMORY_PAGE;
-        machine_child.machine.add_cycles_no_checking(extra_cycles)?;
         // Run the child machine and check result.
         match machine_child.run() {
             Ok(data) => {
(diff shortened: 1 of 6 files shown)


script/src/syscalls/spawn.rs
```diff
@@ -237,24 +218,128 @@ where
                 return Ok(true);
             }
         }
-        // Check result.
+        // Deduct cycles used to build the child machine
```
mohanson, line 221: We need to deduct extra cycles before `load_program`. Consider a scenario: an evil user tries to run a non-RISC-V program, and the syscall spawn will fail at `load_program`. We have initialized the VM instance( includes memory, syscalls, etc. ) for running the program, but the eval user only need to pay 500 cycles.

mohanson (approved):

zhangsoledad (approved):

script/src/verify.rs
```diff
@@ -1029,50 +1026,159 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
+                    ));
                 }
-                _ => Err(ScriptError::VMInternalError(format!("{error:?}"))),
+                _ => return Err(ScriptError::VMInternalError(format!("{error:?}"))),
             },
-        }
+        };
+    }
+
+    let (exit_code, cycles, _) = callee_data.unwrap();
```
eval-exec, line 1174: Is it possible for `callee_data` to be `None`?

script/src/verify.rs
```diff
@@ -1029,50 +1026,159 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
-                )));
-            }
-            load_ret.map_err(|e| ScriptError::VMInternalError(format!("{e:?}")))?;
+            // NOTE: previously, we made a distinction between machines
+            // that completes program loading without errors, and machines
+            // that fail program loading due to cycle limits. For the latter
+            // one, we won't generate any snapshots. Starting from this version,
+            // we will remove this distinction: when loading program exceeds
+            // maximum cycles, the error will be triggered when executing the
+            // first instruction. As as result, now all ResumableMachine will
```
eval-exec, line 1081:
```suggestion
            // first instruction. As a result, now all ResumableMachine will
```

script/src/verify.rs
```diff
@@ -1029,50 +1026,159 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
+            // on which limit to use. The cycles consumed to load a program, is
+            // thus at most 131072 or ~145920, which is far less than the cycle
+            // limit for running a single transaction (70 million or more). In
+            // reality it might be extremely rare that loading a program would
+            // result in exceeding cycle limits. Removing the distinction here,
+            // would help simply the code.
+            // * If you pay attention to the code now, we already have this behavior
+            // in the code: most syscalls use +add_cycles_no_checking+ in the code,
+            // meaning an error would not be immediately generated when cycle limit
+            // is reached, the error would be raise when executing the first instruction
```
eval-exec, line 1096:
```suggestion
            // is reached, the error would be raised when executing the first instruction
```

eval-exec (approved):

script/src/verify.rs
```diff
@@ -45,17 +47,17 @@ use core::sync::atomic::{AtomicBool, Ordering};
...
 impl ChunkState {
-    pub fn suspended(machine: ResumableMachine) -> Self {
-        ChunkState::Suspended(Some(machine))
+    pub fn suspended(machines: Vec<ResumableMachine>, context: Arc<Mutex<MachineContext>>) -> Self {
+        ChunkState::Suspended(machines, context)
     }
 
     pub fn suspended_type_id() -> Self {
-        ChunkState::Suspended(None)
+        ChunkState::Suspended(vec![], Default::default())
```
eval-exec, line 60: In `TransactionScriptsVerifier::verify_group_with_chunk`, if `verifier.verify()` get `Err(ScriptError::ExceededMaximumCycles(_))`, it will return an empty `vms`.

Is is possiable that the empty `vms` received by `TransactionScriptsVerifier::resume_from_snap` and cause unexpected behavior?

script/src/verify.rs
xxuejie: Ideally when we enter this function, we are expecting at least one VM instance, which will ensure `callee_data` is not `None`. I've added a check at the top of this function ensuring at least one VM exists.

script/src/verify.rs
xxuejie: The answer here has 2 sides:

* The only chance when `ChunkState::Suspended` contains an empty `vms` vector, is cycle limit is reached when we are verifying the embedded type ID script. However at the very top of `verify_group_with_chunk`, we are checking if current script is the embedded type ID script, so when we are entering `resume_from_snap` again, `resume_from_snap` will call `verify_group_with_chunk`, which checks against the embedded type ID script with special handling. So having empty `vms` here won't cause any problem.
* That being said, right now we have `resume_from_state` in addition to `resume_from_snap`, and `resume_from_state` is missing the check for th embedded type ID script, nor the empty `vms`, I've fixed this part. Good catch.

eval-exec: Thank you. :heart_decoration:

eval-exec (approved):

XuJiandong (changes requested):

script/src/types.rs
```diff
@@ -112,17 +112,109 @@ pub(crate) type Machine = AsmMachine;
...
+
+/// Data structure captured all environment data for a suspended machine
+#[derive(Clone, Debug)]
+pub enum ResumePoint {
+    Initial,
+    Spawn {
+        callee_peak_memory: u64,
+        callee_memory_limit: u64,
+        content: Vec<u8>,
+        content_length: u64,
```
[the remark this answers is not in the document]
xxuejie: I'm looking at a different angle: this is required to build [this syscall](https://github.com/nervosnetwork/ckb/blob/81ebaecf26f0240060654d8744a52986346293d1/script/src/syscalls/spawn.rs#L166-L169) used by a child. We don't know that whether a child machine has invoked `set_content` or not when suspended, so I believe this is required to build the full environment for a child machine.

xxuejie:
@eval-exec Personally I don't think this PR is an enhancement. To me the previous spawn implementation has a bug when used with the suspend / resume feature. This PR fixes the bug.

script/src/verify.rs
```diff
@@ -566,14 +585,16 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
 
     fn build_state(
         &self,
-        vm: Option<ResumableMachine>,
+        vms: Vec<ResumableMachine>,
```
quake, line 588: build_state is a pure fn, suggest to remove `self` or use `TransactionState::new` directly

script/src/verify.rs
```diff
@@ -934,7 +940,7 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
         &self,
         group: &ScriptGroup,
         max_cycles: Cycle,
-        snap: &Option<(Snapshot, Cycle)>,
+        snaps: Option<&[(Snapshot, Cycle, ResumePoint)]>,
```
quake, line 943: suggest to change use slice directly without an Option wrapper, an empty slice means no snapshoots:
```suggestion
        snaps: &[(Snapshot, Cycle, ResumePoint)],
```

script/src/types.rs
```diff
@@ -112,17 +112,109 @@ pub(crate) type Machine = AsmMachine;
...
+        caller_content_addr: u64,
+        caller_content_length_addr: u64,
+    },
+}
+
+/// Data structure captured all environment data for a suspended machine.
+/// While ResumePoint is used in snapshots for easier serialization, this
+/// contains data that can be used at runtime amongst running machines.
+#[derive(Clone, Debug)]
+pub enum ResumeData {
```
quake, line 142: I suggest to remove this enum, we can replace it by changing the struct `ResumableMachine` to enum, this will make the code shorter and easier to read:
```rust
pub enum ResumableMachine {
    Intial(Machine),
    Spawn(Machine, SpawnData),
}
```

xxuejie: I do suggest we keep this enum, the problem lies at [spawn syscalls](https://github.com/nervosnetwork/ckb/blob/077d20554aaa7df6daf4114fd20c5596657cb2a7/script/src/syscalls/spawn.rs#L246-L255), here we are using `ResumeData` to build the `Machine` that gets included in `ResumableMachine`, moving the data from `ResumeData` into `ResumableMachine` means we will have to pass in a ton of parameters here. Keeping `ResumeData` as its own data structure, IMO, is a more balanced solution.

script/src/verify.rs
xxuejie: Done

script/src/verify.rs
xxuejie: Done

script/src/types.rs
quake: there is a typo in my comment, this is an updated version:
```
pub enum ResumableMachine {
    Intial(Machine),
    Spawn(Machine, ResumeData),
}

pub struct ResumeData {
    callee_peak_memory: u64,
    callee_memory_limit: u64,
    content: Arc<Mutex<Vec<u8>>>,
    content_length: u64,
    caller_exit_code_addr: u64,
    caller_content_addr: u64,
    caller_content_length_addr: u64,
}
```

xxuejie: Done

script/src/verify.rs
```diff
@@ -1029,50 +1008,184 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
+        ));
+    }
+
+    let map_vm_internal_error = |error: VMInternalError| match error {
+        VMInternalError::CyclesExceeded => ScriptError::ExceededMaximumCycles(max_cycles),
+        _ => ScriptError::VMInternalError(format!("{error:?}")),
+    };
+
+    while !machines.is_empty() {
+        let mut machine = machines.pop().unwrap();
```
quake, line 1136:
```suggestion
    while let Some(mut machine) = machines.pop() {
```

script/src/verify.rs
```diff
@@ -1117,21 +1133,26 @@ fn run_vms(
...
             .map_err(map_vm_internal_error)?;
         }
 
         match machine.run() {
             Ok(code) => {
-                callee_data = Some((code, machine.machine.cycles(), resume_data));
+                callee_data = Some((code, machine.cycles()));
+                if let ResumableMachine::Spawn(_, data) = machine {
+                    spawn_data = Some(data);
+                }
```
quake:
```
if let ResumableMachine::Spawn(_, spawn_data) = machine {
  callee_data = Some((code, machine.cycles(), spawn_data));
}
```

script/src/verify.rs
```diff
@@ -1029,50 +1008,184 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
+                        context.suspended_machines.drain(..).collect()
+                    };
+                    // The inner most machine lives at the top of the vector,
+                    // reverse the list for natural order.
+                    new_suspended_machines.reverse();
+                    let mut suspended_machines: Vec<_> = machines.drain(..).collect();
+                    suspended_machines.push(machine);
+                    suspended_machines.append(&mut new_suspended_machines);
+                    return Ok(ChunkState::suspended(
+                        suspended_machines,
```
quake, line 1172: I think we can use the owned `machines` vec directly:
```suggestion
                    machines.push(machine);
                    machines.append(&mut new_suspended_machines);
                    return Ok(ChunkState::suspended(
                        machines,
```

script/src/verify.rs
```diff
@@ -1029,50 +1008,184 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
+            // reality it might be extremely rare that loading a program would
+            // result in exceeding cycle limits. Removing the distinction here,
+            // would help simply the code.
+            // * If you pay attention to the code now, we already have this behavior
+            // in the code: most syscalls use +add_cycles_no_checking+ in the code,
+            // meaning an error would not be immediately generated when cycle limit
+            // is reached, the error would be raised when executing the first instruction
+            // after the syscall. What's more, when spawn is loading a program
+            // to its child machine, it also uses +add_cycles_no_checking+ so it
+            // won't generate errors immediately. This means that all spanwed machines
```
quake, line 1097:
```suggestion
            // won't generate errors immediately. This means that all spawned machines
```

script/src/types.rs
```diff
@@ -112,41 +112,111 @@ pub(crate) type Machine = AsmMachine;
 #[cfg(not(has_asm))]
 pub(crate) type Machine = TraceMachine<CoreMachine>;
 
-pub struct ResumableMachine {
-    machine: Machine,
-    pub(crate) program_bytes_cycles: Option<Cycle>,
+/// Common data that would be shared amongst multiple VM instances.
+/// One sample usage right now, is to capture suspended machines in
+/// a chain of spanwed machines.
```
quake, line 117:
```suggestion
/// a chain of spawned machines.
```

script/src/verify.rs
xxuejie: This won't work, consider a case where we have a single script that does not call spawn, it gets suspended somehow. In case case, we will need `code` and `machine.cycles()` to update the final return result, but `spawn_data` will be missing.

quake: But I see that on line 1138 we did match `(Some, Some)`, in the case you describe, is there a problem too? Can you add a corresponding unit test?

xxuejie: Imagine a case where we have 3 scripts:

```
A -> B -> C
```

A calls B via spawn, B calls C via spawn. And during the execution of C, the VM gets suspended.

The code works like following on line 1183:

* C is the leaf VM, line 1138 will result in a miss. We don't need to update any information on C(which makes sense, C is the leaf VM, it is not any caller VM).
* When C terminates, `callee_data` is updated with exit code and cycles from C, `spawn_data` contains any data C set via `set_content`
* At this stage we will continue run B, line 1138 will result in a match since both `callee_data` and `spawn_data` are available, which makes sense, since B indeed calls spawn, it would require the exit code of a child script, and any data the child script updates via `set_content`. The cycles consumed to run C will also be added to B.
* When B terminates, it updates `callee_data` and `spawn_data` again.
* At this stage we will run A, line 1138 will result in a match. `callee_data` and `spawn_data` will be used to update A to reflect spawn results.
* When A terminates, it updates `callee_data`, but since it is not created from a spawn syscall, it does not update `spawn_data` per line 1153.
* At the final stage, we do `unwrap` on `callee_data`, to get exit code and cycles for the complete execution trace.

Now consider a different case:

```
A
```

We only executes A, A does not call spawn, the VM gets suspended when running A:

* Upon entry, line 1138 will be a miss since `callee_data` and `spawn_data` are both `None`.
* When A completes execution, it will update `callee_data` on line 1152, but it is not created from a spawn syscall, it does not update `spawn_data` per line 1153.
* Finally, we do `unwrap` on `callee_data`, to get the exit code and cycles for the complete execution trace.

In this second case, we will run into a scenario where `callee_data` contains data, but `spawn_data` stays None.

There is never a case when `spawn_data` contains data but `callee_data` is None. `callee_data` always contains the exit code and cycles for the last terminated machine.

Actually all the above behavior has been covered by tests already.

quake: I see, I didn't notice at first that there was an unwrap after the loop, thanks for the detailed explanation.

but I think we can use a simpler variable initialization here to remove the `Option` init / unwrap:

```rust
let (mut exit_code, mut cycles, mut spawn_data) = (0, 0, None);

while let Some(mut machine) = machines.pop() {
    if let Some(callee_spawn_data) = spawn_data.as_ref() {
        update_caller_machine(
...

    match machine.run() {
        Ok(code) => {
            exit_code = code;
            cycles = machine.cycles();
            if let ResumableMachine::Spawn(_, data) = machine {
                spawn_data = Some(data);
            }
```

xxuejie: Done

xxuejie:
> LGTM, please update the PR description.

Already updated.

PR #3999, opened 2023-06-01
Wait tokio tasks finish before CKB process exit

### LINKED ISSUE #3994: Fix ckb graceful shutdown
Because the official documentation is misleading, https://github.com/tokio-rs/tokio/issues/4231, resulting in the current graceful shutdown mechanism is actually incorrect, need to fix and explore how to simplify.

eval-exec:

Issue Number: close #3994 

Problem Summary:

What's Changed:

#### Implementation
- [x] Add a std::threads JoinHandle register to collect all child threads' `JoinHandle`
- [x] broadcast CKB's exit signal to child threads and tokio `CancellationToken`.
- [x] wait all ckb threads finish before process exit.
- [x] wait all ckb tokio spawned tasks finish before process exit.
#### Test
- [x] Unit test for `broadcast_exit_signals`, `wait_all_ckb_services_exit`, `register_thread`, `new_crossbeam_exit_rx` and `new_tokio_exit_rx`
- [x] Integration test: Bats clit test
  Ideas to implement integration test:
   1. start processes by `ckb run`
   2. wait a duration about 10 seconds, then kill ckb process.
   3. wait ckb process exit.
   4. assert there are log like `all ckb threads have been stopped` and `ckb shutdown` exist in the log file.
- [x] Fix bat-cli timeout [issue](https://github.com/nervosnetwork/ckb/actions/runs/5373271361/jobs/9747486128)

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


eval-exec:
https://github.com/nervosnetwork/ckb/blob/93fe7503d561a5b642bfd0893dc0a4c6fe7df7d6/ckb-bin/src/subcommand/run.rs#L68

This line is unreachable, beacause `TxPoolController::save_pool()` never return an `Err`.

eval-exec:
Ideas to implement integration test:
1. start processes by `ckb run`, `ckb miner`
2. wait a duration about 10 seconds, then send SIGINT to it.
3. wait process exit.
4. assert there are `all ckb threads have been stopped` and `all tokio tasks have been stopped` exist in the log file.

util/stop-handler/src/stop_register.rs
```diff
@@ -0,0 +1,78 @@
...
+            Ok(_) => {}
+            Err(TrySendError::Full(_)) => error!("send exit signal to channel failed since the channel is full, this should not happen"),
+            Err(TrySendError::Disconnected(_)) => {
+                info!("broadcast thread: channel is disconnected")
+            }
+        });
+}
+
+/// Register a thread `JoinHandle` to  `CKB_HANDLES`
+pub fn register_thread(name: &str, thread_handle: std::thread::JoinHandle<()>) {
```
chenyukang, line 71: The `name` is only for printing messages?
so it's not an issue when multiple `name` are same.

eval-exec: Yes.

util/stop-handler/Cargo.toml
chenyukang, line 1: Should we add a new version for `ckb-stop-handler`?
I'm not sure whether any other projects are using it.

eval-exec: Yes, we should.

util/stop-handler/src/stop_register.rs
```diff
@@ -0,0 +1,78 @@
...
+    let _ = exit_signal.recv();
+    debug!("received exit signal, broadcasting exit signal to all threads");
+    let mut handles = CKB_HANDLES.lock();
+    for (name, join_handle) in handles.thread_handles.drain(..) {
+        match join_handle.join() {
+            Ok(_) => {
+                debug!("wait thread {} done", name);
+            }
+            Err(e) => {
+                warn!("wait thread {}: ERROR: {:?}", name, e)
```
chenyukang, line 23: Did you verify the error log contains panic message?
https://stackoverflow.com/questions/69387239/what-is-in-err-when-i-try-to-join-a-panicked-thread

util/app-config/src/tests/graceful_shutdown.bats
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bats
+bats_load_library 'bats-assert'
+bats_load_library 'bats-support'
+
```
chenyukang, line 4: can we add a test case, some sub-services terminated before receiving exit notification, and then the `wait_all_ckb_services_exit` also works for this scenario, maybe also double confirm the log messages.

util/stop-handler/src/stop_register.rs
eval-exec: This WARN log didn't print thread's panic stacktrace, I should use:
```rust
    warn!("wait thread {}: ERROR: {:?}",name,e.downcast_ref::<String>())
```

util/app-config/src/tests/graceful_shutdown.bats
eval-exec: Yes, we can.

util/stop-handler/src/stop_register.rs
eval-exec: Update:
If the `panic!` use a formatter, `e.downcast_ref::<&str>()` will get `None`, so I have to use `e.downcast_ref::<String>()`:
```rust
                let msg = e.downcast_ref::<str>().unwrap_or_else(|| {
                    e.downcast_ref::<String>()
                        .map(|s| s.as_str())
                        .unwrap_or("downcast_ref didn't get any error")
                });
                warn!("wait thread {}: ERROR: {:?}", name, msg)
```

zhangsoledad (approved):

PR #4022, opened 2023-06-19
fix: 2023 vm version check

zhangsoledad:

Until the 2023 rule is enabled, rpc should reject Data2 transactions

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/test/src/specs/hardfork/v2023/vm_version2.rs
+++ b/test/src/specs/hardfork/v2023/vm_version2.rs
@@ -1,5 +1,8 @@
 use crate::{
-    util::{cell::gen_spendable, check::is_transaction_committed},
+    util::{
+        cell::gen_spendable,
+        check::{assert_epoch_should_less_than, is_transaction_committed},
+    },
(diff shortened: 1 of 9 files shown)


test/src/specs/hardfork/v2023/vm_version2.rs
```diff
@@ -0,0 +1,408 @@
...
+        &self,
+        inputs: &mut impl Iterator<Item = packed::CellInput>,
+        script: &NewScript,
+        max_vm_version: u8,
+    ) {
+        for vm_version in 0..=RPC_MAX_VM_VERSION {
+            let res = if vm_version <= max_vm_version {
+                ExpectedResult::ShouldBePassed
+            } else if vm_version <= MAX_VM_VERSION {
+                ExpectedResult::IncompatibleVmV2
```
eval-exec, line 333: I think this branch is unreachable.

PR #4034, opened 2023-06-21
fix: get_consensus compatibility

zhangsoledad:

Revert the formatting changes to get_consensus to make it compatible with the previous version

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


miner/src/lib.rs
```diff
@@ -23,6 +24,7 @@ impl From<BlockTemplate> for Work {
     fn from(block_template: BlockTemplate) -> Work {
         let work_id = block_template.work_id;
         let block: Block = block_template.into();
+        info!("block {}", block.transactions().len());
```
eval-exec, line 27: I think it should be a `debug` level log.

eval-exec (approved):

PR #4058, opened 2023-07-05
Fix `Relayer::request_proposal_txs` dedup proposals

eval-exec:

https://github.com/nervosnetwork/ckb/blob/345a88140a9f6d41d57afa76fc9ac0bc52bdd0cf/sync/src/relayer/mod.rs#L235-L243

`proposals.dedup()` only [remove consecutive repeated elements](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup), this is not expected.

What's Changed:

### Related changes

- use [`iterator_tools unique()`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique) to remove duplicate proposals

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/sync/src/relayer/mod.rs
+++ b/sync/src/relayer/mod.rs
@@ -239,9 +239,9 @@ impl Relayer {
         nc: &dyn CKBProtocolContext,
         peer: PeerIndex,
         block_hash_and_number: BlockNumberAndHash,
-        proposals: Vec<packed::ProposalShortId>,
+        proposals: impl Iterator<Item = packed::ProposalShortId>,
     ) {
-        let proposals: Vec<ProposalShortId> = proposals.into_iter().unique().collect_vec();
+        let proposals: Vec<ProposalShortId> = proposals.unique().collect_vec();
         let tx_pool = self.shared.shared().tx_pool_controller();
         let fresh_proposals = match tx_pool.fresh_proposals_filter(proposals) {
             Err(err) => {
--- a/sync/src/relayer/mod.rs
+++ b/sync/src/relayer/mod.rs
@@ -39,6 +39,7 @@ use ckb_types::{
     prelude::*,
 };
 use ckb_util::Mutex;
+use itertools::Itertools;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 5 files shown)


sync/src/relayer/mod.rs
```diff
@@ -238,9 +239,9 @@ impl Relayer {
         nc: &dyn CKBProtocolContext,
         peer: PeerIndex,
         block_hash_and_number: BlockNumberAndHash,
-        mut proposals: Vec<packed::ProposalShortId>,
+        proposals: Vec<packed::ProposalShortId>,
     ) {
-        proposals.dedup();
```
driftluo, line 243: I don't think there is a duplicate here

this fn used in two way：
https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/block_transactions_process.rs#L104
https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/compact_block_process.rs#L80

Will there be no duplicate proposals in the block? @zhangsoledad

eval-exec: https://github.com/nervosnetwork/ckb/blob/345a88140a9f6d41d57afa76fc9ac0bc52bdd0cf/sync/src/relayer/compact_block_process.rs#L62-L63

I couldn't find any duplicate check for `compact_block.proposals()` in the code?

zhangsoledad: > https://github.com/nervosnetwork/ckb/blob/345a88140a9f6d41d57afa76fc9ac0bc52bdd0cf/sync/src/relayer/compact_block_process.rs#L62-L63
> 
> I couldn't find any duplicate check for `compact_block.proposals()` in the code?

CompactBlockVerifier checks whether the proposal is duplicated, and the proposal of a block is also not allowed to be duplicated, and there is a corresponding check, but we do not limit the proposal to two uncles (at most two), so it may still occur here  https://github.com/nervosnetwork/ckb/blob/develop/sync/src/relayer/block_transactions_process.rs#L104
@driftluo  help to confirm.

zhangsoledad: Also from a defensive programming point of view, I would prefer to add processing explicitly.

zhangsoledad: In addition, from the efficiency, it would be better to change this interface to `impl trait`
```rust
pub fn request_proposal_txs(
        &self,
        ...
        proposals: impl Iterator<Item = packed::ProposalShortId>,
    ) 
```

PR #4073, opened 2023-07-20
refactor: split ckb-gen-types from ckb-types

EthanYuan:

Split out Molecule generated types while providing `no-std` support to `ckb-std`, replacing crates `ckb-standalone-types`.

- Separated generated types from `ckb-types`, named as `ckb-gen-types`.
- `ckb-types` re-exports `ckb-gen-types` to ensure compatibility with other crates that reference it.
- `ckb-gen-types` comes with `no-std` support and can fully replace `ckb-standalone-types`, which `ckb-std` relies on.
- The basic principles of extracting conversions and extensions in `ckb-gen-types` are as follows:
    - If the conversion or extension involves `std` types, it will be kept in `ckb-gen-types` and conditionally compiled with the "std" feature.
    - If the conversion or extension involves types from third-party crates, it will also remain in `ckb-gen-types`.
    - If the conversion or extension involves types from `ckb-types::core`, it will be kept in `ckb-types`, and a custom `trait` will be placed in the `prelude` to address `orphan rules`.
- `ckb-gen-types` adds feature controls to several sub-modules in the extension, following the approach used in [`ckb-standalone-types`](https://github.com/nervosnetwork/ckb-std/commit/f15436829c895546751d1df6101cc6b0887bd2b9) to be compatible with `ckb-std`:
  - `calc-hash`: no-std
  - `check-data`: no-std
  - `serialized-size`: no-std
  - `std`: default, enables all in the std environment.
- A feature, `ckb-contract`, has been added to `ckb-hash` to support compatibility with `ckb-std` and allow specifying the use of `blake2b-ref`, following the approach used in `ckb-standalone-types`.
- Made some trade-offs in `ckb-gen-types` (perhaps a better approach could be considered):
    - To maintain compatibility with `ckb-std`,  `core::ScriptTypeHash`  have been kept in `ckb-gen-types`.

What's Changed:

### Related changes

- `ckb-std` update for test `ckb-gen-types`: https://github.com/EthanYuan/ckb-std/tree/update
- `ckb-x64-simulator` (`ckb-std`'s dep) update for test: https://github.com/EthanYuan/ckb-x64-simulator/tree/update

### Check List 

Tests 

- Unit test
- Integration test
  - After replacing `ckb-standalone-types` with `ckb-gen-types`, `ckb-std` passes the tests successfully.

Side effects

- Performance regression

### Release note 

```release-note
```


chenyukang (approved):

util/gen-types/src/generated/blockchain.rs
eval-exec, line 1: The `Makefile` should be modified:
https://github.com/nervosnetwork/ckb/blob/f8aa2f52f62989b39751637462e85503ad81569f/Makefile#L228

eval-exec (approved):

util/gen-types/src/generated/blockchain.rs
EthanYuan: Thanks, fixed.

EthanYuan: When I modify it to:

```makefile
GEN_MOL_IN_DIR := util/gen-types/schemas
GEN_MOL_OUT_DIR := util/gen-types/src/generated
```
It causes failures in the `ci_integration_tests_windows.yaml`. Do you have any suggestions?

EthanYuan: Rerunning the ci again passes the test, and there seems to be some randomness to it.

zhangsoledad (approved):

util/types/src/core/blockchain.rs
```diff
@@ -2,87 +2,15 @@ use ckb_error::OtherError;
...
 use crate::packed;
 
-/// Specifies how the script `code_hash` is used to match the script code and how to run the code.
-/// The hash type is split into the high 7 bits and the low 1 bit,
-/// when the low 1 bit is 1, it indicates the type,
-/// when the low 1 bit is 0, it indicates the data,
-/// and then it relies on the high 7 bits to indicate
-/// that the data actually corresponds to the version.
-#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
-pub enum ScriptHashType {
```
eval-exec, line 12: Why we remove `Hash`?

PR #4074, opened 2023-07-20
Fix `FeeOfMultipleMaxBlockProposalsLimit` failed on MacOS CI action

eval-exec:

This [CI job failed](https://github.com/nervosnetwork/ckb/actions/runs/5596018174/jobs/10232539158), and we can't reproduce it.

What's Changed:

### Related changes
- Try to fix `FeeOfMultipleMaxBlockProposalsLimit` 

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/test/src/specs/mining/fee.rs
+++ b/test/src/specs/mining/fee.rs
@@ -146,7 +146,7 @@ impl Spec for FeeOfMultipleMaxBlockProposalsLimit {
         });
 
         (0..multiple).for_each(|_| {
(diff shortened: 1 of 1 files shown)


chenyukang:
There are similar test-cases such as `ConflictInGap` and `ProposeTransactionButParentNot`? maybe we need to assert the length `transactions` of block in these two.

quake (approved):

PR #4079, opened 2023-07-26
Tx pool Replace-by-fee

### LINKED ISSUE #3734: Support Replace-by-fee in memory pool
CKB only supports Child-pay-for-parent fee-bumping mechanism, but CPFP is not available in some scenarios. For example, when all the cell outputs have a since precondition.

Add opt-in mechanism to replace a transaction in the memory pool.

- The user must set a flag to show that a transaction can be replaced.
- The new transaction must have a higher fee weight and a total fee amount.

#### References

- https://bitcoincore.org/en/faq/optin_rbf/

chenyukang:

Add feature of Replace-by-fee for `tx-pool`.

Fixes #3734

Problem Summary:

What's Changed:

This RBF feature enables users to replace a transaction with a higher fee rate when pending in `tx-pool`, which is useful when the transaction is stuck in the `tx-pool`:
- Add `min_rbf_rate` in `ckb.toml` with default value `1500`, which means the minimum extra fee rate for RBF, the unit is `shannons/KB`
- Add fields `fee` and `min_replace_fee` in `get_transaction`, which means the the minimal fee need to pay for RBF for a specific transaction
- `min_replace_fee = sum(replaced_tx_fee) + (min_rbf_rate * size)`
- The replaced transaction will be removed from `tx-pool` and with the status `Rejected`.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


tx-pool/src/process.rs
```diff
@@ -949,36 +1001,48 @@ impl TxPoolService {
     }
 }
 
-type PreCheckedTx = (Byte32, Arc<ResolvedTransaction>, TxStatus, Capacity, usize);
+type PreCheckedTx = (
```
eval-exec, line 1004: Suggest to add a code comment for every element of this type.

tx-pool/src/process.rs
```diff
@@ -949,36 +1001,48 @@ impl TxPoolService {
...
+    Byte32,
+    Arc<ResolvedTransaction>,
+    TxStatus,
+    Capacity,
+    usize,
+    HashSet<ProposalShortId>,
+);
```
eval-exec, line 1011: Suggest to add a code comment for every element of this type.

tx-pool/src/process.rs
```diff
@@ -202,13 +217,34 @@ impl TxPoolService {
...
+                // Try normal path first, if double-spending check success we don't need RBF check
+                // this make sure RBF won't introduce extra performance cost for hot path
+                let res = resolve_tx(tx_pool, &snapshot, tx.clone(), false);
+                match res {
+                    Ok((rtx, status)) => {
+                        let fee = check_tx_fee(tx_pool, &snapshot, &rtx, tx_size)?;
+                        Ok((tip_hash, rtx, status, fee, tx_size, HashSet::new()))
+                    }
+                    Err(err) => {
+                        if tx_pool.enable_rbf() && matches!(err, Reject::Resolve(_)) {
```
eval-exec, line 233: Is it possible to have a scenario where `err` is `RBFRejected` when we reach this line?

tx-pool/src/pool.rs
```diff
@@ -465,6 +529,114 @@ impl TxPool {
...
+        rtx: &ResolvedTransaction,
+        conflicts: &HashSet<ProposalShortId>,
+        fee: Capacity,
+        tx_size: usize,
+    ) -> Result<(), Reject> {
+        assert!(self.enable_rbf());
+        assert!(!conflicts.is_empty());
+
+        let short_id = rtx.transaction.proposal_short_id();
+        // Rule #4, new tx' fee need to higher than min_rbf_fee computed from the tx_pool configuration
```
eval-exec, line 544:
```suggestion
        // Rule #4, new tx's fee need to higher than min_rbf_fee computed from the tx_pool configuration
```

tx-pool/src/pool.rs
```diff
@@ -84,6 +85,66 @@ impl TxPool {
...
+
+        let extra_rbf_fee = self.config.min_rbf_rate.fee(size as u64);
+        let replaced_sum_fee: Capacity =
+            entries
+                .iter()
+                .map(|c| c.inner.fee)
+                .fold(Capacity::zero(), |acc, x| {
+                    acc.safe_add(x).unwrap_or_else(|_| {
+                        error!("replaced_sum_fee {} overflow by add {}", acc, x);
+                        Capacity::zero()
```
eval-exec, line 133: The `err` is not returned to upper level.

tx-pool/src/pool.rs
```diff
@@ -465,6 +529,114 @@ impl TxPool {
...
+        // Rule #3, new tx's fee need to higher than conflicts, here we only check the root tx
+        if let Some(min_replace_fee) = self.calculate_min_replace_fee(conflicts, tx_size) {
+            if fee < min_replace_fee {
+                return Err(Reject::RBFRejected(format!(
+                    "Tx's current fee is {}, expect it to >= {} to replace old txs",
+                    fee, min_replace_fee,
+                )));
+            }
+        } else {
+            panic!("calculate_min_replace_fee must success");
```
eval-exec, line 554: How about return an `err` to upper level, don't panic.

tx-pool/src/process.rs
chenyukang: No possible, the `resolve_tx_from_pool` always return `Reject::Resolve` for error.

tx-pool/src/pool.rs
chenyukang: Sounds reasonable.

eval-exec (approved):

tx-pool/src/pool.rs
```diff
@@ -465,6 +518,116 @@ impl TxPool {
...
+                "calculate_min_replace_fee failed".to_string(),
+            ));
+        }
+
+        let pool_entries = conflicts
+            .iter()
+            .map(|id| {
+                self.get_pool_entry(id)
+                    .expect("conflict Tx should be in pool")
+            })
```
quake, line 553: To minimize the potential bug impact, I suggest using `filter_map` here instead of `map/expect`

tx-pool/src/pool.rs
```diff
@@ -84,6 +87,53 @@ impl TxPool {
...
+        &self,
+        conflicts: &HashSet<ProposalShortId>,
+        size: usize,
+    ) -> Option<Capacity> {
+        let entries = conflicts
+            .iter()
+            .map(|id| {
+                self.get_pool_entry(id)
+                    .expect("conflict Tx should be in pool")
+            })
```
quake, line 114: To minimize the potential bug impact, I suggest using `filter_map` here instead of `map/expect`

tx-pool/src/pool.rs
```diff
@@ -465,6 +518,116 @@ impl TxPool {
...
+        let conflicts = pool_entries
+            .iter()
+            .map(|e| e.inner.clone())
+            .collect::<Vec<_>>();
+
+        // Rule #2, new tx don't contain any new unconfirmed inputs
+        let mut inputs = HashSet::new();
+        for c in conflicts.iter() {
+            inputs.extend(c.transaction().input_pts_iter());
+        }
```
quake, line 567: no need to clone pool entries, we may use the `vec#iter` directly:

```suggestion
        // Rule #2, new tx don't contain any new unconfirmed inputs
        let mut inputs: HashSet<ckb_types::packed::OutPoint> = HashSet::new();
        for c in pool_entries.iter() {
            inputs.extend(c.inner.transaction().input_pts_iter());
        }
```

```rust
        for conflict in pool_entries.iter() {
            let descendants = self.pool_map.calc_descendants(&conflict.id);
```

tx-pool/src/pool.rs
```diff
@@ -465,6 +518,116 @@ impl TxPool {
...
+        rtx: &ResolvedTransaction,
+        conflicts: &HashSet<ProposalShortId>,
+        fee: Capacity,
+        tx_size: usize,
+    ) -> Result<(), Reject> {
+        assert!(self.enable_rbf());
+        assert!(!conflicts.is_empty());
+
+        let short_id = rtx.transaction.proposal_short_id();
+        // Rule #4, new tx's fee need to higher than min_rbf_fee computed from the tx_pool configuration
```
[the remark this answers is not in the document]
chenyukang: `Rule #1` is the conflicted `tx` need to be declared as `replaceable`, in bitcoin it's `sequence number below MAX-1`.
We plan to not introduce this option for a `tx` currently.

tx-pool/src/pool.rs
chenyukang: I think it's better to use `map/expect` since it's like an assertion, `self.get_pool_entry(id)` returning `None` implies an inconsistency(invalid id or something), even we skip the error the result may be wrong, is it is a worser situation.

chenyukang: ah, let me reconsider it to confirm it won't impact other tx by checking the top-down path.

chenyukang: fixed in https://github.com/nervosnetwork/ckb/pull/4079/commits/95e3596c3943ec84116e6d2be9146e927eb89c7b, thanks.

quake (approved):

PR #4089, opened 2023-08-02
chore(test): add ctrl handler and fix integration test timeout issue

chenyukang:

Problem Summary:

There is a scenario `make integration` will hang for a long time when there is a bug in a `spec`, which may happen since we have a lot functions such as `mine_with_blocking`.

Proper `ctrl` handler will print out the running spec name, so that we may narrow the scope of root cause.

There is a odd scenario `make integration` will hang forever, it's safe to add `timeout` in `CI`, there is a trivial bug for that.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -135,8 +135,7 @@ fn main() {
         for name in cloned_running_names.lock().iter() {
             warn!("spec {} is still not finished", name);
         }
-        // sleep 1 second to wait for the log flush
-        std::thread::sleep(Duration::from_secs(1));
+        log::logger().flush();
         std::process::exit(1);
     })
     .expect("Error setting Ctrl-C handler");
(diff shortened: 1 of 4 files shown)


eval-exec (approved):

test/src/main.rs
```diff
@@ -122,6 +125,22 @@ fn main() {
...
+            "Total {} specs are not finished",
+            cloned_running_names.lock().len()
+        );
+        for name in cloned_running_names.lock().iter() {
+            warn!("spec {} is still not finished", name);
+        }
+        // sleep 1 second to wait for the log flush
+        std::thread::sleep(Duration::from_secs(1));
+        std::process::exit(1);
+    })
```
eval-exec, line 141: `drop(logger_guard);` will flush log data.

chenyukang: won't move `logger_guard`.

zhangsoledad (approved):

PR #4090, opened 2023-08-04
feat: add CellsCommitments softfork

quake:

This PR adds a softfork implementation of CellsCommitments that can be used to prove transaction output (cell) existence and status (live / dead). 

The implementation uses the merkle mountian range (MMR) as internal data structure, the hash of cell status is stored as MMR element. This PR is still in draft, and is created only for early code review. Please noted that no additional unit or integration tests have been added to this PR, as the relevant code logic has already been covered by the original tests, e.g. whether the MMR root was updated correctly when the chain was reorging.

There are 4 more TODOs.

- [ ] Because the transactions in the block will affect the MMR root, the original way of adding transactions by blank block template without modifying the extension field is no longer suitable for the integration test, there are about 30+ integration tests that need to be modified, and some of them have been modified so far, and need to be handled in a unified way.

- [x] lack of 2 RPC to generate and verify MMR proof.

- [ ] DB migration, there may be a need for an efficient way to update MMR for historical data.
 
- [x] the corresponding RFC is being prepared, and will be submitted together with the RFC when this PR is formally submitted. https://github.com/nervosnetwork/rfcs/pull/424

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- There may be performance regression, need more test on mainnet and testnet real data.

### Release note 

```release-note
```

### DIFF
--- a/store/src/cell.rs
+++ b/store/src/cell.rs
@@ -1,6 +1,11 @@
 use crate::{ChainStore, StoreTransaction};
-use ckb_error::Error;
-use ckb_types::{core::BlockView, packed, prelude::*};
+use ckb_error::{Error, InternalErrorKind};
+use ckb_types::{
+    core::{BlockNumber, BlockView},
+    packed,
+    prelude::*,
+    utilities::merkle_mountain_range::{hash_out_point_and_status, CellStatus},
+};
 use std::collections::HashMap;
 
 /**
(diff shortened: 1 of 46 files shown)


store/src/cell.rs
```diff
@@ -26,6 +31,39 @@ use std::collections::HashMap;
...
+            }
+        }
+
+        for out_point in tx.output_pts().into_iter() {
+            let hash = hash_out_point_and_status(&out_point, block_number, BlockNumber::MAX);
+            let mmr_position = cells_root_mmr
+                .push(hash)
+                .map_err(|e| InternalErrorKind::MMR.other(e))?;
+            let cell_status = CellStatus::new(mmr_position, block_number);
+            txn.insert_cells_root_mmr_status(&out_point, &cell_status)?;
```
doitian, line 59: Ensure to add test cases that cells created in the block are consumed in other transactions in the same block.

PR #4093, opened 2023-08-09
fix: data2 value

zhangsoledad:

The hash type is split into the high 7 bits and the low 1 bit, when the low 1 bit is 1, it indicates the type, when the low 1 bit is 0, it indicates the data, and then it relies on the high 7 bits to indicate that the data actually corresponds to the version.
So data2 should be 4.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/util/types/src/core/blockchain.rs
+++ b/util/types/src/core/blockchain.rs
@@ -3,6 +3,11 @@ use ckb_error::OtherError;
 use crate::packed;
 
 /// Specifies how the script `code_hash` is used to match the script code and how to run the code.
+/// The hash type is split into the high 7 bits and the low 1 bit,
+/// when the low 1 bit is 1, it indicates the type,
+/// when the low 1 bit is 0, it indicates the data,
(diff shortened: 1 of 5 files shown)


util/types/src/core/blockchain.rs
```diff
@@ -46,7 +51,7 @@ impl TryFrom<packed::Byte> for ScriptHashType {
 impl ScriptHashType {
     #[inline]
     pub(crate) fn verify_value(v: u8) -> bool {
-        v <= 3
+        v <= 4
```
eval-exec, line 54: if `v` is `3` , this will return `true`.

eval-exec (approved):

eval-exec:
CI action failed, this PR need re-execute `make gen-rpc-doc`.

eval-exec (approved):

Issue #4092, opened 2023-08-08
Lock the epoch length in a devchain using mainnet data


phroi:
#### Background
Currently I'm working on [rescuing dCKB deposits](https://github.com/dckb-rescuer) by writing a new web interface. Given the lack of documentation, it seems impossible to re-deploy correctly those contracts on a brand new devnet.

#### Is your feature request related to a problem? Please describe.
[Following @doitian instructions](https://talk.nervos.org/t/how-to-fork-mainnet-l1-into-a-devnet/7329/5), I'm running a devchain using existing mainnet data.

Now I want to speed up this devnet to test dCKB and NervosDAO interactions, so I followed [the steps previously suggested](https://talk.nervos.org/t/is-there-any-way-to-speed-up-the-blockchain-in-a-way-that-180-epochs-happen-in-a-reasonable-time-frame-in-the-ckb-local-devchain/7163), except for changing the `genesis_epoch_length` as indicated.

The issue is the epoch length doesn't seem much affected by them, as currently one epoch is more than 16 minutes.

 To fully reproduce this issue follow the [steps indicated on the Readme](https://github.com/dckb-rescuer/v1-interface/tree/ed3b384967d24cbbf297ec2a2e41c356bf90668b).

#### Describe the solution you'd like
[@doitian suggested](https://talk.nervos.org/t/how-to-fork-mainnet-l1-into-a-devnet/7329/9): 

> It seems we need a feature to lock the epoch length since a starting epoch for dev environment.

Keep up the great work 💪
Phroi

eval-exec:
Hello, have you tried speeding up the ckb miner? 

You can achieve this by setting lower values for `miner.client.poll_interval` and `miner.workers.value` in `ckb-miner.toml` file , ensuring that `miner.client.poll_interval` is smaller than `miner.workers.value`.

phroi:
@eval-exec the relevant part of `ckb-miner.toml` was like this:
```
[miner.client]
rpc_url = "http://127.0.0.1:8114/"
block_on_submit = true

# block template polling interval in milliseconds
poll_interval = 1000

# enable listen notify mode
# listen = "127.0.0.1:8888"

[[miner.workers]]
worker_type = "Dummy"
delay_type = "Constant"
value = 200 # instead of 5000
```

phroi:
Now using @eval-exec  `ckb-miner.toml` parameters one epoch is down to 4 minutes :partying_face: 

[...]

PR #4097, opened 2023-08-11
BREAK CHANGE! Consensus: Allow `epoch_duration_target` to impact epoch length in `Dummy` mode.

eval-exec:

Related issue: #4092 

Problem Summary:

What's Changed:

### Related changes

- let `epoch_duration_target` affect epoch length when `permanent_difficulty_in_dummy` is enabled in `Dummy` mode.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
Note: This PR introduce a break change. When the CKB is in `Dummy` mode and the `permanent_difficulty_in_dummy` option is set to `true`, the epoch length will be influenced by the `epoch_duration_target` parameter. Specifically, the epoch length will be calculated as `epoch_duration_target / 8`.
```

### DIFF
--- a/.github/workflows/ci_integration_tests_ubuntu.yaml
+++ b/.github/workflows/ci_integration_tests_ubuntu.yaml
@@ -93,7 +93,7 @@ jobs:
     - uses: actions/checkout@v3
     - name: build ckb and run bats cli test
       run: |
-        make build
+        make prod
         make cli-test
       shell: bash
     env:
--- a/spec/src/consensus.rs
+++ b/spec/src/consensus.rs
@@ -804,12 +804,7 @@ impl Consensus {
                         let remainder_reward =
                             Capacity::shannons(primary_epoch_reward % epoch.length());
 
-                        let mut next_epoch_length = epoch.length();
-                        let epoch_length_from_param =
-                            cmp::max(self.epoch_duration_target() / MIN_BLOCK_INTERVAL, 1);
-                        if next_epoch_length != epoch_length_from_param {
-                            next_epoch_length = self.epoch_duration_target() / MIN_BLOCK_INTERVAL;
-                        }
+let next_epoch_length = (self.epoch_duration_target() + MIN_BLOCK_INTERVAL - 1) / MIN_BLOCK_INTERVAL
 
                         let dummy_epoch_ext = epoch
                             .clone()
--- a/spec/src/consensus.rs
+++ b/spec/src/consensus.rs
@@ -804,6 +804,13 @@ impl Consensus {
                         let remainder_reward =
                             Capacity::shannons(primary_epoch_reward % epoch.length());
 
+                        let mut next_epoch_length = epoch.length();
+                        let epoch_length_from_param =
+                            cmp::max(self.epoch_duration_target() / MIN_BLOCK_INTERVAL, 1);
+                        if next_epoch_length != epoch_length_from_param {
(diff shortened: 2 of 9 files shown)


spec/src/consensus.rs
```diff
@@ -804,6 +804,13 @@ impl Consensus {
                         let remainder_reward =
                             Capacity::shannons(primary_epoch_reward % epoch.length());
 
+                        let mut next_epoch_length = epoch.length();
+                        let epoch_length_from_param =
+                            cmp::max(self.epoch_duration_target() / MIN_BLOCK_INTERVAL, 1);
+                        if next_epoch_length != epoch_length_from_param {
+                            next_epoch_length = self.epoch_duration_target() / MIN_BLOCK_INTERVAL;
+                        }
```
quake, line 812: simple int div round up:
```suggestion
let next_epoch_length = (self.epoch_duration_target() + MIN_BLOCK_INTERVAL - 1) / MIN_BLOCK_INTERVAL
```

eval-exec: if `self.epoch_duration()` is 0, `next_epoch_length` will be 0, this is wrong.

eval-exec: How about this?
```rust
let next_epoch_length = 1 + (self.epoch_duration_target() + MIN_BLOCK_INTERVAL) / MIN_BLOCK_INTERVAL;
```

quake: I think we have checked `epoch_duration_target` non-zero somewhere already, it is used as divisor in fn `build_genesis_epoch_ext`.

eval-exec:
`make cli-test` has been successfully executed.
I invite @zhangsoledad to review this PR and provide feedback.

quake (approved):

eval-exec:
https://github.com/nervosnetwork/ckb/blob/1ca9a324418e5e4ff3c299d11094ff093e773abb/test/src/specs/dao/satoshi_dao_occupied.rs#L145

I think previous usage of `epoch_duration_target` is wrong,
`epoch_duration_target` isn't how many blocks in a epoch, it's a duration. I'm going to modify `Some(2)` to `Some(2 * 8)`

eval-exec:
Fixed CI issue.

quake (approved):

eval-exec:
It's strange, all integration tests passed on my machine, but they failed on the GitHub merge queue. Investigating why...

https://github.com/nervosnetwork/ckb/actions/runs/5862389610/job/15894109895

EthanYuan (approved):

.github/workflows/ci_integration_tests_ubuntu.yaml
```diff
@@ -93,7 +93,7 @@ jobs:
     - uses: actions/checkout@v3
     - name: build ckb and run bats cli test
       run: |
-        make build
+        make prod
```
chenyukang, line 96: I'm not sure, why change `build` to `prod`, it's not related to the issue we are resolving in this PR?

eval-exec: `make build` and `make prod` are distinct. The former does not include features such as `deadlock_detection` and `with_sentry`. And CKB packages are released using `make prod`. Therefore, I believe it would be more appropriate to use `make prod` for the `make cli-test` command.

chenyukang: sound reasonable, just a little bit weird to mixed in this PR 😆

quake (approved):

PR #4121, opened 2023-08-28
chore: add histogram metrics for relay compact block verify time

quake:

An efficient block verification process is important for reducing the network uncle rate, especially when the relay protocol receives the latest block,  which allows the miner to update the tip faster and thus work on the latest chain quickly. User may need this metric to monitor and optimize ckb node.

This PR add histogram metrics for relay protocol to record the compact block verification time.

### Check List 

Tests 

https://github.com/nervosnetwork/ckb/blob/19f5362a82da87441daa9ba002a3c4f404cff9de/util/metrics-config/src/lib.rs#L17-L16

### Release note 

```release-note
```

### DIFF
--- a/util/metrics/src/lib.rs
+++ b/util/metrics/src/lib.rs
@@ -7,8 +7,8 @@
 //! [`ckb-metrics-service`]: ../ckb_metrics_service/index.html
 
 use prometheus::{
-    register_histogram_vec, register_int_counter, register_int_gauge, register_int_gauge_vec,
-    HistogramVec, IntCounter, IntGauge, IntGaugeVec,
+    register_histogram, register_histogram_vec, register_int_counter, register_int_gauge,
+    register_int_gauge_vec, Histogram, HistogramVec, IntCounter, IntGauge, IntGaugeVec,
 };
 use prometheus_static_metric::make_static_metric;
 use std::cell::Cell;
@@ -48,6 +48,8 @@ pub struct Metrics {
     pub ckb_freezer_read: IntCounter,
     /// Counter for relay transaction short id collide
     pub ckb_relay_transaction_short_id_collide: IntCounter,
+    /// Histogram for relay compact block verify time
+    pub ckb_relay_cb_verify_time: Histogram,
     /// Counter for relay compact block transaction count
     pub ckb_relay_cb_transaction_count: IntCounter,
     /// Counter for relay compact block reconstruct ok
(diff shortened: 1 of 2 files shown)


eval-exec (approved):

util/metrics/src/lib.rs
```diff
@@ -79,6 +81,11 @@ static METRICS: once_cell::sync::Lazy<Metrics> = once_cell::sync::Lazy::new(|| M
         "The CKB relay transaction short id collide"
     )
     .unwrap(),
+    ckb_relay_cb_verify_time: register_histogram!(
+        "ckb_relay_cb_verify_time",
+        "The CKB relay compact block verify time"
+    )
```
eval-exec, line 87: I think it's neccesery to define buckets values like:
```
    ckb_relay_cb_verify_time: register_histogram!(
        "ckb_relay_cb_verify_time",
        "The CKB relay compact block verify time",
        vec![0.0005, 0.001, 0.0015, 0.002, 0.0025, 0.003, ...]
    )
```

util/metrics/src/lib.rs
```diff
@@ -48,6 +48,8 @@ pub struct Metrics {
     pub ckb_freezer_read: IntCounter,
     /// Counter for relay transaction short id collide
     pub ckb_relay_transaction_short_id_collide: IntCounter,
+    /// Histogram for relay compact block verify time
+    pub ckb_relay_cb_verify_time: Histogram,
```
eval-exec, line 52: How about `ckb_relay_cb_verify_timecost`?

quake: renamed to `ckb_relay_cb_verify_duration`

util/metrics/src/lib.rs
quake: the default buckets is:
```rust
/// The default [`Histogram`] buckets. The default buckets are
/// tailored to broadly measure the response time (in seconds) of a
/// network service. Most likely, however, you will be required to define
/// buckets customized to your use case.
pub const DEFAULT_BUCKETS: &[f64; 11] = &[
    0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
];
```

I think it's better than an equal division buckets in this case.

eval-exec (approved):

doitian (approved):

PR #4126, opened 2023-08-30
feat: notify dummy miner for new work

quake:

We often use the dummy miner mode for local development, but currently the dummy miner only submits new blocks at configured intervals, when other nodes broadcast the new block, the dummy miner will not switch to the latest tip but continue to submit the old block, so it will become an uncle block.

This behavior is fine in single-node mode, but when we need to simulate the hash rate distribution of multiple nodes for testing, there is no way to simulate this behavior with the dummy miner.

Tweak the dummy miner inner loop, if there is new tip (`pow_hash` changed), start working on the new one.

### Check List 

Tests 

```
# init two nodes
./ckb init -c dev --genesis-message dummy-miner-test  --ba-arg miner1-arg
./ckb init -c dev --genesis-message dummy-miner-test  --ba-arg miner2-arg

# add nodes to the whitelist section and config two miners' ckb-miner.toml
# for example, miner1 is 5 sec , miner2 is 9 sec
value = 5000
value = 9000

# start ckb nodes and miners and verify the block cellbase distribution by rpc `get_block_by_number` or check the log manually, it should be distributed as 5 / 14 =~ 35.7% and 9 / 14 =~ 64.3%
```

### Release note 

```release-note
```

### DIFF
--- a/miner/src/worker/dummy.rs
+++ b/miner/src/worker/dummy.rs
@@ -8,7 +8,7 @@ use indicatif::ProgressBar;
 use rand::thread_rng;
 use rand_distr::{self as dist, Distribution as _};
 use std::thread;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 pub struct Dummy {
     delay: Delay,
@@ -94,10 +94,29 @@ impl Dummy {
         }
     }
 
-    fn solve(&self, pow_hash: Byte32, work: Work, nonce: u128) {
-        thread::sleep(self.delay.duration());
-        if let Err(err) = self.nonce_tx.send((pow_hash, work, nonce)) {
-            error!("nonce_tx send error {:?}", err);
+    fn solve(&mut self, mut pow_hash: Byte32, mut work: Work, nonce: u128) {
+        let instant = Instant::now();
+        let delay = self.delay.duration();
+        loop {
+            thread::sleep(Duration::from_millis(10));
+            if instant.elapsed() > delay {
+                if let Err(err) = self.nonce_tx.send((pow_hash, work, nonce)) {
+                    error!("nonce_tx send error {:?}", err);
+                }
+                return;
+            }
+            // if there is new work and pow_hash changed, start working on the new one
+            if let Ok(WorkerMessage::NewWork {
+                pow_hash: new_pow_hash,
+                work: new_work,
+                ..
+            }) = self.worker_rx.try_recv()
+            {
+                if new_pow_hash != pow_hash {
+                    pow_hash = new_pow_hash;
+                    work = new_work;
+                }
+            }
         }
     }
 }


miner/src/worker/dummy.rs
```diff
@@ -94,10 +94,29 @@ impl Dummy {
...
+                pow_hash: new_pow_hash,
+                work: new_work,
+                ..
+            }) = self.worker_rx.try_recv()
+            {
+                if new_pow_hash != pow_hash {
+                    pow_hash = new_pow_hash;
+                    work = new_work;
+                }
+            }
```
eval-exec, line 119: How about this:
```rust
    fn solve(&self, pow_hash: Byte32, work: Work, nonce: u128) {
        select! {
            recv(self.worker_rx) -> msg => {
                if let Ok(WorkerMessage::NewWork {
                        pow_hash: new_pow_hash,
                        work: new_work,
                        ..
                }) = msg {
                    if new_pow_hash != pow_hash {
                        pow_hash = new_pow_hash;
                        work = new_work;
                    }
                }
            },
            default(Duration::from_millis(self.delay.duration())) => {
                if let Err(err) = self.nonce_tx.send((pow_hash, work, nonce)) {
                    error!("nonce_tx send error {:?}", err);
                }
                return;

            }
        }
    }
```

eval-exec (approved):

miner/src/worker/dummy.rs
eval-exec, line 119: How about this:
```rust
    fn solve(&self, pow_hash: Byte32, work: Work, nonce: u128) {
        select! {
            recv(self.worker_rx) -> msg => {
                if let Ok(WorkerMessage::NewWork {
                        pow_hash: new_pow_hash,
                        work: new_work,
                        ..
                }) = msg {
                    if new_pow_hash != pow_hash {
                        pow_hash = new_pow_hash;
                        work = new_work;
                    }
                }
            },
            default(self.delay.duration()) => {
                if let Err(err) = self.nonce_tx.send((pow_hash, work, nonce)) {
                    error!("nonce_tx send error {:?}", err);
                }
                return;

            }
        }
    }
```

quake: The current develop branch use a simple thread sleep, and I wanted to keep the code as unchanged as possible, rather than introduce a tokio runtime.

eval-exec: > rather than introduce a tokio runtime.

OK, but `self.nonce_tx`'s type is `crossbeam_channel::Sender`. We won't need tokio runtime here.

quake: I see, I thought it's tokio's select macro, let me change it and test later, thanks.

quake: after test, I found the code generated by `select` macro is different, the `pow_hash` and `work` vars are not read again, I will keep current code.

PR #4128, opened 2023-08-31
feat(rpc): Introduce the new method `generate_epochs` in the `IntegrationTest` module

EthanYuan:

Problem Summary:

In the context of CKB applications, many contracts are closely tied to `epoch` processes, including ["DAO"](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0023-dao-deposit-withdraw/0023-dao-deposit-withdraw.md) script, ["cheque"](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0039-cheque/0039-cheque.md) script, and others. During the development process, facilitating the acceleration of epochs becomes essential.

Before this PR, there were two methods: 
1. relied on miners rapidly producing blocks
2. required multiple calls to the CKB RPC `generate_block`

Clearly, the latter is more suitable for integration testing (without requiring a miner). However, using `generate_block` poses two issues. Firstly, it requires the calling side to calculate and determine the number of times `generate_block` should be called based on the desired epochs to fast-forward. Secondly, if called too quickly, `generate_block` may produce duplicate block hashes (indicating that no new blocks were generated), necessitating deduplication on the calling side.

The newly designed RPC `generate_epochs` introduces a parameter called `num_epochs`. Upon successful execution, it will return the current epoch target (in the form of `EpochNumberWithFraction`), significantly simplifying integration testing for applications like "DAO".

What's Changed:

add rpc:

```rust
#[rpc(name = "generate_epochs")]
fn generate_epochs(&self, num_epochs: EpochNumberWithFraction) -> Result<EpochNumberWithFraction>;
```
### Check List 

Tests 

- Unit test
  - [dao-related tests](https://github.com/EthanYuan/mercury/blob/dev-0.4-omni-0.108/integration/src/tests/build_dao_transaction.rs#L22) 

### Release note 

```release-note
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -3216,8 +3216,8 @@ Response
 
 
 #### Method `fast_forward_epochs`
-* `fast_forward_epochs(epoch_count)`
-    * `epoch_count`: [`Uint64`](#type-uint64)
+* `fast_forward_epochs(epochs_to_skip)`
+    * `epochs_to_skip`: [`Uint64`](#type-uint64)
 * result: [`Uint64`](#type-uint64)
 
 Fast-forwarding epochs during development, can be useful for scenarios
@@ -3228,7 +3228,7 @@ Returns the updated epoch number after fast forwarding.
 
 ###### Params
 
-*   `epoch_count` - The number of epochs to fast forward.
+*   `epochs_to_skip` - The number of epochs to fast forward.
 
 ###### Examples
 
@@ -3240,7 +3240,7 @@ Returns the updated epoch number after fast forwarding.
   "id": 42,
   "jsonrpc": "2.0",
   "method": "fast_forward_epochs",
-  "params": "0x4"
+  "params": ["0x6"]
 }
 ```
 
@@ -3252,7 +3252,7 @@ Returns the updated epoch number after fast forwarding.
 {
   "id": 42,
   "jsonrpc": "2.0",
-  "result": 0x8,
+  "result": "0x8",
   "error": null
 }
 ```
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -68,6 +68,7 @@ The crate `ckb-rpc`'s minimum supported rustc version is 1.67.1.
         * [Method `process_block_without_verify`](#method-process_block_without_verify)
         * [Method `truncate`](#method-truncate)
         * [Method `generate_block`](#method-generate_block)
+        * [Method `fast_forward_epochs`](#method-fast_forward_epochs)
         * [Method `notify_transaction`](#method-notify_transaction)
         * [Method `generate_block_with_template`](#method-generate_block_with_template)
         * [Method `calculate_dao_field`](#method-calculate_dao_field)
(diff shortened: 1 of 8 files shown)


eval-exec (approved):

rpc/src/module/test.rs
```diff
@@ -170,6 +170,58 @@ pub trait IntegrationTestRpc {
     #[rpc(name = "generate_block")]
     fn generate_block(&self) -> Result<H256>;
 
+    /// Generate epochs during development, can be useful for scenarios
+    ///
```
doitian, line 174:
```suggestion
```

rpc/src/tests/module/test.rs
```diff
@@ -0,0 +1,112 @@
...
+    consensus.genesis_epoch_ext = epoch_ext;
+    consensus.epoch_duration_target = 240;
+    consensus.permanent_difficulty_in_dummy = true;
+
+    setup(consensus)
+}
+
+fn get_current_epoch(suite: &RpcTestSuite) -> EpochNumberWithFraction {
+    let store = suite.shared.store();
+    let tip_block_number = store.get_tip_header().unwrap().data().raw().number();
```
eval-exec, line 107: You can also retrieve the `tip_block_number` by `store.get_tip_header().unwrap().number()`.

EthanYuan: Yes, the latter is better.

rpc/src/module/test.rs
EthanYuan: fixed

rpc/src/tests/module/test.rs
```diff
@@ -0,0 +1,112 @@
...
+        get_current_epoch(&suite),
+        EpochNumberWithFraction::new(3, 5, GENESIS_EPOCH_LENGTH)
+    );
+
+    // generate 3/2 epoch
+    suite.rpc(&RpcTestRequest {
+        id: 42,
+        jsonrpc: "2.0".to_string(),
+        method: "generate_epochs".to_string(),
+        params: vec!["0x20003000000".into()],
```
eval-exec, line 64: How is the hardcoded string `0x20003000000` calculated?

EthanYuan:
```rust
assert_eq!(
    Into::<ckb_jsonrpc_types::Uint64>::into(EpochNumberWithFraction::new(0, 3, 2)).to_string(),
    "0x20003000000".to_string()
);
```

I'll try to add it in the comments, maybe other users will be confused as well.

rpc/src/module/test.rs
```diff
@@ -522,6 +578,33 @@ impl IntegrationTestRpc for IntegrationTestRpcImpl {
         self.process_and_announce_block(block_template.into())
     }
 
+    fn generate_epochs(
+        &self,
+        num_epochs: EpochNumberWithFraction,
+    ) -> Result<EpochNumberWithFraction> {
```
15168316096, line 584: If the request growth epoch is too large, the waiting time will be too long. The more logs generated, the higher the CPU load. Let’s see if we can optimize the CPU usage through asynchronous means.
```
ckb@test-02:~/xueyl/ckb/ckb/generate_epochs/specs$ curl --location 'http://127.0.0.1:8202/' --header 'Content-Type: application/json' --data '{
  "id": 42,
  "jsonrpc": "2.0",
  "method": "generate_epochs",
  "params": ["0xFFFFFFFFFFFFFFFF"]
}'
```
PID USER PR NI VIRT RES SHR S %CPU %MEM TIME+ COMMAND
2624591 ckb 20 0 1811504 782352 60596 S 360.8 2.4 47:56.19 ckb

rpc/README.md
```diff
@@ -3214,6 +3215,64 @@ Response
...
+
+Response
+
+
+```
+{
+  "id": 42,
+  "jsonrpc": "2.0",
+  "result": "0xa0001000003",
+  "error": null
```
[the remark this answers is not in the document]
EthanYuan: It seems that similar issues exist in other RPC method examples, and we can consider removing them together.

PR #4133, opened 2023-09-01
fix: keep peer store address unique

driftluo:

When the same address enters the peer store, we need to reset the random id and ensure that the address is unique

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/peer_store/addr_manager.rs
+++ b/network/src/peer_store/addr_manager.rs
@@ -18,14 +18,17 @@ impl AddrManager {
     /// Add an address information to address manager
     pub fn add(&mut self, mut addr_info: AddrInfo) {
         if let Some(key) = multiaddr_to_socketaddr(&addr_info.addr) {
-            if let Some(exists_last_connected_at_ms) = self
-                .get(&addr_info.addr)
-                .map(|addr| addr.last_connected_at_ms)
-            {
+            if let Some(&id) = self.addr_to_id.get(&key) {
+                let (exist_last_connected_at_ms, random_id_pos) = {
+                    let info = self.id_to_info.get(&id).expect("must exists");
+                    (info.last_connected_at_ms, info.random_id_pos)
+                };
                 // Get time earlier than record time, return directly
-                if addr_info.last_connected_at_ms < exists_last_connected_at_ms {
-                    return;
+                if addr_info.last_connected_at_ms >= exist_last_connected_at_ms {
+                    addr_info.random_id_pos = random_id_pos;
+                    self.id_to_info.insert(id, addr_info);
                 }
+                return;
             }
 
             let id = self.next_id;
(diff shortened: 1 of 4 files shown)


network/src/peer_store/addr_manager.rs
```diff
@@ -29,7 +29,12 @@ impl AddrManager {
             }
 
             let id = self.next_id;
-            self.addr_to_id.insert(key, id);
+            let removed = self.addr_to_id.insert(key, id);
+            if let Some(k) = removed {
```
eval-exec, line 33: How about `if let Some(removed) = self.addr_to_id.insert(key, id)`

eval-exec (approved):

network/src/peer_store/addr_manager.rs
```diff
@@ -18,14 +18,18 @@ impl AddrManager {
...
+                    (info.last_connected_at_ms, info.random_id_pos)
+                };
                 // Get time earlier than record time, return directly
-                if addr_info.last_connected_at_ms < exists_last_connected_at_ms {
+                if addr_info.last_connected_at_ms < exist_last_connected_at_ms {
                     return;
                 }
+                self.swap_random_id(random_id_pos, self.random_ids.len() - 1);
+                self.random_ids.pop();
+                self.id_to_info.remove(&id);
```
quake, line 32: I don't know why we need to remove the old id and re-insert it, can we replace the `id_to_info` entry with same id? I think it's simple and easy to read.
```rust
            if let Some(&id) = self.addr_to_id.get(&key) {
                let exist_last_connected_at_ms = self.id_to_info.get(&id).expect("must exists").last_connected_at_ms;
                if addr_info.last_connected_at_ms > exist_last_connected_at_ms {
                    self.id_to_info.insert(id, addr_info);
                }
                return;
            }
```

quake (changes requested):

network/src/peer_store/addr_manager.rs
driftluo: https://github.com/nervosnetwork/ckb/pull/4133/files#diff-d2a5fc6b3c7ac53b73e2b01fa86a9fa42bf03efbfb7df0d7c78dea00bb37d482L51 

We reset the update id just to implement the requirements of this algorithm, I'm not sure what will happen if we don't reset, it may need to ask @jjyr

quake: The logic here is that when a duplicate key exists, if the `last_connected_at_ms` satisfies the condition, then replace it with the new value, and this logic has nothing to do with generating randomized positions.

driftluo: In the original logic, the random id corresponding to addr_info has been updated. This random id is something related to the random method. If we don't care about this update, we can directly replace addr info

jjyr: It seems has no side effect, the code delete the old addr and re-insert it, I think we can simply replace the entry.

quake (approved):

PR #4191, opened 2023-10-13
performance: tuning rocksdb bloom filter

quake:

We previously used a global lru cache as rocksdb cache (default 128M) and setting the `cache_index_and_filter_blocks` parameter to true, the rocksdb's `IndexBlock` is placed in the cache also (note, we did not use `FilterBlock` in current code). With the increasing of stored data, `DataBlock` and `IndexBlock` will compete for cache expiration, there will be more cache misses, resulting in read performance degradation. 

rocksdb statistics log
```
Block cache LRUCache@0x7f41c443d3d0#3747917 capacity: 128.00 MB seed: 886014176 usage: 117.24 MB table_size: 32768 occupancy: 5795 collections: 155 last_copies: 19 last_secs: 0.001642 secs_since: 0
Block cache entry stats(count,size,portion): DataBlock(5543,23.49 MB,18.3507%) IndexBlock(251,93.22 MB,72.8312%) Misc(1,0.00 KB,0%)
```

This PR tries to adjust the cache configuration, and also uses a bloom filter to improve the performance of prefix seek.

through the mainnet synchronization test, the overall performance has been improved by about 6%.

QQ20231013-110427@2x

and memory usage has also decreased
QQ20231013-110228@2x

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/db/src/db.rs
+++ b/db/src/db.rs
@@ -72,14 +72,16 @@ impl RocksDB {
         for cf in cf_descriptors.iter_mut() {
             let mut block_opts = BlockBasedOptions::default();
             block_opts.set_ribbon_filter(10.0);
-            block_opts.set_cache_index_and_filter_blocks(true);
-            block_opts.set_pin_l0_filter_and_index_blocks_in_cache(true);
             block_opts.set_index_type(BlockBasedIndexType::TwoLevelIndexSearch);
             block_opts.set_partition_filters(true);
             block_opts.set_metadata_block_size(4096);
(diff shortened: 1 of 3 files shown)


zhangsoledad:
https://github.com/nervosnetwork/rust-rocksdb/pull/52

eval-exec (approved):

zhangsoledad (approved):

PR #4199, opened 2023-10-20
Fix orphan pool for long pending tx issues

chenyukang:

Project Godwoken found that some `tx` will stay in `pending` for a long time. After analyzing the log(at most 4 hours), we found `tx` may unexpectedly stay in `orphan pool` for the scenario transaction chain. For example, `tx1 -> tx2`, after `tx1` is submitted `tx2` stayed in `pending` for a long time, actually `tx2` wasn't submitted into `txpool`.

Problem Summary:

Fix the issues in `process_orphan_tx` which may forget some txs in `orphan pool`. There are 3 issues this PR targeted to resolve:

- 1. `find_by_previous` was only returning one `tx` by searching with input(because the `tx` in orphan pool may have conflicts), which may cause issue when there are multiple child `tx` for one parent `tx`, `process_orphan_tx` may can not process all of them.

- 2. The evict time interval was `2 * max_block_interval`, which is `96` seconds, this is too short for the delay of networking, and we didn't send `Reject` to filter, which may make the node don't broadcast them later.

- 3. For the conflicted `tx` scenario, since we can not resolve all inputs in `orphan pool`, we can not add `Replace-by-fee` checking rules for them, so I change the code to use `HashMap<OutPoint, HashSet<ProposalShortId>>`, so we don't handle the conflict in `orphan pool`.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


doitian:
Please choose the release note for the PR.

tx-pool/src/component/orphan.rs
```diff
@@ -122,6 +123,15 @@ impl OrphanPool {
         if self.entries.contains_key(&tx.proposal_short_id()) {
             return;
         }
+
+        // double spend checking
+        if tx
```
driftluo, line 128: Do we need to be concerned about double-spending issues in the orphan pool?

chenyukang: I discussed with @zhangsoledad , he mentioned one reason that `find_orphan_by_previous` return only one `OrphanEntry` is `orphan pool` may contain conflicted `tx` and suggest add double-spending check.

After txpool refactor, we have double-spending check when a `tx` is added to `Pending`, I think it is trivial we add it here. One result I can think of is that conflicted `tx` may broadcasted more.

Do you have more concerns about this?

tx-pool/src/component/orphan.rs
```diff
@@ -1,14 +1,16 @@
...
     core::{Cycle, TransactionView},
     packed::{OutPoint, ProposalShortId},
 };
 use ckb_util::shrink_to_fit;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 const SHRINK_THRESHOLD: usize = 100;
-pub(crate) const ORPHAN_TX_EXPIRE_TIME: u64 = 2 * 48; // double block interval
+pub(crate) const ORPHAN_TX_EXPIRE_TIME: u64 = 100 * MAX_BLOCK_INTERVAL; // double block interval
```
driftluo, line 13: comment should change

tx-pool/src/component/orphan.rs
```diff
@@ -98,46 +105,60 @@ impl OrphanPool {
...
+                .or_insert_with(HashSet::default)
+                .insert(tx.proposal_short_id());
         }
-        self.limit_size();
+
+        self.limit_size()
     }
 
-    pub fn find_by_previous(&self, tx: &TransactionView) -> Option<ProposalShortId> {
+    pub fn find_by_previous(&self, tx: &TransactionView) -> Vec<ProposalShortId> {
```
quake, line 157: I think we may return `Vec<&ProposalShortId>` since it's only used to find orphan tx by id internally, it will remove the `clone` operation in this fn.

tx-pool/src/component/orphan.rs
```diff
@@ -1,14 +1,18 @@
...
 };
 use ckb_util::shrink_to_fit;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 const SHRINK_THRESHOLD: usize = 100;
-pub(crate) const ORPHAN_TX_EXPIRE_TIME: u64 = 2 * 48; // double block interval
+
+/// 100 max block interval
+pub(crate) const ORPHAN_TX_EXPIRE_TIME: u64 = 100 * MAX_BLOCK_INTERVAL;
```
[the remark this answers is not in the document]
chenyukang: one issue with increasing it to 100 block interval is we may store more orphan tx in the pool, but we still have 100 max orphan numbers limit, so it should be ok.

on the other hand, the previous 96 seconds is too short in practice, we have observed this kind of case, so let's keep this change.

quake (approved):

driftluo (approved):

PR #4203, opened 2023-10-25
 persistence softfork cache

zhangsoledad:

In [RFC0043](https://github.com/nervosnetwork/rfcs/blob/master/rfcs/0043-ckb-softfork-activation/0043-ckb-softfork-activation.md) we introduced the signal mechanism to implement softfork-activation, for each softfork deployment its state needs to be accumulated statistics, the current implementation is to use a memory cache to store the state, resulting in each reboot, the state of the deployment needs to be regenerated. when in the case of large amounts of data, the reboot will be lagged for some time.

Introduces a persistent mmap cache so that the state doesn't need to be recalculated each time it is restarted

The persistence path is stored in DATA_DIR/softfork/\<deployment\>/\<cache file\> format.

image

after

image

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


util/types/src/core/mod.rs
```diff
@@ -27,11 +27,14 @@ mod fee_rate;
 mod reward;
 mod transaction_meta;
 mod views;
+
 pub use advanced_builders::{BlockBuilder, HeaderBuilder, TransactionBuilder};
 pub use blockchain::DepType;
 pub use extras::{BlockExt, EpochExt, EpochNumberWithFraction, TransactionInfo};
 pub use fee_rate::FeeRate;
+use once_cell::sync::OnceCell;
```
eval-exec: `OnceCell` have merged into rust's `std` library: https://doc.rust-lang.org/std/cell/struct.OnceCell.html

eval-exec (approved):

util/types/src/core/mod.rs
zhangsoledad: The once_cell in std is still a bit different from the crate we're using now, and the equivalent would be https://doc.rust-lang.org/std/sync/struct.OnceLock.html, which we can decide to replace later.

Issue #4181, opened 2023-10-09
Release v0.112.0


doitian:
RC Date: 2023-10-16
Release Date: 2023-10-30

### Features

-   #4128 **rpc:** Introduce the new method `generate_epochs` in the `IntegrationTest` module (@EthanYuan)
-   #4079: Tx pool Replace-by-fee (@chenyukang)
-   #4108: Add feature of Replace-by-fee for tx-pool (@chenyukang)
-   #4232: RBF will also replace tx not in Pending (@chenyukang)

### Bug Fixes

-   #4172: Fix `ckb-hash` depends `blake2b-ref` multiple times (@eval-exec)
-   #4171 **light-client:** Could not check MMR for fork chains (@yangby-cryptape)
-   #4182: Should only return main chain data (@quake)
-   #4208: Fix orphan pool issue for long pending tx (@chenyukang)

### Improvements

-   #3993: Tx pool rewrite with multi_index_map (@chenyukang)
-   #4146: Upgrade CKB's `rust-toolchain` from `1.67.1` to `1.71.1` (@eval-exec)
-   #4160: Remove useless `--cfg disable_faketime` from `RUSTFLAGS` (@eval-exec)
-   #4158: Remove useless `non_owning_clone` method for `ChainController` and `NetworkController` (@eval-exec)
-   #4161: Remove `#[derive(Clone)]` from `Synchronizer` (@eval-exec)
-   #4113 **(molecule):** Remove deprecated `SyncMessage` union items. (@eval-exec)
-   #4186: Remove `Consensus.dao_type_hash`'s `Option` wrapper (@eval-exec)

doitian:
### Bug Fixes

-   #4182: Should only return main chain data (@quake)

### Improvements

-   #4186: Remove `Consensus.dao_type_hash`'s `Option` wrapper (@eval-exec)

doitian:
New changes in RC2

-   #4208: Fix orphan pool issue for long pending tx (@chenyukang)

chenyukang:
## Breakchange notes

We introduce a huge code refactor on the `transaction pool` module and added a the feature of RBF(Replace by Fee) in this release. 

Here are two points need to be noted:
1. Double-spending check now will work before any transaction is added into the transaction pool, a transaction that conflicts with current transactions in the pool will be rejected, the only way to add the new transaction into the pool is by improving the fee and if RBF checking passed, transaction pool will remove old conflicted transactions, then add the new one into pool.

[...]

PR #4208, opened 2023-10-27
Backport: Fix orphan pool issue for long pending tx

chenyukang:

Project Godwoken found that some `tx` will stay in `pending` for a long time. After analyzing the log(at most 4 hours), we found `tx` may unexpectedly stay in `orphan pool` for the scenario transaction chain. For example, `tx1 -> tx2`, after `tx1` is submitted `tx2` stayed in `pending` for a long time, actually `tx2` wasn't submitted into `txpool`.

Problem Summary:

Fix the issues in `process_orphan_tx` which may forget some txs in `orphan pool`. There are 3 issues this PR targeted to resolve:

- 1. `find_by_previous` was only returning one `tx` by searching with input(because the `tx` in orphan pool may have conflicts), which may cause issue when there are multiple child `tx` for one parent `tx`, `process_orphan_tx` may can not process all of them.

- 2. The evict time interval was `2 * max_block_interval`, which is `96` seconds, this is too short for the delay of networking, and we didn't send `Reject` to filter, which may make the node don't broadcast them later.

- 3. For the conflicted `tx` scenario, since we can not resolve all inputs in `orphan pool`, we can not add `Replace-by-fee` checking rules for them, so I change the code to use `HashMap<OutPoint, HashSet<ProposalShortId>>`, so we don't handle the conflict in `orphan pool`.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


eval-exec (approved):

doitian:
Release RC2 for #4181

doitian (approved):

PR #4210, opened 2023-10-30
Add doc for `get_transactions` and `get_cells` RPC

eval-exec:
Add rpc doc for `get_transactions` and `get_cells` to clarify their pagination style.
When the number of returned objects is less than the requested `limit`, it indicates that it's the last page.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/module/indexer.rs
+++ b/rpc/src/module/indexer.rs
@@ -63,6 +63,7 @@ pub trait IndexerRpc {
     ///
     /// ## Returns
     ///
+    /// > If the number of objects is less than the requested `limit`, it indicates that these are the last page of get_cells.
     /// * objects:
     ///     - output: the fields of an output cell
     ///     - output_data: the cell data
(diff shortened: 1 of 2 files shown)


rpc/src/module/indexer.rs
```diff
@@ -63,6 +63,7 @@ pub trait IndexerRpc {
     ///
     /// ## Returns
     ///
+    /// > If the number of objects is less than the requested `limit`, it indicates that these are the last page of get_cells.
```
doitian, line 66: Why quote (`> `) here? If the text messes up with the following list, try to separate them with a blank line.

```suggestion
    /// If the number of objects is less than the requested `limit`, it indicates that these are the last page of get_cells.
    ///
```

zhangsoledad (approved):

PR #4218, opened 2023-11-02
fix: fix vm version select

driftluo:

1. During synchronization, 2019 version transactions should use the v0 version of the VM
2. rpc should faithfully represent the epoch in which the 2021 hardfork is located(just VM relations)
3. ~cycles due to wrong version vm will be reset to None~(migration will on another pr)

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


script/src/verify/tests/utils.rs
```diff
@@ -135,10 +136,15 @@ impl TransactionScriptsVerifierWithEnv {
         let tmp_dir = TempDir::new().unwrap();
         let db = RocksDB::open_in(&tmp_dir, COLUMNS);
         let store = Arc::new(ChainDB::new(db, Default::default()));
+        let version_1_enabled_at = 10;
```
eval-exec, line 139: Should we assign different values for `version_1_enabled_at` and `version_2_enabled_at`?

eval-exec, line 139: Should we assign different values for `version_1_enabled_at` and `version_2_enabled_at`?

script/src/verify/tests/ckb_latest/features_since_v2019.rs
```diff
@@ -193,6 +193,10 @@ fn check_signature_referenced_via_type_hash() {
 #[test]
 fn check_signature_referenced_via_type_hash_ok_with_multiple_matches() {
     let script_version = SCRIPT_VERSION;
+    if script_version < ScriptVersion::V1 {
+        // This transaction is not supported in the 2019 version
```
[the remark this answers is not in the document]
driftluo: It cannot be done. The transaction is restricted by rfc_0029, not vm version. In the current state, the transaction can be verified by vm.
What we are currently repairing is the vm version, which does not include all judgment conditions that are buryed.

util/launcher/src/migrate.rs
```diff
@@ -29,6 +30,7 @@ impl Migrate {
         migrations.add_migration(Box::new(migrations::AddChainRootMMR)); // TODO(light-client) update the comment: which version?
         migrations.add_migration(Box::new(migrations::AddBlockFilterColumnFamily)); // since v0.105.0
         migrations.add_migration(Box::new(migrations::AddBlockFilterHash)); // since v0.108.0
+        migrations.add_migration(Box::new(migrations::BlockExt2019ToZero::new(hardforks))); // since v0.111.1
```
eval-exec, line 33: Should we add a "Release Note" about the affect about `BlockExt2019ToZero` migration?

driftluo: This will cause ckb to need to manually perform magration, and the release note should have prompts.

rpc/README.md
```diff
@@ -1659,12 +1659,12 @@ Response
...
-            { "rfc": "0031", "epoch_number": "0x0" },
-            { "rfc": "0032", "epoch_number": "0x0" },
-            { "rfc": "0036", "epoch_number": "0x0" },
-            { "rfc": "0038", "epoch_number": "0x0" },
+            { "rfc": "0029", "epoch_number": "0x1526" },
+            { "rfc": "0030", "epoch_number": "0x1526" },
+            { "rfc": "0031", "epoch_number": "0x1526" },
+            { "rfc": "0032", "epoch_number": "0x1526" },
+            { "rfc": "0036", "epoch_number": "0x1526" },
+            { "rfc": "0038", "epoch_number": "0x1526" },
```
zhangsoledad, line 1667: Here as I understand it, it should just restore vm version selection related, not all.

zhangsoledad (changes requested):
I don't really support setting cycles to 0, 0 is too vague, we would have used Option<Cycles>, if not it's None, if not None , Cycles should be an exact value.  why 0？

script/src/verify.rs
```diff
@@ -566,8 +579,10 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
             ScriptHashType::Type => {
```
zhangsoledad, line 579: Do we need to restore the on condition of data1, which was not enabled before 2021 was enable?

zhangsoledad (approved):

PR #4226, opened 2023-11-09
Add migration run in background mode

### LINKED ISSUE #3683: Guide users to re-sync only when migration fails.
https://github.com/Magickbase/neuron-public-issues/issues/71#issuecomment-1297020609

The message is misleading. User may think re-sync is faster than migration.

Remove the following line from the prompt.

```
Another choice is to delete all data, then synchronize them again.
```

Instead, tell users that if the migration fails, try to delete all data and sync again.

chenyukang:

Fix the issue that `db migration` may cost long time.
Fix #3683 by the way.

Problem Summary:

- Add run in background option and strategy for future migrations
- Add migration from #4218 as the first migration run in background

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 
```release-note
```


util/launcher/src/migrations/set_2019_block_cycle_zero.rs
```diff
@@ -0,0 +1,95 @@
...
+        let chain_db = ChainDB::new(db, StoreConfig::default());
+        let limit_epoch = self.hardforks.ckb2021.rfc_0032();
+
+        eprintln!("begin to run block_ext 2019 to zero migrate...");
+        if limit_epoch == 0 {
+            return Ok(chain_db.into_inner());
+        }
+
+        let epoch_number: packed::Uint64 = limit_epoch.pack();
+        if let Some(epoch_hash) = chain_db.get(COLUMN_EPOCH, epoch_number.as_slice()) {
```
driftluo, line 41: There are some special cases not covered here：

If the current node synchronization has not exceeded hardfork epoch, this judgment is problematic.

So getting the first header here still requires some operations.
1. Get the current tip header
2. Compare the epoch of the tip header with the epoch of the hardfork switch, and then determine whether the first header is the tip header or the header corresponding to the previous hash of the hardfork switch.

eval-exec (approved):

db-migration/src/lib.rs
```diff
@@ -1,13 +1,24 @@
...
+use std::collections::VecDeque;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
+use std::sync::Mutex;
+use std::thread;
+use std::thread::JoinHandle;
+
+/// Shutdown flag for background migration.
+pub static SHUTDOWN_BACKGROUND_MIGRATION: once_cell::sync::Lazy<AtomicBool> =
+    once_cell::sync::Lazy::new(|| AtomicBool::new(false));
```
eval-exec, line 21: How about [OnceCell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) in the rust std library?

db-migration/Cargo.toml
```diff
@@ -15,6 +15,9 @@ ckb-db = { path = "../db", version = "= 0.113.0-pre" }
 ckb-logger = { path = "../util/logger", version = "= 0.113.0-pre" }
 ckb-error = { path = "../error", version = "= 0.113.0-pre" }
 ckb-db-schema = { path = "../db-schema", version = "= 0.113.0-pre" }
+ckb-channel = { path = "../util/channel", version = "= 0.113.0-pre" }
+ckb-stop-handler = { path = "../util/stop-handler", version = "= 0.113.0-pre" }
+once_cell = "1.8.0"
```
eval-exec, line 20: If we can use [std::cell::OnceCell](https://doc.rust-lang.org/std/cell/struct.OnceCell.html), then we can remove `once_cell`

db-migration/src/lib.rs
chenyukang: it's not sync, and `once_cell` is used a lot in our project.
A cleaner option is using `once_cell::sync::OnceCell` since we only want to check whether the flag is set.

db-migration/src/lib.rs
```diff
@@ -139,6 +231,37 @@ impl Migrations {
...
+        let tx_clone = tx.clone();
+        thread::spawn(move || {
+            let _ = exit_signal.recv();
+            let res = SHUTDOWN_BACKGROUND_MIGRATION.set(true);
+            let _ = tx_clone.send(Command::Stop);
+            eprintln!("set shutdown flag to true: {:?} version: {}", res, clone);
+        });
+
+        let handler = worker.start();
+        tx.send(Command::Start).expect("send start command");
```
eval-exec, line 258: Do we realy need `Command::Start` variant?

db-migration/src/lib.rs
```diff
@@ -19,7 +31,65 @@ fn internal_error(reason: String) -> Error {
...
+}
+
+impl MigrationWorker {
+    pub fn new(tasks: Arc<Mutex<MigrationTasks>>, db: RocksDB, inbox: Receiver<Command>) -> Self {
+        Self { tasks, db, inbox }
+    }
+
+    pub fn start(self) -> JoinHandle<()> {
+        thread::spawn(move || {
+            if let Ok(Command::Start) = self.inbox.recv() {
```
eval-exec, line 58: I think this line can be removed.

Issue #4237, opened 2023-11-19
Easy daemon mode for ckb run


CipherWang:
#### Is your feature request related to a problem? Please describe.
There is no simple approach to run a ckb node in daemon or background mode. Although there is an introduction to [start a ckb service](https://github.com/nervosnetwork/ckb/tree/develop/devtools/init), it's still too complex for average users.

#### Describe the solution you'd like
Similar to bitcoind, if ckb run provide a simple switch like `ckb run --daemon`, I believe more nodes will appear.

eval-exec:
I read [bitcoind daemon mode](https://bitcoin.stackexchange.com/questions/74116/bitcoind-daemon-mode), learned that: `bitcoind --daemon` just run `bitcoind` in background, it won't auto start after system reboot.

However, using `systemd` can make the process restarts automatically after a reboot

janx:
But assume the existence of `systemd` reduce daemon mode's compatibility? e.g. does it work on Mac?

eval-exec:
MacOS doesn't have `systemd`, the quivalent of systmed on MacOS is called [launchd](https://apple.stackexchange.com/questions/455362/equivalent-to-systemd-timers-on-os-x)

PR #4238, opened 2023-11-19
Fix using `journalctl` to check the log of `ckb.service`

eval-exec:

In the documentation for starting the ckb service via systemd on [here](https://github.com/nervosnetwork/ckb/blob/develop/devtools/init/linux-systemd/README.md), it's noted that the `ckb.service` is handled by the root's systemd. Therefore, to view its logs, you'd need to use `sudo journalctl`.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/devtools/init/linux-systemd/README.md
+++ b/devtools/init/linux-systemd/README.md
@@ -61,7 +61,7 @@ sudo systemctl enable ckb.service
 If ckb doesn't seem to start properly you can view the logs to figure out the problem:
 
 ```bash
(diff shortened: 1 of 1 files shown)


eval-exec:
The guidance on handling the ckb process via `systemd` could be made more user-friendly for non-programmers. 

Ref: #4237

doitian (approved):

quake (approved):

PR #4224, opened 2023-11-09
feat: add rich-indexer which is another built-in indexer based on relational database

EthanYuan:
### What problem does this PR solve?

Problem Summary:

CKB already has a built-in indexer, which is based on a key-value database for indexing. Rich-Indexer is a complement to the original indexer, as it is based on a relational database, providing support for more complex and flexible queries.

Currently, Rich-Indexer supports two types of database drivers. 

1. One is the embedded SQLite, which users can start using without any configuration.
2. The other is PostgreSQL, which can undergo independent, customized configurations for both software and hardware. Additionally, it allows users to engage in secondary development based on it.

### What is changed and how it works?

What's Changed:

Once the Rich-Indexer is activated, the CKB node's RPC based on the Rich-Indexer will gain additional capabilities.

| INDEXER RPC          | Indexer           | Rich-Indexer       |
|---------------------|-------------------|--------------------|
| `get_cells` script args `partial` mode search     | ❌     | ✔️      |
| `get_cells` cell data filter(`prefix\|exact\|partial`)     | ✔️     | ✔️      |
| `get_transactions` script args `partial` mode search | ❌          | ✔️   |
| `get_transactions` cell data filter(`prefix\|exact\|partial`)     | ❌         |  ✔️      |
| `get_cells_capacity` script args `partial` mode search        |    ❌    |    ✔️     |
| `get_cells_capacity` cell data filter(`prefix\|exact\|partial`)        |    ✔️    |   ✔️     |

In addition, Rich-Indexer can also expose full block data in the form of a PostgreSQL database, providing the most basic and flexible indexing format accessible via SQL and compatible with any programming language.

Please refer to the [README](https://github.com/EthanYuan/ckb/tree/indexer-r/util/rich-indexer#readme) for more detailed information

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- Performance regression

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4244, opened 2023-11-22
feat: introduce `init tip` setting for indexer to skip previous block synchronization

EthanYuan:

Problem Summary:

CKB-indexer supports building indexes only for interested blocks. This can be done by setting `block_filter` before this PR. But the filter can only traverse the block from 0 and for the uninterested block also need to save its `Header(BlockNumber, &'a Byte32, bool)`. 

 This pr introduces the `init tip` setting, which can be set directly when the indexer initializes the database for the first time, which brings two benefits:

1. synchronization can be started instantly from any height
2. the previous block does not need to save any data

What's Changed:

- Added commented out configuration item to ckb.toml template (as an example)
- Implement `apply_init_tip` function, which is called in `IndexerService::new`
- Necessary unit tests

### Related changes

This feature will also be used for index-r: 

- #4224  

### Check List 

Tests 

- Unit test
- Manual test
  - general: the indexer can skip the synchronization of the block before the init tip, and the init tip block only saves the block `Header(BlockNumber, &'a Byte32, bool)`
  - tip hash doesn't match, it rolls back to empty
  - when the init tip is set greater than the current node tip, there are no side effects, it waits for the node to synchronize

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -205,5 +205,5 @@ block_uncles_cache_size    = 30
 # cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
 # # The initial tip number and hash can be set as the starting height for building indexes in indexer-r. Effective only during the initial index creation.
 # init_tip_number = 10000
-# # The initial tip hash must match the tip number; otherwise, it will result in a rollback.
+# # The initial tip hash must match the tip number; otherwise, it will result in a rollback to empty.
 # init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -199,3 +199,11 @@ block_uncles_cache_size    = 30
 # [indexer_v2]
 # # Indexing the pending txs in the ckb tx-pool
 # index_tx_pool = false
+# # Customize block filtering rules to index only retained blocks
+# block_filter = "block.header.number.to_uint() >= \"0x0\".to_uint()"
+# # Customize cell filtering rules to index only retained cells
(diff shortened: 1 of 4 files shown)


resource/ckb.toml
```diff
@@ -199,3 +199,11 @@ block_uncles_cache_size    = 30
...
 # # Indexing the pending txs in the ckb tx-pool
 # index_tx_pool = false
+# # Customize block filtering rules to index only retained blocks
+# block_filter = "block.header.number.to_uint() >= \"0x0\".to_uint()"
+# # Customize cell filtering rules to index only retained cells
+# cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
+# # The initial tip number and hash can be set as the starting height for building indexes in indexer-r. Effective only during the initial index creation.
+# init_tip_number = 10000
+# # The initial tip hash must match the tip number; otherwise, it will result in a rollback to empty.
+# init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
```
quake, line 209: suggest to use one block hash parameter only, indexer may use `get_header` rpc to fetch the block number, and start the indexing service only when corresponding header was found.

EthanYuan: Entering a hash manually has the added feature that the block height of the node itself may not have been reached yet and `get_header` may fail.

EthanYuan: Entering a number manually considering a situation where the block height of the node itself may not have been reached and get_header may fail.

quake: please refer to my previous comment:
> start the indexing service only when corresponding header was found.

EthanYuan: Another issue is that if the hash is wrong, then there is no chance of rollback.

quake: indexer service won't starting if the hash is wrong.

EthanYuan: OK, got it.

PR #4247, opened 2023-11-27
feat: introduce `init_tip_hash` setting for indexer to skip previous block synchronization

EthanYuan:

Problem Summary:

CKB-indexer supports building indexes only for interested blocks. This can be done by setting `block_filter` before this PR. But the filter can only traverse the block from 0 and for the uninterested block also need to save its `Header(BlockNumber, &'a Byte32, bool)`. 

 This pr introduces the `init_tip_hash` setting. The init tip is effective as long as it is set higher than the current height of the indexing head (reboot required), as the starting height of the indexing. This provides the following benefits:

1. synchronization can be started instantly from any height block (by block hash)
2. the previous block does not need to save any data

What's Changed:

- Added commented out configuration item to ckb.toml template (as an example)
- Implement `apply_init_tip` function, which is called in `IndexerService::spawn_poll`
- The `reset-data --indexer` option is supported.

### Related changes

This feature will also be used for index-r: 

- #4224  

### Check List 

Tests 

- Manual test
  - general: the indexer can skip the synchronization of the block before the init tip, and the init tip block only saves the block `Header(BlockNumber, &'a Byte32, bool)`
  - the init tip is effective as long as it is set higher than the current height of the indexing head (reboot required)
  - when the header corresponding to the `init_tip_hash` is not found, it will attempt to continue searching, and the indexing service of the indexer will not start

### Release note 

```release-note
```

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -199,3 +199,7 @@ block_uncles_cache_size    = 30
 # [indexer_v2]
 # # Indexing the pending txs in the ckb tx-pool
 # index_tx_pool = false
+# # Customize block filtering rules to index only retained blocks
+# block_filter = "block.header.number.to_uint() >= \"0x0\".to_uint()"
+# # Customize cell filtering rules to index only retained cells
+# cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -203,3 +203,5 @@ block_uncles_cache_size    = 30
 # block_filter = "block.header.number.to_uint() >= \"0x0\".to_uint()"
 # # Customize cell filtering rules to index only retained cells
 # cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
+# # The initial tip can be set at the time of the indexer's first initialization as the starting height for index construction.
+# init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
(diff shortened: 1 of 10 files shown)


rpc/src/service_builder.rs
```diff
@@ -201,14 +202,20 @@ impl<'a> ServiceBuilder<'a> {
...
+                let rpc_methods = IndexerRpcImpl::new(indexer_handle).to_delegate();
+                if self.config.indexer_enable() {
+                    start_indexer(&shared, indexer, indexer_config.index_tx_pool);
+                    self.add_methods(rpc_methods);
+                } else {
+                    self.update_disabled_methods("Indexer", rpc_methods);
+                }
+            }
+            Err(e) => {
+                warn!("Failed to enable indexer: {}", e);
```
eval-exec, line 217: I believe the "error!" function should be used here. The CKB process is expected to terminate if the initialization of the indexer fails.

EthanYuan: In the case where the ckb node itself has not reached the height of the indexer's init_tip, the CKB should not exit.

eval-exec (approved):

resource/ckb.toml
```diff
@@ -196,6 +196,13 @@ block_uncles_cache_size    = 30
 # # Or you may want use more flexible scripts, block template as arg.
 # notify_scripts = ["{cmd} {blocktemplate}"]
 #
+# # CKB built-in indexer settings. Existing indexes can be cleaned up using the `ckb reset-data` subcommand.
```
quake, line 199:
```suggestion
# # CKB built-in indexer settings. Existing indexes can be cleaned up using the `ckb reset-data --indexer` subcommand.
```

quake (approved):

PR #4250, opened 2023-11-28
chore: CKB Text Optimization

EthanYuan:

Problem Summary:

CKB, being an executable program, produces a wide array of output messages during its runtime. These messages include console printouts, logs, error notifications, and text displayed within subcommands. These texts are authored by different developers at different stages of the development process, resulting in variations in language style and the possibility of syntax errors.

So, it is necessary to conduct an overall text review and optimization.

What's Changed:

We performed text extraction and editing based on revision 5975fda. Our professional writer has helped revise the text with existing issues. Many thanks to @linnnsss for the editing.

We utilized the text optimizer tool provided by #4189 to collect and format the text for organization. This made it possible for the writer to review. After the writer's editing, we employed the same tool for backfilling, resulting in the creation of this PR.

### Related changes

- #4189 

### Check List 

Tests 

### Release note 

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -113,7 +113,7 @@ impl NetworkState {
                     })
             })
             .collect();
-        info!("loading the peer store, which may take a few seconds to complete");
+        info!("Loading the peer store. This process may take a few seconds to complete.");
         let peer_store = Mutex::new(PeerStore::load_from_dir_or_default(
             config.peer_store_path(),
         ));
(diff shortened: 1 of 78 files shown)


network/src/network.rs
```diff
@@ -219,7 +219,7 @@ impl NetworkState {
             }
         } else {
             debug!(
-                "Ban session({}) failed: not in peer registry or it is in the whitelist",
+                "Ban session({}) failed: not found in peer registry or on the whitelist",
```
chenyukang, line 222: `not found in peer registry or it already on the whitelist`?

chenyukang (approved):

network/src/network.rs
EthanYuan: > not found in peer registry or it is on the whitelist

It looks more readable this way.

EthanYuan:
The text extracted and modified falls into 4 categories, each saved in:

- thiserror_text_list.yml
- clap_text_list.yml
- log_text_list.yml
- std_output_text_list.yml

They are packaged into a zip file and archived under this PR:

[revised_text_files.zip](https://github.com/nervosnetwork/ckb/files/13507055/revised_text_files.zip)

zhangsoledad (approved):

PR #4258, opened 2023-12-06
Fix concurrency issue for RBF 

chenyukang:

Problem Summary:

When multiple conflicted `tx` are added into `txpool` at the same time, because multiple threads may access same old data, `resolve_tx` may not resolve the conflicted tx to `Dead`, then `check_rbf` will be skipped, spec test `RbfConcurrency` is added to reproduce this scenario.

In this PR, `check_rbf` is moved to `submit_entry`, which will be guarded with a `write` lock, this change will make sure there is no double-spending tx in `Pending`, and with an extra benefit that only `tx` passed `verify_rtx` may invoke `check_rbf`.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/component/edges.rs
+++ b/tx-pool/src/component/edges.rs
@@ -1,5 +1,7 @@
-use ckb_logger::error;
-use ckb_types::packed::{Byte32, OutPoint, ProposalShortId};
+use ckb_types::{
+    core::tx_pool::Reject,
+    packed::{Byte32, OutPoint, ProposalShortId},
+};
 use std::collections::{hash_map::Entry, HashMap, HashSet};
 
 #[derive(Default, Debug, Clone)]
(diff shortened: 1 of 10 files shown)


tx-pool/src/process.rs
```diff
@@ -231,21 +239,25 @@ impl TxPoolService {
...
+                            let (rtx, status) = resolve_tx(tx_pool, &snapshot, tx.clone(), true)?;
+                            let fee = check_tx_fee(tx_pool, &snapshot, &rtx, tx_size)?;
+                            // Try an RBF cheap check, here if the tx is resolved as Dead,
+                            // we assume there must be conflicted happened in txpool now, if there is no conflicted transactions reject it
+                            let conflicts = tx_pool.find_conflict_tx(&rtx.transaction);
                             if conflicts.is_empty() {
+                                info!(
+                                    "{} is resolved as Dead, but there is no conflicted tx",
+                                    rtx.transaction.proposal_short_id()
+                                );
```
eval-exec, line 257: Should we use `error!` here?

driftluo (approved):

quake (changes requested):

tx-pool/src/component/edges.rs
```diff
@@ -28,6 +29,10 @@ impl Edges {
     }
 
     pub(crate) fn insert_input(&mut self, out_point: OutPoint, txid: ProposalShortId) {
+        // inputs is occupied means double speanding happened here
```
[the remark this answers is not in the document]
driftluo: I think this law change will have side effects. It will squeeze in late-entry transactions and destroy the behavior of data. We should not allow wrong behavior to continue.

tx-pool/src/component/edges.rs
```diff
@@ -27,8 +31,22 @@ impl Edges {
...
     }
 
-    pub(crate) fn insert_input(&mut self, out_point: OutPoint, txid: ProposalShortId) {
+    pub(crate) fn insert_input(
+        &mut self,
+        out_point: OutPoint,
+        txid: ProposalShortId,
+    ) -> Result<(), Reject> {
+        // inputs is occupied means double speanding happened here
+        if let Some(old) = self.inputs.get(&out_point) {
```
quake, line 40: suggest to use `hashmap#entry` which will reduce one hash op for the key.
```
match self.inputs.entry(out_point) {
  Entry::Occupied(occupied) => {
    return Err(Reject::RBFRejected(msg));
  }
  Entry::Vacant(vacant) => {
    vacant.insert(txid);
  }
}

quake (approved):

tx-pool/src/component/edges.rs
chenyukang: Good point, fixed!

quake (approved):

PR #4276, opened 2023-12-25
chore: Move  txpool statistics from callbacks to PoolMap

chenyukang:

Problem Summary:

The `total_tx_size` and `total_tx_cycle` are zero when we use `plug_entry` to add entries into `tx_pool`, which caused some confusing when debugging.

Updating statics was implemented in callbacks because we had 3 separate queue in `txpool`, now we have refactor it to the only one `PoolMap`, it's better and cleaner to move this part to `PoolMap`.

What's Changed:

Cleanup `callbacks` to move `txpool` statics part to `PoolMap`.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/pool.rs
+++ b/tx-pool/src/pool.rs
@@ -35,10 +35,6 @@ pub struct TxPool {
     pub(crate) pool_map: PoolMap,
     /// cache for committed transactions hash
     pub(crate) committed_txs_hash_cache: LruCache<ProposalShortId, Byte32>,
-    // sum of all tx_pool tx's virtual sizes.
-    pub(crate) total_tx_size: usize,
-    // sum of all tx_pool tx's cycles.
-    pub(crate) total_tx_cycles: Cycle,
     /// storage snapshot reference
     pub(crate) snapshot: Arc<Snapshot>,
     /// record recent reject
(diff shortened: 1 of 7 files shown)


tx-pool/src/pool.rs
```diff
@@ -323,18 +289,7 @@ impl TxPool {
...
-            }
-            return true;
-        }
-
-        if let Some(entry) = self.pool_map.remove_entry(id) {
-            self.update_statics_for_remove_tx(entry.size, entry.cycles);
-            return true;
-        }
-        false
+        !entries.is_empty()
```
eval-exec, line 292: I believe this change makes the logic inconsistent with the previous commit.

chenyukang: `self.pool_map.remove_entry(id)` will never return `Some(_)`, because the above `self.pool_map.remove_entry_and_descendants(id)` already removed the entry with `id`.

tx-pool/src/pool.rs
```diff
@@ -456,7 +410,7 @@ impl TxPool {
 
     pub(crate) fn drain_all_transactions(&mut self) -> Vec<TransactionView> {
         let mut txs = CommitTxsScanner::new(&self.pool_map)
-            .txs_to_commit(self.total_tx_size, self.total_tx_cycles)
+            .txs_to_commit(self.pool_map.total_tx_size, self.pool_map.total_tx_cycles)
```
quake, line 413: Changing it to `usize::max` better reflects that the intent of this fn is to drain all transactions from the pool.

```suggestion
            .txs_to_commit(usize::MAX, usize::MAX)
```

quake (approved):

PR #4303, opened 2024-01-16
feat(indexer): support cell data filter

EthanYuan:

Problem Summary:

We might want to filter cells based on output data, and the CKB Indexer can handle that easily.

What's Changed:

This pull request extends `IndexerSearchMode` by adding `Partial`, which, when combined with the existing `Prefix` and `Exact` modes, can be used for flexible filtering of cell data.

### Check List 

Tests 

- Unit test
- Manual test

```bash
curl -X POST -H "Content-Type: application/json" -d '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "get_cells",
    "params": [
        {
            "script": {
                "code_hash": "0xbbad126377d45f90a8ee120da988a2d7332c78ba8fd679aab478a19d6c133494",
                "hash_type": "data1",
                "args": "0x"
            },
            "script_type": "type",
            "script_search_mode": "prefix",
            "filter": {
                "output_data": "0xa58618a553",
                "output_data_filter_mode": "partial"
            },
            "with_data": false
        },
        "asc",
        "0x64"
    ]
}' 'http://127.0.0.1:8114'
```

```bash
curl -X POST -H "Content-Type: application/json" -d '{
    "id": 2,
    "jsonrpc": "2.0",
    "method": "get_cells_capacity",
    "params": [
        {
            "script": {
                "code_hash": "0xbbad126377d45f90a8ee120da988a2d7332c78ba8fd679aab478a19d6c133494",
                "hash_type": "data1",
                "args": "0x"
            },
            "script_type": "type",
            "script_search_mode": "prefix",
            "filter": {
                "output_data": "0xa58618a553",
                "output_data_filter_mode": "partial"
            }
        }
    ]
}' 'http://127.0.0.1:8114'
```

### Release note 

```release-note
```

### DIFF
--- a/util/indexer/src/service.rs
+++ b/util/indexer/src/service.rs
@@ -13,7 +13,7 @@ use ckb_async_runtime::{
 use ckb_db_schema::{COLUMN_BLOCK_BODY, COLUMN_BLOCK_HEADER, COLUMN_INDEX, COLUMN_META};
 use ckb_jsonrpc_types::{
     IndexerCell, IndexerCellType, IndexerCellsCapacity, IndexerOrder, IndexerPagination,
-    IndexerScriptSearchMode, IndexerScriptType, IndexerSearchKey, IndexerTip, IndexerTx,
+    IndexerScriptType, IndexerSearchKey, IndexerSearchMode, IndexerTip, IndexerTx,
     IndexerTxWithCell, IndexerTxWithCells, JsonBytes, Uint32,
 };
 use ckb_logger::{error, info};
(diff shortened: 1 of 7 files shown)


eval-exec (approved):

util/indexer/src/service.rs
```diff
@@ -433,6 +444,30 @@ impl IndexerHandle {
...
+                        }
+                        IndexerSearchMode::Exact => {
+                            if output_data.raw_data() != data {
+                                return None;
+                            }
+                        }
+                        IndexerSearchMode::Partial => {
+                            if !output_data
+                                .raw_data()
+                                .windows(data.len())
```
quake, line 462: An attacker may construct a query with zero data length which will cause the rpc server panic. Please add an unit test and check data length before call `windows` fn.

quake (changes requested):

util/indexer/src/service.rs
EthanYuan: OK

EthanYuan: Use `memchr::memmem::find` instead of `windows` for support of empty data.

quake (approved):

PR #4306, opened 2024-01-18
Fix RBF fee and celldeps check rule issue with replaced txs and their descendants

chenyukang:

Problem Summary:

The current RBF fee check rules use the directly conflicted transactions' fee as compare target, but in the child-pay-for-parent scenario, the new transaction with fee large than root transactions but less than all descendant's sum may succeed in RBF rules.

The current celldep check also has a similar issue.

This PR will fix these 2 issue, by adding all the fee of descendants and root transactions and check the celldeps including descendants.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


gpBlockchain:
LGTM

zhangsoledad (approved):

PR #4315, opened 2024-01-22
Don't account for cell dep for MAX_ANCESTORS_COUNT

chenyukang:

We met a scenario that a `cell dep` may have many transactions as descendants, and we can not  submit a new transaction to consume the dep cell.

Problem Summary:

`cell dep` should be handled specially when calculating the entry's `ancestors_count`.
In this PR, we add a `direct_ancestors_count` to store the directed parent-child relationship between transactions.

This is a short-term solution.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/tests/module/pool.rs
+++ b/rpc/src/tests/module/pool.rs
@@ -184,7 +184,8 @@ fn test_send_transaction_exceeded_maximum_ancestors_count() {
         });
     }
 
-    // the default value of pool config `max_ancestors_count` is 125, only 125 txs will be added to committed list of the block template
+    // the default value of pool config `max_ancestors_count` is 125,
+    // only 125 txs will be added to committed list of the block template
     suite.wait_block_template_array_ge("transactions", 1);
 
     let response = suite.rpc(&RpcTestRequest {
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -488,6 +488,7 @@ fn all_specs() -> Vec<Box<dyn Spec>> {
         Box::new(CompactBlockRelayLessThenSharedBestKnown),
         Box::new(InvalidLocatorSize),
         Box::new(SizeLimit),
+        Box::new(TxPoolLimitAncestorCount),
         Box::new(SendDefectedBinary::new(
             "send_defected_binary_reject_known_bugs",
             true,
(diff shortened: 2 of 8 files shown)


zhangsoledad (approved):

quake (approved):

rpc/src/tests/module/pool.rs
```diff
@@ -184,7 +184,8 @@ fn test_send_transaction_exceeded_maximum_ancestors_count() {
         });
     }
 
-    // the default value of pool config `max_ancestors_count` is 125, only 125 txs will be added to committed list of the block template
+    // the default value of pool config `max_ancestors_count` is 125,
```
eval-exec, line 187: But I found:
https://github.com/nervosnetwork/ckb/blob/4851b307b018b6bc86d2f79632223c97364f53c6/resource/ckb.toml#L140-L139

the `max_ancestors_count` is 25.

chenyukang: I think we can change the parameter in `ckb.toml` to 125 later.

quake: I think it's a comment error, we should change the default const value to 25 also.

eval-exec: Should we change `const DEFAULT_MAX_ANCESTORS_COUNT: usize` to 25 too?

chenyukang: it should be 125?
https://github.com/nervosnetwork/ckb/blob/b8ccf92e5c95310f6ea82a3c4575de7b05c03904/util/app-config/src/legacy/tx_pool.rs#L123

25 is originally referred from bitcoin, we use 125.

eval-exec: I think it should be `min` instead of `max`?
```rust
 max_ancestors_count: cmp::min(DEFAULT_MAX_ANCESTORS_COUNT, max_ancestors_count), 
```

chenyukang: no, it's `max`.
I think this commit is not complete, since we change from 25 to 125, we'd better also change the value in `ckb.toml`? 
https://github.com/nervosnetwork/ckb/commit/07d4db9e5ad4a9f0df079bf26e8aebd1c8e9b478

@zhangsoledad

PR #4314, opened 2024-01-20
Fix error message when sent HTTP GET to RPC

chenyukang:
### What problem does this PR solve?

After RPC framework changed, if we sent an GET request to RPC endpoint, the error message has changed to:

```console
Missing request extension: Extension of type `jsonrpc_utils::stream::StreamServerConfig` was not found. Perhaps you forgot to add it? See `axum::Extension`.%
```

Let's keep the same message with the old version.

Issue Number: close #4309

Problem Summary:

### What is changed and how it works?

Add a special handler for RPC endpoint, distinguished from the Websockets configuration.

What's Changed:

### Related changes

- PR to update `owner/repo`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- Performance regression
- Breaking backward compatibility

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4331, opened 2024-01-29
Fix the websockets terminate issue

chenyukang:

Problem Summary:

There is a trivial issue introduced in #4314, the `shutdown` layer didn't applied to websocket after the PR, which makes it will hung up when server terminate websocket.

Fix the shutdown handler, make it's default for HTTP and Websocket, since we will always send exit signal in these two scenario, so we can make the code simpler.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/server.rs
+++ b/rpc/src/server.rs
@@ -88,8 +88,10 @@ impl RpcServer {
     ) -> Result<SocketAddr, AnyError> {
         let stream_config = StreamServerConfig::default()
             .with_keep_alive(true)
-            .with_channel_size(4)
-            .with_pipeline_size(4);
+            .with_pipeline_size(4)
+            .with_shutdown(async move {
+                new_tokio_exit_rx().cancelled().await;
+            });
 
         // HTTP and WS server.
         let post_router = post(handle_jsonrpc::<Option<Session>>);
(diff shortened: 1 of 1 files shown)


rpc/src/server.rs
```diff
@@ -100,24 +102,14 @@ impl RpcServer {
...
             .route("/ping", get(ping_handler))
             .layer(Extension(Arc::clone(rpc)))
             .layer(CorsLayer::permissive())
             .layer(TimeoutLayer::new(Duration::from_secs(30)))
-            .layer(Extension(stream_config.clone()));
-
-        if enable_websocket {
-            let ws_config: StreamServerConfig =
-                stream_config
-                    .with_keep_alive(true)
```
eval-exec, line 115: Is it correct to remove `with_keep_alive(true)` from this PR?

chenyukang: there is already a `with_keep_alive` at line 90.

rpc/src/server.rs
```diff
@@ -88,8 +88,10 @@ impl RpcServer {
     ) -> Result<SocketAddr, AnyError> {
         let stream_config = StreamServerConfig::default()
             .with_keep_alive(true)
-            .with_channel_size(4)
```
gpBlockchain, line 91: Why should with_channel_size be changed to 8, and what impact will it have?
```
impl Default for StreamServerConfig {
    fn default() -> Self {
        Self {
            channel_size: 8,
            pipeline_size: 1,
            keep_alive: false,
            keep_alive_duration: Duration::from_secs(60),
            ping_interval: Duration::from_secs(19),
            shutdown_signal: future::pending().boxed().shared(),
        }
    }
}
```

chenyukang: it's a channel size used in session:
```rust
   let (tx, mut rx) = channel(config.channel_size);
    let session = Session {
        id: SESSION_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed),
        raw_tx: tx,
    };
```

if the size is larger, the channel may hold more temporary messages from requests before the backend functional code handles it, but of course need more memory, it will only make a difference in a very high load scenario.

We don't have a specific reason to make it `4`, so let's use the default?

gpBlockchain:
LGMT

eval-exec (approved):

gpBlockchain (approved):

PR #4339, opened 2024-02-01
Add conflicts cache for tx_pool to record conflicted transactions

chenyukang:

Problem Summary:

Starting from 112.0, we don't allow conflicted transactions stay in txpool at the same time, this may have a regression impact for compact block replay, since we reject any double-spending transaction when RBF is disable or RBF check failed.

We add a new pool called `conflicts_cache` into `TxPool` in this PR, we record the conflicted transaction in these scenarios:

1. RBF is enabled, record those replaced transactions into `conflicts_cache`
2. RBF is enabled, but RBF check rule failed with conflicts, record old transaction into `conflicts_cache`
3. RBF is disabled, but there is a conflict happening when `pre_check`, add the new transaction into `conflicts_cache` after verification.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


tx-pool/src/pool.rs
```diff
@@ -380,11 +388,14 @@ impl TxPool {
...
         proposal_id: &ProposalShortId,
     ) -> Option<TransactionView> {
-        self.get_tx_from_pool(proposal_id).cloned().or_else(|| {
-            self.committed_txs_hash_cache
-                .peek(proposal_id)
-                .and_then(|tx_hash| self.snapshot().get_transaction(tx_hash).map(|(tx, _)| tx))
-        })
+        self.get_tx_from_pool(proposal_id)
+            .cloned()
+            .or_else(|| self.conflicts_cache.peek(proposal_id).cloned())
```
eval-exec, line 393: When using `get_tx_with_cycles`, should it use `peek` or `get`?

> Unlike `get`, `peek` does not update the least recently used (LRU) list, so the position of the key will remain unchanged.

chenyukang: I think we should use `peek` here (the difference is very trivial),
`get_tx_from_pool_or_store` is with `&self`, if use `get` we need `mut self`, 
I guess the same reason applied to `committed_txs_hash_cache.peek(..)` .

sync/src/relayer/mod.rs
```diff
@@ -447,8 +447,20 @@ impl Relayer {
 
         if !short_ids_set.is_empty() {
             let tx_pool = self.shared.shared().tx_pool_controller();
-
+            let short_ids_set_clone = short_ids_set.clone();
```
quake, line 450: suggest to use `log_enabled` and avoids clone when the debug level is disabled.

tx-pool/src/process.rs
```diff
@@ -135,7 +155,7 @@ impl TxPoolService {
                 for id in conflicts.iter() {
                     let removed = tx_pool.pool_map.remove_entry_and_descendants(id);
                     for old in removed {
-                        debug!(
+                        eprintln!(
```
quake, line 158: why use eprintln here?

chenyukang: it's forgotten after debug code, rollbacked.

sync/src/relayer/mod.rs
chenyukang: fixed.

eval-exec (dismissed):

doitian (approved):

zhangsoledad (approved):

PR #4345, opened 2024-02-08
`IndexerService::apply_init_tip` should stop after received exit signal.

eval-exec:
### What problem does this PR solve?

Issue Number: close #4311

Problem Summary:

In the `apply_init_tip` function, there is a loop that should take into account the exit signal:
https://github.com/nervosnetwork/ckb/blob/1ee9c985d61285749128adbbb916ede7742404aa/util/indexer/src/service.rs#L156-L172

 Additionally, once `initial_syncing` is finished and CKB receives an exit signal, there is no need to spawn the `new_block_watcher` polling job.

What's Changed:

### Related changes

- handle exit signal in `apply_init_tip`

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- None

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4348, opened 2024-02-09
IndexerService should use `is_cancelled()` to check if ckb received Ctrl-C signal

eval-exec:
…trl-c signal

Problem Summary:
In #4345, the `tokio::select!` will stuck forever if ckb doesn't received ctrl-C signal. Should use `stop.is_canceled()`.

What's Changed:

### Related changes

- use `stop.is_canceled()` to check if ckb has received exit signal.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None
### Release note 

```release-note
```

### DIFF
--- a/util/indexer/src/service.rs
+++ b/util/indexer/src/service.rs
@@ -238,13 +238,9 @@ impl IndexerService {
         self.async_handle.spawn(async move {
             let _initial_finished = initial_syncing.await;
             info!("initial_syncing finished");
-
-            tokio::select! {
-                _ = stop.cancelled() => {
-                    info!("Indexer received exit signal, cancel new_block_watcher task, exit now");
-                    return;
-                },
-                else => {},
+            if stop.is_cancelled() {
+                info!("Indexer received exit signal, cancel new_block_watcher task, exit now");
+                return;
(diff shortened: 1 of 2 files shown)


quake (approved):

EthanYuan (approved):

util/indexer/src/service.rs
```diff
@@ -238,13 +238,9 @@ impl IndexerService {
...
             info!("initial_syncing finished");
-
-            tokio::select! {
-                _ = stop.cancelled() => {
-                    info!("Indexer received exit signal, cancel new_block_watcher task, exit now");
-                    return;
-                },
-                else => {},
+            if stop.is_cancelled() {
+                info!("Indexer received exit signal, cancel new_block_watcher task, exit now");
```
15168316096, line 242: if use rich-indexer,this content need update? auto get indexer type which is using

EthanYuan: Not needed. The version of CKB that includes rich-indexer has been refactored, and this check is moved to the module `indexer-sync`, which is common to both indexer and rich-indexer.

PR #4352, opened 2024-02-18
[WIP] Add invalidated_tx_count to fix cell dep and cell consume issues

chenyukang:
### What problem does this PR solve?

Problem Summary:

The are conflicts rule between cell dep reference and cell consume relationships.

Assume there are 3 transactions, B cell dep A(and there may be many transactions also cell dep A), C consumes A. 

- If C is committed first, B and all other transactions will be invalidated.
- If there are too many transactions all cell dep A, transaction C can not be submitted to `txpool`.

### What is changed and how it works?

In this PR, we introduce these changes:

- add a new `invalidated_tx_count` to Entry, marked as how many transactions will be invalidated if by this transaction entry.
- Add `invalidators` to store all transactions which may invalidate the current transaction
- Remove `DirectParents` and  `direct_ancestors_count`, it's now replaced by `Parents` and `ancestors_count`
- Fix order based on the relationships of invalidators and parents.

What's Changed:

### Related changes

- PR to update `owner/repo`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- Performance regression
- Breaking backward compatibility

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4363, opened 2024-02-23
Evict possible cell ref txs before submitting cell consuming transaction

chenyukang:

Problem Summary:

The are conflicts rule between cell dep reference and cell consume relationships.

Assume there are 3 transactions, B cell dep A(and there may be many transactions also cell dep A), C consumes A. 

- If C is committed first, B and all other transactions will be invalidated.
- If there are too many transactions all cell dep A, transaction C can not be submitted to `txpool`.

This is an alternative solution with https://github.com/nervosnetwork/ckb/pull/4352
In this PR, we introduce these changes:

- Add a new `check_entry_ancestors_limit ` to find all candidates needed to evict due to cell ref conflicts, based on `evict_score`. 
- Remove `DirectParents` and  `direct_ancestors_count`, we make sure an entry's ancestors number is valid after evicting conflicted cell ref transactions.
- Move `ExceededMaximumAncestorsCount` check to place before inserting, so that we can assume there will no error report when committing a entry.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/component/pool_map.rs
+++ b/tx-pool/src/component/pool_map.rs
@@ -569,16 +569,17 @@ impl PoolMap {
                     break;
                 }
             }
+        } else {
+            return Err(Reject::ExceededMaximumAncestorsCount);
         }
 
-        // if ancestors count is still larger than limitation,
-        // return ExceededMaximumAncestorsCount error directly
+        // some txs in `parents` are removed, now `ancestors` need to re-caculate,
         let ancestors = self
             .links
             .calc_relation_ids(parents.clone(), Relation::Parents);
-        if ancestors.len() + 1 > self.max_ancestors_count {
-            return Err(Reject::ExceededMaximumAncestorsCount);
-        }
+
+        // we can assume the number now is less than `max_ancestors_count`
+        assert!(ancestors.len() < self.max_ancestors_count);
 
         self._record_ancestors(entry, ancestors, parents);
         Ok(evicted)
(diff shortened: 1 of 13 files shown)


tx-pool/src/component/pool_map.rs
```diff
@@ -119,7 +119,7 @@ impl PoolMap {
 
     #[cfg(test)]
     pub(crate) fn add_proposed(&mut self, entry: TxEntry) -> Result<bool, Reject> {
-        self.add_entry(entry, Status::Proposed)
+        self.add_entry(entry, Status::Proposed).map(|_| true)
```
doitian, line 122: Any background for this change? Does it revert the "Direct Parent" patch?

tx-pool/src/component/pool_map.rs
```diff
@@ -189,19 +189,24 @@ impl PoolMap {
             })
     }
 
-    pub(crate) fn add_entry(&mut self, mut entry: TxEntry, status: Status) -> Result<bool, Reject> {
+    pub(crate) fn add_entry(
```
doitian, line 193: Please add a simple doc for the func about the return value.

tx-pool/src/component/pool_map.rs
chenyukang: `add_entry` was changed. Anyway, I just fixed `add_entry` to return `(succ, evicts)`, so that we can still get a flag whether the entry was inserted successfully.

doitian (dismissed):

doitian (approved):

doitian (approved):

PR #4380, opened 2024-03-13
New spawn with scheduler

mohanson:
In this PR, we refactored the implementation of spawn. The refactored syscall API is defined as follows: https://github.com/XuJiandong/ckb-c-stdlib/blob/syscall-spawn/ckb_syscall_apis.h#L54-L68.

Review Introduction: https://github.com/nervosnetwork/ckb/pull/4380#issuecomment-2058070291

Documentation: https://github.com/nervosnetwork/rfcs/pull/436


script/Cargo.toml
```diff
@@ -32,8 +32,8 @@ tokio = { version = "1.35.0", features = ["rt-multi-thread"] }
 
 [dev-dependencies]
 proptest = "1.0"
-ckb-db = { path = "../db", version = "= 0.114.0-pre" }
-ckb-store = { path = "../store", version = "= 0.114.0-pre" }
+ckb-db = { path = "../db", version = "= 0.114.0-pre", features = ["portable"] }
+ckb-store = { path = "../store", version = "= 0.114.0-pre", features = ["portable"] }
```
chenyukang, line 36: why we need to add this.

mohanson: Because @XuJiandong is developed on an ancient machine, the code can only be compiled after adding this feature. It can be deleted in the future.

script/testdata/README.md
```diff
@@ -4,7 +4,7 @@
 
   - Create a directory named `deps`.
 
-  - Clone https://github.com/nervosnetwork/ckb-c-stdlib into `deps`.
+  - Clone https://github.com/XuJiandong/ckb-c-stdlib.git with branch `syscall-spawn` into `deps`.
```
chenyukang, line 7: we may need to add a FIXME to remember to clean this up if we don't want to resolve it right now.

chenyukang:
There is trivial difference for TPS in benchmark testing

- benchmark data for this PR
```console
duration send tx tps 615,duration avg delay 12.958ms,sum tps:607
```

- benchmark data for `yukang-new-verify` branch
```console
duration send tx tps 584,duration avg delay 13.656ms,sum tps:608
```


cc @gpBlockchain, you may want to double confirm it and also run benchmark testing for transactions with large cycles.

XuJiandong:
### Syscall Spawn Review Introduction


The design of the syscall `spawn` function draws inspiration from Unix and
Linux, hence they share the same terminologies: process, pipe, and file
descriptor.

In the context of `ckb-vm`, a process represents the active execution of a
RISC-V binary. This binary can be located within a cell, which is indicated by a
script or `OutPoint`. Additionally, a RISC-V binary can also be found within the
witness during a syscall `spawn` operation. A pipe is established by associating
two file descriptors, each linked to one of its ends. These file descriptors
cannot be duplicated and are exclusively owned by the process. Furthermore, the
file descriptors can only be either read from or written to; they cannot be both
read from and written to simultaneously.

When the spawn serial syscalls (such as ckb_spawn, ckb_read, ckb_write, etc.)
are invoked, the corresponding process will yield, and they will be scheduled to
run again. Some processes can run immediately, while others will enter a "wait"
status (VmState::Wait, VmState::WaitForWrite, VmState::WaitForRead) due to IO
operations.

During the wait status, there are two forms of memory representation. The first
is the "instantiated" status, which holds the full memory of the process. The
second is the "suspended" status, which holds a snapshot of memory. The latter
can save a significant amount of memory. When the process is ready to run, a
process with "suspended" status becomes "instantiated" by unpacking the snapshot
into full memory.

The spawn serial introduces the following [syscalls](https://github.com/XuJiandong/ckb-c-stdlib/blob/c651f745f8a1bb5065e73efa2d84f3318973559a/ckb_syscall_apis.h#L54-L68):

- ckb_spawn
- ckb_wait
- ckb_process_id
- ckb_pipe
- ckb_read
- ckb_write
- ckb_inherited_file_descriptors
- ckb_close

Some usage of syscalls can be found [here](https://github.com/libraries/ckb/blob/9b96f5379369fb96b71da2a6a15803bd6fba6ccd/script/testdata/spawn_utils.h#L71).

The RFC documentation is currently under preparation.

XuJiandong:
Please find the attached coverage report from unit tests:
[tarpaulin-report-0417.html.gz](https://github.com/nervosnetwork/ckb/files/15006643/tarpaulin-report-0417.html.gz)

It is generated via cargo tarpaulin:
```
cargo tarpaulin -o html --tests -- v2023
```

script/src/syscalls/spawn.rs
```diff
@@ -1,95 +1,43 @@
...
-use ckb_types::core::cell::CellMeta;
 use ckb_vm::{
-    cost_model::estimate_cycles,
-    registers::{A0, A1, A2, A3, A4, A5, A7},
-    DefaultMachineBuilder, Error as VMError, Memory, Register, SupportMachine, Syscalls,
+    machine::SupportMachine,
+    memory::Memory,
+    registers::{A0, A1, A2, A3, A4, A7},
+    snapshot2::{DataSource, Snapshot2Context},
+    syscalls::Syscalls,
```
XuJiandong, line 12: test

script/src/syscalls/write.rs
```diff
@@ -0,0 +1,60 @@
...
+            .memory_mut()
+            .load64(&Mac::REG::from_u64(length_addr))?
+            .to_u64();
+
+        // We can only do basic checks here, when the message is actually processed,
+        // more complete checks will be performed.
+        // We will also leave to the actual write operation to test memory permissions.
+        if !fd.is_write() {
+            machine.set_register(A0, Mac::REG::from_u8(INVALID_FD));
+            return Ok(true);
```
eval-exec, line 43: Should it return `Ok` here? Or `Err`?

script/src/verify.rs
```diff
@@ -853,18 +844,21 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
 
         for (idx, (_hash, group)) in self.groups().enumerate().skip(current + 1) {
             let remain_cycles = limit_cycles.checked_sub(current_used).ok_or_else(|| {
-                ScriptError::Other(format!("expect invalid cycles {limit_cycles} {cycles}"))
-                    .source(group)
+                ScriptError::Other(format!(
+                    "here expect invalid cycles {limit_cycles} {cycles}"
```
eval-exec, line 848: Do we need to append `{current_used}` to `format!`'s argument?

script/src/verify.rs
```diff
@@ -853,18 +844,21 @@ impl<DL: CellDataProvider + HeaderProvider + ExtensionProvider + Send + Sync + C
...
             })?;
 
-            match self.verify_group_with_chunk(group, remain_cycles, &[]) {
-                Ok(ChunkState::Completed(used_cycles)) => {
-                    current_used = wrapping_cycles_add(current_used, used_cycles, group)?;
-                    cycles = wrapping_cycles_add(cycles, used_cycles, group)?;
+            match self.verify_group_with_chunk(group, remain_cycles, &None) {
+                Ok(ChunkState::Completed(used_cycles, consumed_cycles)) => {
+                    current_used = wrapping_cycles_add(current_used, consumed_cycles, group)?;
+                    eprintln!("used_cycles: {:?}", used_cycles);
```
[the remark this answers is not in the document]
mohanson: This is a temporarily log and has been removed.

script/src/verify.rs
mohanson: better add it

script/src/syscalls/write.rs
mohanson: It should return Ok(true), which indicates that the syscall has been executed normally. This is the common way of writing syscalls, such as https://github.com/nervosnetwork/ckb/blob/31bcecddc8ad5b7d0a30399ef55a5ce0a359bb7e/script/src/syscalls/load_header.rs# L165-L168

script/src/scheduler.rs
```diff
@@ -0,0 +1,882 @@
...
+                ids.len()
+            )));
+        }
+
+        let mut uninstantiated_ids: Vec<VmId> = ids
+            .iter()
+            .filter(|id| !self.instantiated.contains_key(id))
+            .copied()
+            .collect();
+        while (!uninstantiated_ids.is_empty()) && (self.instantiated.len() < MAX_INSTANTIATED_VMS) {
```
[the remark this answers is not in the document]
mohanson: `self.instantiated.len()` could be equal to `MAX_INSTANTIATED_VMS`. In this case, an instantiated vm will be suspended (from instantiated to uninstantiated), and the vm that needs to be running will be changed from uninstantiated to instantiated.

XuJiandong (changes requested):

script/src/scheduler.rs
```diff
@@ -0,0 +1,882 @@
...
+                        vm_id,
+                        VmState::Wait {
+                            target_vm_id: args.target_id,
+                            exit_code_addr: args.exit_code_addr,
+                        },
+                    );
+                }
+                Message::Pipe(vm_id, args) => {
+                    if self.fds.len() as u64 >= MAX_FDS {
+                        let (_, machine) = self
```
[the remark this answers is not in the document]
mohanson: Yes, this is a bug

chain/src/tests/load_code_with_snapshot.rs
```diff
@@ -113,7 +113,7 @@ fn test_load_code() {
     let tx_status = tx_pool.get_tx_status(tx.hash());
     assert_eq!(
         tx_status.unwrap().unwrap(),
-        (TxStatus::Pending, Some(11174))
+        (TxStatus::Pending, Some(11325))
```
quake, line 116: the lock script's `hash_type` is `Data`, why is the cycle changed?

mohanson: I recompiled the lib used in this testcase with new compilation parameters, you can view the modification in script/testdata/Makefile

quake (changes requested):

script/src/scheduler.rs
```diff
@@ -0,0 +1,875 @@
...
+                Message::Spawn(vm_id, args) => {
+                    // All fds must belong to the correct owner
+                    let mut fds_valid = true;
+                    for fd in &args.fds {
+                        if !(self.fds.contains_key(fd) && (self.fds[fd] == vm_id)) {
+                            self.ensure_vms_instantiated(&[vm_id])?;
+                            let (_, machine) =
+                                self.instantiated.get_mut(&vm_id).ok_or_else(|| {
+                                    Error::Unexpected("Unable to find VM Id".to_string())
+                                })?;
```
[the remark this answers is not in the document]
mohanson: `machine` are mutable borrows, using it as return values will make `self` also mutable borrows, writing codes like this in Rust is terrible.

script/src/syscalls/read.rs
```diff
@@ -0,0 +1,60 @@
...
+        let buffer_addr = machine.registers()[A1].to_u64();
+        let length_addr = machine.registers()[A2].to_u64();
+        let length = machine
+            .memory_mut()
+            .load64(&Mac::REG::from_u64(length_addr))?
+            .to_u64();
+
+        // We can only do basic checks here, when the message is actually processed,
+        // more complete checks will be performed.
+        // We will also leave to the actual write operation to test memory permissions.
```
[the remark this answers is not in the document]
mohanson: A mistake, `We will also leave to the actual write operation to test memory permissions.` should be removed.

script/src/scheduler.rs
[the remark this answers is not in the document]
quake: I don't think it's a problem in rust, please check the patch:
[ensure_vm_instantiated.diff.txt](https://github.com/nervosnetwork/ckb/files/15106702/ensure_vm_instantiated.diff.txt)

After we change all the snippets from `ensure_vms_instantiated` to `ensure_vm_instantiated`, the previous fn can be removed, and we can use a simpler method to resume from full state.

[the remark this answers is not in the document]
mohanson: This works in some places, but not in others, such as in `fn iterate()`.

```rs
let (context, machine) = self.ensure_vm_instantiated(&vm_id_to_run)?; // The function needs to be modified so that it returns context
context.set_base_cycles(self.total_cycles); // cannot use `self.total_cycles` because it was mutably borrowed use of borrowed `*self`
```

PR #4381, opened 2024-03-14
feat: limit ibd orphan pool size

driftluo:

This PR adds a special download mode during ibd to limit the memory usage of the orphan pool.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/sync/src/orphan_block_pool.rs
+++ b/sync/src/orphan_block_pool.rs
@@ -1,4 +1,4 @@
-use ckb_logger::debug;
+use ckb_logger::{debug, error};
 use ckb_types::core::EpochNumber;
 use ckb_types::{core, packed};
 use ckb_util::{parking_lot::RwLock, shrink_to_fit};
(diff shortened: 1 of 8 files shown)


sync/src/orphan_block_pool.rs
```diff
@@ -72,6 +78,7 @@ impl InnerPool {
             }
         }
 
+        self.block_size -= removed.iter().map(|b| b.data().total_size()).sum::<usize>();
```
doitian, line 81: Be more defensive here? Since if we mess up and the block size is wrapped around, we will keep in the special downloading mode for a long time.

```suggestion
        self.block_size = self.block_size.checked_sub(removed.iter().map(|b| b.data().total_size()).sum::<usize>()).unwrap_or_else(|| {
            // report error
            0
        });
```

driftluo: Both addition and subtraction were changed to checked, and then the error log was output, although I think there will be no problem.

doitian (dismissed):

zhangsoledad (approved):

quake (approved):

PR #4387, opened 2024-03-21
Fix type name issue caused from schemars

chenyukang:

There is an issue from schemars: https://github.com/GREsau/schemars/issues/193, which may insert `_and_` and `_for_` in types, we don't want this.

Add a wrapper function `fix_type_name` to workaround it, it will cost much more time to wait schemars fix its issue, or we need to dig into schemars by ourself which will also cost time.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -6037,6 +6037,7 @@ It's an enum value from one of:
   - desc : Descending order
   - asc : Ascending order
 
+<a id="type-indexerpagination_for_indexercell"></a>
 ### Type `IndexerPagination<IndexerCell>`
 IndexerPagination wraps objects array and last_cursor to provide paging
 
@@ -6048,6 +6049,7 @@ IndexerPagination wraps objects array and last_cursor to provide paging
 
 * `objects`: `Array<` [`IndexerCell`](#type-indexercell) `>` - objects collection
 
+<a id="type-indexerpagination_for_indexertx"></a>
 ### Type `IndexerPagination<IndexerTx>`
 IndexerPagination wraps objects array and last_cursor to provide paging
 
@@ -6578,6 +6580,7 @@ The information about an active running protocol.
 
 * `version`: `string` - Active protocol version.
 
+<a id="type-responseformat_for_blockview"></a>
 ### Type `ResponseFormat<BlockView>`
 This is a wrapper for JSON serialization to select the format between Json and Hex.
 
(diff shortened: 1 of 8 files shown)


quake (dismissed):

rpc/README.md
```diff
@@ -171,14 +171,15 @@ The crate `ckb-rpc`'s minimum supported rustc version is 1.71.1.
...
+    * [Type `HardForks`](#type-hardforks)
     * [Type `Header`](#type-header)
     * [Type `HeaderView`](#type-headerview)
     * [Type `IndexerCell`](#type-indexercell)
     * [Type `IndexerCellType`](#type-indexercelltype)
     * [Type `IndexerCellsCapacity`](#type-indexercellscapacity)
     * [Type `IndexerOrder`](#type-indexerorder)
-    * [Type `IndexerPagination_for_IndexerCell`](#type-indexerpagination_for_indexercell)
-    * [Type `IndexerPagination_for_IndexerTx`](#type-indexerpagination_for_indexertx)
+    * [Type `IndexerPagination<IndexerCell>`](#type-indexerpagination_for_indexercell)
```
EthanYuan, line 181: link is broken.

rpc/README.md
```diff
@@ -6566,6 +6578,45 @@ The information about an active running protocol.
...
+
+###### Examples
+
+`ResponseFormat<BlockView>` returns the block in its Json format or molecule serialized Hex format.
+
+#### Fields
+
+`ResponseFormat<HeaderView>` is a JSON object with the following fields.
+
+* `inner`: [`Either<HeaderView | JsonBytes>`](#type-either_for_headerview_and_jsonbytes) - The inner value.
```
EthanYuan, line 6605: The type `Either` has not been generated, but there is a link.

rpc/README.md
chenyukang: fixed.

rpc/README.md
chenyukang: yes, because this is a hack fix for schemars, schemars respect this option:
https://github.com/nervosnetwork/ckb/blob/ed1c72fae1ef50be982158ab7764cdea02de13bc/util/jsonrpc-types/src/lib.rs#L87

but we don't know how to determine whether to show the link, so it's safe to show all links for types(even in small possibility the link is only works as a dead link as `#`).

similar scenarios happened in the previous version of doc generator.

EthanYuan (approved):

eval-exec (approved):

PR #4390, opened 2024-03-25
Limit txpool size when inserting an Entry 

chenyukang:

Problem Summary:

Currently, the `limit_size` is only invoked by `_update_tx_pool_for_reorg`, so we only shrink tx_pool size when generating a new block. Ideally, we should also limit size when inserting a new Entry into pool.

What's Changed:
Invoke `limit_size` in `submit_entry`.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/pool.rs
+++ b/tx-pool/src/pool.rs
@@ -263,7 +263,13 @@ impl TxPool {
     }
 
     // Remove transactions from the pool until total size <= size_limit.
-    pub(crate) fn limit_size(&mut self, callbacks: &Callbacks) {
+    // Return a `Reject` for current inserting entry if it's removed
+    pub(crate) fn limit_size(
+        &mut self,
+        callbacks: &Callbacks,
+        current_entry_id: Option<&ProposalShortId>,
+    ) -> Option<Reject> {
+        let mut ret = None;
         while self.pool_map.total_tx_size > self.config.max_tx_pool_size {
             let next_evict_entry = || {
                 self.pool_map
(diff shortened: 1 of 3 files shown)


tx-pool/src/pool.rs
```diff
@@ -263,7 +263,12 @@ impl TxPool {
     }
 
     // Remove transactions from the pool until total size <= size_limit.
-    pub(crate) fn limit_size(&mut self, callbacks: &Callbacks) {
+    pub(crate) fn limit_size(
+        &mut self,
+        callbacks: &Callbacks,
+        current_entry_id: Option<&ProposalShortId>,
+    ) -> Result<(), Reject> {
```
eval-exec, line 270: Given that `current_entry_id` is an `Option`, should the return type be `Option<Reject>` too?

The primary function of `limit_size` is to remove transactions to ensure the total size fits within the size_limit, so its execution should always be successful and not return an `Err`.

chenyukang: make sense, then we use `map_or` to convert it into `Err` if necessary.

chenyukang:
I ran a benchmark testing for this PR, seems there is no effect on TPS:

```console
duration send tx tps 619,duration avg delay 12.895ms,sum tps:620
```

cc @gpBlockchain @zhangsoledad

zhangsoledad (approved):

PR #4418, opened 2024-04-14
Set ChainService process_block channel size to zero

eval-exec:

On the develop branch, the `ChainService`'s `process_block` channel has a capacity of `DEFAULT_CHANNEL_SIZE`(32), but it does not provide any benefit for `ckb-sync` or `ckb-chain`:
https://github.com/nervosnetwork/ckb/blob/149b5c53e2c6e089d3e4ff6013d1992b56f3b4c6/chain/src/chain.rs#L225

In a scenario where `ckb-sync` sends a block to `ChainService` through `ChainController::process_block`, and at the same time, `ChainService` receives a `Ctrl-C` exit signal.
Then `ChainService`'s `process_block_sender` have one `ProcessBlockRequest` in its buffer:

https://github.com/nervosnetwork/ckb/blob/149b5c53e2c6e089d3e4ff6013d1992b56f3b4c6/chain/src/chain.rs#L41

When `self.process_block_sender` and `self.stop_receiver` are ready at the same time, `ChainService`'s `select!` may choose to execute `self.stop_receiver`:
https://github.com/nervosnetwork/ckb/blob/149b5c53e2c6e089d3e4ff6013d1992b56f3b4c6/chain/src/chain.rs#L237-L272

then the `process_block_receiver` will be dropped.

However, the `process_block_sender` is held by the `Synchronizer`'s `ChainController`, and there is one `ProcessBlockRequest` in `process_block_sender`'s buffer, causing the `ChainController::internal_block_process` to block indefinitely.

In that case, `Synchronizer::received` will block indefinitely at this point:
https://github.com/nervosnetwork/ckb/blob/149b5c53e2c6e089d3e4ff6013d1992b56f3b4c6/sync/src/synchronizer/mod.rs#L784

### Related changes

- Set `ChainService`'s process_block channel capacity size to 0.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -20,7 +20,7 @@ use ckb_types::{
             ResolvedTransaction,
         },
         hardfork::HardForks,
-        service::{Request, DEFAULT_CHANNEL_SIZE},
+        service::Request,
         BlockExt, BlockNumber, BlockView, Cycle, HeaderView,
     },
     packed::{Byte32, ProposalShortId},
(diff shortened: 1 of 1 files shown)


quake (approved):

eval-exec:
> Will it have any negative impact?

There should be no negative impact.

driftluo:
ref：https://github.com/crossbeam-rs/crossbeam/issues/1102
this is a bug for crossbeam channel

PR #4452, opened 2024-05-09
Skip storing all zero cycles in db when disable script

chenyukang:

When run `ckb run` without disabling `assume-valid-target`, all the cycles in blocks are zero, since we skipped verification in this scenario.

Don't store all zeros in db, set `ext.cycles` to None.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


eval-exec (commented):
How about change `ContextualBlockVerifier::verifier`'s return type to `Option<Cycle, ...>`?

zhangsoledad (approved):

PR #4465, opened 2024-05-20
feat: add estimate_fee_rate rpc

quake:

resolve https://github.com/nervosnetwork/ckb/issues/3673

Since CKB transaction confirmation involves a two-step process: 1) propose and 2) commit, it is complex to predict the transaction fee accurately with the expectation that it will be included within a certain block height.

This PR relies on two assumptions and uses a simple strategy to estimate the transaction fee: 1) all transactions in the pool are waiting to be proposed, and 2) no new transactions will be added to the pool. In practice, this simple method should achieve good accuracy fee rate and running performance.

The rpc parameter `target_to_be_committed` is limited to 3~131, because the minimum value of the proposal window is 2, which is equivalent to predicting the height of current_tip + 1 ~ 128.

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```

### DIFF
--- a/rpc/README.md
+++ b/rpc/README.md
@@ -104,6 +104,7 @@ The crate `ckb-rpc`'s minimum supported rustc version is 1.71.1.
     * [Module Pool](#module-pool) [👉 OpenRPC spec](http://playground.open-rpc.org/?uiSchema[appBar][ui:title]=CKB-Pool&uiSchema[appBar][ui:splitView]=false&uiSchema[appBar][ui:examplesDropdown]=false&uiSchema[appBar][ui:logoUrl]=https://raw.githubusercontent.com/nervosnetwork/ckb-rpc-resources/develop/ckb-logo.jpg&schemaUrl=https://raw.githubusercontent.com/nervosnetwork/ckb-rpc-resources/develop/json/pool_rpc_doc.json)
 
         * [Method `send_transaction`](#pool-send_transaction)
+        * [Method `estimate_fee_rate`](#pool-estimate_fee_rate)
         * [Method `test_tx_pool_accept`](#pool-test_tx_pool_accept)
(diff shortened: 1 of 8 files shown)


rpc/README.md
```diff
@@ -4422,6 +4423,53 @@ Response
...
+predict the transaction fee accurately with the expectation that it will be included within a certain block height.
+
+This method relies on two assumptions and uses a simple strategy to estimate the transaction fee: 1) all transactions
+in the pool are waiting to be proposed, and 2) no new transactions will be added to the pool.
+
+In practice, this simple method should achieve good accuracy fee rate and running performance.
+
+###### Params
+
+* `target_to_be_committed` - The target block number to be committed, minimum value is 3 and maximum value is 131.
```
eval-exec, line 4444: > The rpc parameter target_to_be_committed is limited to 3~131, because the minimum value of the proposal window is 2, which is equivalent to predicting the height of current_tip + 1 ~ 128.

128+2=130

eval-exec (approved):

tx-pool/src/component/pool_map.rs
```diff
@@ -329,6 +329,33 @@ impl PoolMap {
...
+                target_blocks -= 1;
+                if target_blocks == 0 {
+                    return entry.inner.fee_rate();
+                }
+                current_block_bytes = entry.inner.size;
+                current_block_cycles = entry.inner.cycles;
+            }
+        }
+
+        min_fee_rate
```
chenyukang, line 356: what is `min_fee_rate` used for in this function? seems it is not referenced in this function.

quake: this fn will return `min_fee_rate` when estimation target is not found.

quake:
merged into #4477

PR #4468, opened 2024-05-23
Add fuzz on network

joii2020:
Use fuzz to test some code of the network.
Found a bug: 
[AddrManager::fetch_random](https://github.com/nervosnetwork/ckb/blob/develop/network/src/peer_store/addr_manager.rs#L44-L77)

When `count` is 0 and `addr_infos.push(addr_info);`(on line 69) has already been executed, checking `if addr_infos.len() == count` will definitely return false. At this point, it will continue executing all of `self.random_ids` before returning. I believe that in this case, when `count` is 0, it should immediately return an empty `Vec`.


network/fuzz/fuzz_targets/fuzz_peer_store.rs
```diff
@@ -0,0 +1,117 @@
...
+    let basic_num = data.get::<u32>();
+
+    let last_connected_time_range = if flags >> 2 & 0b1 == 1 {
+        60_000 // 1 minute
+    } else {
+        1000 * 60 * 60 * 24 * 365
+    };
+
+    let num = basic_num % 16 + (16384) - 8; // ±8
+    let mut rng = thread_rng();
```
XuJiandong, line 74: Don't use rng in fuzzing test. Use random data from input.

network/fuzz/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2024-05-14
```
XuJiandong, line 1: better to choose `nightly`

network/fuzz/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn from_bytes(d: &[u8]) -> ipnetwork::Ipv6Network {
+        Self::from(std::net::Ipv6Addr::from_bytes(d))
+    }
+}
+impl FromBytes<PeerId> for PeerId {
+    fn type_size() -> usize {
+        32
+    }
+    fn from_bytes(d: &[u8]) -> PeerId {
+        PeerId::from_bytes(vec![vec![0x12], vec![0x20], d.to_vec()].concat()).unwrap()
```
XuJiandong, line 141: 0x12, 0x20 can be from random input.

network/fuzz/fuzz_targets/fuzz_peer_store.rs
```diff
@@ -0,0 +1,117 @@
...
+    let now_ms = ckb_systemtime::unix_time_as_millis();
```
XuJiandong, line 75: don't use time `now`

network/fuzz/fuzz_targets/fuzz_peer_store.rs
```diff
@@ -0,0 +1,117 @@
...
+            now_ms - ((rng.next_u32() as u64) % last_connected_time_range),
+            rng.next_u32(),
+        );
+    }
+}
+
+fuzz_target!(|data: &[u8]| {
+    let mut data = BufManager::new(data);
+
+    let now_ms = ckb_systemtime::unix_time_as_millis();
```
XuJiandong, line 97: Don't use time `now`

network/fuzz/fuzz_targets/fuzz_peer_store.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    addr_str += &format!("/tcp/{}", data.get::<u16>());
+
+    addr_str += &format!("/p2p/{}", data.get::<ckb_network::PeerId>().to_base58());
+
+    (addr_str.parse().unwrap(), flags)
+}
+
+fn add_ban_addr(data: &mut BufManager, peer_store: &mut PeerStore) {
+    let now_ms = ckb_systemtime::unix_time_as_millis();
```
XuJiandong, line 33: Don't use time `now`.

XuJiandong (changes requested):

doitian (approved):

XuJiandong (approved):

zhangsoledad (approved):

PR #4469, opened 2024-05-24
[WIP] Add limitation for RPC slow queries

chenyukang:

Currently, we have some RPC API from indexer such as `get_cells`, `get_transactions`, `get_cells_capacity` which may cost a lot of computing resources, it may occupy all the tokio tasks, and ckb can not response to other RPC APIs.

This PR try to fix this issue by adding limitation on computing resources for slow queries
- [add limited count iterator for indexer · nervosnetwork/ckb@a83d416](https://github.com/nervosnetwork/ckb/commit/a83d416f353278f47a8cebf0a90c64ff79ae81c0)
- [add slow\_query\_parallelism to limit slow query · nervosnetwork/ckb@a3bf751](https://github.com/nervosnetwork/ckb/commit/a3bf7519b839fde4857cf22105a5ff6b04fce7ec)
- TODO: limit the number of batch in JSONPRC.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/ckb-bin/src/subcommand/run.rs
+++ b/ckb-bin/src/subcommand/run.rs
@@ -5,7 +5,6 @@ use ckb_build_info::Version;
 use ckb_launcher::Launcher;
 use ckb_logger::info;
 use ckb_stop_handler::{broadcast_exit_signals, wait_all_ckb_services_exit};
-
 use ckb_types::core::cell::setup_system_cell_cache;
 
 pub fn run(args: RunArgs, version: Version, async_handle: Handle) -> Result<(), ExitCode> {
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -207,6 +207,7 @@ block_uncles_cache_size    = 30
 # cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
 # # The initial tip can be set higher than the current indexer tip as the starting height for indexing.
 # init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
+# iterator_next_limit = 100000
 #
 # # CKB rich-indexer has its unique configuration.
 # [indexer_v2.rich_indexer]
--- a/rpc/src/module/indexer.rs
+++ b/rpc/src/module/indexer.rs
@@ -1,3 +1,5 @@
+use std::sync::Arc;
+
 use crate::error::RPCError;
 use async_trait::async_trait;
 use ckb_indexer::IndexerHandle;
@@ -7,6 +9,7 @@ use ckb_jsonrpc_types::{
 };
 use jsonrpc_core::Result;
 use jsonrpc_utils::rpc;
+use tokio::sync::Semaphore;
 
 /// RPC Module Indexer.
 #[rpc(openrpc)]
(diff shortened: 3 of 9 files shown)


eval-exec (approved):

driftluo:
I think there is something wrong with the modification of the iterator. When the user's request value exceeds the default value, it should be forced to change to the upper limit and return the result instead of returning an error directly. This is also the case in eth filter-related RPCs.

Just change the following
```rust
let limit = std::cmp::min(limit.value() as usize, iterator_next_limit);
```

phroi:
@doitian as you are aware, on Testnet  I encountered an issue which renders a node unresponsive for a few minutes, which is very related to this PR underlying issue.

Is there any work being done or planned to address the underlying issue?

Phroi

P.S.: I don't mind adding as comment to this PR the exact description of the issue I stumbled upon.

chenyukang:
@phroi 
yes, you're welcome to add more details.

phroi:
Hey @chenyukang!! One month ago I reported this RPC call as DoS vector on Nervos Bounty without any remediation being started as a result:

```yaml
{
    "id": 42,
    "jsonrpc": "2.0",
    "method": "get_cells_capacity",
    "params": [
        {
            "script": {
                "code_hash": "0x9bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8",
                "hash_type": "type",
                "args": "0x"
            },
            "script_type": "lock"
        }
    ]
}
```

This RPC call matches a large share of Testnet live cells and it was randomly discovered by a LLM while writing tests, so it's pretty easy to stumble upon.

Repeated calls progressively increase node downtime, and attempting to terminate the RPC request does not stop the processing. This means that **any node exposing this method can be made unusable**.

The same issue doesn't appear on a Mainnet node, likely because far fewer live cells match that query. That said, in the future, as L1 state grows, it will impact Mainnet as well.

I can't stress enough the risk for any internet-exposed RPC nodes, yet there's no fundamental mitigation deployed nor documentation, and proposed workarounds are either impractical or [out of date](https://github.com/cryptape/ckb-nginx-proxy).

I would like to formally request the work on this PR to be resumed or at the very least its DoS-able nature to be clearly documented in the [RPC docs](https://github.com/nervosnetwork/ckb/tree/develop/rpc), so that practitioners can understand the trade-offs involved.

Love & Peace, Phroi

chenyukang:
@phroi 
Thanks for the details.

As for the solution to limit RPC cost too much server resources, we already provide some config parameters from PR:
https://github.com/nervosnetwork/ckb/pull/4576
https://github.com/nervosnetwork/ckb/pull/4529

You're right, it's still a potential issue in the long term, especially for RPC `get_cells_capacity`.

We will continue the work on this part to prevent DoS(Testnet need to be fixed), add the related document and Nginx configuration template.

phroi:
Hey @chenyukang!! Yeah, the linked PRs are steps in the right direction, but they do not seem to have effect on `get_cells_capacity`

> You're right, it's still a potential issue in the long term, especially for RPC get_cells_capacity.
>
> We will continue the work on this part to prevent DoS(Testnet need to be fixed), add the related document and Nginx configuration template.

That's exactly what I'm looking for, thank you so much!! 🥰

Additionally, I would evaluate if it makes sense to deprecate that method, cause `get_cells_capacity` results can be calculated from `get_cells`, which is already future-proofed.

BTW please, remember to tag me in the associated PRs, not that I'm qualified for reviewing CKB node code, but I'd like to keep an eye on how/if this issue is actually solved.

Love & Peace, Phroi

PR #1659, opened 2019-09-28
feat: Fee estimate RPC

jjyr:
This PR adds a new RPC [estimate_fee_rate](https://github.com/nervosnetwork/ckb/pull/1659/files#diff-622e6d119ac5d43f7eb41cb596159f9fR907). It takes the basic idea from bitcoin's [estimatesmartfee](https://bitcoincore.org/en/doc/0.16.0/rpc/util/estimatesmartfee/), however, we ignore the magic numbers and tricks from the original code.

We estimate the tx fee rate by track txs that entered tx pool. See details https://github.com/nervosnetwork/ckb/pull/1659/files#diff-ff03764a87b23e747dadc645fcf8df8bR21

PR #4477, opened 2024-06-04
feat: (experimental) optional fee estimator with different algorithms

chaoticlonghair:

> [!WARNING]
> This feature is experimental.
> I have done lots of manual tests before submit this PR, but more production environment tests are still required.

Add an optional fee estimator with different algorithms and a RPC method for it.

#### Description

- I do lots manual tests for several day, I don't sure which algorithm is better, so I just submit both of them.

  Let's make choice after collect more data in production environment (both mainnet and testnet).

- Only 4 levels of priorities for the fee estimator.

  Different algorithms have different APIs, I just use some constants as default parameters to avoid specialized parameters, to make the public API more generic.

Current, there are 2 built-in algorithms:

- `Confirmation Fraction`

  Migrated from [this CKB PR](https://github.com/nervosnetwork/ckb/pull/1659).

  👉🏼 Sometimes, it returns unreasonable result, I haven't figured it out yet, since I couldn't reproduce it manually.

- `Weight-Units Flow`

  A variant of the [Weight-Units Flow Fee Estimator for Bitcoin](https://bitcoiner.live/?tab=info).

  The main difference is I use the block interval to replace the time interval, to avoid unit conversion with random distributions, since our APIs require blocks as the target unit for estimation.

  The details of this algorithm are written in the head section of that source code file.

Merged fallback fee estimates algorithm from #4465.

#### Usage

Append the following content to `ckb.toml`:

```toml
[fee_estimator]
# Specifies the fee estimates algorithm. Current algorithms: ConfirmationFraction, WeightUnitsFlow.
algorithm = "WeightUnitsFlow"
```

And enable JSON-RPC API module `Experiment`.

After the CKB node started, try the following command:

```shell
echo '{ "id": 1, "jsonrpc": "2.0", "method": "estimate_fee_rate", "params": [] }' \
    | curl -s -H "Content-Type: application/json" -d @- "http://localhost:8114" \
    | jq
```

An example of result is:

```json
{
  "jsonrpc": "2.0",
  "result":"0x3e8",
  "id": 1
}
```

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


util/fee-estimator/src/estimator/confirmation_fraction.rs
```diff
@@ -0,0 +1,600 @@
...
+        if self.is_ready {
+            if in_ibd {
+                self.clear();
+                self.is_ready = false;
+            }
+        } else if !in_ibd {
+            self.clear();
+            self.is_ready = true;
+        }
+    }
```
eval-exec, line 452: `update_ibd_state` is executed every time ckb-chain finishes processing a block, is this the expected behavior?

Or should `update_ibd_state` only be executed when the `ibd_state` of ckb changes from `true` to `false`?

util/fee-estimator/src/estimator/weight_units_flow.rs
```diff
@@ -0,0 +1,463 @@
...
+        sorted_current_txs: &[TxStatus],
+    ) -> Result<Option<FeeRate>, Error> {
+        ckb_logger::debug!(
+            "boot: {}, current: {}, target: {target_blocks} blocks",
+            self.boot_tip,
+            self.current_tip,
+        );
+        let historical_blocks = Self::historical_blocks(target_blocks);
+        ckb_logger::debug!("required: {historical_blocks} blocks");
+        if historical_blocks > self.current_tip - self.boot_tip {
```
eval-exec, line 256: Is it possible for current_tip to be lower than boot_tip? How about using `saturating_sub` here?

util/fee-estimator/src/constants.rs
```diff
@@ -0,0 +1,24 @@
+//! The constants for the fee estimator.
+
+use ckb_chain_spec::consensus::{MAX_BLOCK_INTERVAL, MIN_BLOCK_INTERVAL, TX_PROPOSAL_WINDOW};
+use ckb_types::core::{BlockNumber, FeeRate};
+
+/// Average block interval.
+pub(crate) const AVG_BLOCK_INTERVAL: u64 = (MAX_BLOCK_INTERVAL + MIN_BLOCK_INTERVAL) / 2;
```
eval-exec, line 7: The `AVG_BLOCK_INTERVAL` is calculated as `(48 + 8) / 2`, which equals 28 seconds. However, according to https://explorer.nervos.org/charts/block-time-distribution, most block intervals are less than 10 seconds.

util/fee-estimator/src/constants.rs
```diff
@@ -0,0 +1,24 @@
...
+
+/// Max target blocks, about 1 hour.
+pub(crate) const MAX_TARGET: BlockNumber = (60 * 60) / AVG_BLOCK_INTERVAL;
+/// Min target blocks, in next block.
+pub(crate) const MIN_TARGET: BlockNumber = TX_PROPOSAL_WINDOW.closest() + 1;
+
+/// Lowest fee rate.
+pub(crate) const LOWEST_FEE_RATE: FeeRate = FeeRate::from_u64(1000);
```
eval-exec, line 15: Should we reuse `DEFAULT_MIN_FEE_RATE` from `ckb-app-config`?

util/fee-estimator/src/error.rs
```diff
@@ -0,0 +1,20 @@
...
+use thiserror::Error;
+
+/// A list specifying general categories of fee estimator errors.
+#[derive(Error, Debug, PartialEq)]
+pub enum Error {
+    /// Dummy fee estimator is used.
+    #[error("dummy fee estimator is used")]
+    Dummy,
+    /// Not ready for do estimate.
+    #[error("not ready")]
```
eval-exec, line 12: How about given a more verbose errro: `#[error("not ready, node is in IBD mode")]`

util/fee-estimator/src/estimator/confirmation_fraction.rs
eval-exec: I believe that when CKB enters IBD mode, `self.ibd_finished` is set to `true` only once. If CKB loses internet connection after 24 hours, `self.ibd_finished` will always remain =true=.

Cc. @driftluo

```rust
    /// Return whether chain is in initial block download
    pub fn is_initial_block_download(&self) -> bool {
        // Once this function has returned false, it must remain false.
        if self.ibd_finished.load(Ordering::Acquire) {
            false
        } else if unix_time_as_millis().saturating_sub(self.snapshot().tip_header().timestamp())
            > MAX_TIP_AGE
        {
            true
        } else {
            self.ibd_finished.store(true, Ordering::Release);
            false
        }
    }

```

util/fee-estimator/src/estimator/confirmation_fraction.rs
```diff
@@ -0,0 +1,552 @@
...
+struct TxConfirmStat {
+    min_fee_rate: FeeRate,
+    /// per bucket stat
+    bucket_stats: Vec<BucketStat>,
+    /// bucket upper bound fee_rate => bucket index
+    fee_rate_to_bucket: BTreeMap<FeeRate, usize>,
+    /// confirm_blocks => bucket index => confirmed txs count
+    confirm_blocks_to_confirmed_txs: Vec<Vec<f64>>,
+    /// confirm_blocks => bucket index => failed txs count
+    confirm_blocks_to_failed_txs: Vec<Vec<f64>>,
```
eval-exec, line 50: It seems that `confirm_blocks_to_failed_txs`'s length is always `1`?

quake (approved):

eval-exec (approved):

zhangsoledad (approved):

PR #4487, opened 2024-06-20
chore: tweak max_ancestors_count

zhangsoledad:
The current limitation of max_ancestors is too small and affects many application scenarios. After retesting, we have adjusted max_ancestors to a more appropriate yet safe value.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test

Side effects

- Performance regression

### Release note 

```release-note
```


eval-exec (approved):

quake (approved):

chenyukang:
I remember this case also need to fix: `test/src/specs/tx_pool/limit.rs`.

eval-exec (approved):

Issue #4520, opened 2024-07-16
High-Intensity JSON RPC BatchRequest Causes Server Crash and Unpredictable Downtime


Cupnfish:
### Current Behavior
When subjected to high-intensity JSON RPC BatchRequest, the CKB node causes the entire server to crash, rendering it completely inaccessible. The crash occurs after a certain batch size threshold is exceeded, and even after stopping the requests, the server will still crash at an unpredictable time in the future.

### Expected Behavior
I expect the node to send an error message to the client when it cannot handle high-intensity batch sizes, rather than attempting to process the requests and causing the server to crash. Ideally, the node should shut down to prevent affecting the entire server.

### Environment
- **CKB version**: ckb 0.116.1 
- **Chain**: testnet, mainnet (both have experienced the issue)
- **Operating system**: Ubuntu 20.04
- **Arch**: x64
- **Installation**: GitHub Release

### Additional context/Screenshots
The issue was first observed when my indexer node encountered the bug and stopped working. I had used high-frequency batch sizes for a few seconds before stopping the requests. Approximately two hours later, the server suddenly became unresponsive.

eval-exec:
Hello, What are the total memory, available memory, and CPU configuration of this server?

chenyukang:
I think this PR may help to resolve this issue:
https://github.com/nervosnetwork/ckb/pull/4459

chenyukang:
we have another working in progress PR try to limit the resource spent on heavy RPC:
https://github.com/nervosnetwork/ckb/pull/4469

Cupnfish:
@eval-exec 

The server's configuration is not critical. Previously, the server with lower specifications would crash when the batch size reached 1000. With the current server's improved specifications, it crashes when the batch size reaches 2000. The interface causing the crash is `get_block_by_number`.

Cupnfish:
Thank you for your efforts, @chenyukang ! We're eagerly looking forward to the launch of this feature to optimize resource usage. Currently, I've already determined the batch size that our server can tolerate, and to avoid any further crashes, I won't be conducting related tests for now. However, I believe this improvement will have a positive impact on our system, and I'm looking forward to the resolution of this issue.

eval-exec:
@Cupnfish Could you provide the output of `free -h`?

Cupnfish:
@eval-exec 
```
free -h
               total        used        free      shared  buff/cache   available
[...]

PR #4529, opened 2024-07-17
Add jsonrpc batch request limit

chenyukang:

Partially address #4520

Problem Summary:

User may send large batch jsonrpc request, if the request costs much memory it may makes the ckb server hang up.

We should provide a way to support user to config the batch request limit.

[jsonrpsee](https://github.com/paritytech/jsonrpsee/blob/7b87b641ed767d4ad0ff299d66c6369bb1bff3e5/server/src/server.rs#L1289-L1303) support this option, we can also support it with jsonrpc-utils.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


rpc/src/server.rs
```diff
@@ -22,6 +30,10 @@ use tokio_util::codec::{FramedRead, FramedWrite, LinesCodec, LinesCodecError};
 use tower_http::cors::CorsLayer;
 use tower_http::timeout::TimeoutLayer;
 
+const DEFAULT_BATCH_LIMIT: usize = 2000;
+
+static JSONRPC_BATCH_LIMIT: once_cell::sync::OnceCell<usize> = once_cell::sync::OnceCell::new();
```
eval-exec, line 35: Why do we need a `JSONRPC_BATCH_LIMIT`? Can we only use `DEFAULT_BATCH_LIMIT` instead?

eval-exec (dismissed):

rpc/src/server.rs
chenyukang: removed `DEFAULT_BATCH_LIMIT` now, as discussed, we don't add default limit for user.

rpc/Cargo.toml
```diff
@@ -52,6 +52,7 @@ async-stream = "0.3.3"
 ckb-async-runtime = { path = "../util/runtime", version = "= 0.117.0-pre" }
 # issue tracking: https://github.com/GREsau/schemars/pull/251
 schemars = { version = "0.8.19", package = "ckb_schemars" }
+once_cell = "1.19.0"
```
driftluo, line 55: use https://doc.rust-lang.org/nightly/std/sync/struct.OnceLock.html

zhangsoledad (approved):

PR #4291, opened 2024-01-04
New script verify with ckb-vm pause

chenyukang:
### What problem does this PR solve?

New script verify flow based on ckb-vm pause

### What is changed and how it works?

1. Use a job queue for pending transactions waiting for verifying
2. Multiple workers trigger the verification process by picking task from queue 
3. Verification is changed to `async` style
4. Use channel to resume/suspend vm
5. No snapshot from VM machines
6. No `Suspend` state from cache
7. All transactions from remote peer will be added into queue

What's Changed:

### Related changes

- PR to update `owner/repo`:
- Need to cherry-pick to the release branch

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- Performance regression
- Breaking backward compatibility

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4561, opened 2024-08-01
Recover possible transaction in conflicted cache when RBF

chenyukang:

When we replace some transactions in RBF, some old transactions may be okay to be recovered, since there are no conflicts after removing transactions. 

If we don't recover these transactions, it may introduce a scenario of a possible cycling attack in lightning network.

What's Changed:

Add `conflicts_outputs_cache` to track the relationship of `input -> tx` in RBF, then add possible transactions into the verify pool, so that they may be recovered.

This PR is based on the `new-verify` branch, one reason is we can not recursively call `submit_entry` in async. Another reason is `new-verify`'s worker-consumer model is more suitable for retrying submit transactions.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/pool.rs
+++ b/tx-pool/src/pool.rs
@@ -28,6 +28,7 @@ use std::sync::Arc;
 
 const COMMITTED_HASH_CACHE_SIZE: usize = 100_000;
 const CONFLICTES_CACHE_SIZE: usize = 10_000;
+const CONFLICTES_INPUTS_CACHE_SIZE: usize = 30_000;
 const MAX_REPLACEMENT_CANDIDATES: usize = 100;
 
 /// Tx-pool implementation
@@ -44,6 +45,8 @@ pub struct TxPool {
     pub(crate) expiry: u64,
     // conflicted transaction cache
     pub(crate) conflicts_cache: lru::LruCache<ProposalShortId, TransactionView>,
+    // conflicted transaction outputs cache, input -> tx_short_id
+    pub(crate) conflicts_outputs_cache: lru::LruCache<OutPoint, ProposalShortId>,
 }
 
 impl TxPool {
@@ -59,6 +62,7 @@ impl TxPool {
             recent_reject,
             expiry,
             conflicts_cache: LruCache::new(CONFLICTES_CACHE_SIZE),
+            conflicts_outputs_cache: lru::LruCache::new(CONFLICTES_INPUTS_CACHE_SIZE),
         }
     }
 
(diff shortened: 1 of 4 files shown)


chenyukang:
it is ready to be merged, since https://github.com/nervosnetwork/ckb/pull/4291 is merged.

tx-pool/src/pool.rs
```diff
@@ -59,6 +62,7 @@ impl TxPool {
             recent_reject,
             expiry,
             conflicts_cache: LruCache::new(CONFLICTES_CACHE_SIZE),
+            conflicts_outputs_cache: lru::LruCache::new(CONFLICTES_INPUTS_CACHE_SIZE),
```
quake, line 65: we may add two config options for `CONFLICTES_CACHE_SIZE` and `CONFLICTES_INPUTS_CACHE_SIZE`

quake (approved):

chenyukang:
Some tradeoff for current solution:

1. The current solution is when performing RBF (Replace-by-Fee) and replacing other transactions, we collect the inputs from the evicted transactions by the way(exclude the inputs used by the newly submitted transactions), and then check whether the existing conflicted pool contains any transactions spending those remaining inputs without other conflicts, and then add them back. Implementation is trivial here; the key is to avoid introducing performance regression.


2. Bitcoin's RBF has two modes discussed: one is Full-Replace-By-Fee, and the other is First-Seen-Safe, the latter mode can prevent this attack.
    1. First-Seen-Safe requires that the outputs of the replaced transaction must be a subset of the new transaction. This means that for a replacement to succeed, the new transaction cannot conflict with multiple transactions. Therefore, the step `as long as B2’s feerate is higher than B1's and A2's, it will evict B1 and A2 from the transaction pool` cannot be achieved, so the attack will not succeed.
    2. First-Seen-Safe is a more restrictive form of RBF.
    3. First-Seen-Safe RBF was previously discussed in Bitcoin, but it doesn’t seem to have been merged in the end, and the main reason was that no consensus was reached. Some maintainers preferred Full-RBF for simplicity:
        1. [PR #6176 on GitHub](https://github.com/bitcoin/bitcoin/pull/6176)
        2. [Discussion on Bitcoin Development](https://bitcoin-development.narkive.com/NgSiQBmM/elopment-first-seen-safe-replace-by-fee)
        3. [bitcoin wiki](https://en.bitcoin.it/wiki/Transaction_replacement)

PR #4562, opened 2024-08-01
fix: fix sync relayer collaboration

driftluo:

`HeaderProcess` uses the current snapshot to process the one-time message, During the entire processing process, the changes in the db itself are imperceptible. Normally, there is no problem here. But during this process, if a compact block message is received and processed successfully, the chain inserts a new block, and the block's header is in the previous header process message, there will be a concurrency problem.

I use a simple diagram to describe the time series problem here：

```bash
sync thread   |          process headers         |

relay thread    |  compact block process  |
```
The time series here is that the header process holds the old snapshot longer than the compact block process completion time.

This will cause the block status in the old snapshot to differ from the actual status.

https://github.com/nervosnetwork/ckb/blob/f3efbf24f893b9a45531a8863c549d70ec2f3d9b/sync/src/types/mod.rs#L2315-L2335

The actual status should be `BLOCK_VALID`, but because the block cannot be seen in the old snapshot, the returned status will be `UNKNOWN`

https://github.com/nervosnetwork/ckb/blob/f3efbf24f893b9a45531a8863c549d70ec2f3d9b/sync/src/synchronizer/headers_process.rs#L290-L336

The `UNKNOWN` status will cause the header to be reinserted into the header map, which will cause the block's status to become `HEADER_VALID`, the compact blocks received later will all be put into the orphan block pool due to this abnormal state. This will cause synchronization to pause until no compact block messages are received for a while. The sync protocol can then resume synchronization by re-downloading the abnormal block.

This PR forcefully reset the snapshot in get block status to obtain the latest status

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/sync/src/tests/sync_shared.rs
+++ b/sync/src/tests/sync_shared.rs
@@ -1,12 +1,15 @@
 use crate::block_status::BlockStatus;
+use crate::relayer::tests::helper::MockProtocolContext;
+use crate::relayer::CompactBlockProcess;
+use crate::synchronizer::HeadersProcess;
 use crate::tests::util::{build_chain, inherit_block};
-use crate::SyncShared;
+use crate::{Relayer, Status, SyncShared, Synchronizer};
 use ckb_chain::chain::ChainService;
 use ckb_shared::SharedBuilder;
 use ckb_store::{self, ChainStore};
 use ckb_test_chain_utils::always_success_cellbase;
 use ckb_types::core::Capacity;
-use ckb_types::prelude::*;
+use ckb_types::{packed, prelude::*};
 use std::sync::Arc;
 
 #[test]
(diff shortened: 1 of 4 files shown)


sync/src/tests/sync_shared.rs
```diff
@@ -183,3 +186,148 @@ fn test_switch_valid_fork() {
...
+    let status = header_process.execute();
+    assert!(status.is_ok());
+
+    assert_eq!(
+        sync_shared
+            .active_chain()
+            .get_block_status(&new_block.hash()),
+        BlockStatus::BLOCK_VALID
+    )
+}
```
eval-exec, line 247: Suggest let CompactBlockProcess process more child blocks here.

driftluo: `collaboration2` use two child to test it

sync/src/types/mod.rs
```diff
@@ -1506,11 +1506,19 @@ impl SyncShared {
...
-        self.state.may_set_shared_best_header(header_view);
+        // This judgment assumes that the header processing is continuous:
+        //
+        // 1. The header process ensures that header processing is continuous.
+        // 2. When a block jump(can't find its parents) occurs, the compact block
+        //    will skip the processing of the block and use the get header message first.
+        // 3. If there is no block skipping, compact block processing is also sequential
+        if self.state.may_set_shared_best_header(header_view.clone()) {
+            self.state.header_map.insert(header_view);
+        }
```
eval-exec, line 1521: suggest add info level log in else.

driftluo: According to the ckb block difficulty algorithm, any uncle block will trigger this else branch

eval-exec (approved):

PR #4576, opened 2024-08-09
fix: add limit to get cells

### LINKED ISSUE #4520: High-Intensity JSON RPC BatchRequest Causes Server Crash and Unpredictable Downtime

driftluo:

close #4520 related https://github.com/serde-rs/json/issues/635

A large number of data requests will consume more than 10x the memory when serialized in JSON, resulting in huge memory overhead and affecting the operation of the server

We recommend setting the memory usage limit at 2 GB, meaning only 200 MB of data can be requested at a time. The maximum cell data size is 500 KB, and 400 cells can be requested at one time.

In addition, we strongly recommend using this configuration #4529 to limit the upper limit of batch requests.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/resource/ckb.toml
+++ b/resource/ckb.toml
@@ -212,6 +212,12 @@ block_uncles_cache_size    = 30
 # cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
 # # The initial tip can be set higher than the current indexer tip as the starting height for indexing.
 # init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
+# By default, there is no limitation on the size of indexer request
+# However, because serde json serialization consumes too much memory(10x),
+# it may cause the physical machine to become unresponsive.
+# We recommend a consumption limit of 2g, which is 400 as the limit,
+# which is a safer approach
+# request_limit = 400
 #
 # # CKB rich-indexer has its unique configuration.
 # [indexer_v2.rich_indexer]
--- a/util/app-config/src/configs/indexer.rs
+++ b/util/app-config/src/configs/indexer.rs
@@ -35,6 +35,9 @@ pub struct IndexerConfig {
     /// The init tip block hash
     #[serde(default)]
     pub init_tip_hash: Option<H256>,
+    /// limit of indexer reqeust
+    #[serde(default)]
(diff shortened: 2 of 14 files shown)


eval-exec (approved):

resource/ckb.toml
```diff
@@ -212,6 +212,12 @@ block_uncles_cache_size    = 30
 # cell_filter = "let script = output.type;script!=() && script.code_hash == \"0x00000000000000000000000000000000000000000000000000545950455f4944\""
 # # The initial tip can be set higher than the current indexer tip as the starting height for indexing.
 # init_tip_hash = "0x8fbd0ec887159d2814cee475911600e3589849670f5ee1ed9798b38fdeef4e44"
+# By default, there is no limitation on the size of indexer request
+# However, because serde json serialization consumes too much memory(10x),
+# it may cause the physical machine to become unresponsive.
+# We recommend a consumption limit of 2g, which is 400 as the limit,
+# which is a safer approach
+# request_limit = 400
```
chenyukang, line 220: `indexer_request_limit` maybe a better name, as it's limiting the indexer request.

driftluo: This configuration item is under the indexer module. I don’t think it’s necessary to emphasize that it is the indexer request

eval-exec:
@driftluo Please rebase develop

eval-exec (approved):

eval-exec (approved):

PR #4578, opened 2024-08-12
Integration Test: Put SyncChurn to last

eval-exec:
### Related changes

- Put SyncChurn to last in integration test

### Check List 

Tests 

Side effects

- None

### Release note 

```release-note
```


test/src/main.rs
```diff
@@ -591,6 +590,8 @@ fn all_specs() -> Vec<Box<dyn Spec>> {
         Box::new(CheckVmVersion2),
         Box::new(CheckVmBExtension),
         Box::new(RandomlyKill),
+        // SyncChurn : This is a CPU-intensive operation, so we place it last.
```
chenyukang, line 593: I think the run order is random?

eval-exec: :thinking:

PR #4588, opened 2024-08-14
IntegrationTest: Let RandomKill and SyncChurn run in sequence

eval-exec:
### Related changes

- Integration Test: let RandomlyKill and SyncChurn run in sequnece

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -148,11 +148,17 @@ fn main() {
     let mut test_results = Vec::new();
     let mut worker_running = worker_count;
     let mut done_specs = 0;
+    let mut started_sequential = false;
     while worker_running > 0 {
         if max_time > 0 && start_time.elapsed().as_secs() > max_time {
             // shutdown, specs running to long
             workers.shutdown();
             break;
+        }
+
+        if worker_running == 1 && !started_sequential {
+            started_sequential = true;
+            workers.start_sequencial()
         }
 
         let msg = match notify_rx.recv_timeout(Duration::from_secs(5)) {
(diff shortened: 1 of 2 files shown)


test/src/main.rs
```diff
@@ -148,13 +148,19 @@ fn main() {
...
     let mut done_specs = 0;
+    let mut started_sequential = false;
     while worker_running > 0 {
         if max_time > 0 && start_time.elapsed().as_secs() > max_time {
             // shutdown, specs running to long
             workers.shutdown();
             break;
         }
 
+        if worker_running == 1 && !started_sequential {
```
chenyukang, line 159: This seems the last spec job may still have overlapping with the sequential spec job, suppose the `msg` is `Notify::Start`, means last job is starting executing, and we start `workers.start_sequencial()`, one sequential job will start, so there is a overlapping.

chenyukang: add a check `&& msg != Notify::Start` ?

eval-exec: > This seems the last spec job may still have overlapping with the sequential spec job, suppose the `msg` is `Notify::Start`, means last job is starting executing, and we start `workers.start_sequencial()`, one sequential job will start, so there is a overlapping.

`workers.start_sequencial()` will only be executed once.

chenyukang: I know it will be executed once, I mean when `worker_running == 1 && !started_sequential` is true, msg maybe `Notify::Start`, the last spec job is starting to run, not mean the last job has just finished, if we start sequential job at this time, there maybe a overlapping with last job?

chenyukang: seems it's ok, `workers.start_sequencial()` only set a flag.

eval-exec: When `worker_running == 1`, only one worker (the sequential worker) should be running, and all other parallel workers should have exited.

> `msg` could be `Notify::Start`
If `msg` is `Notify::Start`, the sequential worker might be executing `self.run_spec(spec.as_ref(), 0);` at that moment.
A worker can execute only one spec at a time.
Therefore, there is no chance of overlap.

chenyukang: the last running worker must be the `sequencial_worker`...

chenyukang (approved):

PR #4608, opened 2024-08-23
ckb-metrics: set ckb_chain_tip on startup

eval-exec:
### Related changes

- Set metrics::ckb_chain_tip on startup

### Check List 

Tests 

Side effects

- None

### Release note 

```release-note
```


chenyukang:
what is the issue we don't set ckb_chain_tip on startup?
please add more details in the PR description.

PR #4663, opened 2024-09-25
chore: remove deny with sqlx

driftluo:

remove deny with sqlx

By the way, I think that SQL creates tables and indexes that can use `IF NOT EXISTS` to avoid judging whether it has been initialized. Although it is a non-standard statement, but both SQLite and Postgresql support it.

https://github.com/nervosnetwork/ckb/blob/21f49f23a20dfb329638ffb13fa020e069b987c3/util/rich-indexer/resources/create_postgres_table.sql#L1

https://github.com/nervosnetwork/ckb/blob/21f49f23a20dfb329638ffb13fa020e069b987c3/util/rich-indexer/src/store.rs#L68

https://github.com/nervosnetwork/ckb/blob/21f49f23a20dfb329638ffb13fa020e069b987c3/util/rich-indexer/src/store.rs#L77-L79

https://github.com/nervosnetwork/ckb/blob/21f49f23a20dfb329638ffb13fa020e069b987c3/util/rich-indexer/src/store.rs#L251-L266

https://github.com/nervosnetwork/ckb/blob/21f49f23a20dfb329638ffb13fa020e069b987c3/util/rich-indexer/src/store.rs#L204-L221

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/util/rich-indexer/src/indexer/remove.rs
+++ b/util/rich-indexer/src/indexer/remove.rs
@@ -76,7 +76,7 @@ async fn remove_batch_by_blobs(
 
     // execute
     query
-        .execute(tx)
+        .execute(tx.as_mut())
         .await
         .map_err(|err| Error::DB(err.to_string()))?;
 
@@ -98,7 +98,7 @@ async fn reset_spent_cells(tx_id_list: &[i64], tx: &mut Transaction<'_, Any>) ->
         .map_err(|err| Error::DB(err.to_string()))?;
 
     sqlx::query(&query)
-        .execute(&mut *tx)
+        .execute(tx.as_mut())
         .await
         .map_err(|err| Error::DB(err.to_string()))?;
 
@@ -117,7 +117,7 @@ async fn query_uncle_id_list_by_block_id(
             "#,
     )
     .bind(block_id)
-    .fetch_all(tx)
+    .fetch_all(tx.as_mut())
     .await
     .map(|rows| rows.into_iter().map(|row| row.get("uncle_id")).collect())
     .map_err(|err| Error::DB(err.to_string()))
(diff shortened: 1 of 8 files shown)


zhangsoledad (approved):

eval-exec (dismissed):

zhangsoledad (approved):

util/rich-indexer/src/indexer/remove.rs
```diff
@@ -229,11 +229,11 @@ async fn script_exists_in_output(
         "#,
     )
     .bind(script_id)
-    .fetch_one(&mut *tx)
+    .fetch_one(tx.as_mut())
     .await
     .map_err(|err| Error::DB(err.to_string()))?;
 
-    Ok(row_lock.get::<bool, _>(0) || row_type.get::<bool, _>(0))
+    Ok(row_lock.get::<i64, _>(0) == 1 || row_type.get::<i64, _>(0) == 1)
```
eval-exec, line 236: Perhaps we can return early with =Ok(row_lock.get::<i64, _>(0) == 1)= after immediately fetching =row_lock=?

eval-exec (approved):

util/rich-indexer/src/indexer/remove.rs
```diff
@@ -215,7 +215,7 @@ async fn script_exists_in_output(
         "#,
     )
     .bind(script_id)
-    .fetch_one(&mut *tx)
+    .fetch_one(tx.as_mut())
     .await
     .map_err(|err| Error::DB(err.to_string()))?;
```
eval-exec, line 221: How about move `row_lock.get::<i64, _>(0) == 1` checking in this line?

PR #4676, opened 2024-10-14
Remove empty entry for `OrphanPool`

eval-exec:
When an entry in `OrphanPool` is empty, it should be removed.

What's Changed:

### Related changes

- Remove empty entry in `OrphanPool`

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/tx-pool/src/component/orphan.rs
+++ b/tx-pool/src/component/orphan.rs
@@ -75,9 +75,13 @@ impl OrphanPool {
         self.entries.remove(id).map(|entry| {
             debug!("remove orphan tx {}", entry.tx.hash());
             for out_point in entry.tx.input_pts_iter() {
-                self.by_out_point
-                    .get_mut(&out_point)
-                    .map(|set| set.remove(id));
+                if let Some(ids_set) = self.by_out_point.get_mut(&out_point) {
+                    ids_set.remove(id);
+
+                    if ids_set.is_empty() {
+                        self.by_out_point.remove(&out_point);
+                    }
+                }
             }
             entry
         })


tx-pool/src/process.rs
```diff
@@ -751,6 +750,10 @@ impl TxPoolService {
 
         if verify_cache.is_none() {
             // update cache
+            let TransactionView {
```
doitian, line 753: how about just `drop(tx)`;

eval-exec: OK, I will try `drop(tx)`.

tx-pool/src/component/orphan.rs
```diff
@@ -78,6 +78,11 @@ impl OrphanPool {
                 self.by_out_point
                     .get_mut(&out_point)
                     .map(|set| set.remove(id));
+
+                // remove empty by_out_point entry
```
doitian, line 82: I think you should test `self.by_out_point[out_point]` is empty.

eval-exec: Yes, I made a mistake.

chenyukang (dismissed):

util/types/src/core/views.rs
```diff
@@ -298,6 +298,15 @@ macro_rules! define_inner_getter {
 }
 
 impl TransactionView {
+    /// Destruct TransactionView to get wtx_hash
+    pub fn destruct_wtx_hash(self) -> packed::Byte32 {
+        let TransactionView {
+            witness_hash: wtx_hash,
+            ..
+        } = self;
+        wtx_hash
```
doitian, line 307: Not sure but `self.wtx_hash` should just work

doitian, line 307: Not sure but `self.witness_hash` should just work

```suggestion
        self.witness_hash
```

eval-exec: Not sure too. @quake

eval-exec: We have confirmed that holding a witness_hash will not prevent the release of TransactionView.

zhangsoledad (approved):

PR #4683, opened 2024-10-22
feat: small wasm support

driftluo:

After some effort, the ckb-light-client project has been verified and run on the browser. It can perform basic queries and responses and can be embedded in any web page. The binary size is about 20M, but there is still some work to be verified. This PR is the basic support for WASM.

wait https://github.com/nervosnetwork/tentacle/pull/375

Issues to discuss：
- ~I disable the peer-store dump on WASM now~ load/dump peer store on wasm
- I unimplemented soft-fork version bit get/insert on WASM now
- ~I disable the network key persistence on WASM now~ export fn to set network key, if not set, return error

### Check List

Tests 

- Unit test
- Integration test

### Release note

```release-note
```


network/src/network.rs
```diff
@@ -1419,3 +1418,22 @@ pub(crate) async fn async_disconnect_with_message(
...
+
+    iter.find_map(|proto| {
+        if let p2p::multiaddr::Protocol::Ws = proto {
+            Some(TransportType::Ws)
+        } else {
+            None
+        }
+    })
+    .unwrap_or(TransportType::Tcp)
+}
```
quake, line 1439:
```suggestion
pub fn find_type(addr: &Multiaddr) -> TransportType {
    if addr.iter().any(|proto| matches!(proto, Protocol::Ws)) {
        TransportType::Ws
    } else {
        TransportType::Tcp
    }
}
```

network/src/peer_store/peer_store_impl.rs
```diff
@@ -165,20 +170,33 @@ impl PeerStore {
...
+        let filter = |peer_addr: &AddrInfo| {
+            extract_peer_id(&peer_addr.addr)
+                .map(|peer_id| !peers.contains_key(&peer_id))
+                .unwrap_or_default()
+                && peer_addr
+                    .connected(|t| t > addr_expired_ms && t <= now_ms.saturating_sub(DIAL_INTERVAL))
+                && required_flags_filter(required_flags, Flags::from_bits_truncate(peer_addr.flags))
+                && matches!(find_type(&peer_addr.addr), TransportType::Ws)
+        };
+
```
quake, line 197: suggest to move this code block to the fn `addr_manager#fetch_random`

driftluo: These filter conditions are not the same:
1. Get the feeler detection address list, because it ensures that the ws/tcp protocol addresses in the peer store can be updated normally
2. Get the identify attempt to connect to the address list formally, native only needs tcp, wasm only needs ws
3. Get the discovery propagation address list to ensure that any protocol is propagated

quake: the only diff is 
```
#[cfg(not(target_family = "wasm"))]
&& !matches!(find_type(&peer_addr.addr), TransportType::Ws
#[cfg(target_family = "wasm")]
&& matches!(find_type(&peer_addr.addr), TransportType::Ws)
```

other filter condition are same, we may move it to fetch_random, like:

```
  filter.filter(
#[cfg(not(target_family = "wasm"))]
&& !matches!(find_type(&peer_addr.addr), TransportType::Ws
#[cfg(target_family = "wasm")]
&& matches!(find_type(&peer_addr.addr), TransportType::Ws)
)
```

driftluo: `fetch_addrs_to_attempt` on native need `!matches!(find_type(&peer_addr.addr), TransportType::Ws)`, on WASM need `matches!(find_type(&peer_addr.addr), TransportType::Ws)`

The other two can be combined, but this function should not work.

spec/Cargo.toml
```diff
@@ -25,6 +25,8 @@ ckb-error = { path = "../error", version = "= 0.120.0-pre" }
 ckb-traits = { path = "../traits", version = "= 0.120.0-pre" }
 ckb-logger = { path = "../util/logger", version = "= 0.120.0-pre" }
 
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+cacache = { version = "13.0.0", default-features = false, features = ["mmap"] }
```
eval-exec, line 29: It seems that this code is same as L14.

driftluo: rebase issue..

eval-exec (approved):
Do we need to specify a new release target in `.github/workflows/package.yaml`?

driftluo:
> Do we need to specify a new release target in `.github/workflows/package.yaml`?

I don't think it's necessary, ckb itself doesn't release a wasm version

network/src/services/dump_peer_store.rs
```diff
@@ -25,16 +27,24 @@ impl DumpPeerStoreService {
         }
     }
 
+    #[cfg(not(target_family = "wasm"))]
     fn dump_peer_store(&self) {
         let path = self.network_state.config.peer_store_path();
         self.network_state.with_peer_store_mut(|peer_store| {
             if let Err(err) = peer_store.dump_to_dir(&path) {
-                warn!("Dump peer store error, path: {:?} error: {}", path, err);
+                ckb_logger::warn!("Dump peer store error, path: {:?} error: {}", path, err);
```
eval-exec, line 35: How about remove `ckb_logger::`

network/src/services/outbound_peer.rs
```diff
@@ -155,14 +155,20 @@ impl Future for OutboundPeerService {
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
eval-exec, line 155: This PR splited the implementation of `DumpPeerStoreService->poll`, but does't split the implementation of `OutboundPeerService->poll` for `wasm` and `not(wasm)`. Is this expected?

util/systemtime/src/lib.rs
```diff
@@ -4,7 +4,10 @@ mod test_realtime;
 
 #[cfg(feature = "enable_faketime")]
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
-use std::time::Duration;
+#[cfg(not(target_family = "wasm"))]
+pub use std::time::{Duration, Instant, SystemTime};
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
```
eval-exec, line 9: Why `target_os = "unknown"`.

network/src/services/outbound_peer.rs
driftluo: yes, it seems github ci win/macos has some issue on disk，if the DumpPeerStoreService poll doesn't split with `wasm` and `not(wasm)`, it will fail on ci.

The difference between their implementations is whether to perform a dump operation immediately.

util/systemtime/src/lib.rs
driftluo: `web-time` doesn't support `wasm32-wasi*`, it only support `wasm32-unknown-*` on browser

util/runtime/src/brower.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+impl Spawn for Handle {
+    fn spawn_task<F>(&self, future: F)
+    where
+        F: Future<Output = ()> + 'static,
+    {
+        self.spawn(future);
+    }
+}
```
eval-exec, line 27: In `native.rs`, `F` have `Send`:
```rust

impl Spawn for Handle {
    fn spawn_task<F>(&self, future: F)
    where
        F: Future<Output = ()> + Send + 'static,
    {
        self.spawn(future);
    }
}
```
but in `browser.rs` F doesn't have `Send`, is this expected?

driftluo: Yes, wasm does not support multi-threading in browsers

eval-exec (dismissed):

util/rich-indexer/src/indexer_handle/async_indexer_handle/get_cells.rs
```diff
@@ -190,7 +190,7 @@ impl AsyncRichIndexerHandle {
             if let Some(script) = filter.script.as_ref() {
                 query = query
                     .bind(script.code_hash.as_bytes())
-                    .bind(script.hash_type.clone() as i16);
+                    .bind(script.hash_type as i16);
```
quake, line 193: Is this change related to wasm?

util/runtime/src/lib.rs
```diff
@@ -1,159 +1,15 @@
...
-impl Spawn for Handle {
-    fn spawn_task<F>(&self, future: F)
-    where
-        F: Future<Output = ()> + Send + 'static,
-    {
-        self.spawn(future);
-    }
-}
+#[cfg(target_family = "wasm")]
+mod brower;
```
quake, line 13: typo? browser

util/rich-indexer/src/indexer_handle/async_indexer_handle/get_cells.rs
quake: please move unrelated change to another PR

network/src/services/dump_peer_store.rs
quake, line 35: why use fullpath here and remove the warn from `use` in line 2?

util/types/src/core/cell.rs
```diff
@@ -165,7 +165,7 @@ impl CellMeta {
 }
 
 /// TODO(doc): @quake
-#[derive(PartialEq, Debug, Eq)]
+#[derive(PartialEq, Debug, Eq, Clone)]
```
quake, line 168: unrelated change to wasm?

resource/src/lib.rs
```diff
@@ -195,6 +208,9 @@ impl Resource {
                 SourceFiles::new(&BUNDLED_CELL, &BUNDLED).read(bundled)
             }
             Resource::FileSystem { file } => Ok(Box::new(BufReader::new(fs::File::open(file)?))),
+            Resource::Raw { raw } => Ok(Box::new(BufReader::new(Cursor::new(
```
quake, line 211: cursor has wrapped in-memory buffer already:
```suggestion
Resource::Raw { raw } => Ok(Box::new(Cursor::new(raw.clone()))),
```

network/src/services/dump_peer_store.rs
driftluo: wasm target will warn unused import `warn!`

util/types/src/core/cell.rs
driftluo: https://github.com/driftluo/ckb-light-client/blob/wasm-support-with-indexdb/light-client-lib/src/storage/mod.rs#L159-L164
impl this must have

eval-exec (approved):

PR #4699, opened 2024-10-28
Update hardcode assume valid target

eval-exec:
Update hardcode assume valid target
### Related changes

- Update hardcode assume valid target

### Check List 

Tests 

Side effects

- None

### Release note 

```release-note
```


doitian:
Merged back the assume valid target in the latest release

PR #4709, opened 2024-11-12
fix: fix pg sqlite inconsistent types

driftluo:

Fix inconsistent mapping of PG Sqlite types in rust

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


eval-exec (approved):
I'm not sure under what circumstances a 'pg_type' would be classified as BOOLEAN. But approve.

PR #4741, opened 2024-12-05
feat: let old config node open ws listen by default

driftluo:

Added `reuse_tcp_with_ws` option, the default value is true, so that all nodes upgraded to this version will enable the TCP + WS listening mode by default

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


eval-exec (approved):
How about use a big Braces to include L1166 to L1220
```rust
#[cfg(not(target_family = "wasm"))]
{ 
   L1166 -> L1220
   ...
   ...
}

File: network/src/network.rs [hunk 1]
```diff
@@ -1163,30 +1163,63 @@ impl NetworkService {
...
+                                        .expect("Tcp listen must exsit")
+                                        .clone();
+                                    addr.push(Protocol::Ws);
+                                    addr
+                                };
+                                addresses.push(ws_listen);
+                            }
+
+                            addresses
+                        } else {
```
eval-exec, line 1193: I think we can delete these 3 lines.

eval-exec (approved):

zhangsoledad (approved):

zhangsoledad (approved):

PR #4733, opened 2024-12-02
[WIP] Add proxy support for ckb

eval-exec:
- **Using proxy version of tentacle**
- **Add proxy support for ckb**

### What problem does this PR solve?

Issue Number: close #4473

### What is changed and how it works?

What's Changed:

### Related changes

- Add `ProxyConfig` to network config

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- None

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #4759, opened 2024-12-25
`IdentifyMessage::decode` allow failure on Unsupported MultiAddr

eval-exec:

Problem Summary:

When a remote peer's IdentifyMessage contains 1 multiaddr which cannot parse successfully, should ignore that one, instead return an empty public_addrs.

What's Changed:

### Related changes

- let `IdentifyMessage::decode` allow parse failure on Unsupported MultiAddr

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


eval-exec:
This changed should be added in https://github.com/nervosnetwork/ckb/pull/4733

PR #4760, opened 2024-12-25
peer-store, wasm: use `path` as database name in wasm peer-store implementation

Officeyutong:

In the original peer store implementation of wasm , there is only one database (in IndexedDB) to store all peers, different network configurations are distinguished by store names, this can cause one issue: There is no way to use multiple configurations after the database was created, since IndexedDB disallows creating stores unless there is an upgrade for a certain database. 

What's Changed:
This PR updates peer store implementation on wasm, to use database names to distinguish different configurations, and database names comes from `path` section in network configuration, so makes it possible to switch network configurations in the same website

### DIFF
--- a/network/src/peer_store/browser.rs
+++ b/network/src/peer_store/browser.rs
@@ -59,6 +59,8 @@ pub async fn get_db<P: AsRef<Path>>(path: P) -> &'static Storage {
     DB.get_or_init(|| Storage::new(path)).await
 }
 
+const STORE_NAME: &str = "main-store";
+
 #[derive(Clone)]
 pub struct Storage {
     chan: tokio::sync::mpsc::Sender<Request>,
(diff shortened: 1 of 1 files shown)


network/src/peer_store/browser.rs
```diff
@@ -67,15 +69,14 @@ pub struct Storage {
 impl Storage {
     pub async fn new<P: AsRef<Path>>(path: P) -> Self {
         let factory = Factory::new().unwrap();
-        let mut open_request = factory.open("network", Some(1)).unwrap();
-        let store_name = path.as_ref().to_str().unwrap().to_owned();
-        let store_name_clone = store_name.clone();
+        let database_name = path.as_ref().to_str().unwrap().to_owned();
+        let mut open_request = factory.open(&database_name, Some(1)).unwrap();
```
eval-exec, line 73: There is a `secret_key` file in `network` dir. If the secret key in CKB's network directory changes, it may cause the `NodeId` to change as well. Is this expected?

driftluo: No, wasm target‘s secret_key isn't stored on this path. The user manually enters it

eval-exec (approved):

network/src/peer_store/browser.rs
```diff
@@ -67,15 +69,14 @@ pub struct Storage {
 impl Storage {
     pub async fn new<P: AsRef<Path>>(path: P) -> Self {
```
driftluo, line 70: The path needs to be added with the network name. No matter what network is passed in, the same path is used.

Officeyutong: - For the first question, the user can customize store path in network configuration to make different network use different databases
- For the second, IndexedDB limits store creation only to the time of upgrading database, if we want to use different stores for different configurations, we'll need a suitable way to determine database version

driftluo (approved):

PR #4783, opened 2025-01-15
Peer store optimization

driftluo:

1. peer store only retains TCP multiaddr
2. peer store support store DNS
3. `OutboundPeerService` specifying the transport type, In wasm, user can specify ws or wss for dial operation
4. identify/discovery allows no tcp/ip multiaddr spread
5. public_listens allow no tcp/ip multiaddr
6. Keep connected nodes up to date in the peer store
7. Fix feeler flags fetch fail

### Check List 

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -1531,27 +1531,13 @@ pub enum TransportType {
     Wss,
 }
 
-impl<'a> From<TransportType> for p2p::multiaddr::Protocol<'a> {
-    fn from(value: TransportType) -> Self {
-        match value {
-            TransportType::Ws => Protocol::Ws,
-            TransportType::Wss => Protocol::Wss,
-            _ => unreachable!(),
-        }
-    }
-}
-
 pub(crate) fn find_type(addr: &Multiaddr) -> TransportType {
     let mut iter = addr.iter();
 
-    iter.find_map(|proto| {
-        if let Protocol::Ws = proto {
-            Some(TransportType::Ws)
-        } else if let Protocol::Wss = proto {
-            Some(TransportType::Wss)
-        } else {
-            None
-        }
+    iter.find_map(|proto| match proto {
+        Protocol::Ws => Some(TransportType::Ws),
+        Protocol::Wss => Some(TransportType::Wss),
+        _ => None,
     })
     .unwrap_or(TransportType::Tcp)
 }
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -105,15 +105,14 @@ impl NetworkState {
             .iter()
             .chain(config.public_addresses.iter())
             .cloned()
-            .filter_map(|mut addr| {
-                multiaddr_to_socketaddr(&addr)
-                    .filter(|addr| is_reachable(addr.ip()))
-                    .and({
-                        if extract_peer_id(&addr).is_none() {
-                            addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
-                        }
-                        Some(addr)
-                    })
+            .filter_map(|mut addr| match multiaddr_to_socketaddr(&addr) {
+                Some(socket_addr) if !is_reachable(socket_addr.ip()) => None,
+                _ => {
+                    if extract_peer_id(&addr).is_none() {
+                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
+                    }
+                    Some(addr)
+                }
             })
             .collect();
         info!("Loading the peer store. This process may take a few seconds to complete.");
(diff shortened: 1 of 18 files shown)


network/src/network.rs
```diff
@@ -1521,18 +1523,33 @@ pub(crate) async fn async_disconnect_with_message(
...
-    {
-        TransportType::Ws
-    } else {
-        TransportType::Tcp
+impl<'a> From<TransportType> for p2p::multiaddr::Protocol<'a> {
+    fn from(value: TransportType) -> Self {
+        match value {
+            TransportType::Ws => Protocol::Ws,
+            TransportType::Wss => Protocol::Wss,
+            _ => unreachable!(),
```
eval-exec, line 1537: Is this possible that a match arm is a `TransportType::Tcp`?

driftluo: First, there is no such scenario in the current code. 
Second, it is impossible to convert here because TCP needs to contain port information, but the type cannot know it.

This conversion is only performed when OutboundPeerService dial

network/src/services/outbound_peer.rs
```diff
@@ -63,8 +70,15 @@ impl OutboundPeerService {
...
-        for addr in attempt_peers.into_iter().map(|info| info.addr) {
-            self.network_state.dial_feeler(&self.p2p_control, addr);
+        for mut addr in attempt_peers.into_iter().map(|info| info.addr) {
+            self.network_state.dial_feeler(&self.p2p_control, {
+                if !matches!(self.transport_type, TransportType::Tcp) {
+                    addr.push(self.transport_type.into());
+                    addr
+                } else {
+                    addr
+                }
```
eval-exec, line 80: How about:
```suggestion
                if !matches!(self.transport_type, TransportType::Tcp) {
                    addr.push(self.transport_type.into());
                }
                addr
```

eval-exec (dismissed):

network/src/network.rs
```diff
@@ -105,15 +105,20 @@ impl NetworkState {
...
+                    }
+                    Some(addr)
+                }
+                None => {
+                    if extract_peer_id(&addr).is_none() {
+                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
+                    }
+                    Some(addr)
+                }
+                Some(_) => None,
```
eval-exec, line 121: How about:
```rust
            .filter_map(|mut addr| match multiaddr_to_socketaddr(&addr) {
                Some(socket_addr) if !is_reachable(socket_addr.ip()) => None,
                _ => {
                    if extract_peer_id(&addr).is_none() {
                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
                    }
                    Some(addr)
                }
            })
```

eval-exec (dismissed):

network/src/network.rs
quake: If it is only used inside the OutboundPeerService dial fn, it is recommended to use a more intuitive way to write the conversion in the dial fn:
```rust
        for mut addr in peers {
            match &self.transport_type {
                TransportType::Tcp => (),
                TransportType::Ws => addr.push(Protocol::Ws),
                TransportType::Wss => addr.push(Protocol::Wss)
            }

            self.network_state.dial_identify(&self.p2p_control, addr);
        }
```

network/src/network.rs
```diff
@@ -1520,19 +1520,38 @@ pub(crate) async fn async_disconnect_with_message(
...
+    iter.find_map(|proto| {
+        if let Protocol::Ws = proto {
+            Some(TransportType::Ws)
+        } else if let Protocol::Wss = proto {
+            Some(TransportType::Wss)
+        } else {
+            None
+        }
+    })
+    .unwrap_or(TransportType::Tcp)
```
quake, line 1556:
```suggestion
    addr.iter().find_map(|proto| match proto {
        Protocol::Ws => Some(TransportType::Ws),
        Protocol::Wss => Some(TransportType::Wss),
        _ => None,
    }).unwrap_or(TransportType::Tcp)
```

network/src/peer_registry.rs
```diff
@@ -191,10 +192,30 @@ impl PeerRegistry {
...
+                None => false,
+            })
+            .unwrap()
+    }
+
+    /// Get feeler session flags
+    pub fn feeler_flags(&self, addr: &Multiaddr) -> Option<Flags> {
+        extract_peer_id(addr)
+            .map(|peer_id| self.feeler_peers.get(&peer_id).cloned())
+            .flatten()
```
quake, line 216:
```suggestion
extract_peer_id(addr).and_then(|peer_id| self.feeler_peers.get(&peer_id).cloned())
```

network/src/peer_registry.rs
```diff
@@ -191,10 +192,30 @@ impl PeerRegistry {
...
+        extract_peer_id(addr)
+            .map(|peer_id| match self.feeler_peers.get_mut(&peer_id) {
+                Some(i) => {
+                    *i = flags;
+                    true
+                }
+                None => false,
+            })
+            .unwrap()
+    }
```
quake, line 210: remove unwrap
```suggestion
    if let Some(peer_id) = extract_peer_id(addr) {
        if let Some(i) = self.feeler_peers.get_mut(&peer_id) {
            *i = flags;
            return true;
        }
    }
    false
```

gpBlockchain:
LGTM

eval-exec (approved):

network/src/network.rs
```diff
@@ -105,15 +105,14 @@ impl NetworkState {
...
-                        Some(addr)
-                    })
+            .filter_map(|mut addr| match multiaddr_to_socketaddr(&addr) {
+                Some(socket_addr) if !is_reachable(socket_addr.ip()) => None,
+                _ => {
+                    if extract_peer_id(&addr).is_none() {
+                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
+                    }
+                    Some(addr)
+                }
```
eval-exec, line 115: I found this code change's logic is different from previous.

What if a `multiaddr_to_socketaddr` is Some, and it's reachable?
- In previous code, we will check `extrace_peer_id.is_none` to push local_peer_id.
- In this PR's code, we never push local_peer_id.

It this expected? Cc. @driftluo @quake

PR #4784, opened 2025-01-16
fix: `dial_feeler(..)` function dial identify protocol

chaoticlonghair:

The `dial_feeler(..)` function dial with the identify protocol acctually.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -490,7 +490,7 @@ impl NetworkState {
         if let Err(err) = self.dial_inner(
             p2p_control,
             addr.clone(),
-            TargetProtocol::Single(SupportProtocols::Identify.protocol_id()),
+            TargetProtocol::Single(SupportProtocols::Feeler.protocol_id()),
         ) {
             debug!("dial_feeler error {err}");
         } else {


driftluo (changes requested):
all sessions open with identify，then it will open others https://github.com/nervosnetwork/ckb/blob/develop/network/src/protocols/identify/mod.rs#L466-L476

driftluo:
> If dial with the feeler is inside the identify protocol, why don't just call dial_identify(..), with error handling.

dial feeler success should mark this dial is feeler, but dial identify not
```rust
  self.with_peer_registry_mut(|reg| {
      reg.add_feeler(&addr);
  });
```

driftluo:
> dial identify will open identify with all supported protocols inside it, then close feeler, and keep the session connected

no, feeler never opens on dial identify

driftluo:
dial_identify：
- A dial identify to B: connected
- A‘s identify open all protocols except feeler

dial_feeler:
- A mark feeler
- A dial identify to B: connected
- A‘s identify open feeler

driftluo:
> But dial_inner is before add_feeler.

I think a better description here is, "After successfully sending the command, mark the feeler", there is no dial completed at this time

PR #4785, opened 2025-01-17
Improving spawn and exec syscalls

mohanson:

Problem Summary:

The spawn and exec system calls have some usage issues. The overall operation is too complicated and requires additional checks when reading args from the vm. We found that these checks are unnecessary and can be avoided by other means.

> Note: This PR is based on ckb-vm develop branch; need to wait for this PR to be merged: https://github.com/nervosnetwork/ckb-vm/pull/450 and then release ckb-vm v0.24.13

What's Changed:

- Use ckb-vm's new `FlattenedArgsReader` to simplify reading args in spawn and exec syscalls.
- Add exec v2 implementation.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


eval-exec (commented):
Please rebase onto `develop`.

script/src/types.rs
```diff
@@ -441,6 +451,7 @@ pub struct FdArgs {
 
 #[derive(Clone, Debug)]
 pub enum Message {
+    Exec(VmId, ExecArgs),
```
xxuejie, line 454: I would recommend rename this `ExecV2` as well.

xxuejie:
Just a note that a significant portion of the changes here, is due to the fact that the [PR](https://github.com/nervosnetwork/ckb-vm/pull/451) is now applied directly to the develop branch. Per CKB-VM's release schedule, CKB will for now stick to `v0.24.x` series of CKB-VM. So when the PR at CKB-VM is merged, cherry-picked, and release to a proper `v0.24.x` series, many of changes here will be reverted.

script/src/syscalls/exec_v2.rs
```diff
@@ -0,0 +1,121 @@
...
+            machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
+            return Ok(true);
+        }
+        if length > 0 {
+            let end = offset.checked_add(length).ok_or(VMError::MemOutOfBound)?;
+            if end > full_length {
+                machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
+                return Ok(true);
+            }
+        }
```
eval-exec, line 102: What should ExecV2 do when `length == 0`?

script/src/syscalls/exec_v2.rs
```diff
@@ -0,0 +1,121 @@
...
+                return Ok(true);
+            }
+            Err(e) => return Err(e),
+        };
+        if offset >= full_length {
+            machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
+            return Ok(true);
+        }
+        if length > 0 {
+            let end = offset.checked_add(length).ok_or(VMError::MemOutOfBound)?;
```
eval-exec, line 97: both `offset` and `length` are `<= u32::MAX`, so `offset.checked_add(length)` will be always a `Some`

eval-exec: So I think L92->95 can be omitted.

script/src/syscalls/exec_v2.rs
```diff
@@ -0,0 +1,121 @@
...
+            if end > full_length {
+                machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
+                return Ok(true);
+            }
+        }
+
+        let argc = machine.registers()[A4].clone();
+        let argv = machine.registers()[A5].clone();
```
eval-exec, line 105: L53->L55 is:
```rust
        let index = machine.registers()[A0].to_u64();
        let mut source = machine.registers()[A1].to_u64();
        let place = machine.registers()[A2].to_u64();
```
So how about use `.to_u64()` too

```suggestion
        let argc = machine.registers()[A4].to_u64();
        let argv = machine.registers()[A5].to_u64();
```

script/src/scheduler.rs
```diff
@@ -761,26 +791,40 @@ where
...
 
+    // Load the program into an empty vm.
+    fn load_vm_program(
+        &mut self,
+        context: &MachineContext<DL>,
+        machine: &mut Machine,
+        data_piece_id: &DataPieceId,
+        offset: u64,
+        length: u64,
+        args: Option<(u64, u64, u64)>,
```
eval-exec, line 808: How about creating a ne struct:
```rust
struct VmArgs {
  vm_id: u64,
  argc: u64,
  argv: u64,
}
```
to replace the anonymouse turple `(u64, u64, u64)`?  This would improve code readability and make the parameters more self-explanatory.

eval-exec (approved):

script/src/syscalls/exec_v2.rs
mohanson: It will try to load an empty elf file, which will result in elf parsing errors.

script/src/syscalls/exec_v2.rs
mohanson: > both offset and length are <= u32::MAX, so offset.checked_add(length) will be always a Some.

This description is correct, and we can probably remove this redundant check.

L92->L96 is used to determine whether the data offset exceeds the total length of the data. Why can it be omitted?

script/src/syscalls/exec_v2.rs
mohanson: Yes, we don't have to clone the data here.

script/src/scheduler.rs
mohanson: In the previous internal review, Xuejie suggested that I use `Option<(u64, u64, u64)>`. I am not sure whether I should make a struct for them now.

script/src/syscalls/exec_v2.rs
eval-exec: > L92->L96 is used to determine whether the data offset exceeds the total length of the data. Why can it be omitted?

I think
```rust
        if offset >= full_length {
            machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
            return Ok(true);
        }
        if length > 0 {
            let end = offset.checked_add(length).ok_or(VMError::MemOutOfBound)?;
            if end > full_length {
                machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
                return Ok(true);
            }
        }
```
is equal to: 
```rust
        let end = offset.checked_add(length).ok_or(VMError::MemOutOfBound)?;
        if end >= full_length {
            machine.set_register(A0, Mac::REG::from_u8(SLICE_OUT_OF_BOUND));
            return Ok(true);
        }

```

Is above replacement correct?

script/src/scheduler.rs
eval-exec: I'm not sure either.

script/src/syscalls/exec_v2.rs
mohanson: Please forgive me, I made a mistake earlier, **when length == 0, data[offset...] will be read instead of empty bytes;**

The current code behavior is correct.

script/src/syscalls/exec_v2.rs
mohanson: You are right, we can combine the two checks and don't need to check the case where length == 0.

script/src/scheduler.rs
xxuejie: I did re-check offline communication logs, previously this is represented using the following structure:

```
pub enum BootArgsType {
    Static,
    Stream { vm_id: u64, argc: u64, argp: u64 },
}
```

Since the enum only had 2 variant: one that has 3 arguments, while the other is empty. I just believe that this 2-value enum, could in fact simply be an `Option`.

I have no opinion whether it is `Option<(u64, u64, u64)>` or `Option<ArgvPointer>`, both feel the same to me. I never say `(u64, u64, u64)` is better in readability to `ArgvPointer` or another structure, both feel the same to me but I do understand a different opinion might occur.

script/src/scheduler.rs
```diff
@@ -340,6 +339,29 @@ where
...
+                    new_machine.machine.add_cycles_no_checking(old_cycles)?;
+                    self.load_vm_program(
+                        &context,
+                        &mut new_machine,
+                        &args.data_piece_id,
+                        args.offset,
+                        args.length,
+                        Some((vm_id, args.argc, args.argv)),
+                    )?;
+                    self.instantiated.insert(vm_id, (context, new_machine));
```
xxuejie, line 363: Please cap the number of maximum instantiated virtual machines here.

script/src/types.rs
```diff
@@ -350,12 +350,22 @@ pub enum VmState {
     WaitForRead(ReadState),
 }
 
+#[derive(Clone, Debug)]
+pub struct ExecV2Args {
+    pub data_piece_id: DataPieceId,
+    pub offset: u64,
+    pub length: u64,
```
xxuejie, line 357: Notice both `ExecV2Args` and `SpawnArgs` contain `data_piece_id`, `offset` and `length`. All 3 are required to locate a program. We should make them a struct to distinguish them from other parameters.

script/src/syscalls/exec_v2.rs
```diff
@@ -0,0 +1,122 @@
...
+            Err(_) => {
+                machine.set_register(A0, Mac::REG::from_u8(INDEX_OUT_OF_BOUND));
+                return Ok(true);
+            }
+        };
+        let bounds = machine.registers()[A3].to_u64();
+        let offset = bounds >> 32;
+        let length = bounds as u32 as u64;
+
+        // We are fetching the actual cell here for some in-place validation
```
xxuejie, line 77: Is there a chance we can move those validations to the scheduler part, so we only loads the actual program once?

script/src/scheduler.rs
mohanson: Exec will always remove an old running vm instance and then join a new one, so the maximum number of instantiations will not be triggered.

xxuejie: Ah I see, in that sense this code is correct, but can we change the code to the following flow to better illustrate this fact:

```
let machine_n_context = self.instantiated.get_mut(&vm_id)...
{
  let old_machine = &machine_n_context.1;
  ...
}
...
*machine_n_context = (context, new_machine);
```

The latter insert line can be quite confusing and maybe one day we forgot about this underlying logic.

mohanson: In this way of writing, `machine_n_context` borrows `self` as mutable, This will cause `self.create_dummy_vm` and `self.load_vm_program` to no longer be usable

xxuejie: I see, you are right here. In this sence I recommend the following line as a hint:

```
debug_assert!(self.instantiated.contains_key(&vm_id));
self.instantiated.insert(vm_id, (context, new_machine));
```

mohanson: OK. I'll add a comment as well.

script/src/syscalls/exec_v2.rs
mohanson: That should be possible, let me simplify it.

mohanson: In the current implementation, it is not possible to load data only once.

Because in fact, they are two different calls. The first one is `sc.load_data(&data_piece_id, 0, 0)` called in exec_v2, the purpose is to obtain the total length of cell_data or witness, and the second one is `sc.load_data(&location.data_piece_id, location.offset, location.length)`, the purpose is to obtain the actual program to be executed from cell_data or witness.

xxuejie: What if we move the validation code till later part when the length and the data are fetched in the scheduler part? Total length is only there for verification purposes, and loading per `length`, should already do many of the verification work.

Still we might hit other roadblockers, I just wonder if the code can be simplified in v 2

mohanson: The current code is completely translated from

https://github.com/nervosnetwork/ckb/blob/develop/script/src/syscalls/exec.rs#L149-L165

If the offset and length parameters are invalid and the A4 register exceeds 4M, then after moving the verification code, the exec syscall will change from the normal SLICE_OUT_OF_BOUND return value to a VMError.

I think there are more compatibility issues.

mohanson: I tried to do this, but encountered some compatibility issues, and fixing these compatibility issues may involve modules that this PR does not care about.

For example, in DataSource, sc.load_data always returns a Bytes array (possibly an empty array) regardless of offset and length. Therefore, when the developer incorrectly passes in offset and length, in your modification plan, the scheduler will incorrectly execute the elf parsing step and return VMError.

xxuejie: I do understand that we might not be able to do this but still want to follow up first:

> Therefore, when the developer incorrectly passes in offset and length, in your modification plan, the scheduler will incorrectly execute the elf parsing step and return VMError.

Isn't this expected behavior? If one passes in an invalid offset/length, it will indicate and invalid elf

mohanson: Both `spawn` and `exec_v2` contain this validation code. I created a function to avoid duplication of code. Do you think this is OK?

xxuejie: I've been toying with the code here, and this [commit](https://github.com/xxuejie/ckb/commit/d8309bec0cb20d95a351d9c3755ef98089df5a2b) fully captures my opinions. Of course it changes current exec's behavior, so out-of-bound offset/length does not result in errors anymore, they only chop the loaded binary as much as possible(e.g., when offset is bigger than data length, an empty buffer will be returned).

1. In a v2 design, we are allowed to change exec's behavior
2. If we really think about [this piece of code](https://github.com/nervosnetwork/ckb/blob/b44d2d2729cee98ccab8db83fddcd235619db145/script/src/syscalls/exec.rs#L150-L163), I'm not sure if they make sense anymore. when [loading any other type of data](https://github.com/nervosnetwork/ckb/blob/b44d2d2729cee98ccab8db83fddcd235619db145/script/src/syscalls/utils.rs#L8-L24), CKB never generates out-of-bound errors, it simply returns as much data as possible. I personally do not have a convincing reason why loading binaries should be any different.

So on a personal opinion, I would remove those redundant checks on offset/length in exec's V2 design. We might not be able to do the same thing this time for spawn(since it already is activated on testnet), but I do believe it makes sense to alter exec's design now, and also change spawn's design when we have a chance.

mohanson: I have made the changes according to your opinions.

script/src/scheduler.rs
```diff
@@ -340,6 +347,49 @@ where
...
+                    let max_cycles = old_machine.machine.max_cycles();
+                    let (program, _full_length) = {
+                        let mut sc = old_context.snapshot2_context().lock().expect("lock");
+                        match sc.load_data(
+                            &args.location.data_piece_id,
+                            args.location.offset,
+                            args.location.length,
+                        ) {
+                            Ok(val) => val,
+                            Err(Error::SnapshotDataLoadError) => {
```
xxuejie, line 368: Sorry I missed this, for a V2 clean design, can we just return the snapshot data load error here? I think there is no need to mimic v1 behavior.

script/src/syscalls/exec_v2.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    fn ecall(&mut self, machine: &mut Mac) -> Result<bool, VMError> {
+        if machine.registers()[A7].to_u64() != EXEC {
+            return Ok(false);
+        }
+        let index = machine.registers()[A0].to_u64();
+        let mut source = machine.registers()[A1].to_u64();
+        let place = machine.registers()[A2].to_u64();
+        // To keep compatible with the old behavior. When Source is wrong, a
```
xxuejie, line 37: I do want to ask here: if old behavior is not a concern, what a proper solution here should be?

Same question goes for `Place` parsing below

xxuejie (commented):
Just some minor comments on the necessity of following old code behaviors

script/src/scheduler.rs
mohanson: Yes, we should return a VMError

script/src/syscalls/exec_v2.rs
mohanson: I suggest removing this part of the compatibility code.

xxuejie (approved):

mohanson:
@eval-exec   I have no idea why `ckb-sync tests::synchronizer::basic_sync::basic_sync` fails, can you take a look? https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35965802195?pr=4785

eval-exec:
> @eval-exec I have no idea why `pthread lock: Invalid argument`, can you take a look? https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35965802195?pr=4785

Ok, That error occurred before, and it seems to be related to ckb-rocksdb.

eval-exec:
> @eval-exec I have no idea why `pthread lock: Invalid argument`, can you take a look? https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35965802195?pr=4785


Latest unit test failed:
```
──── STDERR:             ckb-script verify::tests::ckb_2023::features_since_v2021::check_typical_secp256k1_blake160_2_in_2_out_tx_with_state
thread 'verify::tests::ckb_2023::features_since_v2021::check_typical_secp256k1_blake160_2_in_2_out_tx_with_state' panicked at script/src/verify/tests/ckb_latest/features_since_v2021.rs:878:9:
step_cycles 3242342
```
https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35970732805?pr=4785

mohanson:
> > @eval-exec I have no idea why `pthread lock: Invalid argument`, can you take a look? https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35965802195?pr=4785
> 
> Latest unit test failed:
> 
> ```
> ──── STDERR:             ckb-script verify::tests::ckb_2023::features_since_v2021::check_typical_secp256k1_blake160_2_in_2_out_tx_with_state
> thread 'verify::tests::ckb_2023::features_since_v2021::check_typical_secp256k1_blake160_2_in_2_out_tx_with_state' panicked at script/src/verify/tests/ckb_latest/features_since_v2021.rs:878:9:
> step_cycles 3242342
> ```
> 
> https://github.com/nervosnetwork/ckb/actions/runs/12885771772/job/35970732805?pr=4785

The incorrect test case has been fixed, please re-execute ci

eval-exec (approved):

xxuejie (approved):

quake (approved):

xxuejie (approved):

PR #4793, opened 2025-01-21
feat: mark dns address connected time on peer store

driftluo:

Mark DNS address connected time on peer store, allow it to spread across multiple nodes. 

The transmission path is as follows:

- DNS address public with identify protocol https://github.com/nervosnetwork/ckb/blob/b44d2d2729cee98ccab8db83fddcd235619db145/network/src/protocols/identify/mod.rs#L520-L530
- discovery protocol will random fetch it from peer store when it had connected during last week https://github.com/nervosnetwork/ckb/blob/b44d2d2729cee98ccab8db83fddcd235619db145/network/src/protocols/discovery/mod.rs#L140-L150
- peer store random fetch https://github.com/nervosnetwork/ckb/blob/b44d2d2729cee98ccab8db83fddcd235619db145/network/src/peer_store/peer_store_impl.rs#L206-L210

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/protocols/identify/mod.rs
+++ b/network/src/protocols/identify/mod.rs
@@ -457,8 +457,15 @@ impl Callback for IdentifyCallback {
                     // and if set it to peer store, it will be broadcast to the entire network,
                     // but this is an unverified address
                     if renew {
+                        let dns_addr = self.network_state.pending_dns_addrs.write().remove(
+                            &extract_peer_id(&context.session.address).expect("must have peerid"),
+                        );
                         self.network_state.with_peer_store_mut(|peer_store| {
                             peer_store.add_outbound_addr(context.session.address.clone(), flags);
+                            if let Some(dns) = dns_addr {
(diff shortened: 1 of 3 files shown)


eval-exec (approved):

network/src/protocols/identify/mod.rs
```diff
@@ -457,8 +457,15 @@ impl Callback for IdentifyCallback {
                     // and if set it to peer store, it will be broadcast to the entire network,
                     // but this is an unverified address
                     if renew {
+                        let dns_addr = self.network_state.pending_dns_addrs.write().remove(
```
quake, line 460: A address will be written to the `pending_dns_addrs` field when the session open, but it is only removed when the identity protocol succeeds, if the identity protocol fails, then this field will keep growing, it is recommended to remove it when the session is closed.

quake (approved):

PR #4795, opened 2025-01-22
`NetRpcImpl::get_peers` return `Remoteaddress.addresses` dedup

eval-exec:

`get_peers` RPC returned `RemoteAddress.addresses` shouldn't contains duplicate `MultiAddr`.

### Related changes

- In `NetRpcImpl::get_peers`, use `HashSet` to remove duplicate `MultiAddr`.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/rpc/src/module/net.rs
+++ b/rpc/src/module/net.rs
@@ -11,6 +11,7 @@ use ckb_systemtime::unix_time_as_millis;
 use ckb_types::prelude::{Pack, Unpack};
 use jsonrpc_core::Result;
 use jsonrpc_utils::rpc;
+use std::collections::HashSet;
 use std::sync::Arc;
 
 const MAX_ADDRS: usize = 50;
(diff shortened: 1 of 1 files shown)


rpc/src/module/net.rs
```diff
@@ -586,7 +587,7 @@ impl NetRpc for NetRpcImpl {
             .connected_peers()
             .iter()
             .map(|(peer_index, peer)| {
-                let mut addresses = vec![&peer.connected_addr];
+                let mut addresses: HashSet<_> = vec![&peer.connected_addr].into_iter().collect();
```
quake, line 590: no need to create vec and convert to hashet, we may use the iterator to collect directly:
```
let addresses: HashSet<_> = std::iter::once(peer.connected_addr).chain(peer.listened_addrs.iter()).collect();
```

driftluo (approved):

doitian (approved):

PR #4800, opened 2025-02-05
feat: impl filter with transport

driftluo:

outbound service impl filter with transport：
TCP -> All
WS -> IP and DNS
WSS -> Only DNS

### Check List

Tests 

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/services/outbound_peer.rs
+++ b/network/src/services/outbound_peer.rs
@@ -55,8 +55,19 @@ impl OutboundPeerService {
 
     fn dial_feeler(&mut self) {
         let now_ms = unix_time_as_millis();
+        let filter = |peer_addr: &AddrInfo| match self.transport_type {
+            TransportType::Tcp => true,
+            TransportType::Ws => peer_addr
+                .addr
+                .iter()
+                .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_) | Protocol::Tcp(_))),
+            TransportType::Wss => peer_addr
+                .addr
+                .iter()
+                .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_))),
+        };
         let attempt_peers = self.network_state.with_peer_store_mut(|peer_store| {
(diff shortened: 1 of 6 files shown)


network/src/services/outbound_peer.rs
```diff
@@ -55,8 +55,23 @@ impl OutboundPeerService {
...
+                    .iter()
+                    .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_) | Protocol::Tcp(_)))
+            }),
+            TransportType::Wss => Box::new(|peer_addr: &AddrInfo| {
+                peer_addr
+                    .addr
+                    .iter()
+                    .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_)))
+            }),
+        };
```
quake, line 72:
```suggestion
        let filter = |peer_addr: &AddrInfo| {
            match self.transport_type {
                TransportType::Tcp => true,
                TransportType::Ws => peer_addr
                    .addr
                    .iter()
                    .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_) | Protocol::Tcp(_))),
                TransportType::Wss => peer_addr
                    .addr
                    .iter()
                    .any(|p| matches!(p, Protocol::Dns4(_) | Protocol::Dns6(_))),
            }
        };
```

network/src/services/outbound_peer.rs
```diff
@@ -97,8 +112,24 @@ impl OutboundPeerService {
 
         let target = &self.network_state.required_flags;
 
+        let filter: Box<dyn Fn(&AddrInfo) -> bool> = match self.transport_type {
```
quake, line 115: same suggestion as previous comment

network/src/services/outbound_peer.rs
driftluo: This is not possible because the filter return value types of each branch are different, so the only way to unify them is to use the dyn trait.

quake (approved):

eval-exec (approved):

PR #4833, opened 2025-03-11
chore: when the user configuration is wrong, exit and output log

driftluo:

when the user configuration is wrong, exit and output log(The wasm side code is not modified, because wasm currently has no listen, and this field configuration is meaningless)

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -107,8 +107,13 @@ impl NetworkState {
             .filter_map(|mut addr| match multiaddr_to_socketaddr(&addr) {
                 Some(socket_addr) if !is_reachable(socket_addr.ip()) => None,
                 _ => {
-                    if extract_peer_id(&addr).is_none() {
-                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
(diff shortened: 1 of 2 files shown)


network/src/network.rs
```diff
@@ -107,8 +107,13 @@ impl NetworkState {
             .filter_map(|mut addr| match multiaddr_to_socketaddr(&addr) {
                 Some(socket_addr) if !is_reachable(socket_addr.ip()) => None,
                 _ => {
-                    if extract_peer_id(&addr).is_none() {
-                        addr.push(Protocol::P2P(Cow::Borrowed(local_peer_id.as_bytes())));
+                    match extract_peer_id(&addr) {
+                        Some(peer_id) if peer_id == local_peer_id => {
```
eval-exec, line 111: I think it should be `!=`?

eval-exec (approved):

eval-exec (approved):

eval-exec (approved):

eval-exec (dismissed):

eval-exec (approved):

PR #4847, opened 2025-03-12
feat: Syscall generator shall be customizable

xxuejie:
This enables ckb-script to be used in more environments. I do remember that there is discussion about moving ckb-script out of CKB for more usages. However, I implemented this change due to 2 factors:

* The change is not so big
* We had similar APIs for debug printer as well

In a later time, we can re-evaluate the necessity to move ckb-script out of CKB completely.

Issue Number: close #xxx 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/script/src/types.rs
+++ b/script/src/types.rs
@@ -9,7 +9,7 @@ use ckb_types::{
     prelude::*,
 };
 use ckb_vm::{
-    ISA_B, ISA_IMC, ISA_MOP,
+    ISA_B, ISA_IMC, ISA_MOP, Syscalls,
     machine::{VERSION0, VERSION1, VERSION2},
 };
 use serde::{Deserialize, Serialize};
@@ -64,6 +64,16 @@ pub(crate) type Machine = AsmMachine;
 pub(crate) type Machine = TraceMachine<CoreMachine>;
 
 pub(crate) type DebugPrinter = Arc<dyn Fn(&Byte32, &str) + Send + Sync>;
+pub(crate) type SyscallGenerator<DL> = Arc<
+    dyn Fn(
+            &VmId,
+            &SgData<DL>,
+            &VmContext<DL>,
+            &DebugContext,
+        ) -> Vec<Box<(dyn Syscalls<CoreMachine>)>>
+        + Send
+        + Sync,
+>;
 
 pub struct DebugContext {
     pub debug_printer: DebugPrinter,
--- a/script/src/types.rs
+++ b/script/src/types.rs
@@ -43,11 +43,11 @@ pub type VmIsa = u8;
 pub type VmVersion = u32;
 
 #[cfg(has_asm)]
-pub(crate) type CoreMachineType = AsmCoreMachine;
+pub type CoreMachineType = AsmCoreMachine;
 #[cfg(all(not(has_asm), not(feature = "flatmemory")))]
-pub(crate) type CoreMachineType = DefaultCoreMachine<u64, WXorXMemory<ckb_vm::SparseMemory<u64>>>;
+pub type CoreMachineType = DefaultCoreMachine<u64, WXorXMemory<ckb_vm::SparseMemory<u64>>>;
 #[cfg(all(not(has_asm), feature = "flatmemory"))]
-pub(crate) type CoreMachineType = DefaultCoreMachine<u64, WXorXMemory<ckb_vm::FlatMemory<u64>>>;
+pub type CoreMachineType = DefaultCoreMachine<u64, WXorXMemory<ckb_vm::FlatMemory<u64>>>;
 
 /// The type of core VM machine when uses ASM.
 #[cfg(has_asm)]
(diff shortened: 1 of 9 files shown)


eval-exec (dismissed):

script/src/types.rs
```diff
@@ -64,6 +64,16 @@ pub(crate) type Machine = AsmMachine;
 pub(crate) type Machine = TraceMachine<CoreMachine>;
 
 pub(crate) type DebugPrinter = Arc<dyn Fn(&Byte32, &str) + Send + Sync>;
+pub(crate) type SyscallGenerator<DL> = Arc<
```
quake: Since we're not capturing environment variables in the syscall generator functions, we can use the simpler function pointer type:
```suggestion
pub(crate) type SyscallGenerator<DL> =
    fn(&VmId, &SgData<DL>, &VmContext<DL>, &DebugContext) -> Vec<Box<(dyn Syscalls<CoreMachine>)>>;
```

script/src/verify.rs
```diff
@@ -112,6 +116,22 @@ where
         self.debug_printer = Arc::new(func);
     }
 
+    /// Sets a new syscall generator for special usages
+    pub fn set_syscall_generator<F>(&mut self, func: F)
```
quake, line 120: this fn can be simplified if applied previous review comment
```
    pub fn set_syscall_generator(&mut self, func: SyscallGenerator<DL>) {
        self.syscall_generator = func;
    }
```

xxuejie: The API is designed this way to mimic the above set_debug_printer method.

script/src/types.rs
xxuejie: It is true that if you look at CKB alone, a function pointer type is enough, but this API is designed to cope with more use cases, where mutable data might be used in this syscall. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=354a0207f2854ac23ebfeb304c93c955) for a minimized example. I don't think a simple function pointer will be enough.

script/src/verify.rs
xxuejie: Also see the above comment, function pointer is not enough for us. However, when `SyscallGenerator` is defined as `Arc<Fn(...) -> ...>`, we could indeed write:

```
pub fn set_syscall_generator(&mut self, func: SyscallGenerator<DL>) {     
    self.syscall_generator = func;
}
```

The current API is written so as to mimic existing `set_debug_printer` method. I do believe they should behave similarly.

script/src/types.rs
quake: I may have misunderstood the intent of this PR and the current develop branch code comment, I originally thought that the `VmContext` would be modified to pass data after this PR.

comment: 
```
/// Mutable data at virtual machine level
#[derive(Clone)]
pub struct VmContext<DL>
```

code:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7eda40440cd7f6cbbacdfea7167127a8

xxuejie: So in current design, `VmContext` contains those data that will be used by the scheduler, to pass information across different virtual machines. It is not tweakable by outside environments.

However, there are other data that might be required to a VM, for example, a particular syscall handler might be included, so one can count the number of syscalls executed by a particular VM, those information can be handy at outside environments. And we don't really know in what form those data will come, so there is not way to set aside place in `VmContext` for it.

And there are indeed 2 places for those information:

* One can keep those inside a particular syscall objects
* Or one can use a closure to pass information to the syscall handlers

While both work, I do believe having a closure here, provides enough flexibility.

quake: I now understand that the purpose of this PR is to make it easier for external code to customize the syscall generator by using TransactionScriptsVerifier, and I feel a minor bad smell for introducing Arc<Fn> just for this requirement.

xxuejie: There is a different angle: do you consider `debug_printer` and `syscall_generator` to be similar here? If so, `debug_printer` has been leveraging `Arc<Fn>` for quite some time, and snippets like [this](https://github.com/nervosnetwork/ckb-standalone-debugger/blob/77074ef8e1b6d97f9ec1985c1e5c97986c83156b/ckb-debugger/src/main.rs#L356-L361) has been written in many places. So my question is: why should syscall generator should be any different?

In fact all we need here is `Box<Fn>` like ckb-vm [does](https://github.com/nervosnetwork/ckb-vm/blob/8f0277d1980adcfa19f7d60bd4c9eb82c5eee970/src/machine/mod.rs#L434-L435), the extra `Arc` is simply due to CKB's [future-based runner](https://github.com/nervosnetwork/ckb/blob/220422e94eeaf6a9307e41c97e31e93086f8dcc2/script/src/verify.rs#L641).

xxuejie: And some more points: I'm actually okay making `syscall_generator` a function pointer. However, I think it bad practice that `debug_printer` is `Arc<Fn>`, but `syscall_generator` is `fn`. Both of them should be `Arc<Fn>`, or both of them should be `fn`.

If we make both `debug_printer` and `syscall_generator` as `fn`, it will break existing code where people simply use a closure as `debug_printer`. I'm not sure if this is enough of a compromise we want.

Let me know which side you lean on.

quake: My understanding is that the purpose of `DebugPrinter` is to build a custom debugger syscall, and after this PR allows customization of SyscallGenerator, we should remove debug_printer from `TransactionScriptsVerifier`, and `skip_pause` should also be removed. That's the point of allowing customization of SyscallGenerator

xxuejie: Yeah that makes sense, let me proceed following this path.

eval-exec:
change the log level of `info!("recv preload_task_rx failed, err: {:?}", err);` from error to info will fix the `cli_bats_test` workflo.

script/src/types.rs
xxuejie: I've updated the code per ideas discussed here, please review again.

quake (approved):

zhangsoledad (approved):

PR #4850, opened 2025-03-12
deps: make syn and serde_json be workspac dependencies, manualy upgrade bitflags from 1.3 to 2.9

eval-exec:

Issue Number: close #4844

### Related changes

- **deps: make sync and serde_json be ckb workspace dependencies**
- **deps: upgrade bitflags from 1.3 to 2.9, fix #4844**

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```


driftluo:
bitflag's serialization scheme has a breaking change(In my impression, the endianness sorting has changed) from 1.0 to 2.0, which will cause errors in identifying and reading data stored in peer stores. You cannot simply upgrade it. You need to change the default serialization scheme to 1.0 mode, otherwise, it will cause problems with ckb connection and discovery.

All other places use internal data. Only the network needs to interact and store data, and consistency must be guaranteed.

driftluo:
Just verify whether the conversion between 1.0 and 2.0 is the same in our usage. I remember it is bitflag to u64, then u64 to bitflag.

eval-exec:
`bitflags`'s value are encoded to ckb-network in bellow three message:
1: 
https://github.com/nervosnetwork/ckb/blob/17bd37a3c292f4fa502ddc37034112c07addfea6/util/gen-types/schemas/extensions.mol#L460-L464
2:
https://github.com/nervosnetwork/ckb/blob/17bd37a3c292f4fa502ddc37034112c07addfea6/util/gen-types/schemas/protocols.mol#L43-L48
3:
https://github.com/nervosnetwork/ckb/blob/17bd37a3c292f4fa502ddc37034112c07addfea6/util/gen-types/schemas/protocols.mol#L64-L67

zhangsoledad (approved):

PR #4861, opened 2025-03-27
feat(rpc): add IPC call

mohanson:

What's Changed:

Added an `ipc_call` method in jsonrpc. The method is described in detail in `rpc/src/module/ipc.rs`'s comments.

```rs
/// Call the ipc method in the script.
///
/// ## Params
///
/// * script_locator:
///     - out_point - Reference to the cell by transaction hash and output index.
///     - type_id_args - Reference to the cell by type id type script args.
/// * req:
///     - version - IPC protocol version. Default is 0.
///     - method_id - IPC protocol method id. Default is 0.
///     - payload_format - The format of payload.
///     - payload - Payload.
/// * env: The transaction environment that the ipc script depends on. The script specified by script_group_type + script_hash will be executed, but the script binary will be replaced by the ipc script.
///     - tx - The transaction.
///     - script_group_type - Script group type.
///     - script_hash - Script hash.
///
/// ## Returns
///
/// * version - IPC protocol version. Default is 0.
/// * error_code - IPC error code.
/// * payload_format - The format of payload.
/// * payload - Payload.
///
/// ## Examples
///
/// Request
///
/// ```json
/// {
///     "id": 2,
///     "jsonrpc": "2.0",
///     "method": "ipc_call",
///     "params": [
///         {
///             "out_point": {
///                 "tx_hash": "0x4e75e1d4dddc0efe5e028168f627e3af436f1002b61038d695b24aa8441ffaf5",
///                 "index": "0x0"
///             }
///         },
///         {
///             "version": "0x0",
///             "method_id": "0x0",
///             "payload_format": "json",
///             "payload": {
///                 "SyscallLoadScript": {}
///             }
///         },
///         {
///             "tx": {
///                 "version": "0x0",
///                 "cell_deps": [
///                     {
///                         "out_point": {
///                             "tx_hash": "0xe7db03de9c534cb63d951f5378d5a4fbf43a2ebf5e48cb626562d188a3697772",
///                             "index": "0x0"
///                         },
///                         "dep_type": "dep_group"
///                     }
///                 ],
///                 "header_deps": [],
///                 "inputs": [
///                     {
///                         "since": "0x0",
///                         "previous_output": {
///                             "tx_hash": "0x953d1a4b95cefca9588b6d0bf7e97084bb51010ce87d1bcbe65c0ccd621781b8",
///                             "index": "0x0"
///                         }
///                     }
///                 ],
///                 "outputs": [],
///                 "outputs_data": [],
///                 "witnesses": [
///                     "0x55000000100000005500000055000000410000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
///                 ]
///             },
///             "script_group_type": "lock",
///             "script_hash": "0x0b1bae4beaf456349c63c3ce67491fc75a1276d7f9eedd7ea84d6a77f9f3f5f7"
///         }
///     ]
/// }
/// ```
///
/// Response
///
/// ```json
/// {
///     "jsonrpc": "2.0",
///     "result": {
///         "error_code": "0x0",
///         "payload": {
///             "SyscallLoadScript": "0x490000001000000030000000310000009bd7e06f3ecf4be0f2fcd2188b23f1b9fcc88e5d4b65a8637b17723bbda3cce8011400000075178f34549c5fe9cd1a0c57aebd01e7ddf9249e"
///         },
///         "payload_format": "json",
///         "version": "0x0"
///     }
///     "id": 2
/// }
/// 
```

Here are some additional points to note:

- Allow calling all syscalls in `ipc script`, such as `exec` and `spawn`.
- How to write an ipc script? Here is a simple example. https://github.com/libraries/ipc/blob/master/ipc_test/src/ipc_test.rs
- There are a few things that are not done yet, I plan to submit them using another PRs
  - Add integration tests. I use [python](https://github.com/libraries/ipc) to write test cases during the development stage, <del>I need more time to rewritten it in Rust.</del>, @gpBlockchain is responsible for migrating integration testing to project https://github.com/cryptape/ckb-py-integration-test

### Check List 

Tests 

- Unit test
- Integration test

### Release note 

```release-note
```


rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+            .lock()
+            .map_err(|e| RPCError::custom_with_error(RPCError::IPC, e))?
+            .write_all(&req.serialize())
+            .map_err(|e| RPCError::custom_with_error(RPCError::IPC, e))?;
+        pipesctx
+            .nw
+            .lock()
+            .map_err(|e| RPCError::custom_with_error(RPCError::IPC, e))?
+            .close();
+        let rep = ResponsePacket::read_from(pipesctx.nr.lock().unwrap().deref_mut()).unwrap();
```
eval-exec, line 770: How about changing `rep` to `resp`, since `rep` can easily be confused with `req`?

mohanson:
> Please execute `make gen-rpc-doc`

Done

util/ipc/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "ckb-ipc"
+version = "0.201.0-pre"
```
eval-exec, line 3:
```suggestion
version.workspace = true
```

util/ipc/src/pipe.rs
```diff
@@ -0,0 +1,112 @@
...
+        if self.rx.is_some() {
+            drop(self.rx.take());
+        }
+    }
+}
+
+impl Read for Pipe {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Error> {
+        if self.buf.is_empty() {
+            match self.rx.as_ref().unwrap().lock().unwrap().recv() {
```
eval-exec, line 81: It would be better to avoid using `.unwrap()` here.

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+        let cell_meta_i = CellMeta {
+            cell_output: packed::CellOutput::new_builder()
+                .lock(
+                    packed::Script::new_builder()
+                        .code_hash(
+                            cell_meta_cell_meta_ipc_script
+                                .cell_output
+                                .type_()
+                                .to_opt()
+                                .unwrap()
```
eval-exec, line 527: It’s better to avoid using `.unwrap()` here and elsewhere — proper error handling is preferred.

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+                                .calc_script_hash(),
+                        )
+                        .hash_type(ckb_types::core::ScriptHashType::Type.into())
+                        .build(),
+                )
+                .build_exact_capacity(Capacity::zero())
+                .map_err(|e| RPCError::custom_with_error(RPCError::IPC, e))?,
+            out_point: packed::OutPoint::new(
+                packed::Byte32::new_unchecked(vec![0x00; 32].into()),
+                1,
```
eval-exec, line 537: How about using `packed::Byte32::zero()`

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+            pipesfin.nw.lock().unwrap().close();
+            pipesfin.nr.lock().unwrap().close();
+        });
+        let req = RequestPacket::new(
+            req.version.value() as u8,
+            req.method_id.value(),
+            match fmt {
+                IpcPayloadFormat::Hex => {
+                    let str = req
+                        .payload
```
eval-exec, line 745: Do we need to check the size of `req.payload`?

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+            version: Uint64::from(resp.version() as u64),
+            error_code: Uint64::from(resp.error_code()),
+            payload_format: fmt.clone(),
+            payload: match fmt {
+                IpcPayloadFormat::Hex => Value::String(format!("0x{}", hex_string(resp.payload()))),
+                IpcPayloadFormat::Json => {
+                    serde_from_str::<Value>(String::from_utf8_lossy(resp.payload()).as_ref())
+                        .map_err(|e| RPCError::custom_with_error(RPCError::IPC, e))?
+                }
+            },
```
eval-exec, line 781: Maybe we should change the `payload`'s type to `String` instead of `serde_json::Value`?

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+            pipesctx.clone(),
+        );
+        std::thread::spawn(move || {
+            // Ignore its return value as it is unimportant.
+            script_verifier.verify_single(
+                match env.script_group_type {
+                    ScriptGroupType::Lock => ckb_script::ScriptGroupType::Lock,
+                    ScriptGroupType::Type => ckb_script::ScriptGroupType::Type,
+                },
+                &env.script_hash.0.pack(),
```
eval-exec, line 724:
```suggestion
                &env.script_hash.pack(),
```

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,784 @@
...
+                    .objects;
+                Ok(object_vec[0]
+                    .output_data
+                    .clone()
+                    .ok_or_else(|| RPCError::custom(RPCError::IPC, "Unexpected"))
+                    .map(|e| e.into_bytes())?)
+            }
+            None => Err(RPCError::custom(
+                RPCError::IPCIndexerIsDisabled,
+                "Query by type id requires enabling indexer",
```
eval-exec, line 667: `                "Query by type id requires enabling Indexer in `[rpc.modules]`",`

rpc/src/module/ipc.rs
mohanson: Done

rpc/src/module/ipc.rs
mohanson: CKB's jsonrpc has set `max_request_body_size = 10M`, I don't have much idea whether we need to set the size limit additionally.

rpc/src/module/ipc.rs
mohanson: According to the design document, Payload needs to support two formats. The first is json object, for example:

```json
{
  Payload: {
    "MethodName": { "Parameter": "Value" } 
  }
}
```

The second is Hex mode, for example

```
Payload: "0x0000000001111111122222..."
```

It seems that only `serde_json::Value` can be used to define its type.

gpBlockchain:
LGTM

zhangsoledad (approved):

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,883 @@
...
+impl HeaderProvider for Resource {
+    fn get_header(&self, hash: &ckb_types::packed::Byte32) -> Option<ckb_types::core::HeaderView> {
+        self.shared.snapshot().get_header(hash)
+    }
+}
+
+#[derive(Clone)]
+struct PipesCtx {
+    nr: Arc<Mutex<Pipe>>, // Native Reader
+    nw: Arc<Mutex<Pipe>>, // Native Writer
```
chenyukang, line 124: prefer to use longer variable names, such as `reader`, `vm_reader`.

rpc/src/service_builder.rs
```diff
@@ -76,6 +78,17 @@ impl<'a> ServiceBuilder<'a> {
...
     }
 
+    /// Mounts methods from module IPC if it is enabled in the config.
+    pub fn enable_ipc(mut self, shared: Shared, indexer_rpc_impl: Option<IndexerRpcImpl>) -> Self {
+        let methods = IpcRpcImpl::new(
+            shared,
+            indexer_rpc_impl,
+            self.config.ipc_call_limit_cycles.unwrap_or(100000000),
+            self.config.ipc_call_limit_time.unwrap_or(5),
+        );
```
chenyukang, line 88: prefer to set default value in config fields.

rpc/src/tests/setup.rs
```diff
@@ -191,39 +191,44 @@ pub(crate) fn setup_rpc_test_suite(height: u64, consensus: Option<Consensus>) ->
...
+    let builder = builder.enable_integration_test(
+        shared.clone(),
+        network_controller.clone(),
+        chain_controller.clone(),
+        vec![],
+        vec![],
+    );
+    let builder = builder.enable_debug();
+    let builder = builder.enable_alert(alert_verifier, alert_notifier, network_controller);
+    let indexer_rpc_impl = builder.indexer_rpc_impl.clone();
```
chenyukang, line 230: any specific reason to don't use chain call style here?

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,883 @@
...
+                            format!("Get type id args failed: {}", type_id_args),
+                        )
+                    })?
+                    .output_data
+                    .clone()
+                    .ok_or_else(|| RPCError::custom(RPCError::IPC, "Unexpected"))
+                    .map(|e| e.into_bytes())?)
+            }
+            None => Err(RPCError::custom(
+                RPCError::IPCIndexerIsDisabled,
```
chenyukang, line 705: seems from the code, if IPC module is not enabled, `ipc_script_data_by_type_id_args` will never be called here. 
it's not necessary to add a RPCError::IPCIndexerIsDisabled code for it.

chenyukang: never mind, it's not IPC module here.

chenyukang: so if IPC module is enabled, it will require Index module required, could we just validate this rule when parsed config file.
otherwise user can only figure out this rule when error happened.

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,883 @@
...
+            async move {
+                // Ignore its return value as it is unimportant.
+                let _ = script_verifier
+                    .create_scheduler(&script_group)
+                    .map(|mut scheduler| {
+                        // Make the vm can always be stopped.
+                        debug_assert_eq!(scheduler.consumed_cycles(), 0);
+                        let tic = std::time::SystemTime::now();
+                        loop {
+                            let last_cycles = limit_cycles as u64 - scheduler.consumed_cycles();
```
chenyukang, line 801: do we have a risk of overflow here?

rpc/src/module/ipc.rs
```diff
@@ -0,0 +1,883 @@
...
+                            let step_cycles = std::cmp::min(upper_cycles, last_cycles);
+                            let result = scheduler.run(RunMode::LimitCycles(step_cycles as u64));
+                            if let Err(ckb_vm::Error::CyclesExceeded) = result {
+                                if step_cycles < upper_cycles {
+                                    break;
+                                }
+                                let toc = tic.elapsed().map(|e| e.as_secs()).unwrap_or(u64::MAX);
+                                if toc >= limit_time as u64 {
```
chenyukang, line 809: if timeout, whether we should report an error for endpoint?

rpc/src/module/ipc.rs
mohanson: I added defensive code to avoid the payload being too large.

rpc/src/tests/setup.rs
mohanson: Here I need to clone a copy of indexer_rpc_impl, which cannot be achieved using chain call.

rpc/src/module/ipc.rs
mohanson: When the `script_locator` parameter of ipc_call is `type_id_args`, it will depend on indexer;
When it is `out_point`, it will not depend on indexer.

So it can only be checked at runtime.

rpc/src/module/ipc.rs
mohanson: We are better off using the `saturating_sub` method

PR #4880, opened 2025-04-29
test: Attempted fixes on unstable tests

xxuejie:
Some of our sync tests have been slightly unstable: they pass in 95%+ of the time, however from time to time, they would break, if we try to rerun them in CI, in most cases those tests would then pass.

I managed to dig into those tests with a little help from rr, it seems that 2 timings would affect those unstable tests:

1. When a test finishes, C++'s destructors decide to destruct rocksdb's global Timer structure.
2. When all the threads spawned by ChainService terminates, Rust decides to drop ChainService. However, a RocksDB instance is buried deep in ChainService. When the RocksDB instance is destructed, C++ code would eventually run Timer::Cancel on the global Timer structure.

If timing 2 happens before timing 1, everything is fine, the test will pass. However from time to time, timing 1 might happen before timing 2, in this case unit tests would fail, and we see errors relating to pthread.

This change is an attempt to fix this issue.

Please do note that it attempts to fix those unstable tests we know of in ckb-sync package, however there might or might not be other unstable tests in other CKB packages. Later changes can be used to fix more unstable tests, following the change made in this commit.

Issue Number: close #xxx 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```

### DIFF
--- a/chain/src/init.rs
+++ b/chain/src/init.rs
@@ -150,3 +150,31 @@ pub fn build_chain_services(
 
     (chain_controller, chain_service_thread)
 }
+
+/// This structure restricts the scope of chain service, and forces chain
+/// service threads to terminate before dropping the structure.
+/// The content of this struct will always be present, the reason we
+/// wrap them in an option, is that we will need to consume them in
+/// Drop trait impl of this struct.
+pub struct ChainServiceScope(Option<(ChainController, thread::JoinHandle<()>)>);
+
+impl ChainServiceScope {
+    /// Creates a new ChainServiceScope structure
+    pub fn new(builder: ChainServicesBuilder) -> Self {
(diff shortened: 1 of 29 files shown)


eval-exec (dismissed):

driftluo (dismissed):

xxuejie:
@driftluo @eval-exec I've done a grep on the codebase, and apply the same fix to more potential unstable tests, please review the code again when you are available, thanks.

driftluo (approved):

eval-exec (approved):

zhangsoledad (approved):

PR #4920, opened 2025-07-03
feat: Enhance scriptHashType serialization logic

zhangsoledad:

This PR addresses issues in the serialization of `ScriptHashType`:

1. Mandatory Upgrades for Third-Party Libraries:
Each time a Hardfork introduces a new `Data` type, third-party libraries that rely on CKB types are forced to upgrade. This creates additional maintenance overhead for developers and ecosystem participants.

2. Obscured HashType Design:
The current implementation does not clearly reflect the design rules of `ScriptHashType`. The `ScriptHashType` is structured as follows:
The low 1 bit indicates the type, 1 represents 'type', 0 represents 'data'.
The high 7 bits specify the version of the data.

3. Ambiguous Error Classification in Verification:
During input Verification, there is no clear distinction between consensus errors and serialization errors. 

The newly implemented serialization will be extended to support all Data types, including those not yet enabled, such as Data3 to Data127. This means that at the serialization level, there will no longer be a need to introduce incompatible updates. When a hard fork introduces new Data types in the future, third parties will not be broken. Additionally, the verification of whether the input ScriptHashType is allowed by the current consensus will be performed at the consensus layer, rather than at the serialization level as before.

### Check List 

Tests 

- Unit test

### Release note 

```release-note
```

### DIFF
--- a/test/src/specs/hardfork/v2023/vm_version2.rs
+++ b/test/src/specs/hardfork/v2023/vm_version2.rs
@@ -1,9 +1,6 @@
 use crate::{
     Node, Spec,
-    util::{
-        cell::gen_spendable,
-        check::{assert_epoch_should_less_than, is_transaction_committed},
-    },
+    util::{cell::gen_spendable, check::is_transaction_committed},
     utils::{assert_send_transaction_fail, wait_until},
 };
 use ckb_jsonrpc_types as rpc;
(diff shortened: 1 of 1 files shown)


eval-exec (dismissed):
Need to execute `make gen-rpc-doc` to update `rpc/README.md`

eval-exec (approved):

PR #4933, opened 2025-08-05
CI: add `--locked` for cargo build

eval-exec:
### Related changes

- add `--locked` for all `cargo build`

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ submodule-init:
 
 .PHONY: integration
 integration: submodule-init ## Run integration tests in "test" dir.
-	cargo build --bin ckb --release --features ${CKB_FEATURES}
+	cargo build --locked --bin ckb --release --features ${CKB_FEATURES}
 	RUST_BACKTRACE=1 RUST_LOG=${INTEGRATION_RUST_LOG} test/run.sh -- --bin "${CARGO_TARGET_DIR}/release/${BINARY_NAME}" ${CKB_TEST_ARGS}
 
 .PHONY: integration-release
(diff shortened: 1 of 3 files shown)


chenyukang (approved):

driftluo:
The cargo.lock file exists, so there is no need to add a --locked command.

eval-exec:
> The cargo.lock file exists, so there is no need to add a --locked command.

```bash
cargo build --help | grep locked
      --locked                Assert that `Cargo.lock` will remain unchanged
```

`cargo build` maybe modify the `Cargo.lock`'s content. If we add `--locked`, then we can ensure the `ckb` binary is using the deps version from `Cargo.lock`

eval-exec:
> The cargo.lock file exists, so there is no need to add a --locked command.

```bash
cargo build --help | grep locked
      --locked                Assert that `Cargo.lock` will remain unchanged
```

`cargo build` maybe modify the `Cargo.lock`'s content. If we add `--locked`, then we can ensure the `ckb` binary is using the deps version from `Cargo.lock`

driftluo:
```
--locked
           Asserts that the exact same dependencies and versions are used as when the existing Cargo.lock file was originally generated. Cargo will exit with an error when either of the following scenarios
           arises:

           •   The lock file is missing.

           •   Cargo attempted to change the lock file due to a different dependency resolution.

           It may be used in environments where deterministic builds are desired, such as in CI pipelines.
```

Cargo build will only modify the version in cargo.lock if it does not exist or there is a dependency problem with the version in lock. This --lock only prevents these two situations. Although I don't think it will happen, it is not a problem.

PR #4954, opened 2025-09-14
rich-indexer: Do not insert uncle block into database

eval-exec:

For rich-indexer, we don't really need to insert uncle block related info into database. This PR remove the related logic.
### Related changes

- Remove the uncle block logic for rich-indexer

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
Note: Rich Indexer: Uncle Block Tracking Removed

The rich-indexer no longer tracks uncle blocks. The `block_association_uncle` table has been removed from the schema.

**Action Required:** If you have an existing rich-indexer database, manually drop the unused table: DROP TABLE block_association_uncle;
```

### DIFF
--- a/util/rich-indexer/resources/create_postgres_index.sql
+++ b/util/rich-indexer/resources/create_postgres_index.sql
@@ -1,7 +1,6 @@
 CREATE INDEX IF NOT EXISTS "index_block_table_block_hash" ON "block" ("block_hash");
 
 CREATE INDEX IF NOT EXISTS "index_block_association_proposal_table_block_id" ON "block_association_proposal" ("block_id");
-CREATE INDEX IF NOT EXISTS "index_block_association_uncle_table_block_id" ON "block_association_uncle" ("block_id");
 
 CREATE INDEX IF NOT EXISTS "index_tx_table_tx_hash" ON "ckb_transaction" ("tx_hash");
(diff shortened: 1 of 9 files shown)


driftluo (approved):

doitian:
Suggest adding a release note that user can delete the table by running SQL `DROP table xxx`

eval-exec:
> Suggest adding a release note that user can delete the table by running SQL `DROP table xxx`

Done

chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `origin/develop...bd3b56830b98062d2944bace5e65ddceb3d3e35f`

Validation:
- Focused static tracing plus a minimal SQLite model showed that upgraded rich-indexer databases can retain legacy uncle rows. After the new rollback path deletes only the canonical block row, `ORDER BY id DESC` tip selection can return a stale uncle row.
- A focused `cargo test -p ckb-rich-indexer test_rollback_block_9` attempt did not complete locally because RocksDB binding generation failed with a missing `stdbool.h` header.

Notes:
- Please preserve compatibility cleanup for legacy `block_association_uncle` data, add a migration that removes old uncle rows, or make tip selection explicitly canonical rather than max row id.

PR #4956, opened 2025-09-17
Replace `SledBackend` by `ckb-rocksdb`

eval-exec:

This PR want to load/store Headers from/into ckb-rocksdb

Proposal: [xxx](url) 
Design doc: TODO

What's Changed:

### Related changes

- Remove `SledBackend`
- Load/Store `packed::HeaderView` from `ckb-rocksdb`
- Create a new `COLUMN: COLUMN_HEADER_INDEX` to store `number`->`hash` relation for `Headers`
- in processing headers, use insert headers by 1 batch insert transaction instead of 2000x sequencial insert transaction

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- Performance regression

### Release note 

```release-note
```


chainTe:
## Security Review

Result: 3 confirmed finding(s)
Reviewed range: `2f0cbbdacb484b69b2bbc7575f0fa19bf3a278a2..95c62175132b678df7bde3e13b890b4d7c5dd9c9`

Validation:
- Existing broad tests were not run locally; CI is expected to cover them after the branch is buildable.
- Runtime validation was blocked because `shared/src/types/header_map/memory.rs` contains unresolved conflict markers.
- Static source tracing found security-relevant sync regressions.

Notes:
- Mainnet `MIN_CHAIN_WORK_500K` is changed to zero, which disables the intended minimum-work gate before block download/assume-valid progression.
- `COLUMN_HEADER_INDEX` is keyed only by block number, but `get_ancestor_internal(base, number)` trusts it before proving the returned hash is an ancestor of `base`; valid side-branch headers can poison branch-specific ancestor lookups.
- Block fetching can now proceed from height-only/zero-work best-known headers because the previous total-difficulty superiority check is commented out.

PR #4979, opened 2025-10-17
feat: change compress

driftluo:

If the compressed data is larger than the original data, roll back to the original data

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```


driftluo:
merge to #4986

PR #4986, opened 2025-10-22
perf: improve compress impl

driftluo:

Optimize message compression and incorporate it into the codec to reduce the consumption of a full message copy for each communication

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/compress.rs
+++ b/network/src/compress.rs
@@ -138,25 +138,26 @@ impl LengthDelimitedCodecWithCompress {
             protocol_id,
         }
     }
+
+    fn process(&self, data: &[u8], flag: u8, dst: &mut BytesMut) -> Result<(), io::Error> {
+        let len = data.len() + 1;
+        if len > self.length_delimited.max_frame_length() {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "data too large",
+            ));
+        }
+        dst.reserve(4 + len);
+        dst.put_uint(len as u64, 4);
+        dst.put_u8(flag);
+        dst.extend_from_slice(data);
+        Ok(())
+    }
 }
 
 impl tokio_util::codec::Encoder<Bytes> for LengthDelimitedCodecWithCompress {
     type Error = io::Error;
     fn encode(&mut self, data: Bytes, dst: &mut BytesMut) -> Result<(), io::Error> {
-        let process = |data: &[u8], flag: u8, dst: &mut BytesMut| {
-            let len = data.len() + 1;
-            if len > self.length_delimited.max_frame_length() {
-                return Err(io::Error::new(
-                    io::ErrorKind::InvalidInput,
-                    "data too large",
-                ));
-            }
-            dst.reserve(4 + len);
-            dst.put_uint(len as u64, 4);
-            dst.put_u8(flag);
-            dst.extend_from_slice(data);
-            Ok(())
-        };
         if self.enable_compress && data.len() > COMPRESSION_SIZE_THRESHOLD {
             match SnapEncoder::new().compress_vec(&data) {
                 Ok(res) => {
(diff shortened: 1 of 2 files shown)


eval-exec (dismissed):

eval-exec (dismissed):

network/src/compress.rs
```diff
@@ -116,3 +117,142 @@ pub fn compress(src: Bytes) -> Bytes {
...
+                            if decompressed_bytes_len > MAX_UNCOMPRESSED_LEN {
+                                debug!(
+                                    "The limit for uncompressed bytes len is exceeded. limit: {}, len: {}",
+                                    MAX_UNCOMPRESSED_LEN, decompressed_bytes_len
+                                );
+                                return Err(io::ErrorKind::InvalidData.into());
+                            }
+                            let mut buf = vec![0; decompressed_bytes_len];
+                            match SnapDecoder::new().decompress(&data[1..], &mut buf) {
+                                Ok(_) => Ok(Some(BytesMut::from(Bytes::from(buf)))),
```
quake: here is a data copy when convert buf to Bytes, can we use decompress_vec to write the data to BytesMut directly?

driftluo: > Convert self into BytesMut.

> If bytes is unique for the entire original buffer, this will return a BytesMut with the contents of bytes without copying. If bytes is not unique for the entire original buffer, this will make a copy of bytes subset of the original buffer in a new BytesMut.

From the code implementation perspective, there is no copying, but relying on the internal implementation behavior is not good. I will change to：
```rust
let mut buf = BytesMut::zeroed(decompressed_bytes_len);
match SnapDecoder::new().decompress(&data[1..], &mut buf) {
    Ok(_) => Ok(Some(buf)),
    Err(e) => {
        debug!("snappy decompress error: {:?}", e);
        Err(io::ErrorKind::InvalidData.into())
    }
}
```

eval-exec (approved):

network/src/compress.rs
```diff
@@ -116,3 +117,142 @@ pub fn compress(src: Bytes) -> Bytes {
...
+            enable_compress,
+            protocol_id,
+        }
+    }
+}
+
+impl tokio_util::codec::Encoder<Bytes> for LengthDelimitedCodecWithCompress {
+    type Error = io::Error;
+    fn encode(&mut self, data: Bytes, dst: &mut BytesMut) -> Result<(), io::Error> {
+        let process = |data: &[u8], flag: u8, dst: &mut BytesMut| {
```
quake: Defining this logic as a helper method instead of an inline closure would make the code easier to read and reuse, and keeps the encode function more concise.

```suggestion
impl LengthDelimitedCodecWithCompress {
    fn process(&self, data: &[u8], flag: u8, dst: &mut BytesMut) -> { ... }
}
```

network/src/compress.rs
```diff
@@ -116,3 +117,142 @@ pub fn compress(src: Bytes) -> Bytes {
...
+impl tokio_util::codec::Decoder for LengthDelimitedCodecWithCompress {
+    type Item = BytesMut;
+    type Error = io::Error;
+    fn decode(&mut self, src: &mut BytesMut) -> Result<Option<BytesMut>, io::Error> {
+        if src.is_empty() {
+            return Ok(None);
+        }
+        match self.length_delimited.decode(src)? {
+            Some(mut data) => {
+                if (data[0] & COMPRESS_FLAG) != 0 {
```
quake, line 232: Accessing data[0] and data[1..] can panic if data is empty. A malicious peer could send a crafted frame with no payload (for example just the 4-byte length header like [0, 0, 0, 0]), need to check data length first. Also, please add a corresponding unit test to cover this edge case and verify that malformed or empty frames are handled safely without crashing.

```rust
    let mut buf = BytesMut::from(vec![0, 0, 0, 0]);
    let decoded = codec_with_compress.decode(&mut buf);
```

quake (changes requested):

network/src/compress.rs
```diff
@@ -116,3 +117,147 @@ pub fn compress(src: Bytes) -> Bytes {
...
+        protocol_id: p2p::ProtocolId,
+    ) -> Self {
+        Self {
+            length_delimited,
+            enable_compress,
+            protocol_id,
+        }
+    }
+
+    fn process(&mut self, data: &[u8], flag: u8, dst: &mut BytesMut) -> Result<(), io::Error> {
```
quake:
```suggestion
    fn process(&self, data: &[u8], flag: u8, dst: &mut BytesMut) -> Result<(), io::Error> {
```

quake (approved):

PR #4991, opened 2025-10-30
deps: checkout Cargo.lock from 56ca5f7 then sync

doitian:

Problem Summary: Performance degradation from https://github.com/nervosnetwork/ckb/compare/56ca5f7...1abea92

What's Changed:

Checkout Cargo.lock from 56ca5f7 then run `cargo check` to sync the lock file with `Cargo.toml`.

### Release note 

```release-note
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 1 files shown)


doitian:
Closed because #4992 is a more precise reverting.

PR #4997, opened 2025-11-06
windows: Hint windows user to install latest vc-redist if host vc-redist is lower than 14.44

### LINKED ISSUE #4995: v0.203.0 binary quits on Windows 10
**v0.203.0 quits with this computer:** 
image

_Originally posted by @joshyates1980 in https://github.com/nervosnetwork/ckb/discussions/4987#discussioncomment-14862005_

eval-exec:

This PR want to fix #4995

- Hint windows user to install/upgrade latest Visual C++ Redistribution if user's host miss it or version is lower than 14.44.

### Check List 

Tests 

- Unit test
- Integration test

Side effects

- None

### Release note 

```release-note
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,9 +10,75 @@ fn main() {
     #[cfg(feature = "tokio-trace")]
     console_subscriber::init();
 
+    #[cfg(target_os = "windows")]
+    check_msvc_version();
+
     let version = get_version();
     if let Some(exit_code) = run_app(version).err() {
         ::std::process::exit(exit_code.into());
+    }
+}
+
+#[cfg(target_os = "windows")]
+fn check_msvc_version() {
+    use winreg::RegKey;
+    use winreg::enums::*;
+    // if users msvc version less than 14.44, print a warning
+
+    fn get_vc_redist_version(arch: &str) -> io::Result<Option<String>> {
+        // arch: "x64" or "x86"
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -10,12 +10,78 @@ fn main() {
...
+        if parts.len() >= 4 {
+            let major = parts[0].parse().ok()?;
+            let minor = parts[1].parse().ok()?;
+            let bld = parts[2].parse().ok()?;
+            let rbld = parts[3].parse().ok()?;
+            Some((major, minor, bld, rbld))
+        } else {
+            None
+        }
+    }
```
chenyukang, line 55:
```suggestion
fn version_string_to_tuple(ver: &str) -> Option<(u32, u32, u32, u32)> {
    let mut parts = ver.split('.').filter_map(|s| s.parse::<u32>().ok());
    
    Some((
        parts.next()?,
        parts.next()?,
        parts.next()?,
        parts.next()?,
    ))
}
```

src/main.rs
```diff
@@ -10,12 +10,78 @@ fn main() {
...
+    fn is_version_at_least(current: &str, threshold: &str) -> bool {
```
chenyukang, line 56: how about use https://crates.io/crates/version-compare 
so we don't need to write code to parse and compare by ourself.

PR #5001, opened 2025-11-12
feat: relay use async send msg

driftluo:

async send avoids the problem of messages failing to be sent due to insufficient capacity during congestion.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
```

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -1509,15 +1509,12 @@ impl NetworkController {
     fn broadcast_inner<S: Spawn>(
         &self,
         quick: bool,
-        target: Option<SessionId>,
+        target: TargetSession,
         proto_id: ProtocolId,
         data: Bytes,
         handle: &S,
     ) {
         let async_control: ServiceAsyncControl = self.p2p_control.clone().into();
-        let target = target
-            .map(TargetSession::Single)
-            .unwrap_or(TargetSession::All);
 
         handle.spawn_task(async move {
             if quick {
@@ -1552,7 +1549,7 @@ impl NetworkController {
 
     /// Broadcast a message to all connected peers
     pub fn broadcast_with_handle<S: Spawn>(&self, proto_id: ProtocolId, data: Bytes, handle: &S) {
-        self.broadcast_inner(false, None, proto_id, data, handle)
+        self.broadcast_inner(false, TargetSession::All, proto_id, data, handle)
     }
 
     /// Broadcast a message to all connected peers through quick queue
@@ -1562,7 +1559,7 @@ impl NetworkController {
         data: Bytes,
         handle: &S,
     ) {
-        self.broadcast_inner(true, None, proto_id, data, handle)
+        self.broadcast_inner(true, TargetSession::All, proto_id, data, handle)
     }
 
     /// Send message to one connected peer
@@ -1573,7 +1570,13 @@ impl NetworkController {
         data: Bytes,
         handle: &S,
     ) {
-        self.broadcast_inner(false, Some(session_id), proto_id, data, handle)
+        self.broadcast_inner(
+            false,
+            TargetSession::Single(session_id),
+            proto_id,
+            data,
+            handle,
+        )
     }
 
     /// network message processing controller, always true, if false, discard any received messages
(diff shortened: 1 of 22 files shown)


eval-exec:
Please rebase onto develop branch.

eval-exec:
Ci failed: https://github.com/nervosnetwork/ckb/actions/runs/19323339311/job/55268951692?pr=5001

chain/src/init.rs
```diff
@@ -90,7 +90,7 @@ pub fn build_chain_services(
         })
         .expect("start preload_unverified_block should ok");
 
-    let (process_block_tx, process_block_rx) = channel::bounded(0);
+    let (process_block_tx, process_block_rx) = channel::bounded(24);
```
quake, line 93: How is this fixed value calculated?

network/src/network.rs
```diff
@@ -1429,6 +1434,30 @@ impl NetworkController {
         }
     }
 
+    fn broadcast_inner<S: Spawn>(
+        &self,
+        quick: bool,
+        target: Option<SessionId>,
```
quake:
```suggestion
        target: TargetSession,
```

use TargetSession directly.

sync/src/utils.rs
```diff
@@ -111,13 +147,22 @@ pub(crate) fn send_message_to<Message: Entity>(
     send_message(protocol_id, nc, peer_index, message)
 }
 
-pub(crate) fn quick_send_message_to<Message: Entity>(
-    nc: &dyn CKBProtocolContext,
+pub(crate) async fn send_message_to_async<Message: Entity>(
```
quake: suggest to follow `async fn async_xxx` naming convention
```suggestion
pub(crate) async fn async_send_message_to<Message: Entity>(
```

network/src/protocols/mod.rs
```diff
@@ -94,6 +101,13 @@ pub trait CKBProtocolContext: Send {
     fn quick_send_message_to(&self, peer_index: PeerIndex, data: Bytes) -> Result<(), Error>;
     /// Filter broadcast message through quick queue
     fn quick_filter_broadcast(&self, target: TargetSession, data: Bytes) -> Result<(), Error>;
+    /// Filter broadcast message through quick queue and specify protocol id
+    fn quick_filter_broadcast_with_proto(
```
quake, line 115: I cannot find this new fn usage in the PR

sync/src/relayer/mod.rs
```diff
@@ -721,66 +744,65 @@ fn build_and_broadcast_compact_block(
...
+            .build()
+    };
+    let light_client_peers: HashSet<PeerIndex> = nc
+        .connected_peers()
+        .into_iter()
+        .filter_map(|index| nc.get_peer(index).map(|peer| (index, peer)))
+        .filter(|(_id, peer)| peer.if_lightclient_subscribed)
+        .map(|(id, _)| id)
+        .collect();
+    if let Err(err) = handle.block_on(nc.async_quick_filter_broadcast_with_proto(
```
quake: light client won't relay the compact block to other peers, the message priority should be normal.

sync/src/relayer/mod.rs
```diff
@@ -558,21 +574,24 @@ impl Relayer {
...
-                let status = quick_send_message_to(nc, peer_index, &message);
-                if !status.is_ok() {
-                    ckb_logger::error!(
-                        "send RelayBlockProposal to {}, status: {:?}",
-                        peer_index,
-                        status
-                    );
-                }
-            };
+        async fn send_block_proposals(
```
quake: extract it to a helper fn and reuse with https://github.com/nervosnetwork/ckb/blob/5c6bfca16eb5b1b933128e531589be0dead8b2ba/sync/src/relayer/get_block_proposal_process.rs#L99

chain/src/init.rs
driftluo: There's no calculation basis here; it's just fixing 0 into a bounded value. Using 0 here is actually unreasonable.

@eval-exec and I discussed why it was written as 0 here, but there was no reason why it had to be 0.

The blocks received here come from three sources: RPC mining, sync protocol synchronization, and relay protocol synchronization. A value of 0 means that only one source of data can be received at a time; the other source must wait for processing to complete, but this waiting is not mandatory.

Therefore, a small value was filled in here.

Issue #5011, opened 2025-11-25
Limitations of indexer-related RPCs


driftluo:
## Goal or context for the feature

Previously, due to the large amount of data, some queries were already experiencing excessively long processing times. The most directly affected APIs were `get_cells`, `get_transactions`, and `get_cells_capacity`. 

While the two PRs(https://github.com/nervosnetwork/ckb/pull/4576, https://github.com/nervosnetwork/ckb/pull/4529) introduced configurable limits for `get_transactions` and `get_cells_capacity`, `get_cells` did not improve this. It does not support request limit restrictions, and its problem lies in the fact that it may need to iterate through a large number of cells with a small return value. This is not entirely consistent with the issues of the other two RPC APIs. Furthermore, users can construct an empty filter to iterate through all cells, resulting in significant resource waste and potentially enabling DoS attacks.

relations：https://github.com/nervosnetwork/ckb/pull/4469

## Describe the solution you'd like

The cost of these interfaces varies depending on the host machine's configuration. We plan to impose some time consumption limits on the interfaces, roughly as shown below:

```rust
struct TimeoutIterator<I> {
    inner: I,
    start_time: Instant,
    timeout: Duration,
    timed_out: bool,
}

impl<I: Iterator> Iterator for TimeoutIterator<I> {
    type Item = I::Item;

    fn next(&mut self) -> Option<Self::Item> {
        if self.start_time.elapsed() > self.timeout {
            self.timed_out = true;
            return None;
        }
        self.inner.next()
    }
}
```

During the iteration, time will be controlled. A single query will have a default maximum time interval of 5-10 seconds. If a timeout occurs, the query will be forcibly terminated, and an error will be returned.

Perhaps nginx should also be equipped with corresponding interface query timeout configurations, or provide a similar template with default configurations and timeout configurations for whether RPC interfaces are public or private.

cc：@chenyukang @phroi

quake:
Returning None when a timeout occurs will cause the outer code to interpret the iteration as having completed normally. This can lead to incomplete results being processed or reported. To properly signal that a timeout has occurred, the iterator should return a Result<Item, TimeoutError> instead, so that the caller can distinguish between a normal end of iteration and an early termination due to a timeout.

quake:
one timeout field should be enough:
```
struct TimeoutIterator<I> {
    inner: I,
    timeout: Instant,
}

fn next(&mut self) -> Option<Self::Item> {
        if Instant::now() > self.timeout{
[...]

PR #5012, opened 2025-11-25
feat: add default time cost limit on indexer rpc

driftluo:

Relation Issue Number:  #5011

As data volume increases, the cost of some RPCs increases linearly, while the return value remains relatively unchanged. Imposing hard timeout limits on these RPCs reduces the likelihood of attacks.

### Check List

Tests

- Unit test
- Integration test

### Release note

```release-note
Add a new config option `timeout_limit` under `indexer_v2` to timeout indexer RPC in 10 seconds by default.
```

### DIFF
--- a/sync/src/relayer/tests/compact_block_process.rs
+++ b/sync/src/relayer/tests/compact_block_process.rs
@@ -162,6 +162,7 @@ fn test_unknow_parent() {
     let message = packed::SyncMessage::new_builder().set(content).build();
     let data = message.as_bytes();
 
+    std::thread::sleep(std::time::Duration::from_millis(100));
     // send_getheaders_to_peer
     assert!(nc.has_sent(SupportProtocols::Sync.protocol_id(), peer_index, data));
 }
--- a/util/indexer/src/service.rs
+++ b/util/indexer/src/service.rs
@@ -16,13 +16,50 @@ use ckb_types::{H256, core, packed, prelude::*};
 use memchr::memmem;
 use rocksdb::{Direction, IteratorMode, prelude::*};
 
-use std::convert::TryInto;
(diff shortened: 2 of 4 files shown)


chenyukang:
cc @phroi

sync/src/relayer/tests/compact_block_process.rs
```diff
@@ -162,6 +162,7 @@ fn test_unknow_parent() {
     let message = packed::SyncMessage::new_builder().set(content).build();
     let data = message.as_bytes();
 
+    std::thread::sleep(std::time::Duration::from_millis(100));
     // send_getheaders_to_peer
```
chenyukang, line 166: this is for what?

util/indexer/src/service.rs
```diff
@@ -198,14 +239,14 @@ impl IndexerHandle {
...
-        let iter = snapshot.iterator(mode).skip(skip);
+        let mut iter = TimeoutIterator::new(snapshot.iterator(mode).skip(skip), self.timeout_limit);
 
         let mut last_key = Vec::new();
         let pool = self
             .pool
             .as_ref()
             .map(|pool| pool.read().expect("acquire lock"));
-        let cells = iter
+        let cells = (&mut iter)
```
quake:
```suggestion
        let cells = iter.by_ref()
```

sync/src/relayer/tests/compact_block_process.rs
driftluo: https://github.com/nervosnetwork/ckb/pull/5001 We need to give some time to test asynchronous sending.

doitian (approved):

doitian:
Added a release note on the new added config option.

PR #5041, opened 2025-12-15
fix root crate publishing

doitian:

Problem Summary: `cargo publish` fails because of the dependency on the local crate `ckb-test`.

What's Changed:

- Remove `ckb-test` from dependencies of `ckb`
- Run `ckb-test` using `cargo run -p ckb-test`

### Check List 

Tests 

    - The integration workflows should pass
    - `cargo publish -n` should succeed

### Release note 

```release-note
```

### DIFF
--- a/test/src/main.rs
+++ b/test/src/main.rs
@@ -1,5 +1,19 @@
 #![allow(missing_docs)]
 
+use std::env;
+use std::path::Path;
+
 fn main() {
+    // Check if working directory is in test/ by looking for template/ckb.toml
+    if !Path::new("template/ckb.toml").exists() {
+        // Try test/template/ckb.toml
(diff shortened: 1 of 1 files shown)


eval-exec (commented):
How about add `publish = false` in `ckb-test` crate's Cargo.toml ?

https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

doitian:
> How about add `publish = false` in `ckb-test` crate's Cargo.toml ?
> 
> https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

`cargo publish` ignores that. It just tells cargo not publishing ckb-test, but it still requires ckb-test because it is listed as a dependency.

PR #5044, opened 2025-12-15
fix ci integration tests on Windows

doitian:

Problem Summary: Workflow `ci_integration_tests_windows` fails on develop because the path to the ckb executable is not properly quoted.

What's Changed: Quote the CKB executable path in Makefile

### Check List 

Tests 

- Integration test
    - Run `make integration` in Windows using make from mingw

### Release note 

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,13 +83,13 @@ integration: integration-release
 .PHONY: integration-release
 integration-release: submodule-init
 	cargo build --locked --bin ckb ${VERBOSE} --release --features "deadlock_detection"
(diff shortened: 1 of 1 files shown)


doitian:
Please review as soon as possible. This will block merging into develop because the workflow `ci_integration_tests_windows.yaml` will fail after time out (about 1 hour). When a PR has passed required ubuntu test suites but is not merged immediately, it will be blocked by the failed `ci_integration_tests_windows.yaml`.

Issue #5051, opened 2025-12-17
Replace Self-Hosted Runners with GitHub-Hosted Custom Runners


doitian:
The GitHub-hosted runners have lower maintenance cost and will become cheaper.

From GitHub:

> - On January 1, 2026, you will receive up to a 39% reduction in the net price of GitHub-hosted runners.
> - On March 1, 2026, we are introducing a new $0.002 per-minute GitHub Actions cloud platform charge that will apply to self-hosted runner usage. Any usage subject to this charge will count toward the minutes included in your plan.

We have already remove self-hosted Windows runners in #5043 and will remove macOS runners in #5050 . Now we only have self-hosted Linux runners for `ci_*.yaml` and `coverage_report.yaml` workflows.

Actions:

- [x] Decide the GitHub-hosted runner configuration and compare it with self-hosted runners
- [x] Create GitHub-hosted runners
- [x] Update workflows running on self-hosted runners to use the custom GitHub-hosted runners
- [ ] Remove Self-hosted runners from GitHub
- [ ] Delete Self-hosted runners or relocate them for other usage.

doitian:
Benchmark shows Blacksmith is a better choice.

Workflow | Blacksmith (16C) | Self-hosted Runner (16C) | GitHub Hosted Runner (16C)
-- | -- | -- | --
ci_cargo_deny_ubuntu.yaml | 2m 17s | 3m 21s | 8m 53s
ci_linters_ubuntu.yaml | 2m 47s | 11m 48s | 7m 10s
ci_benchmarks_ubuntu.yaml | 3m 25s | 11m 13s | 8m 42s
ci_aarch64_build_ubuntu.yaml | 5m 24s | 17m 13s | 16m 23s
ci_quick_checks_ubuntu.yaml | 5m 49s | 17m 9s | 13m 2s
ci_unit_tests_ubuntu.yaml | 6m 15s | 18m 45s | 13m 27s
ci_integration_tests_ubuntu.yaml | 8m 27s | 36m 19s | 21m 20s
Estimated Monthly Expenses (USD) | 300-500 | ≈2000 | 400-600

doitian:
Related: #4567

PR #5055, opened 2025-12-18
ci: stop using dtolnay/rust-toolchain

doitian:

Problem Summary: There's little benefit to use the action. Call `rustup` directly.

What's Changed: Avoid the dependency by calling `rustup` directly. This also avoid
specifying the toolchain version because it has already been specified
in `rust-toolchain.toml`.

### Check List 

Tests 

    Watch workflow running status

### Release note 

```release-note
```


doitian:
rustup is not installed in self-hosted runners. We may remove it after #5051

PR #4993, opened 2025-10-31
Upgrade rust-toolchain from 1.85.0 to 1.92.0

eval-exec:
### What problem does this PR solve?

Wait https://github.com/nervosnetwork/ckb-docker-builder/pull/34

### Related changes

- Upgrade rust toolchain from 1.85.0 to 1.91.0

### Check List 

Tests 

- Unit test
- Integration test
- Manual test (add detailed scripts or steps below)
- No code ci-runs-only: [ quick_checks,linters ]

Side effects

- Performance regression
- Breaking backward compatibility

### Release note 

```release-note
Title Only: Include only the PR title in the release note.
```

PR #5076, opened 2025-12-25
feat: bump crates MSRV to 1.92.0

doitian:

Problem Summary: MSRV is missing in crates Cargo.toml. It may break user projects because some crates start use syntax only available in 1.92.0.

What's Changed:

- Add MSRV into Cargo.toml for all crates
- Mark the PR with `feat:` so release-plz will bump the minor version for all crates. This should be done in #4993 but have not. If you bump MSRV in a patch release (e.g., 1.1.1 → 1.1.2), users automatically get the broken build when they run cargo update, with no easy way to opt out.

### Check List 

Tests 

- No code

### Release note 

```release-note
```

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2024"
 description = "CKB Benches"
 homepage = "https://github.com/nervosnetwork/ckb"
 repository = "https://github.com/nervosnetwork/ckb"
+rust-version = "1.92.0"
 
 [dependencies]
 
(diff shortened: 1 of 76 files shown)


zhangsoledad (approved):

eval-exec (approved):

eval-exec:
I have never seen `GetBlockFilterCheckPoints` failed before.

https://github.com/nervosnetwork/ckb/actions/runs/20567422672/job/59067957318

PR #5079, opened 2026-01-04
spec: support data field in IssuedCell

Officeyutong:
Table `IssuedCell` in spec cannot specify output data of a cell

This PR adds a field named `data` and can be ignored (to keep compatibility with previous spec), used to specify data of cells

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression


doitian:
If you add `data`, maybe you also need `type`.

Officeyutong:
> If you add `data`, maybe you also need `type`.

Field `type` is added now

quake (dismissed):

PR #5157, opened 2026-03-24
fix!: bump crate versions for breaking ckb_schemars to schemars migration

doitian:

Problem Summary:

The switch from `ckb_schemars` to `schemars` in #5128 was a breaking change that was incorrectly released as patch/minor bumps by release-plz (#5129). This causes a semver resolution issue:

- `ckb-fixed-hash v1.1.0` on crates.io depends on `ckb-fixed-hash-core = "^1"` (`>=1.0.0, <2.0.0`)
- Cargo resolves this to `ckb-fixed-hash-core v1.1.1`, which depends on `schemars` instead of `ckb_schemars`
- Downstream consumers expecting `ckb_schemars` are silently broken

The same issue applies to `ckb-jsonrpc-types` and `ckb-rpc`.

Bump crates with the breaking `schemars` dependency change to major version 2.0.0, so Cargo's semver resolver cannot mix old consumers with new dependencies:

| Crate | Old Version | New Version | Reason |
|---|---|---|---|
| `ckb-fixed-hash-core` | `1.1.1` | `2.0.0` | Breaking: `ckb_schemars` → `schemars` |
| `ckb-jsonrpc-types` | `1.1.1` | `2.0.0` | Breaking: `ckb_schemars` → `schemars` |
| `ckb-rpc` | `1.2.1` | `2.0.0` | Breaking: `ckb_schemars` → `schemars` |
| `ckb-fixed-hash` | `1.1.0` | `1.2.0` | Depends on `ckb-fixed-hash-core` |
| `ckb-fixed-hash-macros` | `1.1.0` | `1.2.0` | Depends on `ckb-fixed-hash-core` |

**After publishing, the following versions should be yanked from crates.io:**

- `ckb-fixed-hash-core v1.1.1`
- `ckb-jsonrpc-types v1.1.1`
- `ckb-rpc v1.2.1`

### Related changes

- Need to yank broken versions from crates.io after this is published

### Check List

Tests

- No code

Side effects

---
*This PR description was generated with the assistance of AI.*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1032,7 +1032,7 @@ dependencies = [
 
 [[package]]
 name = "ckb-fixed-hash"
-version = "1.1.0"
+version = "1.2.0"
 dependencies = [
  "ckb-fixed-hash-core",
  "ckb-fixed-hash-macros",
@@ -1040,7 +1040,7 @@ dependencies = [
 
 [[package]]
 name = "ckb-fixed-hash-core"
-version = "1.1.1"
+version = "2.0.0"
 dependencies = [
  "faster-hex",
  "schemars",
@@ -1051,7 +1051,7 @@ dependencies = [
 
 [[package]]
 name = "ckb-fixed-hash-macros"
-version = "1.1.0"
+version = "1.2.0"
 dependencies = [
  "ckb-fixed-hash-core",
  "proc-macro2",
@@ -1156,7 +1156,7 @@ dependencies = [
 
 [[package]]
 name = "ckb-jsonrpc-types"
-version = "1.1.1"
+version = "2.0.0"
 dependencies = [
  "ckb-types",
  "faster-hex",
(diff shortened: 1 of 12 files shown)


eval-exec (approved):

doitian:
release-plz will handle other indirect version bumping.

doitian:
Pending for discussion. Convert to draft to prevent merging.

doitian:
The breaking bumping means downstream projects must update dependency on ckb-rpc, ckb-jsonrpc-type, and ckb-fixed-hash-core to `version = "2"` to use the latest version.

quake:
not an issue, the downstream breaking is caused by ckb_mock_tx_types locked ckb-jsonrpc-types version to `=1.1.0`

https://github.com/nervosnetwork/ckb-vm-contrib/blob/main/Cargo.toml#L23

PR #5163, opened 2026-04-07
test: ensure temporary directories are cleaned up after tests

### LINKED ISSUE #4177: Unit tests and Integration tests do not clean up temporary files
## Bug Report
When `make test` and `make integration` finished, there are many temporary dirs created by ckb exists in `TMPDIR`.

### Expected Behavior
I expect temporary files created by unit tests and integration tests should be cleaned up.

### Environment

- **CKB version**:  [latest develop branch](https://github.com/nervosnetwork/ckb/tree/45104e08c18852260282f23858ce21afae4c9f34)
- **Chain**: /
- **Operating system**: `NixOS 23.11`
- **Arch**: x64 

### Additional context/Screenshots
This bug can reproduced by these steps:
```
mkdir /tmp/ckb_integration_test
export TMPDIR=/tmp/ckb_integration_test
make integration
# ...
# after make integration finish.  there are many temporary files created by integration test exists
❯ ls /tmp/ckb_integration_test/
 ckb-it-BlockSyncDuplicatedAndReconnect-net-uTXWNX             ckb-it-GetBlockFilterHashes-net-WrFZMk
 ckb-it-BlockSyncNonAncestorBestBlocks-net-DlHaBN              ckb-it-GetBlockFilters-net-BWcRPK
 ckb-it-BlockSyncOrphanBlocks-net-zZyivE                       ckb-it-GetBlockFiltersNotReachBatch-net-E6d6Nt
 ckb-it-BlockSyncRelayerCollaboration-net-koWF3K               ckb-it-GetBlocksTimeout-net-7292x1
 ckb-it-BlockTransactionsRelayParentOfOrphanBlock-net-gADiyy   ckb-it-HeaderSyncCycle-net-nh3SUA
 ckb-it-CompactBlockEmpty-net-0mnHuF                           ckb-it-InvalidLocatorSize-net-ngAWP3
 ckb-it-CompactBlockEmptyParentUnknown-net-kSbUtY              ckb-it-LastCommonHeaderForPeerWithWorseChain-net-0fPqAX
 ckb-it-CompactBlockLoseGetBlockTransactions-net-GvEHtP        ckb-it-MalformedMessage-net-gLGwqk
 ckb-it-CompactBlockMissingFreshTxs-net-2LLVdx                 ckb-it-MalformedMessageWithWhitelist-net-j4usQO
 ckb-it-CompactBlockMissingNotFreshTxs-net-KXXrh9              ckb-it-MissingUncleRequest-net-

...
...
```

And unit test have same issue:
```bash
mkdir /tmp/ckb_unit_test
export TMPDIR=/tmp/ckb_unit_test
make test
# ...
# after make test finish.  there are many temporary files created by integration test exists
❯ ls /tmp/ckb_unit_test/
 ckb-tmp-0cbhKs   ckb-tmp-c817WZ   ckb-tmp-H8Dndr   ckb-tmp-mjeZRA   ckb-tmp-omZCMf   ckb-tmp-UBezHw   ckb-tmp-ySywRL
 ckb-tmp-0Fw1AB   ckb-tmp-cTYvUD   ckb-tmp-HHEqzd   ckb-tmp-MOzYHI   ckb-tmp-P9rxMW   ckb-tmp-UIbdzL   ckb-tmp-z9Kpmw

...
...
```

jetjinser:

After running `make test`, temporary directories with the `ckb-tmp-` prefix remained in `$TMPDIR` because background tasks held `Arc<Shared>` references, preventing `TempDir` cleanup.
This is fixed by broadcasting the global exit signal or passing weak Arc.

For `make integration`, the old `temp_path` function would create a `TempDir`, immediately `drop` it, and return a plain `PathBuf`. Obviously the `TempDir` was gone right after that function returned, but the test code would go ahead and recreate the directory at that same path later—and nobody ever cleaned it up. So `ckb-it-*` dirs just piled up.
The fix is to stop throwing away the `TempDir` handle: the `Worker` now owns one root `TempDir` for the whole run, and everything else gets a subpath under it. When the root goes out of scope, the entire tree gets wiped automatically.

Issue Number: close #4177。

Unit tests (`make test`): `RpcTestSuite` got a `Drop` impl that calls `ckb_stop_handler::broadcast_exit_signals()` so background services actually stop before `Shared`(`TempDir`) goes away.

Integration tests (`make integration`): instead of `temp_path` throwing away the `TempDir`, the `Worker` creates a root `TempDir` upfront and passes the path down to `Node::new` and `Net::new`. All test subdirs live under that root, so cleanup is automatic and works.

### Related changes

This PR should not have any impact except test.

### DIFF
--- a/test/src/net.rs
+++ b/test/src/net.rs
@@ -1,5 +1,5 @@
 use crate::Node;
-use crate::utils::{find_available_port, message_name, temp_path, wait_until};
+use crate::utils::{TempPathBuf, find_available_port, message_name, wait_until};
 use ckb_app_config::NetworkConfig;
 use ckb_async_runtime::{Runtime, new_global_runtime};
 use ckb_chain_spec::consensus::Consensus;
@@ -12,7 +12,6 @@ use ckb_network::{
 };
 use ckb_util::Mutex;
 use std::collections::HashMap;
-use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -20,12 +19,15 @@ pub type NetMessage = (PeerIndex, ProtocolId, Bytes);
 
 pub struct Net {
     p2p_port: u16,
-    working_dir: PathBuf,
     protocols: Vec<SupportProtocols>,
     controller: NetworkController,
     register_rx: Receiver<(String, PeerIndex, Receiver<NetMessage>)>,
     receivers: HashMap<String, (PeerIndex, Receiver<NetMessage>)>,
+    // _async_runtime MUST be declared before _working_dir
+    // to ensure the runtime stops and releases file handles
+    // before the TempDir attempts to clean up.
     _async_runtime: Runtime,
+    _working_dir: TempPathBuf,
 }
 
 impl Net {
@@ -35,13 +37,13 @@ impl Net {
             "Net cannot initialize with empty protocols"
         );
         let p2p_port = find_available_port();
-        let working_dir = temp_path(spec_name, "net");
+        let working_dir = TempPathBuf::new(spec_name, "net");
 
         let p2p_listen = format!("/ip4/127.0.0.1/tcp/{p2p_port}").parse().unwrap();
         let network_state = Arc::new(
             NetworkState::from_config(NetworkConfig {
                 listen_addresses: vec![p2p_listen],
-                path: (&working_dir).into(),
+                path: working_dir.to_path_buf(),
                 max_peers: 128,
                 max_outbound_peers: 128,
                 discovery_local_address: true,
@@ -79,17 +81,13 @@ impl Net {
         .unwrap();
         Self {
             p2p_port,
-            working_dir,
             protocols,
             controller,
             register_rx,
             receivers: Default::default(),
             _async_runtime: async_runtime,
+            _working_dir: working_dir,
         }
-    }
-
-    pub fn working_dir(&self) -> &PathBuf {
-        &self.working_dir
     }
 
     pub fn p2p_listen(&self) -> String {
--- a/test/src/worker.rs
+++ b/test/src/worker.rs
@@ -1,3 +1,4 @@
+use crate::utils::TempPathBuf;
 use crate::{Spec, utils::nodes_panicked};
 use ckb_channel::{Receiver, Sender, unbounded};
 use ckb_logger::{error, info};
@@ -24,18 +25,20 @@ pub enum Notify {
     Done {
         spec_name: String,
         seconds: u64,
-        node_paths: Vec<PathBuf>,
+        node_paths: Vec<TempPathBuf>,
     },
     Error {
         spec_error: Box<dyn Any + Send>,
         spec_name: String,
         seconds: u64,
         node_log_paths: Vec<PathBuf>,
+        node_paths: Vec<TempPathBuf>,
     },
     Panick {
         spec_name: String,
         seconds: u64,
         node_log_paths: Vec<PathBuf>,
+        node_paths: Vec<TempPathBuf>,
     },
     Stop,
 }
@@ -163,9 +166,10 @@ impl Worker {
             .unwrap();
 
         let mut nodes = spec.before_run();
+        // Used to extend the lifecycle of TempPathBuf
         let node_paths = nodes
             .iter()
-            .map(|node| node.working_dir())
+            .map(|node| node.owned_working_dir())
             .collect::<Vec<_>>();
         let node_log_paths = nodes.iter().map(|node| node.log_path()).collect::<Vec<_>>();
         let result = panic::catch_unwind(panic::AssertUnwindSafe(|| {
@@ -184,6 +188,7 @@ impl Worker {
                     spec_name: spec.name().to_string(),
                     seconds: now.elapsed().as_secs(),
                     node_log_paths,
+                    node_paths,
                 })
                 .unwrap();
         } else if let Some(spec_error) = spec_error {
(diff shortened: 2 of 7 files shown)


quake:
thanks for the PR, good catch!
 
I have a small suggestion:

Net::new doesn't need a root_dir parameter. Its working_dir is only used internally for NetworkConfig::path (secret key, peer store, etc.), and Net::working_dir() is never called externally.

Instead of threading root_dir through Spec::before_run - Node - Net (which adds a root_dir field/method to Node and touches ~30 call sites), just let Net own a TempDir directly:
```rust
pub struct Net {
    // ...
    _temp_dir: TempDir, // auto-cleanup on drop
}
impl Net {
    // signature unchanged — no root_dir needed
    pub fn new(spec_name: &str, consensus: &Consensus, protocols: Vec<SupportProtocols>) -> Self {
        let temp_dir = tempfile::Builder::new()
            .prefix(&format!("ckb-it-{}-net-", spec_name))
            .tempdir()
            .expect("create tempdir failed");
        let working_dir = temp_dir.path().to_path_buf();
        // ...
    }
}
```
This eliminates the root_dir field and root_dir() method on Node, keeps Net::new's original signature, and avoids changing all ~30 call sites.

jetjinser:
> Instead of threading root_dir through Spec::before_run - Node - Net (which adds a root_dir field/method to Node and touches ~30 call sites), just let Net own a TempDir directly:
> 
> ```rust
> pub struct Net {
>     // ...
>     _temp_dir: TempDir, // auto-cleanup on drop
> }
> impl Net {
>     // signature unchanged — no root_dir needed
>     pub fn new(spec_name: &str, consensus: &Consensus, protocols: Vec<SupportProtocols>) -> Self {
>         let temp_dir = tempfile::Builder::new()
>             .prefix(&format!("ckb-it-{}-net-", spec_name))
>             .tempdir()
>             .expect("create tempdir failed");
>         let working_dir = temp_dir.path().to_path_buf();
>         // ...
>     }
> }
> ```
> 
> This eliminates the root_dir field and root_dir() method on Node, keeps Net::new's original signature, and avoids changing all ~30 call sites.

Thanks for the review and advice!

However, we cannot simply bind the lifetime of `TempDir` to `Net`. The `NetworkStat` owns a `PathBuf` from the `TempDir`, which `DumpPeerStoreService` uses to keep active file handles open. Since `DumpPeerStoreService` is dropped *after* `Net`, these unreleased handles prevent the `TempDir` from cleaning up the directory properly (i.e., it fails silently).

Ideally, this would be the best approach for cleanup—in fact, it’s exactly what I attempted before implementing the current solution in this PR. 

Please let me know if I’ve overlooked any details or if you have further suggestions!

quake:
> However, we cannot simply bind the lifetime of `TempDir` to `Net`. The `NetworkStat` owns a `PathBuf` from the `TempDir`, which `DumpPeerStoreService` uses to keep active file handles open. Since `DumpPeerStoreService` is dropped _after_ `Net`, these unreleased handles prevent the `TempDir` from cleaning up the directory properly (i.e., it fails silently).
> 
> Ideally, this would be the best approach for cleanup—in fact, it’s exactly what I attempted before implementing the current solution in this PR.
> 
> Please let me know if I’ve overlooked any details or if you have further suggestions!

_async_runtime is declared before _working_dir, so the runtime drops first. This shuts down all spawned tasks including DumpPeerStoreService, which runs its Drop impl (writing peer store to disk) while the directory still exists. Only after that does _working_dir: TempDir drop and clean up the directory.

even if there's a async timing issue, it probably doesn't matter here. DumpPeerStoreService::drop() writes peer store data, which is irrelevant in a test context, we're about to delete the entire directory anyway. A failed write just produces a warn-level log.

Have you tried the simpler approach locally, letting Net own its own TempDir internally (keeping the current Net::new signature unchanged)?. If you haven't tried it, I can give it a shot on my machine to verify it works.

jetjinser:
> _async_runtime is declared before _working_dir, so the runtime drops first. This shuts down all spawned tasks including DumpPeerStoreService, which runs its Drop impl (writing peer store to disk) while the directory still exists. Only after that does _working_dir: TempDir drop and clean up the directory.

Yes! I completely forgot about that.

Based on your suggestion, I pushed a new commit.

test/src/worker.rs
```diff
@@ -163,9 +166,10 @@ impl Worker {
             .unwrap();
 
         let mut nodes = spec.before_run();
+        // Used to extend the lifecycle of TempPathBuf
         let node_paths = nodes
             .iter()
-            .map(|node| node.working_dir())
+            .map(|node| node.owned_working_dir())
```
jetjinser, line 172: Here is the reason for the changes to the `Notify` struct: `node_paths` is now passed along with `Notify` to ensure that the `TempDir` within the `Node` isn't dropped prematurely when the current function ends. This prevents potential errors in subsequent operations.

quake (approved):

test/src/net.rs
```diff
@@ -12,20 +12,22 @@ use ckb_network::{
...
 pub struct Net {
     p2p_port: u16,
-    working_dir: PathBuf,
     protocols: Vec<SupportProtocols>,
     controller: NetworkController,
     register_rx: Receiver<(String, PeerIndex, Receiver<NetMessage>)>,
     receivers: HashMap<String, (PeerIndex, Receiver<NetMessage>)>,
+    // _async_runtime MUST be declared before _working_dir
+    // to ensure the runtime stops and releases file handles
+    // before the TempDir attempts to clean up.
```
eval-exec, line 28: Maybe we can combine `Runtime` and `TempPathBuf` to one struct? Cc @quake

PR #5168, opened 2026-04-11
feat(network/onion): detect liveness via TcpStream close

### LINKED ISSUE #5161: network/onion: detect Tor liveness via socket close instead of polling
`OnionService` polls `GETINFO uptime` every 3 seconds to check if Tor is alive (https://github.com/nervosnetwork/ckb/blob/3d0da7c3c1f0d6901a6637966b7cb0ee83d96a33/util/onion/src/onion_service.rs#L130-L149). This wastes control-port traffic and adds a 3-second blind window before CKB notices Tor is gone.

React to the control-connection close instead of polling — either by detecting when the `TcpStream` closes, or by subscribing to `SETEVENTS STATUS_CLIENT`.

For reference, Bitcoin Core uses a libevent `BEV_EVENT_EOF` callback to detect the close instantly and then triggers a reconnect handler:

- Event callback: https://github.com/bitcoin/bitcoin/blob/faa18dceba1dd69703a252f751c233d227164689/src/torcontrol.cpp#L133-L148
- Reconnect handler: https://github.com/bitcoin/bitcoin/blob/faa18dceba1dd69703a252f751c233d227164689/src/torcontrol.cpp#L634-L652

jetjinser:

Issue Number: close #5161 
Follow-up to #5271.

Problem Summary:

The onion service currently polls the Tor daemon every 3 seconds via `get_uptime()` to detect connection loss. This adds unnecessary latency (up to 3s for failure detection) and constant I/O overhead. A passive, event-driven approach would react instantly and reduce resource usage.

What's Changed:

- An `oneshot` channel notifies `TorController` when either copy direction terminates (EOF or error).
- Replaced the 3-second ticker loop in `OnionService` with a branch awaiting `wait_for_disconnect()`.

### Check List 

Tests 

- [ ] Unit test
- [ ] Integration test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

Manual test steps:
1. `ckb run` with config listen_on_onion = true.
2. Observe the onion service starts successfully.
3. Kill/exit the Tor process.
4. Verify the node logs `"Tor control connection lost"` within milliseconds, not after a multi-second delay.
5. Restart the Tor daemon and verify the onion service can be re-established.

Side effects

- ~~Performance regression~~
- ~~Breaking backward compatibility~~

### DIFF
--- a/util/onion/src/tor_controller.rs
+++ b/util/onion/src/tor_controller.rs
@@ -101,6 +101,16 @@ impl TorController {
         Ok(())
     }
 
+    /// Waits asynchronously until the underlying TCP connection to the Tor
+    /// controller is severed (either cleanly closed or due to an I/O error).
+    ///
+    /// Returns `Some(())` when the connection is lost, or `None` if the internal
+    /// notification channel was closed unexpectedly (which should not occur during
+    /// normal operation).
+    pub async fn wait_for_disconnect(self) -> bool {
+        self.inner.wait_for_disconnect().await
+    }
+
     /// Add a new v3 onion service to the Tor server.
     pub async fn add_onion_v3(
         &mut self,
--- a/util/onion/src/onion_service.rs
+++ b/util/onion/src/onion_service.rs
@@ -128,22 +128,15 @@ impl OnionService {
         );
 
         self.handle.spawn(async move {
-            let mut ticker = tokio::time::interval(tokio::time::Duration::from_secs(3));
-            loop {
-                tokio::select! {
-                    _ = ticker.tick() => {
-                        let uptime = tor_controller.get_uptime().await;
-                        if let Err(err) = uptime {
-                            error!("Failed to get tor server uptime: {:?}", err);
-                            drop(tor_server_alive_tx);
-                            return;
-                        }
-                    }
-                    _ = stop_rx.cancelled() => {
-                        info!("OnionService received stop signal, exiting...");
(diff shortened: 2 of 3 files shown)


util/onion/src/tor_controller.rs
```diff
@@ -40,15 +53,45 @@ impl TorController {
...
+                        Err(e) => error!("Tor TCP write error: {}", e),
+                    }
+                    _ = disconnect_tx.send(());
+                }
+            }
+
+            debug!("Tor DuplexStream exited.");
+        });
+
+        let mut utc: UnauthenticatedConn<_> = UnauthenticatedConn::new(client);
```
eval-exec: > `UnauthenticatedConn<_>`

Could  `UnauthenticatedConn` replace `_` with the real inner type?

jetjinser: done :)

eval-exec:
I'm thinking we might not need to depend on torut after all. :thinking:

jetjinser:
> I'm thinking we might not need to depend on torut after all. 🤔

Sounds reasonable, I'll try to do so

jetjinser:
@eval-exec Hi, I made a PR for this attempt: #5271

jetjinser:
Since #5271 has been merged, I pushed a simpler implementation and updated the PR description.
Please take a look @eval-exec :)

eval-exec (approved):

PR #5219, opened 2026-05-29
Fix some public security issues

Officeyutong:

This PR will fix the following public security issues:
- An overflow in consensus.rs which will happen after 256 years
- Two overflows in peer_store_impl & net.rs set_ban. There overflows can be triggered by passing specific arguments, but has little impact.
- Compact-block relay accepted headers as valid before checking the consensus epoch difficulty, allowing low-difficulty invalid compact blocks to poison relay state and trigger reconstruction requests.
- Identify accepted untrusted DNS listen addresses into the peer store, allowing malicious peers to influence later outbound dialing targets.
- RecentReject failed to update its in-memory key count after writes, allowing remote rejected transactions to bypass the configured count limit and grow the reject database.
- ConnectionRequest trusted an unauthenticated from peer id, allowing forged hole-punching state to trigger outbound NAT traversal attempts to attacker-controlled addresses.
- HeaderVerifier failed to enforce the consensus block header version, allowing nonzero-version blocks to be accepted through some ingress paths while SendHeaders rejected them.
- ShortIdsVerifier failed to bound compact-block short IDs, allowing oversized compact blocks to amplify relay reconstruction work and pending-state growth.

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression

### DIFF
--- a/network/src/protocols/hole_punching/component/connection_request.rs
+++ b/network/src/protocols/hole_punching/component/connection_request.rs
@@ -112,6 +112,19 @@ impl<'a> ConnectionRequestProcess<'a> {
             Ok(content) => content,
             Err(status) => return status,
         };
+        if content.route.is_empty() {
+            let authenticated_peer_id = self
+                .protocol
+                .network_state
+                .peer_registry
+                .read()
+                .get_peer(self.peer)
+                .and_then(|peer| extract_peer_id(&peer.connected_addr));
+            if authenticated_peer_id.as_ref() != Some(&content.from) {
+                return StatusCode::InvalidFromPeerId
+                    .with_context("the from peer id does not match the session peer id");
+            }
+        }
         if content.listen_addrs.len() > ADDRS_COUNT_LIMIT || content.listen_addrs.is_empty() {
             return StatusCode::InvalidListenAddrLen
                 .with_context("the listen address count is too large or empty");
@@ -304,5 +317,102 @@ impl<'a> ConnectionRequestProcess<'a> {
                 }
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{borrow::Cow, sync::Arc};
+
+    use ckb_app_config::NetworkConfig;
+    use p2p::{
+        builder::ServiceBuilder,
+        multiaddr::{Multiaddr, Protocol},
+        secio::NoopKeyProvider,
+        service::SessionType,
+    };
+
+    use crate::{
+        NetworkState,
+        peer_store::PeerStore,
+        protocols::hole_punching::{
+            HolePunching,
+            status::{BAD_MESSAGE_BAN_TIME, StatusCode},
+        },
+    };
+
+    use super::*;
+
+    #[tokio::test]
+    async fn rejects_forged_from_for_listen_address_without_peer_id() {
+        let forged_from = PeerId::random();
+        let authenticated_peer = PeerId::random();
+        let target = PeerId::random();
+        let session_id = 1.into();
+        let tmp_dir = tempfile::tempdir().expect("create tempdir");
+        let remote_addr: Multiaddr = format!(
+            "/ip4/127.0.0.1/tcp/42/p2p/{}",
+            authenticated_peer.to_base58()
+        )
+        .parse()
+        .unwrap();
+        let listen_addr: Multiaddr = "/ip4/192.0.2.1/tcp/8333".parse().unwrap();
+
+        let network_state = Arc::new(
+            NetworkState::from_config(NetworkConfig {
+                max_peers: 4,
+                max_outbound_peers: 1,
+                path: tmp_dir.path().to_path_buf(),
+                public_addresses: vec![{
+                    let mut addr: Multiaddr = "/ip4/127.0.0.1/tcp/8115".parse().unwrap();
+                    addr.push(Protocol::P2P(Cow::Borrowed(target.as_bytes())));
+                    addr
+                }],
+                ..Default::default()
+            })
+            .expect("init network state"),
+        );
+        network_state
+            .peer_registry
+            .write()
+            .accept_peer(
+                remote_addr,
+                session_id,
+                SessionType::Inbound,
+                &mut PeerStore::default(),
+            )
+            .expect("accept peer");
+
+        let mut protocol = HolePunching::new(Arc::clone(&network_state));
+        let service = ServiceBuilder::<NoopKeyProvider>::default().build(());
+        let p2p_control = service.control().clone().into();
+
+        let listen_addrs = packed::AddressVec::new_builder()
+            .push(
+                packed::Address::new_builder()
+                    .bytes(listen_addr.to_vec())
+                    .build(),
+            )
+            .build();
+        let request = packed::ConnectionRequest::new_builder()
+            .from(forged_from.as_bytes())
+            .to(network_state.local_peer_id().as_bytes())
+            .max_hops(MAX_HOPS)
+            .listen_addrs(listen_addrs)
+            .build();
+
+        let status = ConnectionRequestProcess::new(
+            request.as_reader(),
+            &mut protocol,
+            session_id,
+            &p2p_control,
+            0,
+        )
+        .execute()
+        .await;
+
+        assert_eq!(status.code(), StatusCode::InvalidFromPeerId);
+        assert_eq!(status.should_ban(), Some(BAD_MESSAGE_BAN_TIME));
+        assert!(!protocol.pending_delivered.contains_key(&forged_from));
     }
 }
--- a/network/src/protocols/hole_punching/component/connection_request.rs
+++ b/network/src/protocols/hole_punching/component/connection_request.rs
@@ -401,14 +401,13 @@ mod tests {
 
     fn listen_addrs_without_peer_id() -> packed::AddressVec {
         let listen_addr: Multiaddr = "/ip4/192.0.2.1/tcp/8333".parse().unwrap();
-        let listen_addrs = packed::AddressVec::new_builder()
+        packed::AddressVec::new_builder()
             .push(
                 packed::Address::new_builder()
                     .bytes(listen_addr.to_vec())
                     .build(),
             )
-            .build();
-        listen_addrs
+            .build()
     }
 
     #[tokio::test]
--- a/network/src/protocols/hole_punching/component/connection_request.rs
+++ b/network/src/protocols/hole_punching/component/connection_request.rs
@@ -112,18 +112,17 @@ impl<'a> ConnectionRequestProcess<'a> {
             Ok(content) => content,
             Err(status) => return status,
         };
-        if content.route.is_empty() {
-            let authenticated_peer_id = self
-                .protocol
-                .network_state
-                .peer_registry
-                .read()
-                .get_peer(self.peer)
-                .and_then(|peer| extract_peer_id(&peer.connected_addr));
-            if authenticated_peer_id.as_ref() != Some(&content.from) {
-                return StatusCode::InvalidFromPeerId
-                    .with_context("the from peer id does not match the session peer id");
-            }
+        let authenticated_peer_id = self
+            .protocol
+            .network_state
+            .peer_registry
+            .read()
+            .get_peer(self.peer)
+            .and_then(|peer| extract_peer_id(&peer.connected_addr));
+        let expected_session_peer_id = content.route.last().unwrap_or(&content.from);
+        if authenticated_peer_id.as_ref() != Some(expected_session_peer_id) {
+            return StatusCode::InvalidFromPeerId
+                .with_context("the message route does not match the session peer id");
         }
         if content.listen_addrs.len() > ADDRS_COUNT_LIMIT || content.listen_addrs.is_empty() {
             return StatusCode::InvalidListenAddrLen
@@ -343,11 +342,15 @@ mod tests {
 
     use super::*;
 
-    #[tokio::test]
-    async fn rejects_forged_from_for_listen_address_without_peer_id() {
-        let forged_from = PeerId::random();
-        let authenticated_peer = PeerId::random();
-        let target = PeerId::random();
+    struct TestContext {
+        _tmp_dir: tempfile::TempDir,
+        network_state: Arc<NetworkState>,
+        protocol: HolePunching,
+        p2p_control: ServiceAsyncControl,
+        session_id: PeerIndex,
+    }
+
+    fn build_test_context(authenticated_peer: &PeerId) -> TestContext {
         let session_id = 1.into();
         let tmp_dir = tempfile::tempdir().expect("create tempdir");
         let remote_addr: Multiaddr = format!(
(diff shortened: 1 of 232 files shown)


sync/src/relayer/compact_block_process.rs
```diff
@@ -338,6 +339,28 @@ async fn contextual_check(
         }
     }
 
+    let parent_hash = compact_block_header.parent_hash();
+    if let Some(parent) = shared.store().get_block_header(&parent_hash) {
```
chenyukang: This still skips the epoch/compact-target validation when the parent is known only through `header_map`. The earlier `get_header_index_view(..., store_first)` path accepts such parents, but this new check only looks in the block store; in that case an invalid compact block can still pass contextual_check and enter pending/request missing transactions. Please validate against the parent header already accepted above, or reject when the expected epoch cannot be derived, and add a regression for a header-map parent with a wrong compact target.

network/src/protocols/hole_punching/component/connection_request.rs
```diff
@@ -112,6 +112,19 @@ impl<'a> ConnectionRequestProcess<'a> {
             Ok(content) => content,
             Err(status) => return status,
         };
+        if content.route.is_empty() {
```
chenyukang: Checking the session peer only when `route` is empty leaves a forged-from bypass: a direct peer can send a `ConnectionRequest` with a non-empty, unauthenticated route and an arbitrary `from`, and this branch will skip the binding check. The direct sender should be authenticated as `route.last().unwrap_or(from)` (or non-empty externally supplied routes should be rejected unless the last hop matches the session peer). Please add a regression for forged `from` with a non-empty route.

network/src/protocols/identify/mod.rs
```diff
@@ -29,6 +29,12 @@ const CHECK_TIMEOUT_INTERVAL: u64 = 1;
 const DEFAULT_TIMEOUT: u64 = 8;
 const MAX_ADDRS: usize = 10;
 
+pub(super) fn is_remote_listen_addr_allowed(addr: &Multiaddr, global_ip_only: bool) -> bool {
+    multiaddr_to_socketaddr(addr)
+        .map(|socket_addr| !global_ip_only || is_reachable(socket_addr.ip()))
+        .unwrap_or(false)
```
chenyukang: This rejects all remote listen addresses that cannot be converted to a `SocketAddr`, which also drops `/onion3` addresses. The identify protocol still explicitly allows advertising local `/onion3` listen addresses below, and the peer store has onion-specific handling, so this looks like a compatibility regression while fixing DNS loopback injection. Please keep rejecting unsafe DNS addresses without filtering out the existing onion address support.

tx-pool/src/component/recent_reject.rs
eval-exec, line 65: How about this for better readability:
```rust
self.total_keys_num = self.total_keys_num.saturating_add(1);

if self.total_keys_num > self.count_limit {
    self.shrink()?;
}
```

chainTe:
Security review finding:

The compact-block epoch/compact-target guard added in `contextual_check` only runs when the parent is already stored in the chain DB:

```rust
if let Some(parent) = shared.store().get_block_header(&parent_hash) {
    ... compare expected_epoch / expected_compact_target ...
}
```

But this same function accepts parents from `shared.get_header_index_view(...)`, which can come from `header_map` when the parent header was received but the block is not stored yet. In that case the new guard is skipped entirely. The ordinary `HeaderVerifier` only checks PoW, number, continuous epoch, and timestamp; it does not verify the exact consensus epoch fraction or compact target. After this bypass, the compact-block path calls `insert_valid_header`, then on missing transactions stores the compact block in `pending_compact_blocks` and sends `GetBlockTransactions`.

A peer can therefore send a header-map parent and then a child compact block with an invalid compact target/epoch that still passes the lighter header checks, causing the node to retain/request data for an invalid compact block until later full-block validation. This appears to leave the compact-block DoS fix incomplete for non-stored parents.

Suggested fix: resolve the parent through the same source that made `parent.is_some()` true, and perform the exact epoch/compact-target validation for header-map parents too, or require this check before queuing/requesting missing compact-block data. Please also add a test variant of `test_compact_block_accepts_parent_from_header_map` where the child has an invalid compact target/epoch and assert it is rejected and not inserted into `pending_compact_blocks`.

network/src/protocols/identify/mod.rs
Officeyutong: Fixed. Now `/onion3` procotol is explicitly allowed.

network/src/protocols/hole_punching/component/connection_request.rs
Officeyutong: Fixed

sync/src/relayer/compact_block_process.rs
Officeyutong: Fixed

tx-pool/src/component/recent_reject.rs
Officeyutong: Fixed

Officeyutong:
> Security review finding:
> 
> The compact-block epoch/compact-target guard added in `contextual_check` only runs when the parent is already stored in the chain DB:
> 
> ```rust
> if let Some(parent) = shared.store().get_block_header(&parent_hash) {
>     ... compare expected_epoch / expected_compact_target ...
> }
> ```
> 
> But this same function accepts parents from `shared.get_header_index_view(...)`, which can come from `header_map` when the parent header was received but the block is not stored yet. In that case the new guard is skipped entirely. The ordinary `HeaderVerifier` only checks PoW, number, continuous epoch, and timestamp; it does not verify the exact consensus epoch fraction or compact target. After this bypass, the compact-block path calls `insert_valid_header`, then on missing transactions stores the compact block in `pending_compact_blocks` and sends `GetBlockTransactions`.
> 
> A peer can therefore send a header-map parent and then a child compact block with an invalid compact target/epoch that still passes the lighter header checks, causing the node to retain/request data for an invalid compact block until later full-block validation. This appears to leave the compact-block DoS fix incomplete for non-stored parents.
> 
> Suggested fix: resolve the parent through the same source that made `parent.is_some()` true, and perform the exact epoch/compact-target validation for header-map parents too, or require this check before queuing/requesting missing compact-block data. Please also add a test variant of `test_compact_block_accepts_parent_from_header_map` where the child has an invalid compact target/epoch and assert it is rejected and not inserted into `pending_compact_blocks`.

This is already addressed in the current branch. contextual_check now validates the compact block against the parent resolved by get_header_index_view, including header-map-only parents, and rejects the compact block if the expected epoch cannot be derived. I also added test_invalid_compact_target_with_parent_from_header_map_does_not_enter_pending to cover the non-stored parent case and assert it does not enter pending or send GetBlockTransactions.

sync/src/relayer/compact_block_process.rs
```diff
@@ -338,6 +339,46 @@ async fn contextual_check(
...
+            None
+        }
+    };
+
+    let Some(epoch) = maybe_epoch else {
+        shared
+            .shared()
+            .insert_block_status(block_hash.clone(), BlockStatus::BLOCK_INVALID);
+        return StatusCode::CompactBlockHasInvalidHeader.with_context(format!(
+            "{block_hash} failed to derive expected epoch from parent {parent_hash}"
```
chenyukang, line 362: it's better to return temporary error here?

```diff
diff --git a/sync/src/relayer/compact_block_process.rs b/sync/src/relayer/compact_block_process.rs
index d03794c0c..ce784f3b5 100644
--- a/sync/src/relayer/compact_block_process.rs
+++ b/sync/src/relayer/compact_block_process.rs
@@ -358,10 +358,7 @@ async fn contextual_check(
     };
 
     let Some(epoch) = maybe_epoch else {
-        shared
-            .shared()
-            .insert_block_status(block_hash.clone(), BlockStatus::BLOCK_INVALID);
-        return StatusCode::CompactBlockHasInvalidHeader.with_context(format!(
+        return StatusCode::CompactBlockRequiresParent.with_context(format!(
             "{block_hash} failed to derive expected epoch from parent {parent_hash}"
         ));
     };
diff --git a/sync/src/relayer/tests/compact_block_process.rs b/sync/src/relayer/tests/compact_block_process.rs
index 6421fbd29..64447edfd 100644
--- a/sync/src/relayer/tests/compact_block_process.rs
+++ b/sync/src/relayer/tests/compact_block_process.rs
@@ -47,6 +47,27 @@ fn new_non_tail_header_map_parent(relayer: &crate::relayer::Relayer) -> HeaderVi
     panic!("failed to build a header-map parent with derivable next epoch");
 }
 
+fn new_tail_header_map_parent(relayer: &crate::relayer::Relayer) -> HeaderView {
+    let shared = relayer.shared.shared();
+    let tip_number = relayer.shared.active_chain().tip_header().number();
+
+    for number in (0..=tip_number).rev() {
+        let stored_parent = shared
+            .store()
+            .get_block_hash(number)
+            .and_then(|block_hash| shared.store().get_block_header(&block_hash))
+            .unwrap();
+        let header = new_header_builder(shared, &stored_parent).build();
+        let epoch = header.epoch();
+
+        if epoch.index() + 1 == epoch.length() {
+            return header;
+        }
+    }
+
+    panic!("failed to build a tail header-map parent");
+}
+
 fn child_epoch_after(parent_epoch: EpochNumberWithFraction) -> EpochNumberWithFraction {
     assert!(parent_epoch.index() + 1 < parent_epoch.length());
     EpochNumberWithFraction::new(
@@ -563,6 +584,93 @@ fn test_invalid_compact_target_with_parent_from_header_map_does_not_enter_pendin
     assert!(!nc.has_sent(SupportProtocols::RelayV3.protocol_id(), peer_index, data));
 }
 
+#[test]
+fn test_compact_block_with_tail_parent_from_header_map_does_not_mark_invalid() {
+    let (_chain, relayer, _) = build_chain(20);
+    let header_map_parent = new_tail_header_map_parent(&relayer);
+    let parent_epoch = header_map_parent.epoch();
+    let child_epoch =
+        EpochNumberWithFraction::new(parent_epoch.number() + 1, 0, parent_epoch.length());
+    let child_header = HeaderBuilder::default()
+        .parent_hash(header_map_parent.hash())
+        .number(header_map_parent.number() + 1)
+        .timestamp(header_map_parent.timestamp() + 1)
+        .epoch(child_epoch)
+        .compact_target(header_map_parent.compact_target())
+        .build();
+
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+    let peer_index: PeerIndex = 100.into();
+    relayer
+        .shared()
+        .insert_valid_header(peer_index, &header_map_parent);
+    assert!(
+        relayer
+            .shared()
+            .store()
+            .get_block_header(&header_map_parent.hash())
+            .is_none()
+    );
+
+    let block = BlockBuilder::default()
+        .header(child_header)
+        .transaction(TransactionBuilder::default().build())
+        .transaction(
+            TransactionBuilder::default()
+                .output(
+                    CellOutputBuilder::default()
+                        .capacity(Capacity::bytes(1).unwrap())
+                        .build(),
+                )
+                .output_data(Bytes::new())
+                .build(),
+        )
+        .build();
+    let block_hash = block.header().hash();
+
+    let mut prefilled_transactions_indexes = HashSet::new();
+    prefilled_transactions_indexes.insert(0);
+    let compact_block = CompactBlock::build_from_block(&block, &prefilled_transactions_indexes);
+
+    let mock_protocol_context = MockProtocolContext::new(SupportProtocols::RelayV3);
+    let nc = Arc::new(mock_protocol_context);
+    let compact_block_process = CompactBlockProcess::new(
+        compact_block.as_reader(),
+        &relayer,
+        Arc::<MockProtocolContext>::clone(&nc),
+        peer_index,
+    );
+
+    assert_eq!(
+        rt.block_on(compact_block_process.execute()),
+        StatusCode::CompactBlockRequiresParent.into()
+    );
+    assert_ne!(
+        relayer
+            .shared()
+            .active_chain()
+            .get_block_status(&block_hash),
+        BlockStatus::BLOCK_INVALID
+    );
+
+    let pending_compact_blocks = rt.block_on(relayer.shared.state().pending_compact_blocks());
+    assert!(pending_compact_blocks.get(&block_hash).is_none());
+
+    let content = packed::GetBlockTransactions::new_builder()
+        .block_hash(block_hash)
+        .indexes([1u32])
+        .uncle_indexes(Vec::<u32>::new())
+        .build();
+    let message = packed::RelayMessage::new_builder().set(content).build();
+    let data = message.as_bytes();
+
+    std::thread::sleep(std::time::Duration::from_millis(100));
+    assert!(!nc.has_sent(SupportProtocols::RelayV3.protocol_id(), peer_index, data));
+}
+
 #[test]
 fn test_invalid_difficulty_compact_block_does_not_enter_pending() {
     let (_chain, relayer, _) = build_chain(5);

```

Officeyutong: Now the block won't be marked as invalid

gpBlockchain:
Correction: the earlier comment reviewed a superseded revision range and its no-findings conclusion should not be relied on. The security review is complete for exact range `63223540a46674529007bfa34f745aa04548a78e..4388d53515545ecf418c26237c809f40abde0671`; findings and evidence are being coordinated through the private security-review record. All 22 changed files were reviewed with exact-revision source traces. Broad test suites and live-network exercises were not run.

zhangsoledad (dismissed):

chenyukang:
> @gpBlockchain wrote:
> Security review completed for PR #5219 at head a9fbf4b3554fe7582c4020237840fd4e689d7bf4 against base 20d76a939fdc3df084e6b6d77659243cafa80b5a. I found no reportable security issues in the reviewed diff. Targeted checks covered the P2P hole-punching/identify changes, compact-block validation, header version validation, reward halving behavior, peer ban overflow handling, and tx-pool recent-reject accounting. The RPC set_ban regression tests were attempted but remained blocked by an RPC test-harness EOF during response decode before the method assertions ran, so CI or a known-good RPC test environment should still cover that path.

This PR will be merged after security release

gpBlockchain:
Security review of this PR (`63223540a`..`4388d5351`) found three remaining issues in the claimed public-issue fixes:

1. Compact-block header-map epoch/difficulty fallback can mark an honest block invalid and ban the sender when expected epoch context is taken from the global epoch-number store rather than the parent itself. If the parent body is missing, prefer the existing temporary parent-required path instead of permanently classifying the hash invalid.
2. Hole-punch `ConnectionRequest` now binds the last route hop to the session peer, but `from` is still unauthenticated when the route is non-empty. A connected peer can still populate hole-punch state under a forged initiator id. Bind `from` as well, or add a test where `route` is the session peer and `from` is not.
3. Identify now rejects pure DNS listen addresses, but the onion3 allow rule matches any onion3 component, so mixed DNS+onion3 addresses can still enter the peer store. Allow onion3-only addresses.

The consensus reward shift, ban-time overflow, RecentReject count, short-id bound, and header-version consistency changes look complete.

Existing broad tests were not run locally; CI is expected to cover them.

PR #5236, opened 2026-06-10
fix: Verify DAO lock size before tx-pool script execution

Officeyutong:

Queued tx-pool verification could execute transaction scripts before running `DaoScriptSizeVerifier`. For remote transactions, the script execution limit comes from the peer-declared cycle count, bounded by `max_block_cycles`. This allowed a peer to relay a resolved DAO transaction with a lock-size mismatch and an otherwise expensive script, causing the verify worker to spend cycles before the transaction was rejected by the DAO lock-size rule.

Run the cheap contextual checks without script execution first, then run `DaoScriptSizeVerifier`, and only execute scripts after the DAO lock-size rule passes. This keeps the existing verification behavior while rejecting malformed DAO withdrawal transactions before they can consume script verification cycles.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression


tx-pool/src/util.rs
```diff
@@ -99,32 +99,44 @@ pub(crate) async fn verify_rtx(
...
+                .and_then(|_| {
                     DaoScriptSizeVerifier::new(
-                        rtx,
+                        Arc::clone(&rtx),
                         snapshot.cloned_consensus(),
                         snapshot.as_data_loader(),
                     )
                     .verify()?;
-                    Ok(result)
+                    verifier.verify(max_tx_verify_cycles, false)
```
eval-exec: We don't realy need to make these changes?

eval-exec:
I think the ideal shape would be move DaoScriptSizeVerifier into ContextualTransactionVerifier , What do you think? @zhangsoledad

zhangsoledad (approved):

PR #5237, opened 2026-06-10
fix: Respect tx-pool suspend commands before popping verify tasks

Officeyutong:

Tx-pool verify workers cache the current chunk command in `self.status`, but that cached status was only updated when the outer `tokio::select!` received a command notification. If a queue notification triggered processing first, or if a suspend command arrived while the worker was already inside `process_inner()`, the worker could keep looping with a stale `Resume` status and continue popping queued transactions. This weakened the suspend mechanism used during block processing and could let remote transaction traffic keep verification workers consuming CPU.

Refresh the worker command status from the watch receiver before each `process_inner()` iteration and again before popping a transaction from the verify queue. The command notification branch now uses `borrow_and_update()` to keep the receiver state consistent. With this change, a worker observes the latest Suspend command before taking more queued verification work.

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression

### DIFF
--- a/tx-pool/src/verify_mgr.rs
+++ b/tx-pool/src/verify_mgr.rs
@@ -108,11 +108,11 @@ impl Worker {
 
     async fn process_inner(&mut self) {
         loop {
-            self.refresh_status();
             if self.exit_signal.is_cancelled() {
                 info!("Verify worker::process_inner exit_signal is cancelled");
                 return;
             }
+            self.refresh_status();
             if self.status != ChunkCommand::Resume {
                 return;
             }
--- a/tx-pool/src/verify_mgr.rs
+++ b/tx-pool/src/verify_mgr.rs
@@ -85,14 +85,14 @@ impl Worker {
 
     async fn run(mut self) {
         let queue_ready = self.tasks.read().await.subscribe();
-        self.status = self.command_rx.borrow().to_owned();
+        self.refresh_status();
         loop {
             tokio::select! {
                 _ = self.exit_signal.cancelled() => {
                     break;
                 }
                 _ = self.command_rx.changed() => {
-                    self.status = self.command_rx.borrow().to_owned();
+                    self.status = self.command_rx.borrow_and_update().to_owned();
                     self.process_inner().await;
                 }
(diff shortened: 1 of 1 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding
Reviewed range: `8f6cacf976899b324b2d23a24e5018b54f3502ec..cc67ffdf6825a4dd0972f9160204afb4c0b32a08`

Validation:
- Reviewed `tx-pool/src/verify_mgr.rs` and traced the supporting tx-pool command, queue, cached verification, and chain suspend/resume paths.

Notes:
- The patch narrows the stale-status window, but a suspend command can still arrive after the final status refresh and before the worker removes a queued transaction. If that popped transaction has a verify-cache hit, the cached `verify_rtx` path does not consult `command_rx` before `_process_tx` reaches `submit_entry`, so one transaction can still be submitted during the intended suspend window.
- Please recheck the command after acquiring the queue write lock and before `pop_front`, and make cached/fast `_process_tx` paths honor a pending suspend before submission without dropping a popped transaction.

tx-pool/src/verify_mgr.rs
```diff
@@ -116,6 +121,11 @@ impl Worker {
                 return;
             }
 
+            self.refresh_status();
```
chenyukang, line 124: I believe this 4 lines of change is redundant, since we have line of 111 and 116 in same block.

chenyukang: or we should remove line 111 and 116 check?

eval-exec: I think we should move L111 to L115?

Officeyutong: Moved the first status refresh after the exit-signal check. I kept the second refresh/check before `pop_front`, because `is_empty().await` can yield and a suspend command may arrive during that await; we need to re-read the watch state before taking a transaction from the queue.

chenyukang (approved):

PR #5240, opened 2026-06-16
fix: Charge pipe I/O cycles before scheduler-side copies

Officeyutong:

Issue Number: close #xxx 

Problem Summary:
Pipe read/write transferred-byte cycles were added to the participating VM-local cycle counters during `process_io()`. If a writer stayed blocked in `WaitForWrite` and the root VM terminated before that writer became runnable again, those pending VM-local cycles could be discarded when non-root VMs were purged.

The host-side pipe copy also happened before the scheduler consumed and enforced the transfer cycles for the current iteration, so large pipe transfers could perform memory-copy work before the remaining cycle budget reflected that cost.

Proposal: [xxx](url) 

What's Changed:

Charge pipe transfer cycles directly to the scheduler iteration counter in `process_io()` before performing the memory copy. `process_io()` now receives the current remaining cycle limit and returns `CyclesExceeded` if the transfer would exceed that budget.

The scheduler then consumes the updated `iteration_cycles` in the same `iterate_outer()` iteration, so pipe I/O cycles cannot be lost when a blocked VM is later purged. The existing exact-cycle spawn test was updated for the corrected accounting timing.
### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression

### DIFF
--- a/script/src/scheduler.rs
+++ b/script/src/scheduler.rs
@@ -420,12 +420,6 @@ where
         limit_cycles: Cycle,
     ) -> Result<(VmId, Cycle), Error> {
         let iterate_return = self.iterate_inner(pause.clone(), limit_cycles);
-        self.consume_cycles(self.iteration_cycles)?;
-        let remaining_cycles = limit_cycles
-            .checked_sub(self.iteration_cycles)
-            .ok_or(Error::CyclesExceeded)?;
-        // Clear iteration cycles intentionally after each run
-        self.iteration_cycles = 0;
         // Process all pending VM reads & writes. Notice ideally, this invocation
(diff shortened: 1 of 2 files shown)


mohanson:
This change affects the script's cycles statistics, which may cause network forks.

XuJiandong (commented):
The cycles are part of consensus

Officeyutong:
> The cycles are part of consensus

So it's better now to make these changes?

XuJiandong:
It can be changed in next hard fork.

PR #5242, opened 2026-06-16
fix: Propagate JSONL import parse and verify failures

Officeyutong:

Issue Number: close #xxx 

Problem Summary:
`ckb import` parsed JSONL blocks in a background thread, but several failure paths were not propagated to the import result.

Malformed JSON on the first line used `expect` and could panic. Later malformed lines were parsed inside a spawned thread whose `JoinHandle` was dropped, so parser panics or errors were not reported to the main import path. Line read errors were also discarded with `filter_map(Result::ok)`. After submitting blocks, the import loop only waited for the largest block number to appear in the database, while verification errors were only printed from the callback. This could lead to panic, silent partial import, or an indefinite wait when a submitted block failed verification.

Proposal: [xxx](url) 

What's Changed:

Return structured errors for first-line parsing, make the parser thread return `Result`, and join it from the main import path. JSONL read, parse, and send failures are now propagated instead of being ignored or hidden in a detached thread.

Verification callbacks now report their result through a channel. The import command waits for callbacks from all submitted non-genesis blocks and returns an error for verification failure or `Ok(false)`. This replaces the previous largest-height polling loop, so failed imports terminate with an explicit error instead of hanging.
### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression


eval-exec:
@codex Review.

PR #5265, opened 2026-07-01
refactor(tx-pool): introduce pipeline architecture for concurrent resolve and verify

zhangsoledad:
## Summary

- Refactor tx-pool transaction processing from a serial async model to a multi-stage pipeline (`classify → pre_check → resolve → verify → submit`)
- Achieve **72% latency reduction** and **3.6× throughput improvement** for secp256k1 transactions (1-peer, 8 workers)
- Add configurable verify queue ordering (`arrival_time` or `fee_rate`)
- Fix multiple correctness issues found during audit (RBF livelock, replacement failure recovery, shutdown drain, reorg panic recovery)
- Pipeline is gated behind `#[cfg(feature = "pipeline")]`; the legacy sync path is preserved for compatibility and will be retired in a future release

https://github.com/nervosnetwork/ckb/blob/462250e599680fe19e2cb87526458abeb4f8afa3/tx-pool/pipeline.md

### Why

The tx-pool accumulated many features over the years — RBF, orphan recovery, chunk-based verification, fee estimation, pre-resolve — each bolted onto the original serial processing model. The result was code that was hard to reason about (resolve, verify, and submit stages interleaved on a single async path), hard to parallelize (CPU-intensive script verification blocked the async runtime), and hard to extend (adding new ordering strategies or entry points required touching deeply coupled code paths).

### What

The pipeline architecture separates the transaction lifecycle into clearly bounded stages, each with its own concurrency model:

```
classify → pre_check → resolve → verify → submit
  (sync)   (parallel)  (serial)  (parallel) (serial, write lock)
```

Key design decisions:

**Unified classifier** — `classify_and_enqueue_tx` is the single entry point for all transactions (remote submissions, local submissions, reorg re-adds). It checks the `FlightTracker` across all pipeline stages and routes dependent transactions directly to `OrderedResolveQueue`, avoiding a failing pre-resolve and orphan-pool churn.

**Parallel pre-check and verification** — `PreCheckQueue` workers resolve transactions concurrently. `VerifyQueue` workers verify scripts concurrently on the tokio blocking pool (`block_in_place`). IO-bound snapshot reads and CPU-bound script verification no longer block the async runtime.

**Configurable verify ordering** — The verify queue supports `arrival_time` (FIFO, default) or `fee_rate` (highest fee rate first) ordering via the `verify_ordering` config field. This enables miners to prioritize high-fee transactions for faster pool entry.

**Bounded backpressure** — Fire-and-forget `tokio::spawn` replaced by bounded mpsc channels (`DeferredTask`). Service actor loop capped by `Arc<Semaphore>` to prevent unbounded task accumulation.

**Push-based dependent wake-up** — `OrderedResolveQueue` maintains an `output_dependents` reverse index. When a parent tx enters the pool, its dependent children are extracted directly and re-enqueued, bypassing the FIFO scan.

**Lock-free conflict detection fast path** — `pre_check` skips the tx_pool read lock when all inputs resolve from the chain snapshot.

### Correctness fixes

During implementation and audit, several correctness issues were identified and fixed:

- **RBF livelock**: when a replacement fails and both the entry and displaced tx are recovered, they can cycle indefinitely. Fixed by filtering the entry from the recovered set.
- **RBF replacement failure recovery**: if process_rbf removes old txs but the replacement is later rejected, the old txs are recovered from the conflicts cache.
- **Shutdown drain**: acquire all semaphore permits before save_pool to ensure in-flight tasks complete.
- **Reorg panic recovery**: wrap reorg handler in catch_unwind with automatic respawn.
- **check_rbf_descendants deduplication**: shared descendants were counted twice. Fixed with HashSet dedup.
- **VerifyQueue proposal promotion**: proposal duplicates now promote existing entries in-place.
- **OrderedResolveQueue O(1) lookups**: replaced linear scans with HashMap + lazy tombstone deletion.

### Performance results

Benchmark: Apple M-series (arm64), 100 tx/batch, 30 samples, Criterion 95% CI.

**secp256k1 transactions** (the primary production workload):

| Scenario | Sync | Pipeline | Improvement |
|----------|------|----------|-------------|
| 1 peer, 8 workers | 183 ms | 51 ms | **-72%** |
| 1 peer, 4 workers | 182 ms | 62 ms | **-66%** |
| 4 peers, 8 workers | 63 ms | 51 ms | **-19%** |

Throughput: **546 → 1,955 tx/s** (+258%) at 1 peer, 8 workers.

### Feature flag

The pipeline is gated behind `#[cfg(feature = "pipeline")]`. The legacy sync path is preserved unchanged. The plan is to merge pipeline into the default code path and retire the legacy path in a future release once sufficient production validation has been completed.

### Test coverage

- 81 tx-pool unit tests (all pass with and without pipeline feature)
- 174 integration tests (all pass)
- `make clippy` clean
- New unit tests for verify queue fee-rate ordering (5 tests)
- New integration tests for RBF orphan recovery and concurrency stability


network/src/network.rs
```diff
@@ -340,7 +340,7 @@ impl NetworkState {
     pub(crate) fn public_addrs(&self, count: usize) -> Vec<Multiaddr> {
         let public_addrs = self.public_addrs.read();
         if public_addrs.len() <= count {
-            return public_addrs.iter().cloned().collect();
+            public_addrs.iter().cloned().collect()
```
chenyukang, line 343: why change this? from different version of rust clippy?

script/src/syscalls/load_cell_data.rs
```diff
@@ -65,36 +65,38 @@ where
...
-                    size,
-                    size_addr.to_u64(),
-                ) {
-                    Ok(val) => val,
-                    Err(VMError::SnapshotDataLoadError) => {
-                        // This comes from TxData results in an out of bound error, to
-                        // mimic current behavior, we would return INDEX_OUT_OF_BOUND error.
-                        machine.set_register(A0, Mac::REG::from_u8(INDEX_OUT_OF_BOUND));
-                        return Ok(());
+                    Ok(())
```
chenyukang, line 68: I don't know why we need to remove `return`, but it seems to reduce the readability of the code.

with `return` we can clearly know this can return from function, but expression as return value seems not so clear.

@Officeyutong @eval-exec
what's your opinions on this coding style, seems there is a clippy rules we can choose whether apply this lint check.

chenyukang:
correct me if anything wrong in this diagram 
@zhangsoledad 

```mermaid
flowchart TD
  A["RPC / Network message"] --> B["TxPoolService dispatcher"]
  B --> C["basic validity check"]
  C --> D{"depends on in-flight tx?"}
  D -->|"yes"| E["OrderedResolveQueue"]
  D -->|"no"| F["PreCheckQueue / pre_check"]
  F --> G{"resolved?"}
  G -->|"yes"| H["VerifyQueue"]
  G -->|"missing input"| E
  E --> I["single ordered resolver"]
  I --> H
  H --> J["verify workers"]
  J --> K["submit_entry under tx_pool write lock"]
  K --> L["post_submit_side_effects / after_process"]
```

tx-pool/src/block_assembler/mod.rs
```diff
@@ -328,53 +327,90 @@ impl BlockAssembler {
...
+            new_current.template.transactions.len(),
+        );
+
+        self.try_swap_template(new_current, Some(version)).await;
+    }
+
     pub(crate) async fn update_uncles(&self) {
-        let mut current = self.current.lock().await;
+        let current = self.current.read().await.clone();
+        let version = self.version.load(Ordering::SeqCst);
```
chenyukang: [P1] This version check can still be bypassed by a stale partial update. `update_uncles`/`update_proposals`/`update_transactions` clone `current` and only then read `version`; if a forced `reset_template` swaps `current` and increments `version` in that gap, the old partial update will carry the new version into `try_swap_template` and can overwrite the reset template. Please read `current` and its version under the same lock, or store the version in `CurrentTemplate`.

tx-pool/src/component/rbf_candidates.rs
```diff
@@ -0,0 +1,211 @@
...
+                    "input {:?} already has higher-fee RBF candidate {}",
+                    input, existing_id
+                ));
+            }
+        }
+
+        let mut displaced: Vec<ProposalShortId> = Vec::new();
+        for input in conflict_inputs {
+            if let Some((existing_fee, existing_id)) =
+                self.by_input.insert(input.clone(), (fee, id.clone()))
```
chenyukang: [P1] `register` only overwrites entries for the conflict inputs in the new candidate. If a displaced candidate covered multiple inputs and this new candidate overlaps only one of them, the displaced id remains in `by_input`/`by_id` for the other inputs even after the caller removes it from `VerifyQueue`. Later replacements for those other inputs can then be rejected by a non-existent in-flight candidate. Please remove each displaced candidate from all indexed inputs while holding the RBF lock, or have `register` fully unregister displaced ids before returning them.

tx-pool/src/service/builder.rs
```diff
@@ -0,0 +1,729 @@
...
+                        // complete before persisting the pool state.  The
+                        // semaphore bounds concurrent message handlers at
+                        // max_workers * MESSAGE_CONCURRENCY_MULTIPLIER, so
+                        // acquiring all permits guarantees no handler is still
+                        // running.
+                        let _ = semaphore
+                            .acquire_many(max_workers as u32 * MESSAGE_CONCURRENCY_MULTIPLIER as u32)
+                            .await;
+                        info!("TxPool is saving, please wait...");
+                        service.save_pool().await;
```
chenyukang, line 482: [P2] This drain only waits for message-dispatcher tasks. The pipeline workers (`pre_check_queue`, ordered resolver, verify workers, and deferred worker) can still be processing already dequeued jobs; a verify worker can finish `verify_and_submit_tx` after `save_pool`, leaving an accepted tx unpersisted on graceful shutdown. Please quiesce/join or explicitly drain all pipeline workers/deferred tasks before persisting.

tx-pool/src/service.rs
```diff
@@ -1194,11 +394,11 @@ impl TxPoolService {
     ) -> HashMap<ProposalShortId, TransactionView> {
         let mut txs = HashMap::with_capacity(short_ids.len());
         {
-            let verify_queue = self.verify_queue.read().await;
+            let verify_queue = self.queues.verify_queue.read().await;
```
chenyukang: [P2] `exclude_existing_proposal` now treats txs in `pre_check_queue` and `ordered_resolve_queue` as already known, but compact-block fetching only returns txs from `verify_queue`, orphan, and the main pool. A node can suppress requesting a proposal because it is "known" in pre-check/ordered, then fail to serve/fetch it while it stays there. Please either include pre-check/ordered txs in this lookup, or avoid filtering stages that are not fetchable.

chenyukang (commented):
I left a few inline findings from the tx-pool pipeline review.

script/src/syscalls/load_cell_data.rs
zhangsoledad: This was changed by a newer version of Clippy

zhangsoledad:
> correct me if anything wrong in this diagram @zhangsoledad
> 
> ```mermaid
> flowchart TD
>   A["RPC / Network message"] --> B["TxPoolService dispatcher"]
>   B --> C["basic validity check"]
>   C --> D{"depends on in-flight tx?"}
>   D -->|"yes"| E["OrderedResolveQueue"]
>   D -->|"no"| F["PreCheckQueue / pre_check"]
>   F --> G{"resolved?"}
>   G -->|"yes"| H["VerifyQueue"]
>   G -->|"missing input"| E
>   E --> I["single ordered resolver"]
>   I --> H
>   H --> J["verify workers"]
>   J --> K["submit_entry under tx_pool write lock"]
>   K --> L["post_submit_side_effects / after_process"]
> ```



> correct me if anything wrong in this diagram @zhangsoledad
> 
> ```mermaid
> flowchart TD
>   A["RPC / Network message"] --> B["TxPoolService dispatcher"]
>   B --> C["basic validity check"]
>   C --> D{"depends on in-flight tx?"}
>   D -->|"yes"| E["OrderedResolveQueue"]
>   D -->|"no"| F["PreCheckQueue / pre_check"]
>   F --> G{"resolved?"}
>   G -->|"yes"| H["VerifyQueue"]
>   G -->|"missing input"| E
>   E --> I["single ordered resolver"]
>   I --> H
>   H --> J["verify workers"]
>   J --> K["submit_entry under tx_pool write lock"]
>   K --> L["post_submit_side_effects / after_process"]
> ```

also check here https://github.com/nervosnetwork/ckb/blob/7d8383826ffaff4de9a4b7ea0afa707b5cc4cf3e/tx-pool/pipeline.md.

tx-pool/src/block_assembler/mod.rs
```diff
@@ -12,19 +16,16 @@ pub use candidate_uncles::CandidateUncles;
 use ckb_app_config::BlockAssemblerConfig;
 use ckb_dao::DaoCalculator;
 use ckb_error::{AnyError, InternalErrorKind};
-use ckb_jsonrpc_types::{
-    BlockTemplate as JsonBlockTemplate, CellbaseTemplate, TransactionTemplate, UncleTemplate,
```
chenyukang, line 16: There are some code refactor on `block_assembler` happened in this PR, which make code review harder? is it more reasonable to seperate the refactor in another following PR?

gpBlockchain:
[P1] Successful RBF re-enqueues the replaced descendants as `pending`

The existing `test_send_transaction_duplicate_input_with_son_tx` exposes a semantic regression here. For a chain `T1 -> T2 -> ...`, replacing `T1` with `R` should remove the whole old chain and leave every old transaction immediately `rejected` with an `RBFRejected` reason.

However, `process_rbf` removes the full conflict closure and records every removed entry in `conflicts_cache`. In the same successful replacement, `prepare_rbf_replacement` builds `available_inputs` from the inputs of all removed entries and calls `get_conflicted_txs_from_inputs`. Descendants such as `T2` are then selected as “recoverable”: their input (`T1:0`) is no longer spent by an in-pool transaction, even though the parent `T1` was removed and its output is no longer live.

Those descendants are sent through `DeferredTask::RecoverTxs` into `OrderedResolveQueue`. The raw pool correctly contains only `R`, but `get_transaction(T2)` searches pipeline queues first and therefore reports `pending` instead of the recorded RBF rejection. After retries it may become `rejected/Unknown`, which is also not the intended RBF result.

On a *successful* replacement, recovery must exclude the entries removed by that same replacement; they should remain cached as rejected candidates for a later replacement that genuinely frees their inputs. For a *failed* replacement, we still need to re-enqueue the complete removed chain in dependency order. Please add a regression test covering both the immediate `RBFRejected` status of every displaced descendant and the existing failed-replacement recovery path.

chenyukang:
I think the current tests cover the final recovery behavior, but I don't see a test that directly locks down the “parent is already in-flight, child arrives before the parent is committed to the main tx-pool” semantics.

In particular, `depends_on_pipeline()` routes a transaction that depends on an in-flight tx into `OrderedResolveQueue`. However, for remote transactions, if the ordered resolver pops the child before the parent has landed in the pool, `handle_orphan()` still calls `handle_missing_input_orphan()`, so the child may enter the orphan pool and later be recovered when the parent succeeds. The delayed retry behavior seems to apply only to local transactions.

Could we add a regression test that makes the intended behavior explicit?

- If the intended behavior is that a remote child should not enter the orphan pool while its parent is in-flight, the test can park the parent in the verify queue, submit the remote child, run the ordered resolver, and assert the child remains in the ordered/delayed path rather than moving to orphan.
- If the intended behavior is that a remote child may enter orphan but must be recovered after the parent is accepted, then a test should assert that path explicitly.

This would make the PR’s ordering guarantee much clearer and prevent future changes from accidentally changing the boundary between `OrderedResolveQueue` and the orphan pool.

tx-pool/src/persisted.rs
```diff
@@ -54,38 +59,75 @@ impl TxPool {
...
 
-        let mut file = OpenOptions::new()
-            .create(true)
-            .write(true)
-            .truncate(true)
-            .open(&persisted_data_file)
-            .map_err(|err| {
+        // Step 1: Collect transactions WITHOUT draining the pool.
+        // If anything fails below, the in-memory pool remains intact.
+        let all_txs = self.get_all_txs();
```
chenyukang: Before this change, `save_into_file()` used `drain_all_transactions()`, and the proposed txs were collected through `TxSelector::txs_to_commit()`, which effectively emits ancestor packages before their children. After this PR, `save_into_file()` calls `get_all_txs()`, which just iterates `pool_map.iter()`.

However, `load_persisted_data()` still says “assume that all txs are sorted” and reloads them one by one through `submit_local_tx()`. If the persisted file contains a child before its parent, the child is loaded as a local tx, hits missing input, and may be rejected instead of being kept in the orphan pool. When the parent is loaded later, the child is already gone.

I think we should preserve the parent-before-child ordering before reloading persisted txs. A simple option is to call the existing dependency sorter before importing persisted txs, e.g. sort the loaded vector with `TxPoolService::sort_txs_by_dependencies()` in `load_persisted_data()`.

Could we also add regression tests for this?

1. Persisted txs are recovered correctly even when the persisted vector is child-first.
2. A dependent tx chain already in the main pool survives save + restart with both parent and child restored.

This would make the restart behavior explicit and protect against future changes to `pool_map.iter()` or persistence ordering.

script/src/syscalls/load_cell_data.rs
eval-exec: I like explicitly writing return

eval-exec: I like explicitly return and early return. https://doc.rust-lang.org/rust-by-example/error/result/early_returns.html

gpBlockchain:
[P1] Reorg-recovered txs can stay RPC-pending but never be re-proposed/mined

I think `test_reorg_recovers_dependent_pending_tree` exposes a real reorg/block-template regression in this PR.

The failing scenario is:

1. node1 commits a dependent tx tree: parent -> child -> grandchild.
2. node2 builds a longer blank fork.
3. node1 reorgs to node2's chain.
4. The three txs are recovered into tx-pool: `tx_pool_info.pending == 3`, `orphan == 0`, and `get_transaction(...)` reports them as `pending`.
5. But normal mining through `get_block_template` never commits the grandchild; after 100 blocks it is still RPC `pending`.

The important clue is that this is not a verification/rejection failure. Before the reorg, node1 normally produced proposals and committed the txs. After the reorg, every mined block had only the cellbase tx and `proposals: 0`; there were no further block-template candidate tx lines. So the recovered txs are visible through RPC, but the block assembler never selects them again.

The failure mode points to an internal status/template mismatch around `Gap`:

- `map_pool_status` maps every non-`Proposed` tx to RPC `Pending`, so RPC cannot distinguish real `Pending` from internal `Gap`.
- `pending_size()` includes both `pending_count + gap_count`, so pool stats can also look healthy.
- `PoolMap::get_proposals()` only iterates `Status::Pending`.
- `TxSelector` only commits `Status::Proposed`.
- `status_to_block_assembler_message(Status::Gap)` returns `None`.

That means a recovered tx stuck in internal `Gap` can look pending externally, while being neither re-proposed nor committed.

The reorg path seems able to create exactly this stranded state. In `update_tx_pool_for_reorg`, mine mode handles `Gap -> Proposed` when `snapshot.proposals().contains_proposed(...)`, and `Pending -> Gap/Proposed`, but I do not see a corresponding `Gap -> Pending` demotion when the proposal is no longer valid in the new snapshot. Combined with `update_block_assembler_before_tx_pool_reorg()` resetting the template with detached blocks as candidate uncles, `update_full()` may continue to exclude proposal IDs already present in the current template's uncles, leaving recovered txs without a path back into normal proposal packaging.

The existing Rust spec `test/src/specs/tx_pool/reorg_recovers_dependent.rs` does not catch this because after reorg it manually submits a `propose(node, &[...])` block. That verifies recovery, but bypasses normal block-template proposal selection. The integration failure uses normal mining, which is where the regression appears.

Suggested fix direction:

- During reorg reconciliation, re-evaluate existing `Gap` entries against the new snapshot.
- If a `Gap` tx is no longer in the new proposal/gap window, demote it back to `Pending` and notify the block assembler so it can be proposed again.
- Add a regression test that mines through normal `get_block_template` after reorg, and optionally assert `get_pool_tx_detail_info(tx_hash).entry_status` to distinguish real `pending` from internal `gap`.

test/src/specs/tx_pool/replace.rs
```diff
@@ -906,19 +905,37 @@ impl Spec for RbfConcurrency {
             conflicts.push(tx2);
         }
 
-        // make 5 threads to set_transaction concurrently
-        let mut handles = vec![];
+        // Submit transactions sequentially with small delays so that the
+        // asynchronous pipeline processes each replacement before the next
+        // one arrives.  Concurrent submission causes a livelock with
+        // DeferredTask::RecoverTxs: when an RBF replacement fails, the
+        // displaced tx is re-enqueued, which re-triggers the same conflict.
```
chenyukang, line 912: this test was using concurrency to sending multiple tx for some reason.
we can not change it to sequence mode?

maybe we should fix this implementation to handle concurrency scenario?

chenyukang: this may happen in realworld?

tx-pool/src/service/builder.rs
```diff
@@ -0,0 +1,619 @@
...
+                            .quiesce(Duration::from_secs(PIPELINE_SHUTDOWN_TIMEOUT_SECONDS))
+                            .await;
+
+                        info!("TxPool is saving, please wait...");
+                        service.save_pool().await;
+                        info!("TxPool process_service exit now");
+                        break
+                    },
+                    else => {
+                        // The message channel closed without a cancel signal
```
chenyukang, line 487: The `else` branch is also a shutdown-like exit path: it is reached when all message senders are dropped and `receiver.recv()` returns `None` without the cancellation token firing. However, closing the channel only means no new messages will arrive; it does not guarantee that previously spawned message handlers or pipeline workers have completed.

Since the cancellation branch now drains the semaphore permits and quiesces background workers before `save_pool()`, should the `else` branch share the same drain/quiesce/save path as well? Otherwise the shutdown-drain guarantee only applies to signal-driven shutdown, but not to the channel-closed fallback path.

tx-pool/src/pool/rbf.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        // Record every removed entry (direct conflicts and their descendants)
+        // in the conflicts cache so that they can all be recovered if the
+        // replacement fails or if their inputs become available again.
+        //
+        // The original pipeline source is not retained once a transaction has
+        // entered the pool, so recovered entries fall back to `TxSource::Local`.
+        for old in &all_removed {
+            self.record_conflict(old.transaction().clone(), TxSource::Local);
```
chenyukang, line 337: Treating it as `Local` is a reasonable “internal recovery” policy, but it does change the recovery semantics: recovered remote/proposal txs will be reprocessed without the original peer attribution, declared cycles, remote reject/ban behavior, and proposal-source priority.

but if we do need to keep the source, that involve more change, I'm not sure, just point out this maybe need to consider carefully.

tx-pool/src/service/workers.rs
```diff
@@ -0,0 +1,600 @@
...
+                    // idempotent — the write-lock section (snapshot swap,
+                    // committed/detached-proposal removal, status
+                    // migration, expiry, size limit) is a no-op when
+                    // repeated, and the lock-free retain recovery re-adds
+                    // each tx independently (duplicates return Ok(false)).
+                    // Without the retry a single panic would silently drop
+                    // the whole reorg: detached transactions would never
+                    // be re-added and the pool would stay half-updated
+                    // against the new tip.
+                    let outcome = crate::worker::run_with_one_retry(|| {
```
chenyukang, line 279: why there is no retry in old code?
we need to retry here because there logic and states are more complex in new arch and we need to keep it safe here?
if there is a panic happened and there is no attack path for it, this means it's a logic bug happened in the code,
i'm afraid this kind of issue is hiden if we recovery here.
at least we should add some error log for unexpect inner panic?

i'm not sure everything is idempotent here, there are also some callbacks.

tx-pool/src/component/pre_check_queue.rs
```diff
@@ -0,0 +1,539 @@
...
+            tokio::select! {
+                _ = self.ready.notified() => {}
+                _ = self.cancel.cancelled() => return None,
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
chenyukang, line 308: move tests code to test directory

PR #5294, opened 2026-07-27
refactor: simplify ancestor eviction loop

eval-exec:
## What changed

- Replace the `if let`/`else` branch in the ancestor-eviction loop with a `let ... else` guard.
- Keep the eviction, parent cleanup, and ancestor-count recomputation in the main loop body.

## Why

The iterator-exhaustion case is an early exit. Expressing it as a guard reduces nesting and keeps the successful eviction path linear.

## Impact

This is a behavior-preserving refactor. The stale-parent cleanup and ancestor-count recomputation introduced in #5293 remain unchanged.

## Validation

- `cargo fmt --all -- --check`
- `cargo nextest run -p ckb-tx-pool test_max_ancestors_evicts_stale_parent_descendant`

### DIFF
--- a/tx-pool/src/component/pool_map.rs
+++ b/tx-pool/src/component/pool_map.rs
@@ -617,20 +617,20 @@ impl PoolMap {
 
         let mut iter = evict_candidates.iter();
         while ancestors_count > self.max_ancestors_count {
-            if let Some(next_id) = iter.next() {
-                let removed = self.remove_entry_and_descendants(next_id);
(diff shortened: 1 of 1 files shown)


eval-exec:
What do you think of this style change?

Previous:
```rust
{
    if let Some(next_id) = iter.next() {
        // main work
    } else {
        break;
    }
}
```
  The PR handles the exit condition first:
```rust
{
    let Some(next_id) = iter.next() else {
        break;
    };
  
    // main work
}
```

chenyukang (approved):

PR #5302, opened 2026-07-29
fix(network/onion): bound and drain Tor control lines

eval-exec:
## Summary

- bound the completed Tor control-line queue and apply backpressure when consumers fall behind
- drain unsolicited control lines while waiting for the controller connection to close
- add regression coverage for queue saturation and clean connection closure

## Why

After onion service setup, connection liveness waits for the Tor controller to disconnect. The background reader continued forwarding control lines to a queue while that wait path had no active line consumer, allowing sustained unsolicited output to grow process memory.

The bounded queue caps pending completed lines. Draining during the liveness wait lets the reader continue to EOF or an I/O error so disconnect detection remains responsive under backpressure.

## Validation

- `cargo fmt --all --check`
- `cargo test -p ckb-onion`
- `cargo clippy -p ckb-onion --tests` (passes with two pre-existing `needless_return` warnings in `ckb-network`)

### DIFF
--- a/util/onion/src/tor_connection.rs
+++ b/util/onion/src/tor_connection.rs
@@ -263,12 +263,7 @@ impl TorConnection {
             tokio::select! {
                 biased;
                 result = &mut self.disconnect_rx => return result.is_ok(),
-                line = self.line_rx.recv() => {
-                    if line.is_none() {
-                        // The reader dropped its sender, which only happens when
-                        // the stream hit EOF or an error: the connection is gone.
-                        return true;
-                    }
+                Some(_) = self.line_rx.recv() => {
                     // Discard the unsolicited control line.
                 }
             }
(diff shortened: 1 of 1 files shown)


util/onion/src/tor_connection.rs
```diff
@@ -231,7 +235,7 @@ impl TorConnection {
                             line.pop();
                         }
                         let taken_line = std::mem::take(&mut line);
-                        if line_tx.send(taken_line).is_err() {
+                        if line_tx.send(taken_line).await.is_err() {
```
chenyukang, line 238: **[P2] Cap the size of each control line**

`read_line` keeps extending its internal buffer until it sees LF or EOF. A hostile controller can therefore stream one unterminated line indefinitely. Because `line_tx.send()` is reached only after `read_line` returns, the new 256-entry channel bound does not limit this allocation.

Please use a bounded line decoder that closes the connection on an oversized line, and add an unterminated-line regression test. The current test covers only many small, completed lines.

PR #5311, opened 2026-08-25
fix(tx-pool): bound verify queue transaction count

Officeyutong:

Problem Summary:

The verify queue currently limits admission using only the sum of each
transaction's serialized block size, with a 256 MB limit.

Serialized size does not include the fixed per-entry memory overhead from
`TransactionView`, `VerifyEntry`, and the indexes maintained by
`MultiIndexMap`. A remote peer can therefore relay a large number of small
transactions and cause actual memory consumption to substantially exceed the
configured byte limit.

In testing, 120,000 small transactions accounted for approximately 29 MB of
serialized data but increased the node RSS by approximately 124 MB.

Proposal: [xxx](url) 

What's Changed:

### Related changes

- Retain the existing 256 MB serialized-size limit.
- Add an independent verify queue limit of 100,000 transactions.
- Reject new transactions with `Reject::Full` when either limit is reached.
- Add a regression test covering the transaction-count boundary.

The two limits protect against different cases:

- The byte limit bounds queues containing large transactions.
- The count limit bounds fixed per-entry overhead from large numbers of small
  transactions.

### Check List 

Tests 

- `cargo fmt --check`
- `cargo test -p ckb-tx-pool component::tests::chunk --lib`
- `cargo test -p ckb-tx-pool --lib`
- `git diff --check`

Side effects

- Performance regression

### DIFF
--- a/tx-pool/src/component/tests/chunk.rs
+++ b/tx-pool/src/component/tests/chunk.rs
@@ -91,6 +91,22 @@ async fn verify_queue_basic() {
     assert!(!queue.contains_key(&id));
 }
 
+#[test]
+fn verify_queue_rejects_when_transaction_count_limit_is_reached() {
+    let mut queue = VerifyQueue::new(MAX_TX_VERIFY_CYCLES);
+    queue.set_max_transactions_for_test(1);
+
+    let tx0 = TransactionBuilder::default().build();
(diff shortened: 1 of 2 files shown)


chenyukang:
@zhangsoledad @eval-exec 
do you think this is a issue need to be fixed, I may think 100_000 as default number is somehow small.

cc @gpBlockchain

PR #5316, opened 2026-09-02
refactor(tx-pool): tx-pool: make state changes transactional

zhangsoledad:
# tx-pool: make state changes transactional

A UTXO transaction expresses a state transition as data: which existing outputs to consume and which new outputs to create. Validation establishes whether that transition is allowed before the ledger applies it.

This PR applies the same idea to the transaction pool itself. A **Plan** describes the pool records before and after a proposed change, the original observations that justify it, and the notifications it requires. **Apply** checks those observations against current state, then commits the complete change. The pool uses the design principle behind the transactions it processes to manage its own state transitions.

Replacing a transaction shows why this matters. It can remove descendants, change dependency relationships, release capacity, and require notifications. If a replacement cannot be admitted, removing the incumbent is not a successful partial result. Expressing the whole change as a Plan lets policy evaluate its intended outcome before mutation. Apply then either commits the related records, indexes, accounting, and notification obligations together, or rejects the stale or invalid plan without a partial update.

Admission, replacement, eviction, and chain reconciliation share this boundary. Dry-run can check the same admission decision without executing it, callbacks can consume committed event data after pool guards are released, and independent operations can still commit concurrently. The analogy concerns how state changes are represented and validated; chain consensus and transaction verification retain their existing roles.

## Why the existing structure became difficult to maintain

A transaction pool maintains several related facts: which transactions it retains, which are accepted, which cells they use, how they depend on one another, and what work remains outstanding. These facts also change when the chain changes. On `develop`, understanding one operation requires following updates across the pool, orphan storage, verification queues, mutable graph statistics, and callbacks embedded in mutation paths.

The historical reorg fixes illustrate the maintenance cost. [#3706](https://github.com/nervosnetwork/ckb/pull/3706) handled a parent returning to a pool that already contained its children. [#3713](https://github.com/nervosnetwork/ckb/pull/3713) then made that relationship maintenance apply to every insertion source. [#3886](https://github.com/nervosnetwork/ckb/pull/3886) addressed a downstream packing panic where a descendant relationship did not lead to an available entry.

Those fixes repaired real problems. Together, they show why diagnosis was hard: a failure in block construction could originate in an earlier insertion or recovery path, and repairing one path required finding every other place that relied on the same assumption. The graph, its cached accounting, and its consumers had to agree through each intermediate update.

This refactor makes a complete pool transition the unit of correctness. The implementation gives related changes one owner and one point of commitment, so that future fixes can establish an invariant at the boundary responsible for preserving it.

## Start with a transaction whose state is explicit

Store represents each retained transaction with one current immutable record, called its **owner**, keyed by the raw transaction hash. The record says whether the transaction needs resolution or verification, is waiting for a dependency, is accepted, or is retained as replacement history. Queues and indexes refer to this state; they do not independently decide whether a transaction belongs to the accepted pool.

Changing a transaction's state installs a new owner instance. If a transaction is removed and later reinserted with the same hash, a worker holding the earlier instance can detect that its work has been superseded. Matching the hash and chain tip is insufficient to make an old result current.

This also separates retained work from accepted transactions. Pending, Gap, and Proposed are derived from accepted owners and the corresponding chain proposal view. They describe proposal-window status, separately from resolution, verification, and acceptance. The ordinary public transaction-status response maps Gap to Pending; detailed pool-entry queries can still expose gap. Store owns the pool's records and indexes; canonical chain state and consensus verification remain authoritative for their respective facts.

## Follow a replacement through the architecture

RBF brings the central problem into focus. Suppose A is accepted, C depends on A, and B attempts to replace A. Accepting B requires a coherent decision about the spender, the complete victim set, surviving dependencies, replacement fees, and available capacity. Removing A successfully is only one part of that operation.

The new admission path prepares this decision outside authority guards. It records the **read set**: the original owners, the input's spender, and the complete relationships used to discover and validate affected transactions. Reads can also record absence. This matters because a previously missing producer, spender, or dependent can appear while work is in flight.

The **Plan** combines those observations with the intended changes. Its write set is expressed as before/after owner edits, including B and every affected incumbent. Index updates, queue changes, resource charges, and the required shard support are derived from those edits. Required effects are part of the same Plan.

```mermaid
graph TD
    R["Observe current owners, dependencies and chain context"] --> P["Prepare Plan<br/>Original reads + owner edits + required effects"]
    P --> A["Apply<br/>Revalidate premises and reserve capacity"]
    G["Ordered shard guards<br/>Protect the relevant read/write facts"] -.-> A
    A --> C["Commit the complete transition"]
    C --> O["Current owners<br/>Indexes, queues and charges"]
    C --> E["Committed effect obligations"]
    O --> Q["Queries and block templates"]
    E --> U["Publish after guards are released"]
```

**Apply** is the point where this prepared argument is checked against live state. It acquires the relevant guards, validates the original observations and permissions, and reserves the exact capacity required by the final change. All fallible policy and capacity checks precede mutation. Owners, their projections, accounting, and effect obligations then commit together.

If another transaction D starts depending on A before Apply, the original complete relationship observation becomes stale. The operation makes no partial change and can prepare a new decision with D included.

```mermaid
sequenceDiagram
    participant W as Admission for B
    participant S as Store
    participant X as Concurrent admission
    W->>S: Observe A, C, spender and complete dependencies
    S-->>W: Original observations
    W->>W: Prepare replacement Plan
    X->>S: Accept D, which depends on A
    W->>S: Apply original Plan
    S-->>W: Stale, no membership or effect changes
    W->>S: Read current owners and dependencies again
    S-->>W: Updated observations
    W->>W: Prepare revised replacement Plan
    W->>S: Apply revised Plan
    S->>S: Commit candidate, victims, indexes, charges and effects
    Note over S: Release guards before publication
```

This also makes rejection meaningful. A failed replacement leaves incumbents intact. Capacity policy runs over the proposed result, so a candidate that would evict itself does not first evict other transactions. Shared descendants contribute once to replacement accounting, and backing is checked again after capacity trimming changes the victim set. Dry-run uses the same policy and final validation boundary.

The same model addresses the historical late-parent problem. Admission discovers existing children and includes their changed parent relationships in the Plan. Ancestor and descendant totals are computed from captured immutable membership, avoiding long-lived transitive statistics that every insertion and removal path must incrementally repair. Packing works on that captured graph, with source validity checked when publishing a template.

## Give committed changes one publication path

A complete transition also needs a clear answer to when callbacks run. On `develop`, callback invocations appear in admission, replacement, eviction, and reorg processing, interleaved with state changes. Their callers must understand both the mutation sequence and the assumptions of external code.

Here, decision code describes the events required by a transition. Apply commits their obligations into a bounded outbox alongside the state change. A common publisher invokes callbacks after authority guards are released, using immutable event snapshots. Callback production remains close to the policy that defines the event; callback execution, ordering, and endpoint failure handling converge in one place.

That distinction makes failures easier to investigate. An incorrect event can be traced to its Plan. A committed event awaiting delivery can be traced through the publisher. A caller disappearing after commitment does not erase the obligation. Callback reads are supported, and direct mutating reentry fails promptly instead of waiting on publication of its own effects.

External endpoints still have to return for ordered publication and shutdown to advance. The outbox provides an in-process obligation boundary; external delivery keeps its declared reliability semantics.

## Concurrency with progress that can be explained

Store applies the same commit rules across 256 ownership shards. For ordinary work, Apply acquires ordered guards for the owners, dependencies, and peer permissions its decision touches. Independent changes can reach their final commit sections together, including ordinary transactions that share only a read-only cell dependency. Cross-shard replacement still commits as one coupled transition, without putting unrelated admissions behind a global membership writer.

This keeps synchronization tied to the facts that actually conflict. Physical routing collisions and short shared accounting operations still create contention, while full captures and chain changes have broader scope. Preparation can repeat work, and captured graph computation costs CPU and temporary memory. Production-path tests establish overlap inside the final commit sections, beyond merely running VM verification in parallel; benchmark measurements assess the resulting tradeoffs.

The same explicit ownership helps analyze **liveness**. A waiting transaction retains the missing dependency keys that can wake it. A worker owns its job, result, and active reservation through settlement. Capacity waits retain accountable work and retry on relevant state changes. Stale work either replans, returns to resolution, or observes its successor. These paths identify both the reason for waiting and the component responsible for enabling progress.

Progress also requires the operation that releases a wait to remain runnable. Block verification exercises this directly. It can pause pool computation while reconciliation and publication continue, because those operations are needed to complete chain processing. A running VM can suspend cooperatively with its execution state retained, then continue when the block backlog drains. A pause request, a suspended execution, and a completed shutdown are separate observable conditions.

Deterministic tests exercise dependency changes, commit overlap, block priority, suspension/resumption, publication, and joined shutdown. Each targets a concrete ownership or waiting boundary. The reverse-fanout comparison below adds an observed completion difference under pressure, with its measurement scope kept explicit.

## Make the next bug and the next feature easier to handle

For a graph inconsistency, investigation starts with the current owners and the admission or reconciliation logic that prepares their relationship changes. For a stale result, it follows the decision's original observations and invalidation path. For a notification problem, it distinguishes event creation from committed delivery. The common boundaries give maintainers a smaller set of explanations to test.

They also make useful regressions easier to construct. A test can prepare a removal, insert a late child, and establish that Apply refuses without changing membership or accounting. Another can admit a late parent and inspect the child's resulting ancestry. These directly encode the assumptions that previously had to be reconstructed from a long reorg and packing history. Existing tests cover both examples.

These boundaries also give future features a clear place to express the changes that must agree. A `rolling_min_fee_rate` would couple capacity eviction with any resulting fee-floor update: a failed speculative eviction must leave the live floor unchanged. This requires new policy state and observations, plus integration with the existing fee check, within the established preparation and commitment model.

For package acceptance, a low-fee parent and a high-fee child need resolution in a temporary package view and an appropriate combined fee decision. Where the chosen semantics require new members to succeed together, their edits and affected incumbents can use the existing multi-owner Plan and Apply. Decay rules, package resolution, fee policy, resource limits, and public result semantics still need implementation. The common machinery already provides a place to validate the resulting decision and commit its consequences.

## Reviewing the change

A useful first pass is to follow the replacement example through `authority/model.rs`, `authority/membership.rs`, and `authority/store.rs`, then into `authority/notice.rs`. This establishes the state model and its commitment boundary before considering execution details. Review whether the observed facts and affected set are complete. The contracts and membership tests show how stale decisions and refused multi-owner changes are exercised.

The second pass follows the surrounding lifetimes: `authority/jobs.rs` and `authority/service/` for computation and settlement; `authority/chain.rs` for invalidation and recovery; and `authority/template.rs` / `authority/packing.rs` for consumers. These paths are under `tx-pool/src/`. The architecture, commit, review, and maintenance guides under `tx-pool/docs/` connect the implementation to representative tests and caller contracts.

An earlier frozen benchmark provides a concrete completion contrast. In the same reverse-fanout workload with 5,752 transactions, eight workers, and four peers:

| Implementation | Observed outcome |
|---|---|
| `develop` | Timed out after accepting 101 of 5,752 transactions. |
| Refactored pool | Accepted all 5,752 and produced all 5,752 relay-success results, with no duplicate callback or relay-success results. |

This records a practical progress difference under that workload. The comparison used candidate `190a8140d55c2ba74e59d3083a58e60925d68c5a` and develop production `f75d609f9bfdd0a3160e61d21a38a46c01e45949` with its benchmark adapter. It was a single A/B pilot; baseline non-completion prevented a paired throughput estimate. The timeout establishes the observed failure to finish within the test budget. Final-head benchmark measurements remain pending.

Validation includes 280 transaction-pool Nextest tests, 22 tests for the final notification/configuration changes, a complete macOS native integration run with 177 passing cases, and strict workspace Clippy, formatting, and RPC documentation checks. Passing evidence is reused where the relevant implementation is unchanged; overlapping scopes are not added into a unique test count. Release-platform CI and final performance measurements retain their own scope.

Existing node configuration and v1/v2 pool-file reads remain supported; restored transactions are processed against current state. Consensus checks continue through the canonical verification crates and the official CKB-VM dependency. The Rust crate API changes intentionally: callers migrate from mutable pool access to the controller, fallible construction, and immutable callback snapshots. The maintenance guide records those changes, query semantics, and the best-effort transaction-notification boundary, including the optional pending-indexer limitation after a lost removal event.


chainTe:
Security review completed against `17d7db5bb423a1b2177e14a132a41d5a91a515f3..ad355d65afe2eca471587a744004be98da89aee2`. I found four new issues: one high-severity transaction-pool authority deadlock, two medium-severity freshness races in chain reconciliation and dependency settlement, and one low-severity stale-snapshot race in `clear_tx_pool`.

The blocking themes are: avoid routed lock acquisition while an all-shard read cut is live; fence or version all causal sources across chain receipt capture and apply; distinguish reserved ApplySequence values from a committed-prefix watermark; and make clear-current-pool use execution-time snapshot state or reject stale intent.

`cargo fmt --all -- --check` and the tx-pool security-manifest checker pass. A targeted lock-fairness test could not compile in the review environment because Clang could not locate `stdbool.h`; CI should run the full test suite and deterministic concurrency canaries should be added for each race before merge.

zhangsoledad:
> Security review completed against `17d7db5bb423a1b2177e14a132a41d5a91a515f3..ad355d65afe2eca471587a744004be98da89aee2`. I found four new issues: one high-severity transaction-pool authority deadlock, two medium-severity freshness races in chain reconciliation and dependency settlement, and one low-severity stale-snapshot race in `clear_tx_pool`.
> 
> The blocking themes are: avoid routed lock acquisition while an all-shard read cut is live; fence or version all causal sources across chain receipt capture and apply; distinguish reserved ApplySequence values from a committed-prefix watermark; and make clear-current-pool use execution-time snapshot state or reject stale intent.
> 
> `cargo fmt --all -- --check` and the tx-pool security-manifest checker pass. A targeted lock-fairness test could not compile in the review environment because Clang could not locate `stdbool.h`; CI should run the full test suite and deterministic concurrency canaries should be added for each race before merge.

This is a known open issue: C2 is tracked in HELD_SHARD_CUT_AND_POLICY_READ_COHERENCE, one of eight open terminal-correctness clusters. We are first minimizing the architecture and retiring the composite/alternate routes so the lock graph and correctness TCB are smaller. We will then reopen C2 against the frozen simplified source and close it root-wise with deterministic production-bound, no-sleep interleaving evidence and a lock-order DAG. Until then, C2 and terminal correctness remain explicitly open; we will not apply a finding-specific patch.

resource/ckb.toml
```diff
@@ -208,12 +208,43 @@ enable_deprecated_rpc = false # {{
...
+max_ancestors_count = 1_000
+
+# Queued verification selects higher fee rates first, with arrival-time ties.
+# Use "arrival_time" for FIFO selection; completion can occur out of order.
+# verify_ordering = "fee_rate"
+# The historical name controls the remote small/large declared-cycle boundary,
+# not a maximum transaction verification budget.
 max_tx_verify_cycles = 70_000_000
-max_ancestors_count = 25
+# Worker count defaults to max(3 * CPU cores / 4, 1). Runtime capacity can
```
chenyukang, line 226: now we remove the limit of max_ancestors_count with default value of 25?

rpc/src/error.rs
```diff
@@ -180,7 +180,7 @@ impl RPCError {
             Reject::ExceededMaximumAncestorsCount => {
                 RPCError::PoolRejectedTransactionByMaxAncestorsCountLimit
             }
-            Reject::Full(_) => RPCError::PoolIsFull,
+            Reject::Full(_) | Reject::ExcessiveVerifyTime => RPCError::PoolIsFull,
```
chenyukang: `Reject::ExcessiveVerifyTime` seems better seperate from PoolIsFull.

chenyukang: Reject::ExcessiveVerifyTime seems better seperate from PoolIsFull.

tx-pool/src/service.rs
```diff
@@ -1,1293 +1,171 @@
...
-                    else => break,
-                }
-            }
-        });
-        self.started.store(true, Ordering::Release);
-        if let Err(err) = self.tx_pool_controller.load_persisted_data(txs) {
-            error!("Failed to import persistent txs, cause: {}", err);
-        }
+impl<A> Notify<A> {
+    pub(crate) fn new(arguments: A) -> Self {
```
chenyukang, line 96: why move the code position of `TxPoolController` from this file? seems like a refactoring work, which makes code review harder for this kind of refactoring mixed with new code.

util/gen-types/src/extension/rust_core_traits.rs
```diff
@@ -179,12 +179,19 @@ impl_cmp_partial_ord!(CellDep);
 impl ::core::cmp::Ord for packed::OutPoint {
     #[inline]
     fn cmp(&self, other: &Self) -> ::core::cmp::Ordering {
-        let tx_hash_order = self.tx_hash().cmp(&other.tx_hash());
+        let self_reader = self.as_reader();
```
chenyukang, line 182: why change this?

util/app-config/src/legacy/tx_pool.rs
```diff
@@ -12,8 +12,18 @@ const DEFAULT_MIN_FEE_RATE: FeeRate = FeeRate::from_u64(1000);
 const DEFAULT_MIN_RBF_RATE: FeeRate = FeeRate::from_u64(1500);
 // default max tx verify cycles
 const DEFAULT_MAX_TX_VERIFY_CYCLES: Cycle = TWO_IN_TWO_OUT_CYCLES * 20;
+const DEFAULT_MIN_TX_VERIFY_TIME_MS: u32 = 250;
+const DEFAULT_TX_VERIFY_CYCLES_PER_MS: u64 = 10_000;
+// A tx-pool attempt may execute at most one consensus minimum block-interval
```
chenyukang: seems the whole verify with quota and limits is a seperated implementation out of this tx-pool refactor?

resource/ckb.toml
zhangsoledad: This setting hasn't had any effect for a long time. The value is now defined as a constant in the node, so the setting should be removed from the configuration.

util/gen-types/src/extension/rust_core_traits.rs
zhangsoledad: > why change this?

It’s a general-purpose micro-optimization for OutPoint comparisons, with accompanying tests.

PR #5319, opened 2026-09-07
fix(network): stop onion peers from banning the shared Tor proxy address

Officeyutong:

Problem Summary:

When `[network.onion].listen_on_onion` is enabled, the launcher derives a local
`p2p_listen_address` (normally `127.0.0.1:<p2p port>`) and registers an `ADD_ONION`
listener that forwards every hidden-service connection to that local socket
(`util/launcher/src/lib.rs:341`). Those sessions are then handled as ordinary TCP
sessions, so each onion peer's `connected_addr` is the Tor daemon's local socket rather
than a unique remote identity.

`ban_addr` used to map that address to an IP network and ban it:

```rust
if let Some(addr) = multiaddr_to_socketaddr(addr) {
    let network = ip_to_network(addr.ip());
    self.ban_network(network, timeout_ms, ban_reason)
}
```

So a single malicious onion peer sending a malformed transaction or protocol message
could get `127.0.0.1/32` banned for days, after which `PeerRegistry::accept_peer`
rejects **every** onion inbound peer forwarded by Tor from that same address. An
unauthenticated remote attacker could therefore disable the node's onion inbound
service until the ban expires or is cleared manually.

A second, related gap: an **outbound** `/onion3` peer is dialled through SOCKS5, so its
connected address carries no IP at all. `multiaddr_to_socketaddr` returns `None` for it,
which meant such a peer was never banned for anything.

What's Changed:

**1. Identity bans for addresses that do not identify a single peer**
(`network/src/peer_store/peer_store_impl.rs`)

`ban_addr` now decides based on whether the connected address actually attributes the
misconduct to one peer:

```rust
match multiaddr_to_socketaddr(addr) {
    // The address carries an IP that belongs to this peer alone, ban the network.
    Some(socket_addr) if !self.shared_proxy_addrs.contains(&socket_addr.ip()) => {
        self.ban_network(ip_to_network(socket_addr.ip()), timeout_ms, ban_reason);
    }
    // Shared proxy IP (inbound Tor forwarding) or no IP at all (outbound /onion3).
    _ => self.ban_peer_identity(addr, timeout_ms, &ban_reason),
}
```

`BanList` gains a `PeerId`-keyed map alongside the existing `IpNetwork` map
(`network/src/peer_store/ban_list.rs`). `is_addr_banned` checks both. Identity bans are
memory-only and capped at `MAX_BANNED_PEERS = 16384` with LRU eviction, since a peer
behind a proxy can rotate identities freely.

**2. The exemption is scoped to nodes that actually run an onion service**
(`util/app-config/src/configs/network.rs`)

`trusted_proxies` means "may send a HAProxy PROXY header", and it defaults to loopback.
Reusing it directly would have disabled IP banning for *every* localhost peer on *every*
node. Instead a new accessor gates it on onion listening being genuinely active,
mirroring the conditions the launcher checks before issuing `ADD_ONION`:

```rust
pub fn onion_listen_enabled(&self) -> bool {
    self.onion.listen_on_onion
        && (self.onion.onion_server.is_some() || self.proxy.proxy_url.is_some())
}

pub fn shared_proxy_addrs(&self) -> &[IpAddr] {
    if self.onion_listen_enabled() { &self.trusted_proxies } else { &[] }
}
```

With onion listening off, banning is byte-for-byte the behaviour it was before this PR.
The `PeerStore` field is named `shared_proxy_addrs` rather than `trusted_proxies` to keep
the two concepts from being conflated again.

The `/onion3` fallback in point 1 is deliberately **not** gated on this config: when
`multiaddr_to_socketaddr` returns `None` there is no IP to ban in the first place, so an
identity ban is strictly better than the previous no-op regardless of configuration.

**3. Docs** — `resource/ckb.toml` explains when `trusted_proxies` switches banning to
peer identity, and that identity bans are memory-only and not listed by
`get_banned_addresses`.

Not affected: the `set_ban` RPC takes an `IpNetwork` and goes through
`NetworkController::ban`, never `ban_addr`, so operator-issued IP bans work exactly as
before and still override identity-based handling. `clear_banned_addresses` clears both
lists.

Known limitation, stated explicitly: a peer reached over Tor can rotate its `PeerId` at
no cost, so an identity ban does not stop a determined attacker. The point of this PR is
to remove the self-inflicted outage — one peer can no longer take down onion inbound for
everyone else. This matches how Bitcoin Core handles Tor peers.

### Related changes

### Check List

Tests

- Unit test

New tests in `network/src/tests/peer_store.rs` and `network/src/tests/peer_registry.rs`:

| Test | Covers |
| --- | --- |
| `test_onion_proxy_ban_disconnects_only_offending_session` | With onion on, `ban_session` drops only the offender; other onion peers stay connected and new ones are still admitted |
| `test_ban_stays_ip_based_when_onion_listening_is_disabled` | With onion off, a ban still produces `127.0.0.1/32` and still covers the whole loopback network — pins the no-regression guarantee |
| `test_proxy_identity_ban_expiry_and_address_isolation` | Identity ban follows the peer across addresses, expires, and is cleared by `clear_ban_list`; over IPv4, IPv6 and a non-loopback proxy IP |
| `test_proxy_behaviour_ban_and_explicit_ip_ban` | Score-based `report()` bans reach the identity path; a manual `ban_network` on the same IP still applies |
| `test_untrusted_address_still_banned_by_ip` | Addresses outside the shared set keep IP banning |
| `test_proxy_identity_bans_are_bounded` | `MAX_BANNED_PEERS` cap and LRU eviction |
| `test_outbound_onion_addr_is_banned_by_identity` | Outbound `/onion3` peers are now bannable; the same identity stays banned on clearnet, other identities behind the same onion service are not |
| `test_outbound_onion_behaviour_ban` | `report()` reaches the identity ban for `/onion3` too |
| `test_ban_addr_without_ip_or_peer_id_is_a_noop` | Address with neither IP nor `PeerId` does not panic and bans nothing |
| `test_proxy_ban_admission_isolated_by_peer_id` | `accept_peer` admission: banned identity rejected, siblings admitted, whitelist exemption preserved |

`cargo test -p ckb-network --lib -- --test-threads=1` → 65 passed.
`cargo clippy --workspace --all-targets` → no new warnings.

Integration tests are unaffected: `listen_on_onion` defaults to `false`, only
`test/src/specs/tor/*` enables it, and none of those specs touch the ban list. The specs
that do assert on `get_banned_addresses` (`malformed_message`, `transaction_relay`,
`too_many_unknown_transactions`, `sync_invalid`, and `Node::connect_and_wait_ban`) keep
running through the unchanged IP-ban path.

Side effects

- No performance regression
- No breaking backward compatibility. `get_banned_addresses` output is unchanged for all
  non-onion nodes. On a node with onion listening enabled, automatic bans of onion peers
  no longer appear there, because there is no meaningful IP to report — those are logged
  at `debug` level instead.

### DIFF
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -121,9 +121,10 @@ impl NetworkState {
             .collect();
         info!("Loading the peer store. This process may take a few seconds to complete.");
 
-        let peer_store = Mutex::new(PeerStore::load_from_dir_or_default(
-            config.peer_store_path(),
-        ));
+        let peer_store = Mutex::new(
+            PeerStore::load_from_dir_or_default(config.peer_store_path())
+                .with_shared_proxy_addrs(config.shared_proxy_addrs()),
+        );
         info!("Loaded the peer store.");
 
         if let Some(ref proxy_url) = config.proxy.proxy_url {
@@ -178,7 +179,11 @@ impl NetworkState {
             })
             .collect();
         info!("Loading the peer store. This process may take a few seconds to complete.");
-        let peer_store = Mutex::new(PeerStore::load_from_idb(config.peer_store_path()).await);
+        let peer_store = Mutex::new(
+            PeerStore::load_from_idb(config.peer_store_path())
+                .await
+                .with_shared_proxy_addrs(config.shared_proxy_addrs()),
+        );
         let bootnodes = config.bootnodes();
 
         let peer_registry = PeerRegistry::new(
--- a/network/src/network.rs
+++ b/network/src/network.rs
@@ -983,7 +983,9 @@ impl NetworkService {
             .set_channel_size(config.channel_size())
             .timeout(Duration::from_secs(5))
             .onion_timeout(Duration::from_secs(120))
-            .trusted_proxies(config.trusted_proxies.clone());
+            // Tor forwards client-controlled bytes without authenticating proxy
+            // metadata. Preserve the socket source used by the identity-ban policy.
+            .trusted_proxies(config.forwarding_metadata_proxies());
 
         #[cfg(not(target_family = "wasm"))]
         {
--- a/network/src/peer_store/ban_list.rs
+++ b/network/src/peer_store/ban_list.rs
@@ -1,17 +1,23 @@
 //! Ban list
 use crate::peer_store::Multiaddr;
 use crate::peer_store::types::{BannedAddr, ip_to_network};
+use crate::{PeerId, extract_peer_id};
 use ckb_systemtime::unix_time_as_millis;
+use ckb_util::LinkedHashMap;
 use ipnetwork::IpNetwork;
 use p2p::utils::multiaddr_to_socketaddr;
 use std::collections::HashMap;
 use std::net::IpAddr;
 
 pub(crate) const CLEAR_INTERVAL_COUNTER: usize = 1024;
+// Proxy clients can change identities, so bound these temporary bans.
+pub(crate) const MAX_BANNED_PEERS: usize = 16384;
 
 /// Ban list
 pub struct BanList {
     inner: HashMap<IpNetwork, BannedAddr>,
+    // Session identity bans are intentionally memory-only; the persisted/RPC ban list is IP based.
+    peers: LinkedHashMap<PeerId, u64>,
     insert_count: usize,
 }
 
@@ -26,6 +32,7 @@ impl BanList {
     pub fn new() -> Self {
         BanList {
             inner: HashMap::default(),
+            peers: LinkedHashMap::default(),
             insert_count: 0,
         }
     }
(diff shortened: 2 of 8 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding (Medium)
Reviewed range: `f75d609f9bfdd0a3160e61d21a38a46c01e45949..451e6dcffe03eb2a22e54108efdd1f1bd13f7141`
Review model: `Codex (GPT-5)`

Validation:
- Targeted CKB proxy tests passed (6/6), and targeted tests at the exact locked Tentacle revision passed for both PROXY-protocol and WebSocket forwarding-address handling (5/5).

Notes:
- Onion-forwarded connections arrive from trusted loopback, but Tentacle may replace that transport address from peer-supplied PROXY-protocol or `X-Forwarded-For` metadata before CKB's new shared-proxy check. A remote onion peer can therefore make a later ban target a peer-selected IP instead of only its authenticated `PeerId`.
- Please preserve authenticated transport/onion-forwarder provenance separately from any claimed client address, and add end-to-end regression coverage for both raw TCP and WebSocket onion paths.

PR #5322, opened 2026-09-08
fix(sync): rate limit transaction fetches by hash count

Officeyutong:

Problem Summary:

`GetRelayTransactions` accepts up to 32,767 hashes per request, but the existing limiter charges one unit per message. Large requests therefore consume more allocation and tx-pool lookup work without consuming additional rate-limit quota.

What's Changed:

- Add a per-peer hash-count limiter with a refill rate of 32,767 hashes per second and the same burst capacity, allowing one maximum-sized request.
- Check the budget before constructing the hash set or submitting a tx-pool query. Reject requests exceeding the budget with `TooManyRequests`.
- Preserve the existing per-peer, per-message-type limiter.
- Return early for empty requests and clean up expired limiter entries on disconnect.

### Check List

Tests:

- Add unit tests covering weighted accounting, peer isolation, remaining budget after rejection, empty requests, and maximum-batch admission.
- `cargo nextest run -p ckb-sync --lib`: 85 tests passed.
- `make clippy`: passed.
- `cargo fmt --all -- --check`: passed.

Side effects:

- Sustained large fetch requests may now be throttled even when they satisfy the message-count limit.
- The initial hash budget has not been calibrated through performance benchmarks.

### DIFF
--- a/sync/src/relayer/get_transactions_process.rs
+++ b/sync/src/relayer/get_transactions_process.rs
@@ -39,6 +39,20 @@ impl<'a> GetTransactionsProcess<'a> {
             }
         }
 
+        // The batch bound above makes this conversion safe. Empty requests still
+        // consume the outer message quota, but do not require any pool lookup.
+        let Some(hash_count) = std::num::NonZeroU32::new(message_len as u32) else {
+            return Status::ok();
+        };
+        if !matches!(
+            self.relayer
+                .tx_fetch_rate_limiter
+                .check_key_n(&self.peer, hash_count),
+            Ok(Ok(_))
+        ) {
+            return StatusCode::TooManyRequests.with_context("GetRelayTransactions hashes");
+        }
+
         let tx_hashes = self.message.tx_hashes();
 
         trace_target!(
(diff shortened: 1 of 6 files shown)


chainTe:
## Security Review

Result: 1 confirmed finding (medium severity)
Reviewed range: `f75d609f9bfdd0a3160e61d21a38a46c01e45949..d0be25484facc147eeac05121bb67bf3833deb98`
Review model: `GPT-5 (Codex)`

Validation:
- The two PR-specific `test_fetch_budget` unit tests pass.
- Focused source tracing covered RelayV3 dispatch, Governor batch semantics, tx-pool lookup, packed response construction, and network limits.

Finding:
- The new limiter admits 32,767 hashes per peer session each second, but `GetTransactionsProcess` fetches and packs every resident match before splitting the result into 1 MiB messages. On the default 180 MB transaction-pool limit, one roughly 1 MiB request can therefore induce response work proportional to the whole pool; separate or reconnected sessions receive fresh buckets. Please meter returned serialized bytes and node-wide aggregate work, cap/stream results before materializing the full response, and stop immediately after an intermediate send failure.

PR #5325, opened 2026-09-09
fix: prevent overflow in various calculations and add validation 

chenyukang:

Prevent overflow in numberic calculations and add validation for spec parameters.

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

### Related changes

- PR to update `owner/repo`:

### Check List 

Tests 

- Unit test
- Integration test
- No code

Side effects

- Performance regression

### DIFF
--- a/spec/src/versionbits/mod.rs
+++ b/spec/src/versionbits/mod.rs
@@ -420,7 +420,11 @@ impl<'a> Versionbits<'a> {
 
     /// return bit mask corresponding deployment
     pub fn mask(&self) -> u32 {
-        1u32 << self.deployment().bit as u32
+        let bit = u32::from(self.deployment().bit);
+        debug_assert!(bit < VERSIONBITS_NUM_BITS);
+        1u32.checked_shl(bit)
+            .filter(|_| bit < VERSIONBITS_NUM_BITS)
+            .unwrap_or(0)
     }
 }
 
--- a/chain/src/utils/orphan_block_pool.rs
+++ b/chain/src/utils/orphan_block_pool.rs
@@ -115,7 +115,7 @@ impl InnerPool {
             .get(parent_hash)
             .and_then(|map| {
                 map.iter().next().map(|(_, lonely_block)| {
-                    lonely_block.epoch_number() + EXPIRED_EPOCH < tip_epoch
+                    tip_epoch.saturating_sub(lonely_block.epoch_number()) > EXPIRED_EPOCH
                 })
             })
             .unwrap_or_default()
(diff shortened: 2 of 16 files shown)


spec/src/versionbits/mod.rs
```diff
@@ -420,7 +420,13 @@ impl<'a> Versionbits<'a> {
 
     /// return bit mask corresponding deployment
     pub fn mask(&self) -> u32 {
-        1u32 << self.deployment().bit as u32
```
eval-exec, line 423: how about using `.checked_shl`: https://doc.rust-lang.org/std/primitive.u32.html#method.checked_shl

chenyukang: fixed.

eval-exec (approved):

Xcodes-chain:
Test analysis and py integration update:

- Added py integration regression PR: https://github.com/nervosnetwork/ckb-py-integration-test/pull/137
- Coverage added for consensus/spec parameter validation from this PR:
  - `genesis_epoch_length = 0` is rejected with `InvalidParams` and no panic.
  - `epoch_duration_target = 0` is rejected with `InvalidParams` and no panic.
  - `orphan_rate_target = [1, 0]` is rejected with `InvalidParams` and no panic.
  - large `genesis_epoch_length * orphan_rate_target numerator` overflow is rejected with `InvalidParams` and no panic.
  - a minimal valid boundary spec (`genesis_epoch_length = 1`, `epoch_duration_target = 1`) still starts and mines a block.

Local validation against this PR build (`d0909b3`):

```bash
cargo test -p ckb-chain-spec test_invalid_consensus_params_are_rejected --locked
cargo test -p ckb-types epoch_ext --locked
python3 -m black --check test_cases/v209/test_5325_overflow_hardening.py
python3 -m pytest -vv test_cases/v209/test_5325_overflow_hardening.py
CKB_PR5325_ENABLE=1 python3 -m pytest -vv test_cases/v209/test_5325_overflow_hardening.py
```

Results:
- Rust targeted tests passed.
- Default py run: 5 skipped, because this PR is not yet in the default CKB binary.
- PR #5325 enabled py run: 5 passed.

Finding / integration note:
- The Rust unit test can directly construct `u64::MAX`, but the external TOML spec parser rejects `18446744073709551615` first with `invalid number` before consensus validation runs. The integration overflow case uses `9223372036854775807 * 4294967295`, which is TOML-parseable and still exercises the checked multiplication overflow path.
- No CKB functional issue found in the PR build during the local py integration validation.

PR #5327, opened 2026-09-11
perf(script): reuse loaded cell data in syscall

zhangsoledad:

`load_cell_data` loads the cell data to check the source and get the remaining length. For a nonzero-size request, it then calls `Snapshot2Context::store_bytes`, which loads the same data again. We already have the bytes, so this second lookup is unnecessary.

I'm splitting this optimization out of #5316 so it can be reviewed and merged independently of the tx-pool refactor.

Reuse the loaded bytes when writing VM memory, preserving the existing length write-back, truncation, page tracking order, return code, and cycle accounting.

Add a regression test for VM V0, V1, and V2 that counts provider reads and compares the result against `Snapshot2Context::store_bytes`. It covers empty and partial reads, page boundaries, overlapping buffers, invalid indices and pointers, oversized offsets and lengths, and frozen pages. Errors, registers, cycles, and snapshot pages must match.

The old implementation fails the read-count assertion with two reads instead of one.

### Check List

- [x] Counting and differential regression tests.
- [x] Strict workspace Clippy with all targets and the Makefile feature/lint flags.
- [x] Formatting and `git diff --check`.

### DIFF
--- a/script/src/syscalls/load_cell_data.rs
+++ b/script/src/syscalls/load_cell_data.rs
@@ -58,7 +58,7 @@ where
             .map_err(|e| VMError::Unexpected(e.to_string()))?;
 
         match sc.load_data(&data_piece_id, offset, u64::MAX) {
-            Ok((cell, _)) => {
+            Ok((mut cell, _)) => {
                 let size = machine.memory_mut().load64(&size_addr)?.to_u64();
                 if size == 0 {
                     machine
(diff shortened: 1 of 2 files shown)


mohanson (approved):
No differences in success or failure paths were found.

script/src/syscalls/load_cell_data.rs
```diff
@@ -67,23 +67,16 @@ where
...
-                    Ok(val) => val,
-                    Err(VMError::SnapshotDataLoadError) => {
-                        // This comes from TxData results in an out of bound error, to
-                        // mimic current behavior, we would return INDEX_OUT_OF_BOUND error.
-                        machine.set_register(A0, Mac::REG::from_u8(INDEX_OUT_OF_BOUND));
-                        return Ok(());
-                    }
-                    Err(e) => return Err(e),
-                };
+                // Keep Snapshot2Context::store_bytes ordering while reusing the
```
chenyukang, line 70: it's unwrap `store_bytes` and skip an extra `load_data`, how much perf improvement we can get?

